PR #33, opened 2022-04-04
DMS-53: Adding support for Basic Table Partitioning for Oracle & PostgreSQL

sanyamsinghal:
This PR adds support for basic table partitioning(List, Hash, Range) for Oracle and PostgreSQL

Following code changes were done to support the above features.
1. PostgreSQL: Only change was to look for table_data.sql files mentioned in toc.dat during the progress bar initialization. Changed the progress bar struct array to use struct map making it similar to data import code.

2. Oracle: Added `OracleGetAllPartitionNames()` to fetch partition names along with table list from source(Now ignores Index Tables starting with DR$ and Temporary tables for the current session).
So, updated the system query will look like this: `SELECT table_name FROM all_tables WHERE owner = schema AND TEMPORARY = 'N' AND table_name NOT LIKE 'DR$%%'`

3. Using Full Tables Names for queries like `count(*)` to support different users accessing the schema rather than the owner. For example, if the migration user for oracle is `admin`(superuser) but the schema to export is `sakila`

4. Internal Important Changes:
     i. The progress bar name will be "tableName" by default, "tableName PARTITION(partitionName)" in case of partitions, "schema.tableName" in case more than one schema in a database(postgres)
    ii. For all queries made to source, now we are using fullTableName: "schema.table" for Oracle and postgres, "database.table" for MySQL


yb_migrate/src/migration/common.go
```diff
@@ -309,20 +317,27 @@ func SelectVersionQuery(dbType string, dbConnStr string) string {
...
+
+	saveExportedRowCount(exportDir, tablesProgressMetadata)
+}
+
+func renameDataFiles(tablesProgressMetadata *map[string]*utils.TableProgressMetadata) {
+	for _, tableProgressMetadata := range *tablesProgressMetadata {
+		oldFilePath := tableProgressMetadata.InProgressFilePath
+		newFilePath := tableProgressMetadata.FinalFilePath
 		if utils.FileOrFolderExists(oldFilePath) {
 			os.Rename(oldFilePath, newFilePath)
```
kneeraj, line 335: Please handle the returned err of os.Rename

kneeraj (approved):
LGTM

One nit comment. Please take care.

PR #39, opened 2022-04-08
DMS-33: Enhance dependencies installer script of centos - to acknowledge for ora2pg license & check dependencies version

sanyamsinghal:


installer_scripts/install_centos7.sh
```diff
@@ -2,142 +2,171 @@
...
 if [ $? -ne 0 ]; then
     echo "pg_dump not found."
     while true; do
-        read -p "Install Postgres11? (Y/N): " yn
+        read -p "Install Postgres14? (Y/N): " yn
         case $yn in
             [Yy]* ) 
-            echo "Installing Postgress11."
-            sudo yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm >> mylog.log
+            echo "Installing Postgres14"
```
rahulb-yb, line 92: Are we sure that we want to install PG14 on the client machine? I'm not sure which versions work with yb_migrate

rahulb-yb (approved):
One small concern mentioned in a comment, other than that LGTM

installer_scripts/install_centos7.sh
sanyamsinghal: in case its not installed then only we are going to do that. Other we are checking for minimum version which is 10 i think rightnow. PG_DUMP_MIN_VERSION

PR #54, opened 2022-04-13
Disable BLOB columns data export

sanyamsinghal:
In Oracle/MySQL, Blobs will be considered in schema but no data will be transfer for Blobs columns, so they will be empty.
In PostgreSQL, `bytea` and `text` will behave normally as other datatypes but Large Object`lo` datatype will not be exported.

Testing:
1. Tested standard sakila schema for the following changes with all Oracle and PostgreSQL.
2. Tested the same for ds2 schema having partitioned tables with Oracle and PostgreSQL.
3. Also verified the correctness of the corresponding new log lines for these changes.


yb_migrate/src/migration/postgres.go
```diff
@@ -33,28 +33,6 @@ import (
 
 var commandNotFoundRegexp *regexp.Regexp = regexp.MustCompile(`(?i)not[ ]+found[ ]+in[ ]+\$PATH`)
 
-// TODO: check for pgdump and psql/ysqlsh - installed/set-in-path
-func CheckToolsRequiredForPostgresExport() {
```
kneeraj, line 37: Why this code was required to be removed for Blob removal capability?

kneeraj (approved):
LGTM

A couple of clarifications. Can you also update the PR with how you tested it and for what all source types you tested this. Thanks.

yb_migrate/src/utils/utils.go
```diff
@@ -212,7 +212,7 @@ func CheckToolsRequiredInstalledOrNot(source *Source) {
 	case "oracle":
 		toolsRequired = []string{"ora2pg", "sqlplus"}
 	case "postgresql":
-		toolsRequired = []string{"pg_dump", "strings"}
+		toolsRequired = []string{"pg_dump", "strings", "pg_restore"}
```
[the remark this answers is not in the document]
sanyamsinghal: Its not related to this PR. In general, we are using pg_restore so there should be a check for it.

yb_migrate/src/migration/postgres.go
sanyamsinghal: This function is stale/not-used now. Instead we are using `CheckToolsRequiredInstalledOrNot()` everywhere.

PR #73, opened 2022-04-27
Refactor source db access.

amit-yb:


yb_migrate/src/srcdb/source.go
```diff
@@ -31,6 +37,26 @@ type Source struct {
...
+
+	sourceDB SourceDB
+}
+
+func (s *Source) DB() SourceDB {
+	if s.sourceDB != nil {
+		return s.sourceDB
+	}
+
+	s.Lock()
```
sanyamsinghal, line 51: why Lock() is needed here?

yb_migrate/src/srcdb/source.go
```diff
@@ -104,3 +130,108 @@ func (s *Source) ParseURI() {
...
+		return tls.Config{
+			RootCAs:            rootCertPool,
+			Certificates:       clientCert,
+			InsecureSkipVerify: false,
+			ServerName:         source.Host,
+		}
+	}
+}
+
+func generateSSLQueryStringIfNotExists(s *Source) string {
```
sanyamsinghal, line 207: nit: it is postgres specific, can be moved to `srcdb/postgres.go`

yb_migrate/src/srcdb/source.go
```diff
@@ -104,3 +130,108 @@ func (s *Source) ParseURI() {
 
 	}
 }
+
+func parseSSLString(source *Source) {
```
sanyamsinghal, line 137: parseSSLString() and createTLSConf() is for mysql only, can be moved to srcdb/mysql.go

yb_migrate/src/srcdb/utils.go
```diff
@@ -0,0 +1,12 @@
+package srcdb
+
+import (
+	"fmt"
+	"log"
+	"os"
+)
+
+func ErrExit(formatString string, args ...interface{}) {
```
sanyamsinghal, line 10: is there some issue in using this function from utils package directly?

yb_migrate/src/srcdb/mysql.go
```diff
@@ -0,0 +1,53 @@
...
+
+	db *sql.DB
+}
+
+func newMySQL(s *Source) *MySQL {
+	return &MySQL{source: s}
+}
+
+func (ms *MySQL) Connect() error {
+	db, err := sql.Open("mysql", ms.getConnectionString())
```
sanyamsinghal, line 22: i think we can not using this connection outside the Connect() function.
so its good to have `defer db.Close()`

sanyamsinghal: similarly for other sources

sanyamsinghal (approved):
LGTM, left some comments please respond/take-care before merge

yb_migrate/src/srcdb/mysql.go
amit-yb: That's intentional. We want all source db access to happen behind the `SourceDB` implementation. Our eventual plan is to encapsulate ALL source db access inside the `srcdb` package.

That will make it easy to add support for other databases.

yb_migrate/src/srcdb/source.go
amit-yb: This is a standard implementation of a Singleton pattern. If two go-routines call `source.DB()` at (almost) the same time, there is a possibility that we may instantiate the `SourceDB` twice. That's prevented by the lock.

yb_migrate/src/srcdb/source.go
amit-yb: These functions are only accessing the `Source` type. Hence I kept them closer to the `Source` type. I will consider making them methods of the `Source` type. For now, I will leave them as they are here.

yb_migrate/src/srcdb/source.go
amit-yb: Same as above.

yb_migrate/src/srcdb/utils.go
amit-yb: As of now, `utils` is importing `srcdb` for the `Source` type. Importing `utils` in the `srcdb` package will create a circular import.

If I manage to remove `srcdb` imports from the `utils` package, I will consider using `utils` in this package.

sanyamsinghal: got it

PR #75, opened 2022-04-28
Bug fix: import data restartability in case of oracle

sanyamsinghal:


yb_migrate/cmd/importData.go
```diff
@@ -380,8 +379,11 @@ func checkPrimaryKey(tableName string) bool {
 		table = strings.Split(tableName, ".")[0]
 	}
 
+	/* currently object names for yugabytedb is implemented as case-sensitive i.e. lower-case
+	but in case of oracle exported data files(which we use for to extract tablename)
+	so eg. file EMPLOYEE_data.sql -> table EMPLOYEE which needs to converted for using further */
 	checkPKSql := fmt.Sprintf(`SELECT * FROM information_schema.table_constraints
-	WHERE constraint_type = 'PRIMARY KEY' AND table_name = '%s' AND table_schema = '%s';`, table, schema)
+	WHERE constraint_type = 'PRIMARY KEY' AND table_name = '%s' AND table_schema = '%s';`, strings.ToLower(table), schema)
```
amit-yb, line 386: This will break case-sensitive PG migration.

sanyamsinghal: Just pushed the updated. Have a look @amit-yb

yb_migrate/cmd/importData.go
```diff
@@ -380,6 +379,14 @@ func checkPrimaryKey(tableName string) bool {
 		table = strings.Split(tableName, ".")[0]
 	}
 
+	sourceDBType := ExtractMetaInfo(exportDir).SourceDBType
+	if sourceDBType == ORACLE {
+		table = strings.ToLower(table)
+	}
+
```
amit-yb, line 386: Can you please confirm the behaviour for MySQL as well?

amit-yb (approved):
LGTM with an important comment.

yb_migrate/cmd/importData.go
sanyamsinghal: yes tested that, MySQL exports the `table_data.sql` file into lowercase, so this is not needed.

PR #78, opened 2022-04-29
Refactor target db access into a separate `tgtdb` package

amit-yb:


rahulb-yb (approved):
Looks fine, please test against RF3 cluster for import once before merging- P.S. the TargetDB struct is a good addition! Thanks for this

PR #79, opened 2022-05-04
Rename `generateReport` command to `analyze-schema`.

amit-yb:


rahulb-yb (approved):

sanyamsinghal (approved):
LGTM but i think we should update the current documentation also super soon to avoid confusion.

PR #82, opened 2022-05-04
Remerge the `"Run generateReport on exported schema."` patch after fixing regressions.

amit-yb:


yb_migrate/cmd/exportSchema.go
```diff
@@ -61,25 +80,17 @@ func exportSchema() {
 
 	switch source.DBType {
 	case ORACLE:
-		utils.PrintIfTrue("preparing Ora2Pg for schema export from Oracle\n", source.VerboseMode, !source.GenerateReportMode)
```
sanyamsinghal, line 64: anyspecific reason to remove these lines? we can still have them with verbose mode

sanyamsinghal (approved):

yb_migrate/cmd/exportSchema.go
amit-yb: These lines provide no valuable information to the user. And the terms `ora2pg` and `pg_dump` are internal implementation details.

PR #84, opened 2022-05-05
Added case for language c regex in generateReport

rahulb-yb:


amit-yb (approved):

yb_migrate/cmd/analyzeSchema.go
```diff
@@ -681,7 +687,7 @@ func generateTxtReport(Report utils.Report) string {
 	}
 	for i := 0; i < len(Report.Issues); i++ {
 		txtstring += "Error in Object " + Report.Issues[i].ObjectType + ":\n"
```
sanyamsinghal, line 689: i think colon(:) related edit can be done to this line also.
```
txtstring += "Error in Object: " + Report.Issues[i].ObjectType + "\n"
```

yb_migrate/cmd/analyzeSchema.go
```diff
@@ -394,6 +396,10 @@ func checkDDL(sqlStmtArray [][]string, fpath string) {
 		} else if spc := alterTblSpcRegex.FindStringSubmatch(line[0]); spc != nil {
 			reportCase(fpath, "ALTER TABLESPACE not supported yet.",
 				"https://github.com/YugaByte/yugabyte-db/issues/1153", "", "TABLESPACE", spc[1], line[1])
+		} else if tbl := cLangRegex.FindStringSubmatch(line[0]); tbl != nil {
+			reportCase(fpath, "LANGUAGE C not supported yet.",
```
sanyamsinghal, line 400: can we imporve the message here? Instead of saying not supported, how about mentioning for non-superuser case.

yb_migrate/cmd/analyzeSchema.go
rahulb-yb: This is the first line for each issue, which is why I had not removed it as it looks well-formatted

sanyamsinghal: ohk got it

sanyamsinghal (approved):

PR #90, opened 2022-05-09
Added --load-balancer flag in import command

sanyamsinghal:
This flag will be useful for data import execution and to decide correct default for number of parallel jobs

### DIFF
--- a/yb_migrate/cmd/importData.go
+++ b/yb_migrate/cmd/importData.go
@@ -63,6 +63,7 @@ var allTables []string
 var usePublicIp bool
 var targetEndpoints string
 var copyTableFromCommands = make(map[string]string)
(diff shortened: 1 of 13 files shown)


yb_migrate/cmd/importData.go
```diff
@@ -63,6 +63,7 @@ var allTables []string
 var usePublicIp bool
 var targetEndpoints string
 var copyTableFromCommands = make(map[string]string)
+var LoadBalancerUsed bool // specifies whether load balancer is used in front of yb servers
```
amit-yb, line 66: Start the variable name with a lower-case letter `loadBalancerUsed`.

amit-yb (approved):

yb_migrate/cmd/importData.go
sanyamsinghal: thanks, missed it.

PR #94, opened 2022-05-10
Added logging for ora2pg and pg_dump STDERR, STDOUT 

rahulb-yb:


yb_migrate/src/migration/postgres.go
```diff
@@ -50,7 +50,11 @@ func PgDumpExtractSchema(source *srcdb.Source, exportDir string) {
 	log.Infof("Running command: %s", preparePgdumpCommandString)
 	preparedYsqldumpCommand := exec.Command("/bin/bash", "-c", preparePgdumpCommandString)
 
-	err := preparedYsqldumpCommand.Run()
+	stdout, err := preparedYsqldumpCommand.CombinedOutput()
+	//pg_dump formats its stdout messages, %s is sufficient.
+	if string(stdout) != "" {
+		log.Infof("%s", string(stdout))
+	}
```
amit-yb, line 57: Can we use a mechanism similar to other places i.e. the one that uses `bytes.Buffer` for STDOUT and STDERR.

yb_migrate/src/migration/oracle.go
```diff
@@ -193,28 +194,21 @@ func Ora2PgExportDataOffline(ctx context.Context, source *srcdb.Source, exportDi
...
-
-	err = exportDataCommand.Start()
+	err := exportDataCommand.Start()
 	fmt.Println("starting ora2pg for data export...")
+	if outbuf.String() != "" {
+		log.Infof("ora2pg STDOUT: %s", outbuf.String())
+	}
 	if err != nil {
-		utils.ErrExit("Error while starting ora2pg for data export: %v", err)
+		utils.ErrExit("Error while starting ora2pg for data export: %v\n%s", err, errbuf.String())
```
amit-yb, line 211: You will also need to log the `errbuf` if the `exportDataCommand.Wait()` returns error on line 216.

yb_migrate/src/migration/postgres.go
```diff
@@ -243,7 +251,7 @@ func PgDumpExportDataOffline(ctx context.Context, source *srcdb.Source, exportDi
 	// Wait for pg_dump to complete before renaming of data files.
 	err = proc.Wait()
 	if err != nil {
-		utils.PrintAndLog("pg_dump failed to export data with error: %v\n%s", err, buf.String())
+		utils.PrintAndLog("pg_dump failed to export data with error: %v\n%s", err, errbuf.String())
```
amit-yb, line 254: The `errbuf` contents should only go to the log.

Point the user to the log file in the message sent on the STDOUT.

yb_migrate/src/migration/oracle.go
```diff
@@ -77,6 +77,7 @@ func Ora2PgExtractSchema(source *srcdb.Source, exportDir string) {
 					log.Infof("ERROR in output scanner goroutine: %s", line)
 					runtime.Goexit()
 				}
+				log.Infof("ora2pg STDOUT: %s", outScanner.Text())
```
amit-yb, line 80: Save the result of the `outScanner.Text()` in a local variable and use it everywhere.

amit-yb: I recommend using the `bytes.Buffer` approach (to capture STDOUT/STDERR) here as well.

yb_migrate/src/migration/postgres.go
rahulb-yb: The reason I have not opted for `bytes.Buffer` over here (and other places) is because the command is/was being executed like so: `command.Run()`, which does not need a corresponding `command.Wait()`. Wherever there is/was a `command.Start()', we can choose to opt for the `bytes.Buffer` approach. Instead I have replaced `command.Run()` with `command.CombinedOutput()`, which provides both STDOUT and STDERR, while also not needing to call `command.Wait()`

yb_migrate/src/migration/oracle.go
sanyamsinghal: nit: add this line in `else` condition, since if statement already have log

yb_migrate/src/migration/oracle.go
```diff
@@ -196,26 +196,27 @@ func Ora2PgExportDataOffline(ctx context.Context, source *srcdb.Source, exportDi
...
+	// }
+	// defer stdErrFile.Close()
 
-	exportDataCommand.Stdout = stdOutFile
-	exportDataCommand.Stderr = stdErrFile
+	// exportDataCommand.Stdout = stdOutFile
+	// exportDataCommand.Stderr = stdErrFile
 
-	err = exportDataCommand.Start()
+	stdout, err := exportDataCommand.CombinedOutput()
```
sanyamsinghal: @rahulb-yb have you tested these changes?
You have changed `Start()` call to `CombinedOutput()`
This is going to break code/tool i believe

Start() - will start the execution of command and don't wait command to exit/complete

So with CombinedOutput(), your current goroutine will stay at this point and line 221 will not be executed till the end which in turn will cause the progress bar and everything to stall.

yb_migrate/src/migration/oracle.go
```diff
@@ -193,28 +194,21 @@ func Ora2PgExportDataOffline(ctx context.Context, source *srcdb.Source, exportDi
 		source.NumConnections, projectDirPath, configFilePath)
 
 	//Exporting all the tables in the schema
-	exportDataCommand := exec.Command("/bin/bash", "-c", exportDataCommandString)
+	exportDataCommand := exec.CommandContext(ctx, "/bin/bash", "-c", exportDataCommandString)
```
sanyamsinghal, line 197: Nice catch.

yb_migrate/src/migration/oracle.go
sanyamsinghal: +1

yb_migrate/src/migration/oracle.go
rahulb-yb: Is there a need to save outScanner.Text() in a local variable? It is used sparingly in the code segment

sanyamsinghal (changes requested):

yb_migrate/src/migration/oracle.go
sanyamsinghal: I was reviewing commit by commit.
I think at the end you have corrected this.

yb_migrate/src/migration/oracle.go
rahulb-yb: if condition logs whenever an error is caught, I have purposely kept this log statement ahead of the error-catching If statement due to the `runtime.Goexit()` statement within the if condition

yb_migrate/src/migration/oracle.go
rahulb-yb: yeah, I was attempting to understand and compare at first with main, so I temporarily had this as an initial commit

yb_migrate/src/migration/oracle.go
rahulb-yb: I have made this change for readability in the latest commit 👍🏽

yb_migrate/src/migration/oracle.go
sanyamsinghal: > nit: add complete path for yb_migrate.log fmt.Printf("pg_dump failed to export data with error: %v. Refer to'%s/yb_migrate.log' for further details.", err, exportDir)

sorry, I missed a space after `Refer to` 
fmt.Printf("pg_dump failed to export data with error: %v. Refer to '%s/yb_migrate.log' for further details.", err, exportDir)

sanyamsinghal (approved):
LGTM! Assuming you have testing the changes for all three sources, so that nothing breaks.

PR #95, opened 2022-05-10
Add GetAllPartitionNames() to the SourceDB interface.

amit-yb:


yb_migrate/src/srcdb/mysql.go
```diff
@@ -72,6 +72,10 @@ func (ms *MySQL) GetAllTableNames() []string {
 	return tableNames
 }
 
+func (ms *MySQL) GetAllPartitionNames(tableName string) []string {
+	panic("Not Implemented")
```
sanyamsinghal, line 76: @amit-yb is it okay to panic() as it shows the complete stack trace? or we should use utils.ErrExit() here

sanyamsinghal (approved):
LGTM, please respond to the comment before merging.

yb_migrate/src/srcdb/mysql.go
amit-yb: In its current form this function should never be called. If it gets called for PG or MySQL, we will get this panic message right away during testing. So, users will never see this panic.

sanyamsinghal: got it, thanks!

PR #99, opened 2022-05-12
 Modified isDataLine() to make check based on source-db-type and positioning of line in data file

sanyamsinghal:


sanyamsinghal:
Raised a another PR(#100) with the suggested new changes

PR #127, opened 2022-05-16
Added no_header flag in ora2pg calling commands

sanyamsinghal:
- This will remove unnecessary header comments and SET client_encoding statements

### DIFF
--- a/yb_migrate/cmd/importSchemaYugabyteDB.go
+++ b/yb_migrate/cmd/importSchemaYugabyteDB.go
@@ -82,6 +82,13 @@ func YugabyteDBImportSchema(target *tgtdb.Target, exportDir string) {
 			if err != nil {
 				utils.ErrExit("Failed to run %q on target DB: %s", setSchemaQuery, err)
 			}
+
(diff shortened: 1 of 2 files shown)


amit-yb (approved):

yb_migrate/cmd/importSchemaYugabyteDB.go
```diff
@@ -82,6 +82,13 @@ func YugabyteDBImportSchema(target *tgtdb.Target, exportDir string) {
 			if err != nil {
 				utils.ErrExit("Failed to run %q on target DB: %s", setSchemaQuery, err)
 			}
+
+			setClientEncQuery := IMPORT_SESSION_SETTERS[0]
```
rahulb-yb, line 86: what does this part do, just curious

sanyamsinghal: @rahulb-yb the default `Set client_encoding to 'UTF8'` will be removed by using flag from each schema file, so it is required to be added additionally.

sanyamsinghal:
Raised a new PR #129 with the same changes

Issue #60, opened 2022-04-14
PG13: unrecognized configuration parameter "default_table_access_method" (SQLSTATE 42704)


amit-yb:
While migrating a DB from Postgres 13.6, `import schema` fails with the following error:
```
importing     SCHEMA      
    ERROR: unrecognized configuration parameter "default_table_access_method" (SQLSTATE 42704)
    STATEMENT: SET default_table_access_method = heap;
ERROR: unrecognized configuration parameter "default_table_access_method" (SQLSTATE 42704)
```
`yb_migrate` should drop the line before attempting schema import. Until yb_migrate does that users are advised to manually look for the lines in the `$EXPORT_DIR/schema/***` and delete them.

rahulb-yb:
Added commit to resolve this issue: #128

PR #128, opened 2022-05-17
Added support for PG13

rahulb-yb:
PR for #60


amit-yb (changes requested):
All you have to do is skip the lines having `"SET default_table_access_method"` prefix. Don't make it "PG13" aware. Tomorrow, there will many other version and cases.

yb_migrate/src/srcdb/pg_dump_extract_schema.go
```diff
@@ -189,3 +189,7 @@ func extractSqlStatements(schemaFileLines []string, index *int) string {
 func isSqlComment(line string) bool {
 	return len(line) >= 2 && line[:2] == "--"
 }
+
+func isPG13statement(line string) bool {
```
amit-yb, line 193: Rename this function to `shouldSkipLine(line string)`.

sanyamsinghal (approved):
LGTM

amit-yb (approved):

PR #134, opened 2022-05-24
Add MySQL migration test.

amit-yb:


installer_scripts/common.sh
```diff
@@ -36,8 +36,10 @@ update_yb_migrate_bashrc() {
 
 install_yb_migrate() {
 	output "Installing yb_migrate."
-	$GO install github.com/yugabyte/yb-db-migration/yb_migrate@latest
+	pushd ../yb_migrate
+	$GO install
```
sanyamsinghal, line 40: any specific reason for this change? Both ways are fine, right?

migtests/scripts/functions.sh
```diff
@@ -92,7 +98,7 @@ import_schema() {
 }
 
 import_data() {
-	yb_migrate import data --export-dir ${EXPORT_DIR} \
+	yes | yb_migrate import data --export-dir ${EXPORT_DIR} \
```
sanyamsinghal, line 101: where do we require yes in import data also?

sanyamsinghal (approved):
LGTM! Left some minor comments, please respond before merging.

installer_scripts/common.sh
amit-yb: With the earlier command, the pre-commit tests were always installing `yb_migrate` from the `main` branch. What we actually need is to test the migration with the changes in the PR.

migtests/scripts/functions.sh
amit-yb: For MySQL, when we try to import in the `public` schema, yb_migrate prompts for confirmation.

sanyamsinghal: ok, i thought it is only for import schema.

installer_scripts/common.sh
sanyamsinghal: Thanks.

PR #141, opened 2022-05-27
export-schema: log ora2pg STDOUT and STDERR.

amit-yb:


yb_migrate/src/srcdb/ora2pg_extract_schema.go
```diff
@@ -18,77 +16,52 @@ func ora2pgExtractSchema(source *Source, exportDir string) {
...
 
-		err := exportSchemaObjectCommand.Start()
+		err := cmd.Start()
 		if err != nil {
-			utils.PrintAndLog("Error while starting export: %v", err)
-			exportSchemaObjectCommand.Process.Kill()
-			continue
+			utils.WaitChannel <- 1 //stop waiting with exit code 1
+			<-utils.WaitChannel
+			utils.ErrExit("Failed to initiate %s export: %v\n%s", objectType, err, errbuf.String())
```
sanyamsinghal, line 53: Have you verified that we don't need to wait/sleep after line 52 and before line 53

sanyamsinghal (commented):
The potential/possible issue with this change in logging ora2pg stderr and stdout i see is - if lets say ora2pg hang because of some error then code will not be executed after `cmd.Wait();` at line 56 onwards and in that case we will miss stdout and stderr(line 57-58) in our logs. As far i can recall there were few instances of that behaviour that was the reason for going with a goroutine running periodically for stderr and stdout. cc @rahulb-yb

rahulb-yb (commented):
+1 to Sanyam's statement regarding the goroutine requirement, there was a delay which was needed before checking whether the stdout and stderr buffers were empty, using a static, one-time check would give unexpected behaviour- despite of there being errors, the stderr channel would appear to be empty! Please test and let us know

PR #147, opened 2022-06-07
Call-home functionality 

rahulb-yb:
Some kinks are yet to be worked out:
- JSON formatting [DONE]
- Command piping [DONE]
- Figuring out how to populate certain fields (TargetClusterLocation, TargetDBCores, SourceCloudDBType)


yb_migrate/cmd/analyzeSchema.go
```diff
@@ -784,6 +784,16 @@ func analyzeSchema() {
 		utils.ErrExit("failed to write report to %q: %s", reportPath, err)
 	}
 	fmt.Printf("-- find schema analysis report at: %s\n", reportPath)
+
+	//add diagnostic flag check here...
+	utils.InitJSON(exportDir)
```
sanyamsinghal, line 789: instead of using utils package, create a new package `callhome` or `diagnostics` where all the related function are placed.
You can refer to src or target packages or [datafile package](https://github.com/yugabyte/yb-voyager/pull/143) in my current PR

yb_migrate/src/utils/diagnostics.go
```diff
@@ -0,0 +1,143 @@
...
+
+//Fill in primary-key based fields, if needed
+func InitJSON(exportdir string) {
+	var err error
+	jsonFilePath = filepath.Join(exportdir, "metainfo", "diagnostics.json")
+	_, err = os.OpenFile(jsonFilePath, os.O_RDWR|os.O_CREATE, 0644)
+	if err != nil {
+		ErrExit("Error while creating/opening diagnostics.json file: %v", err)
+	}
+	jsonBuf, err = os.ReadFile(jsonFilePath)
```
sanyamsinghal, line 53: Option1:
as you have already done `os.OpenFile()`, use file variable returned and create reader for reading from the file as this `ReadFile` function will again internally open the file.

```
reader := bufio.NewReader(file) 
or 
scanner := bufio.NewScanner(file)
```

or 
Option2: why are doing `OpenFile()` to check errors, if there is any error `ReadFile()` should report that right?

yb_migrate/src/utils/diagnostics.go
```diff
@@ -0,0 +1,143 @@
...
+//call-home json formats
+var (
+	jsonFilePath    string
+	sendReq         sendRequest
+	jsonBuf         []byte
+	SendDiagnostics bool
+)
+
+type sendRequest struct {
+	MigrationUuid         uuid.UUID `json:"UUID"`                    //done
```
sanyamsinghal, line 26: since this struct is private, no use of making the struct members public. 
convert them to private by changing: `MigrationUuid` to `migrationUuid`

yb_migrate/src/utils/diagnostics.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+}
+
+//Getter method for updating payload
+func GetPayload() *sendRequest {
+	return &sendReq
+}
+
+//Save payload to diagnostics.json
```
sanyamsinghal, line 80: nit: add space after `//` in comments

yb_migrate/src/utils/diagnostics.go
```diff
@@ -0,0 +1,143 @@
...
+
+	files, err := filepath.Glob(datadirfiles)
+	if err != nil {
+		ErrExit("Error while matching files in data dir for diagnostics: %v", err)
+	}
+	var fileInfo fs.FileInfo
+	var totalSize int64
+	var maxFileSize int64
+	for _, file := range files {
+		fileInfo, _ = os.Stat(file)
```
sanyamsinghal, line 130: handle the error returned by `os.Stat()` call

yb_migrate/src/utils/diagnostics.go
```diff
@@ -0,0 +1,143 @@
...
+		if err != nil {
+			ErrExit("Error while reading HTTP response from call-home server: %v", err)
+		}
+		log.Infof("HTTP response after sending diagnostic.json: %s\n", string(body))
+	}
+}
+
+func UpdateDataSize(exportdir string) {
+	datadirfiles := filepath.Join(exportdir, "data", "*_data*")
+	// totalSizeCmd := exec.Command("du", "-s", datadir)
```
sanyamsinghal, line 115: remove the commented code.

yb_migrate/src/utils/diagnostics.go
```diff
@@ -0,0 +1,143 @@
...
+	// if err != nil {
+	// 	ErrExit("Error while executing command for diagnostics data: %s\n%v", totalSizeCmd, err)
+	// }
+	// totalSize := strings.Split(string(stdout), "\t")[0]
+
+	files, err := filepath.Glob(datadirfiles)
+	if err != nil {
+		ErrExit("Error while matching files in data dir for diagnostics: %v", err)
+	}
+	var fileInfo fs.FileInfo
```
sanyamsinghal, line 126: why declare beforehand? Just use `:=` wherever possible, which will reduce the extra lines.

yb_migrate/cmd/analyzeSchema.go
sanyamsinghal, line 789: Its okay to name it: `InitJson()`

Can we move the call to `InitJson()` function to somewhere to `root.go` or `main.go`, like we do initialization for logging instead of doing this everytime.

yb_migrate/cmd/common.go
```diff
@@ -135,6 +135,11 @@ func getMappingForTableNameVsTableFileName(dataDirPath string) map[string]string
 }
 
 func UpdateTableRowCount(source *srcdb.Source, exportDir string, tablesProgressMetadata map[string]*utils.TableProgressMetadata) {
+	var maxTableLines, totalTableLines int64
+
+	utils.InitJSON(exportDir)
```
sanyamsinghal, line 140: consider moving this call to a common location

yb_migrate/cmd/analyzeSchema.go
sanyamsinghal: It will also make code more readable currently utils doesn't signify anything. diagnostics or callhome will be more meaningful

yb_migrate/src/utils/diagnostics.go
```diff
@@ -0,0 +1,143 @@
...
+
+//call-home json formats
+var (
+	jsonFilePath    string
+	sendReq         sendRequest
+	jsonBuf         []byte
+	SendDiagnostics bool
+)
+
+type sendRequest struct {
```
sanyamsinghal, line 25: consider naming as `Payload` since this is a datatype not a functionality.

sanyamsinghal, line 25: If you name the struct as payload then you can make some of these function as method of that struct like: `Send` instead of `SendPayload()` , `Get` instead of `GetPayload`. There you can this struct as public.

I am suggestion something like this:

```
type struct Payload {
....
}

func (payload *Payload) Send() {}

func (payload *Payload) Get() {} 

func (payload *Payload) Init() {} 

```

yb_migrate/cmd/root.go
```diff
@@ -75,6 +75,9 @@ func init() {
 
 	rootCmd.PersistentFlags().BoolVarP(&utils.DoNotPrompt, "yes", "y", false,
 		"assume answer as yes for all questions during migration (default false)")
+
+	rootCmd.PersistentFlags().BoolVar(&utils.SendDiagnostics, "send-diagnostics", true,
+		"send diagnostics data")
```
sanyamsinghal, line 80: you can make the help message richer:
```
Enable or disable the "send-diagnostics" feature that sends analytics data to Yugabyte.
```

sanyamsinghal (changes requested):

yb_migrate/cmd/analyzeSchema.go
rahulb-yb: Done

yb_migrate/src/utils/diagnostics.go
rahulb-yb: Cannot do this, members of struct need to start with capital letters in order to be exported as json: https://stackoverflow.com/questions/50319404/has-json-tag-but-not-exported

yb_migrate/src/utils/diagnostics.go
rahulb-yb: done

yb_migrate/src/utils/diagnostics.go
rahulb-yb: done

yb_migrate/src/utils/diagnostics.go
rahulb-yb: done

yb_migrate/src/utils/diagnostics.go
rahulb-yb: done

yb_migrate/src/utils/diagnostics.go
rahulb-yb: makes sense, done

rahulb-yb: Keeping the struct public would destroy the purpose of using an explicit getter method, but I have renamed for readability purposes

yb_migrate/cmd/common.go
rahulb-yb: Moved to root

yb_migrate/cmd/analyzeSchema.go
rahulb-yb: Moved to root

yb_migrate/cmd/root.go
rahulb-yb: Done

yb_migrate/src/utils/diagnostics.go
rahulb-yb: 1. Will it suffice to immediately close the file? I was searching for something similar to `touch`, where the file is created if not exists, but not truncated if it does exist
2. I am not using OpenFile() to check errors, there are different possibilities of errors occurring in each scenario, which I am catching in the code

yb_migrate/cmd/analyzeSchema.go
rahulb-yb: I will have to change this, during a fresh start, the meta info directory does not exist- attempting to init the local json fails

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -784,6 +785,14 @@ func analyzeSchema() {
 		utils.ErrExit("failed to write report to %q: %s", reportPath, err)
 	}
 	fmt.Printf("-- find schema analysis report at: %s\n", reportPath)
+
+	callhome.InitJSON(exportDir)
+	payload := callhome.GetPayload()
+	issues, _ := json.Marshal(reportStruct.Issues)
```
amit-yb, line 791: Do not ignore errors. Log an ERROR message and return from the callhome functionality.

Same comment applies to other similar occurrences.

yb-voyager/cmd/exportSchema.go
```diff
@@ -72,12 +73,19 @@ func exportSchema() {
...
+	sourceDBVersion := source.DB().GetVersion()
 
-	fmt.Printf("%s version: %s\n", source.DBType, source.DB().GetVersion())
+	fmt.Printf("%s version: %s\n", source.DBType, sourceDBVersion)
 
 	CreateMigrationProjectIfNotExists(source.DBType, exportDir)
 	source.DB().ExportSchema(exportDir)
 	utils.PrintAndLog("\nExported schema files created under directory: %s\n", exportDir+"/schema")
+
+	callhome.InitJSON(exportDir)
```
amit-yb, line 84: Every occurrence of `callhome.GetPayload()` is preceded by `callhome.InitJSON()`. Can we move the call to the `initJSON()` inside the `GetPayload()`? If the "json" is already initialised, `GetPayload()` will simply skip reinitialising the json.

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -0,0 +1,137 @@
...
+//call-home json formats
+var (
+	jsonFilePath    string
+	Payload         payload
+	jsonBuf         []byte
+	SendDiagnostics bool
+)
+
+type payload struct {
+	MigrationUuid         uuid.UUID `json:"UUID"`                    //done
```
amit-yb, line 26: Delete the trailing comments `//done`...

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -0,0 +1,137 @@
...
+	return &Payload
+}
+
+// Send http request to flask servers after saving locally
+func PackAndSendPayload(exportdir string) {
+	if !SendDiagnostics {
+	} else {
+		//Pack locally
+		jsonBuf, _ = json.Marshal(Payload)
+		os.WriteFile(jsonFilePath, jsonBuf, 0644)
```
amit-yb, line 87: Do not ignore errors!

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// Getter method for updating payload
+func GetPayload() *payload {
+	return &Payload
+}
+
+// Send http request to flask servers after saving locally
+func PackAndSendPayload(exportdir string) {
+	if !SendDiagnostics {
```
amit-yb, line 83: You can use an early return from the function to reduce the indentation:
```
func PackAndSendPayload(exportDir string) {
    if !SendDiagnostics {
        return
    }
    // Pack locally.
    jsonBuf, err := json.Marshal(Payload)
    ...
```

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -0,0 +1,137 @@
...
+		jsonBuf, _ = json.Marshal(Payload)
+		os.WriteFile(jsonFilePath, jsonBuf, 0644)
+
+		//Send request
+		Payload.LastUpdatedTime = time.Now().Format("2006-01-02 15:04:05")
+		postBody, _ := json.Marshal(Payload)
+		requestBody := bytes.NewBuffer(postBody)
+
+		log.Infof("Payload being sent for diagnostic usage: %s\n", string(postBody))
+		resp, err := http.Post("http://10.150.5.149:5000/", "application/json", requestBody)
```
amit-yb, line 95: Define a const `CALL_HOME_SERVICE_HOST = "..."` and `CALL_HOME_SEVICE_PORT`.

I assume you're going to change the IP to the public static IP or a domain name. The 10.* address can be reached only from Yugabyte network.

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -0,0 +1,137 @@
...
+
+		if err != nil {
+			utils.ErrExit("Error while sending diagnostic data: %v", err)
```
amit-yb, line 98: No need to exit--just log an ERROR message.

callhome functionality must be "best-effort". In other words, voyager must never fail just because it couldn't public diagnostic info. That means, there should be no occurrence to `utils.ErrExit()` in any changes made by this PR.

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -0,0 +1,137 @@
...
+		if maxFileSize < fileInfo.Size() {
+			maxFileSize = fileInfo.Size()
+		}
+		totalSize += fileInfo.Size()
+	}
+
+	Payload.TotalSize = totalSize
+	Payload.LargestTableSize = maxFileSize
+
+	PackAndSendPayload(exportdir)
```
amit-yb, line 135: This should be moved out of this function and must be called from one level up.

yb_migrate/src/utils/diagnostics.go
sanyamsinghal: 1. os.Create()
2. but isn't that handled/reported by ReadFile() itself?

yb_migrate/cmd/analyzeSchema.go
sanyamsinghal: Feel free to restructure the code to cater this behaviour.

yb_migrate/src/utils/diagnostics.go
sanyamsinghal: yeah ignore `type struct Payload` i meant `type struct payload`

yb-voyager/cmd/exportData.go
```diff
@@ -164,6 +165,8 @@ func exportDataOffline() bool {
 	source.DB().ExportDataPostProcessing(exportDir, tablesProgressMetadata)
 	printExportedRowCount(datafile.OpenDescriptor(exportDir).TableRowCount)
 
+	callhome.UpdateDataSize(exportDir)
+
```
sanyamsinghal, line 169: nit: remove extra new line

yb-voyager/cmd/exportSchema.go
```diff
@@ -72,12 +73,19 @@ func exportSchema() {
...
+	payload := callhome.GetPayload()
+	payload.SourceDBType = source.DBType
+	payload.SourceDBVersion = sourceDBVersion
+	callhome.PackAndSendPayload(exportDir)
```
sanyamsinghal, line 88: @rahulb-yb what is the expected time for this function call to complete - from same network and different(public) network?
Can we consider making it a goroutine to avoid delays to user experience?

sanyamsinghal: Or are we waiting for the response to come, can we do something some just send the request and don't wait for the response? 
referred: [StackOverflow link](https://stackoverflow.com/questions/2190854/sending-post-requests-without-waiting-for-response#:~:text=You%20cannot%20just%20send%20data,will%20wait%20for%20that%20response.)

yb-voyager/cmd/analyzeSchema.go
rahulb-yb: I hadn't cached the error as there was no scenario in which an error would occur over here- I will still add this functionality for consistency in code 👍🏽

yb-voyager/cmd/exportSchema.go
rahulb-yb: Good catch, I was searching for a global initialisation elsewhere in the code, but we run into the risk that the meta info folder was never created- will add this change 👍🏽

yb-voyager/src/callhome/diagnostics.go
rahulb-yb: Will do this, just noticed it will make the code readability more consistent 👍🏽

yb-voyager/cmd/exportSchema.go
rahulb-yb: I do not have an estimate over the algorithm/time complexity used, in my current tests it seems to not affect performance significantly- Even if I did use a goroutine, we would have to ask the main function to wait on the goroutine to finish to as to safely update the diagnostics.json file, and actually send the data

I will also make sure once again that I have only packed/sent the diagnostics data after all the crucial steps have been completed corresponding to each phase of usage of yb-voyager

sanyamsinghal: yeah in the first option you will still have to wait for the goroutine to complete. Check if there is a easy way to not wait for the response of POST request made.

rahulb-yb: As far as my knowledge of computer networking goes, HTTP requests are always met with a response...
Any other solution would involve not using the standard libraries available afaik

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -0,0 +1,159 @@
...
+//call-home json formats
+var (
+	jsonFilePath    string
+	Payload         payload
+	jsonBuf         []byte
+	SendDiagnostics bool
+)
+
+const(
+	CALL_HOME_SERVICE_HOST      = "34.83.149.226"
```
amit-yb, line 26: I assume this is a static IP address.
Note that ephemeral external IP addresses of VMs change on reboot.

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -0,0 +1,159 @@
...
+			log.Errorf("Error while generating new UUID for diagnostics.json: %v", err)
+			return
+		}
+	}
+
+}
+
+// Getter method for updating payload
+func GetPayload(exportDir string) *payload {
+	initJSON(exportDir)
```
amit-yb, line 87: `initJSON()` should get called only once irrespective of how many times `GetPayload()` is called.
Call `initJSON()` only if `Payload.MigrationUuid == uuid.Nil`.

yb-voyager/src/callhome/diagnostics.go
rahulb-yb: We have reserved a static IP address for the VM
All details can be found here: https://console.cloud.google.com/compute/instancesDetail/zones/us-west1-b/instances/yb-voyager-callhome?project=yugabyte&pageState=(%22duration%22:(%22groupValue%22:%22PT1H%22,%22customValue%22:null))

yb-voyager/src/callhome/diagnostics.go
rahulb-yb: GetPayload() is always called once and no more for each phase of the migration, can still make this change 👍🏽

amit-yb: It is better to add the suggested check so that the code will continue working even if `GetPayload()` is called multiple times (in future).

yb-voyager/cmd/exportSchema.go
sanyamsinghal: ok, just having a idea about how much time it is taking will be good. Also check if we have or can have some timeout for HTTP requests so that its not waiting too long for this.

sanyamsinghal (approved):
LGTM, considering all the suggestions have been taken care of and tested.

yb-voyager/cmd/exportSchema.go
rahulb-yb: I do not see any reason why we would face an HTTP timeout, unless someone stops the already running services- just did some testing and sending the request (and receiving the response) seems to take not more than ~150ms. This was done with a fully populated payload, and also includes the time taken to "save" the json file locally

yb-voyager/cmd/analyzeSchema.go
amit-yb: Please log the error and return. There can be many funny ways in which code can fail.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -784,6 +785,13 @@ func analyzeSchema() {
 		utils.ErrExit("failed to write report to %q: %s", reportPath, err)
 	}
 	fmt.Printf("-- find schema analysis report at: %s\n", reportPath)
+
+	payload := callhome.GetPayload(exportDir)
+	issues, _ := json.Marshal(reportStruct.Issues)
+	payload.Issues = string(issues)
+	dbobjects, _ := json.Marshal(reportStruct.Summary.DBObjects)
```
amit-yb, line 792: Same applies here as well.

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"github.com/google/uuid"
+	log "github.com/sirupsen/logrus"
+)
+
+//call-home json formats
+var (
+	jsonFilePath    string
+	Payload         payload
+	jsonBuf         []byte
```
amit-yb, line 21: Why do you need the `jsonFilePath` and `jsonBuf` at the package level? Can we make them local to the function?

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -0,0 +1,162 @@
...
+		return
+	}
+	//Pack locally
+	var err error
+	jsonBuf, err = json.Marshal(Payload)
+	if err != nil {
+		log.Errorf("Error while packing diagnostics json: %v", err)
+		return
+	}
+	os.WriteFile(jsonFilePath, jsonBuf, 0644)
```
amit-yb, line 106: Log error returned by `os.WriteFile()` and return.

amit-yb (approved):
LGTM. Given a few minor comments. Fix them and merge the PR.

yb-voyager/src/callhome/diagnostics.go
rahulb-yb: I am reusing the values in different functions

PR #153, opened 2022-06-14
Bug fix: progress_amount stored as extended file attributes is deleted on editing.

sanyamsinghal:
- Removing the use of file extended attributes and adding additional column in split name for bytesRead. So new format for split name will be - 'tableName.splitNum.offsetEnd.batchSize.bytesRead.[CPD]'

Test Plan:
1. Tested the import data command wrt resume functionality in couple of scenarios for postgresql

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -32,7 +32,6 @@ import (
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
-	"github.com/pkg/xattr"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/datafile"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/tgtdb"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -948,26 +940,18 @@ func getInProgressFilePath(task *SplitFileImportTask) string {
...
 	dir := filepath.Dir(path)
-	parts := strings.Split(base, ".")
 
-	if len(parts) > 5 { //case when filename has schema also
-		return fmt.Sprintf("%s/%s.%s.%s.%s.%s.P", dir, parts[0], parts[1], parts[2], parts[3], parts[4])
-	} else {
-		return fmt.Sprintf("%s/%s.%s.%s.%s.P", dir, parts[0], parts[1], parts[2], parts[3])
-	}
+	base = base[0:len(base)-1] + "P" // *.C -> *.P
+	return fmt.Sprintf("%s/%s", dir, base)
```
amit-yb, line 945: Nit: You don't need to split the `path` into `dir` and `base` to implement this logic. Just following should be enough:
```
path := task.SplitFilePath
return path[0:len(path)-1] + "P"
```

Same applies for the `getDoneFilePath()` as well.

yb-voyager/cmd/importData.go
```diff
@@ -1022,26 +1006,23 @@ func getCopyCommand(table string) string {
...
-		utils.ErrExit("get extended attribute of file %q: %v", filePath, err)
+	splitName := filepath.Base(filePath)
+	parts := strings.Split(splitName, ".")
+
+	var p int64
+	var err error
+	if dataFileDescriptor.TableRowCount != nil {
+		p, err = strconv.ParseInt(parts[len(parts)-3], 10, 64)
+	} else {
+		p, err = strconv.ParseInt(parts[len(parts)-2], 10, 64)
```
amit-yb, line 1018: Can you please add a comment describing what this if-else block is doing and why?

amit-yb (approved):
LGTM with couple of minor comments. Fix them and merge the PR.

yb-voyager/cmd/importData.go
sanyamsinghal: done

yb-voyager/cmd/importData.go
sanyamsinghal: done

PR #156, opened 2022-06-16
fixed the regression in import data status command due to recent changes in #153

sanyamsinghal:


yb-voyager/cmd/root.go
```diff
@@ -72,7 +72,7 @@ func init() {
 		"enable verbose mode for the console output")
 
 	rootCmd.PersistentFlags().StringVarP(&exportDir, "export-dir", "e", "",
-		"export directory (default is current working directory")
+		"export directory (default is current working directory)")
```
amit-yb, line 75: `--export-dir` is a required argument. Required arguments never have default values.

sanyamsinghal: done, updated the help message.

PR #157, opened 2022-06-16
Diagnostics env var+added yb-voyager version to callhome

rahulb-yb:
-Can disable send-diagnostics flag using environment var `YB_VOYAGER_SEND_DIAGNOSTICS`
-Added yb-voyager version to call-home


amit-yb (approved):
LGTM with a small but important comment. Please make the suggested change and merge the PR.

PR #162, opened 2022-06-20
Add test case for `import data file`.

amit-yb:


sanyamsinghal (approved):

sanyamsinghal (commented):
I suppose your data file is csv based on the DATA_FILE_NAME. If so, you should use `--format csv` or avoid as default is that only

amit-yb:
> I suppose your data file is csv based on the DATA_FILE_NAME. If so, you should use `--format csv` or avoid as default is that only

The file has `text` format. Let me see if I can change the extension of the file.

amit-yb:
> > I suppose your data file is csv based on the DATA_FILE_NAME. If so, you should use `--format csv` or avoid as default is that only
> 
> The file has `text` format. Let me see if I can change the extension of the file.

Changed name of the file to `OneMRows.text.gz`.

PR #166, opened 2022-06-24
Initialise logging only if export-dir exists.

amit-yb:
Without this change, yb-voyager would panic if --export-dir doesn't exist.


kneeraj (approved):
lgtm

Issue #160, opened 2022-06-17
Improving export data performance in initialising expected row counts


sanyamsinghal:
Currently in the `export data` command at the beginning, we find out expected row counts using `SELECT COUNT(*)` queries which can be very slow in case of big tables.

So the proposal is to use approximate row counts provided by the databases and use that.

For example, sample queries for Oracle/MySQL/PostgreSQL:
1. Oracle: `SELECT NUM_ROWS FROM ALL_TABLES WHERE TABLE_NAME = 'TABLE_NAME_IN_UPPERCASE';`
2. MySQL: `SELECT table_rows from information_schema.tables
where table_schema = 'database_name' and table_name = 'table_name' ;`
3. PostgreSQL: `SELECT reltuples::bigint AS estimate FROM pg_class WHERE oid = 'myschema.mytable'::regclass;`

PR #172, opened 2022-07-04
Improving export data speed by initialzing progress bars with approx count instead of exact count

sanyamsinghal:
Now we will be using the approx count provided by stats/catalog tables of database to initialize the progress bars, later on replacing that with the actual row count.
With this change, there will almost no delay at the start of the data export and will be much faster as compared to the previous.

Test Plan followed:
        1. Stale/Old row count available in catalog tables, differing by 40-100% - **done**
        2. If table is truncated but stats contain old row count - **done** (truncate command blocked if data export is already y started by pg_dump)
        3. Big table with no info in stats table - **done**(it is in the behaviour of mpb i.e. if PB initialised with zero, doesn’t increment until something non-zero is present to compare)
        4. Each db type with table partitions present - **done** for Oracle and PostgreSQL
        5. Multiple Schemas migration in case of postgres - **done**

Fixing the issue #160

### DIFF
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -137,14 +137,9 @@ func getMappingForTableNameVsTableFileName(dataDirPath string) map[string]string
 
 func UpdateTableApproxRowCount(source *srcdb.Source, exportDir string, tablesProgressMetadata map[string]*utils.TableProgressMetadata) {
 	var maxTableLines, totalTableLines int64
-
 	payload := callhome.GetPayload(exportDir)
 
 	utils.PrintAndLog("calculating approx num of rows to export for each table...")
-	if !source.VerboseMode {
-		go utils.Wait()
-	}
-
 	sortedKeys := utils.GetSortedKeys(tablesProgressMetadata)
 	for _, key := range sortedKeys {
 		approxRowCount := source.DB().GetTableApproxRowCount(tablesProgressMetadata[key])
@@ -156,10 +151,6 @@ func UpdateTableApproxRowCount(source *srcdb.Source, exportDir string, tablesPro
 		tablesProgressMetadata[key].CountTotalRows = approxRowCount
 	}
 
-	if !source.VerboseMode {
-		utils.WaitChannel <- 0
-		<-utils.WaitChannel
-	}
 	payload.LargestTableRows = maxTableLines
 	payload.TotalRows = totalTableLines
 	log.Tracef("After updating total approx row count, TablesProgressMetadata: %+v", tablesProgressMetadata)
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -135,53 +135,34 @@ func getMappingForTableNameVsTableFileName(dataDirPath string) map[string]string
 	return tableNameVsFileNameMap
 }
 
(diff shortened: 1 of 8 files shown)


yb-voyager/src/srcdb/oracle.go
```diff
@@ -37,12 +37,33 @@ func (ora *Oracle) GetTableRowCount(tableName string) int64 {
...
+		utils.ErrExit("Failed to query %q for row count of %q: %s", query, tableName, err)
 	}
 	log.Infof("Table %q has %v rows.", tableName, rowCount)
 	return rowCount
 }
 
+func (ora *Oracle) GetTableApproxRowCount(tableProgressMetadata *utils.TableProgressMetadata) int64 {
+	var approxRowCount sql.NullInt64 // handles case: value of the row is null, default for int64 is 0
+	var query string
+	if !tableProgressMetadata.IsPartition {
```
amit-yb, line 49: Looks like you will also need to accept the `IsPartition` and `ParentTableName` too as arguments.

yb-voyager/src/srcdb/ora2pg_export_data.go
```diff
@@ -89,8 +89,8 @@ func ora2pgExportDataOffline(ctx context.Context, source *Source, exportDir stri
 	exportDataStart <- true
 
 	err = exportDataCommand.Wait()
-	log.Infof("ora2pg STDOUT: %s", outbuf.String())
-	log.Errorf("ora2pg STDERR: %s", errbuf.String())
+	log.Infof("ora2pg STDOUT: %q", outbuf.String())
```
amit-yb, line 92: This will make the log messages unreadable e.g. all new lines will be replaced by `\n`. Any particular reason for changing the format specifier here?

yb-voyager/cmd/common.go
```diff
@@ -135,53 +135,34 @@ func getMappingForTableNameVsTableFileName(dataDirPath string) map[string]string
...
+func UpdateTableApproxRowCount(source *srcdb.Source, exportDir string, tablesProgressMetadata map[string]*utils.TableProgressMetadata) {
 	var maxTableLines, totalTableLines int64
 
 	payload := callhome.GetPayload(exportDir)
 
-	fmt.Println("calculating num of rows to export for each table...")
+	utils.PrintAndLog("calculating approx num of rows to export for each table...")
 	if !source.VerboseMode {
 		go utils.Wait()
 	}
```
amit-yb, line 146: If we can now get row count very quickly, we should get rid of the spinner from this code-path.

yb-voyager/src/srcdb/postgres.go
```diff
@@ -31,18 +32,38 @@ func (pg *PostgreSQL) CheckRequiredToolsAreInstalled() {
...
 	if err != nil {
-		utils.ErrExit("Failed to query row count of %q: %s", tableName, err)
+		utils.ErrExit("Failed to query %q for row count of %q: %s", query, tableName, err)
 	}
 	log.Infof("Table %q has %v rows.", tableName, rowCount)
 	return rowCount
 }
 
+func (pg *PostgreSQL) GetTableApproxRowCount(tableProgressMetadata *utils.TableProgressMetadata) int64 {
+	var approxRowCount sql.NullInt64 // handles case: value of the row is null, default for int64 is 0
```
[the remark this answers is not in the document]
sanyamsinghal: yes let me work on it as a separate PR

PR #173, opened 2022-07-05
Custom settings for oracle and YB session variables

rahulb-yb:
Changes in this patch:

- Added an override for base template ora2pg.conf for advanced users
- Added an override for YB session variables for advanced users
- Callhome now upserts the version number of yb-voyager (this change was needed to be made from client-side)


yb-voyager/src/srcdb/ora2pg_extract_schema.go
```diff
@@ -13,8 +15,15 @@ import (
 
 func ora2pgExtractSchema(source *Source, exportDir string) {
 	schemaDirPath := exportDir + "/schema"
+	customConfigFilePath := exportDir + "/metainfo/base_ora2pg.conf"
 	configFilePath := exportDir + "/temp/.ora2pg.conf"
-	source.PopulateOra2pgConfigFile(configFilePath)
+	if _, err := os.Stat(customConfigFilePath); errors.Is(err, os.ErrNotExist) {
```
sanyamsinghal, line 20: use `utils.FileOrFolderExists()` function we have in our code for this check

yb-voyager/src/srcdb/ora2pg_extract_schema.go
```diff
@@ -13,8 +15,15 @@ import (
 
 func ora2pgExtractSchema(source *Source, exportDir string) {
 	schemaDirPath := exportDir + "/schema"
+	customConfigFilePath := exportDir + "/metainfo/base_ora2pg.conf"
```
sanyamsinghal, line 18: prefer using `filepath.Join()` as a best practice and no need to worry about `/`

yb-voyager/cmd/importData.go
```diff
@@ -903,6 +905,35 @@ func doOneImport(task *SplitFileImportTask, targetChan chan *tgtdb.Target) {
 	}
 }
 
+func addCustomSessionVars(conn *pgx.Conn) bool {
+	sessionVarsPath := exportDir + "/metainfo/yb_session_vars.sql"
+	if _, err := os.Stat(sessionVarsPath); errors.Is(err, os.ErrNotExist) {
```
sanyamsinghal, line 910: similarly here also use `utils.FileOrFolderExists()`

yb-voyager/cmd/importData.go
```diff
@@ -903,6 +905,35 @@ func doOneImport(task *SplitFileImportTask, targetChan chan *tgtdb.Target) {
 	}
 }
 
+func addCustomSessionVars(conn *pgx.Conn) bool {
+	sessionVarsPath := exportDir + "/metainfo/yb_session_vars.sql"
```
sanyamsinghal, line 909: Is there some specific reason for this naming format?
If not i will prefer using: `ybSessionVars.sql` as we are already following camel case format for other vars: `exportSchemaDone`, `exportDataDone`, `dataFileDescriptor.json` etc..

yb-voyager/cmd/importData.go
```diff
@@ -903,6 +905,35 @@ func doOneImport(task *SplitFileImportTask, targetChan chan *tgtdb.Target) {
...
+	if _, err := os.Stat(sessionVarsPath); errors.Is(err, os.ErrNotExist) {
+		return false
+	}
+	utils.PrintAndLog("Using custom session variables from %s.", sessionVarsPath)
```
sanyamsinghal, line 913: nit: `"Using custom session variables for data import from %s."`

yb-voyager/cmd/importData.go
```diff
@@ -903,6 +905,35 @@ func doOneImport(task *SplitFileImportTask, targetChan chan *tgtdb.Target) {
...
+	varsFile, err := os.Open(sessionVarsPath)
+	if err != nil {
+		utils.ErrExit("Error while opening yb_session_vars.sql: %v", err)
+	}
+	defer varsFile.Close()
+	fileScanner := bufio.NewScanner(varsFile)
+	fileScanner.Split(bufio.ScanLines)
```
sanyamsinghal, line 920: what does this line means?

yb-voyager/cmd/importData.go
```diff
@@ -903,6 +905,35 @@ func doOneImport(task *SplitFileImportTask, targetChan chan *tgtdb.Target) {
...
+
+	var curLine string
+	setVarRegex := regexp.MustCompile(`(?i)SET `)
```
sanyamsinghal, line 923: is this regex pat ensuring that `SET` statement is at the beginning of the string, I think its not?

Also trim spaces from the `curLine` (both front and back) so that you can safely check for `SET` at the beginning of line

yb-voyager/cmd/importData.go
```diff
@@ -903,6 +905,35 @@ func doOneImport(task *SplitFileImportTask, targetChan chan *tgtdb.Target) {
...
+	for fileScanner.Scan() {
+		curLine = fileScanner.Text()
+		if !setVarRegex.MatchString(curLine) {
+			utils.ErrExit("Only SET statements allowed in yb_session_vars.sql. Found: %s.", curLine)
+		}
+		_, err := conn.Exec(context.Background(), curLine)
+		if err != nil {
+			utils.ErrExit("Error while modifying custom session variables: %v")
```
sanyamsinghal, line 931: Missed `err` argument. Also make log msg a more detailed:
```
utils.ErrExit("Error while running custom session variables statement(%q): %v", curLine, err)
```

yb-voyager/cmd/importData.go
```diff
@@ -839,15 +840,16 @@ func doOneImport(task *SplitFileImportTask, targetChan chan *tgtdb.Target) {
 
 			dbVersion := targetServer.DB().GetVersion()
 
-			for i, statement := range IMPORT_SESSION_SETTERS {
-				if checkSessionVariableSupported(i, dbVersion) {
-					_, err := conn.Exec(context.Background(), statement)
-					if err != nil {
-						utils.ErrExit("import file %q: run query %q on %q: %s", inProgressFilePath, statement, targetServer.Host, err)
+			if !addCustomSessionVars(conn) {
```
sanyamsinghal, line 843: I will prefer to have both `if` and `else` condition here so make the code more readable instead of doing it secretly in one function.

1. IF condition: check `utils.FileOrFolderExists(customFilePath)` if true then call `useCustomSessionVars` fuction
2. Else condition: current behaviour.

sanyamsinghal (changes requested):
Left review comments, take-care/respond to them before merging.
Also have you tested the Custom ora2pg config file scenario properly, i think it is still using the embedded sample conf file

yb-voyager/cmd/importData.go
rahulb-yb: Split the file using the delimiter as \n

yb-voyager/cmd/importData.go
rahulb-yb: Will make this change

yb-voyager/cmd/importData.go
sanyamsinghal: Afaik this is not required, `fileScanner.Text()` function does split on newline char by itself.

yb-voyager/cmd/importData.go
```diff
@@ -828,15 +828,19 @@ func doOneImport(task *SplitFileImportTask, targetChan chan *tgtdb.Target) {
...
+			sessionVarsPath := filepath.Join(exportDir, "metainfo", "ybSessionVars.sql")
+			if utils.FileOrFolderExists(sessionVarsPath) {
+				addCustomSessionVars(conn, sessionVarsPath)
+			} else {
+				for i, statement := range IMPORT_SESSION_SETTERS {
+					if checkSessionVariableSupported(i, dbVersion) {
+						_, err := conn.Exec(context.Background(), statement)
+						if err != nil {
+							utils.ErrExit("import file %q: run query %q on %q: %s", inProgressFilePath, statement, targetServer.Host, err)
+						}
```
amit-yb, line 840: We should always run this loop first; and then execute the ybSessionVars.sql iff it exists.

yb-voyager/cmd/importData.go
```diff
@@ -892,6 +896,30 @@ func doOneImport(task *SplitFileImportTask, targetChan chan *tgtdb.Target) {
...
+	}
+	defer varsFile.Close()
+	fileScanner := bufio.NewScanner(varsFile)
+
+	var curLine string
+	setVarRegex := regexp.MustCompile(`(?i)^SET `)
+	for fileScanner.Scan() {
+		curLine = strings.TrimSpace(fileScanner.Text())
+		if !setVarRegex.MatchString(curLine) {
+			utils.ErrExit("Only SET statements allowed in yb_session_vars.sql. Found: %s.", curLine)
```
amit-yb, line 913: Fix the file name from yb_session_vars.sql to ybSessionVars.sql.

yb-voyager/cmd/importData.go
```diff
@@ -829,7 +828,10 @@ func doOneImport(task *SplitFileImportTask, targetChan chan *tgtdb.Target) {
 
 			dbVersion := targetServer.DB().GetVersion()
 
-			if !addCustomSessionVars(conn) {
+			sessionVarsPath := filepath.Join(exportDir, "metainfo", "ybSessionVars.sql")
+			if utils.FileOrFolderExists(sessionVarsPath) {
+				addCustomSessionVars(conn, sessionVarsPath)
```
sanyamsinghal, line 833: nit: `addCustomSessionVars` -> `setCustomSessionVars`

yb-voyager/src/srcdb/ora2pg_extract_schema.go
```diff
@@ -14,15 +13,15 @@ import (
...
-	customConfigFilePath := exportDir + "/metainfo/base_ora2pg.conf"
-	configFilePath := exportDir + "/temp/.ora2pg.conf"
-	if _, err := os.Stat(customConfigFilePath); errors.Is(err, os.ErrNotExist) {
-		source.PopulateOra2pgConfigFile(configFilePath)
-	} else {
-		utils.PrintAndLog("Using custom base_ora2pg.conf for migration.")
+	schemaDirPath := filepath.Join(exportDir, "schema")
+	customConfigFilePath := filepath.Join(exportDir, "metainfo", "base_ora2pg.conf")
+	configFilePath := filepath.Join(exportDir, "temp", ".ora2pg.conf")
+	if utils.FileOrFolderExists(customConfigFilePath) {
```
sanyamsinghal, line 19: we can reduce  some extra code by avoiding else statement
```
if utils.FileOrFolderExists(customConfigFilePath) {
       utils.PrintAndLog("Using custom base_ora2pg.conf for schema migration.")
       configFilePath = customConfigFilePath
}
source.PopulateOra2pgConfigFile(configFilePath)
```

yb-voyager/src/srcdb/ora2pg_export_data.go
```diff
@@ -63,6 +63,14 @@ func ora2pgExportDataOffline(ctx context.Context, source *Source, exportDir stri
...
 
 	configFilePath := filepath.Join(exportDir, "temp", ".ora2pg.conf")
+	customConfigFilePath := filepath.Join(exportDir, "metainfo", "base_ora2pg.conf")
+	if utils.FileOrFolderExists(customConfigFilePath) {
+		utils.PrintAndLog("Using custom base_ora2pg.conf for data migration.")
+		source.PopulateOra2pgConfigFile(customConfigFilePath)
+		configFilePath = customConfigFilePath
+	} else {
+		source.PopulateOra2pgConfigFile(configFilePath)
+	}
```
amit-yb, line 73: This is too complicated. All we need to do is create, if it doesn't exists, the base_ora2pg.conf (from the value of the `Ora2pgConfigFile` variable). The `PopulateOra2pgConfigFile()` always takes the full path of the `temp/.ora2pg.conf`; loads the contents from the `base_ora2pg.conf`; and generates the `temp/.ora2pg.conf` after adding all the migration parameters to the contents loaded from the base_ora2pg.conf.

yb-voyager/src/srcdb/source.go
```diff
@@ -151,12 +151,21 @@ func generateSSLQueryStringIfNotExists(s *Source) string {
...
 //go:embed data/sample-ora2pg.conf
-var SampleOra2pgConfigFile string
+var Ora2pgConfigFile string
 
 func (source *Source) PopulateOra2pgConfigFile(configFilePath string) {
 	sourceDSN := source.getSourceDSN()
-
-	lines := strings.Split(string(SampleOra2pgConfigFile), "\n")
+	var SampleOra2pgConfigFile string
+	if strings.HasSuffix(configFilePath, "base_ora2pg.conf") {
```
sanyamsinghal, line 159: You can use `filepath.Base()` to extract the file name from file path.
Prefer either `-` or `_` for both conf files

yb-voyager/src/srcdb/ora2pg_export_data.go
```diff
@@ -63,6 +63,14 @@ func ora2pgExportDataOffline(ctx context.Context, source *Source, exportDir stri
...
 	configFilePath := filepath.Join(exportDir, "temp", ".ora2pg.conf")
+	customConfigFilePath := filepath.Join(exportDir, "metainfo", "base-ora2pg.conf")
+	if utils.FileOrFolderExists(customConfigFilePath) {
+		utils.PrintAndLog("Using custom base-ora2pg.conf for data migration.")
+		source.PopulateOra2pgConfigFile(customConfigFilePath)
+		configFilePath = customConfigFilePath
+	} else {
+		source.PopulateOra2pgConfigFile(configFilePath)
+	}
 	source.PopulateOra2pgConfigFile(configFilePath)
```
amit-yb, line 74: `PopulateOra2pgConfigFile()` is getting called twice?

rahulb-yb: This was not intentional, I don't think my changes reflected due to some internet issues I was having at the time of pushing the relevant patch

installer_scripts/install-yb-voyager
```diff
@@ -299,6 +304,1280 @@ update_bashrc() {
 	done
 }
 
+insert_yb_session_vars_file() {
+	sudo touch /etc/ybSessionVariables.sql
+	sudo sh -c 'printf "SET client_encoding TO '\''UTF8'\'';\nSET session_replication_role TO replica;" > /etc/ybSessionVariables.sql'
+}
+
+insert_ora2pg_conf() {
+conf_file_name="${HOME}/.yb-voyager-base-ora2pg.conf"
```
sanyamsinghal, line 313: To copy the embeded ora2pg.conf file you can try these command from shell script:
```
repoAbsPath=$(git rev-parse --show-toplevel)
cp $repoAbsPath/yb-voyager/src/srcdb/data/sample-ora2pg.conf /etc/path/to/destination/
```

rahulb-yb: Will try this, thanks for letting me know

rahulb-yb: Please review the rest of the PR, I believe we discussed later on as to why this approach may not be feasible either

installer_scripts/install-yb-voyager
```diff
@@ -299,6 +304,20 @@ update_bashrc() {
 	done
 }
 
+insert_yb_session_vars_file() {
+	sudo touch /etc/ybSessionVariables.sql
+	sudo sh -c 'printf "SET client_encoding TO '\''UTF8'\'';\nSET session_replication_role TO replica;" > /etc/ybSessionVariables.sql'
+}
```
amit-yb, line 310: Create a `yb-voyager/files/ybSessionVariables.sql` file and commit it in the git repo. Then you should be able to copy it just like ora2pg.conf file.

The file should be placed at path `/etc/yb-voyager/ybSessionVariables.sql`.

If the file already exists, then this `create_yb_session_vars_file()` function should be a no-op. Else it should download the file from the repo.

Also add the following two lines in the file:
```
SET yb_disable_transactional_writes to false;
SET yb_enable_upsert_mode to true;
```

installer_scripts/install-yb-voyager
```diff
@@ -299,6 +304,20 @@ update_bashrc() {
...
+
+insert_ora2pg_conf() {
+conf_file_name="/etc/yb-voyager-base-ora2pg.conf"
+sudo touch $conf_file_name
+if ! grep -qxF "####################  Ora2Pg Configuration file   #####################" $conf_file_name; then 
+	sudo wget -q https://github.com/yugabyte/yb-voyager/raw/main/yb-voyager/src/srcdb/data/sample-ora2pg.conf
+	sudo mv sample-ora2pg.conf $conf_file_name
+fi
+}
```
amit-yb, line 319: Properly indent the code.
The file path should be `/etc/yb-voyager/base-ora2pg.conf`.
Right at the beginning of this function, you should check whether the already exists. If yes, then return immediately without doing anything.
Then you won't need the `grep -qxF ...` check.

You should also consider using `-O filename` option of wget so that you won't need the additional step to `mv` the file to the final location.

yb-voyager/src/tgtdb/conn_pool.go
```diff
@@ -121,6 +126,29 @@ func (pool *ConnectionPool) getNextUriIndex() int {
...
+				utils.ErrExit("Only SET statements allowed in ybSessionVariables.sql. Found: %s.", curLine)
+			}
+			_, err := conn.Exec(context.Background(), curLine)
+			if err != nil && !strings.Contains(err.Error(), "unrecognized configuration parameter") {
+				return err
+			}
+		}
+		return nil
+	}
 	for k, v := range pool.params.SessionVars {
```
amit-yb, line 152: Once you add the two statements to the ybSessionVariables.sql file, we no longer need this loop.

yb-voyager/src/tgtdb/conn_pool.go
```diff
@@ -121,6 +126,29 @@ func (pool *ConnectionPool) getNextUriIndex() int {
 }
 
 func (pool *ConnectionPool) setSessionVars(conn *pgx.Conn) error {
+	sessionVarsPath := filepath.Join("/", "etc", "ybSessionVariables.sql")
```
amit-yb, line 129: I would suggest the following:
- Replace the `SessionVars` map from the ConnectionPoolParams with a `SessionInitScript []string`.
- Rename `ConnectionPool.setSessionVars()` as `ConnectionPool.initSession()`.
- The `initSession()` simply runs the list of instructions specified in the `pool.params.SessionInitScript`.
- The importData() path where we instantiate the ConnectionPool, should read the `/etc/yb-voyager/sessionInitScript.sql` into a list of lines and pass it in the `ConnectionPoolParams`.

yb-voyager/src/tgtdb/conn_pool.go
rahulb-yb: I am keeping this loop as a failsafe in the corner case that the file does not exist for some reason

yb-voyager/src/tgtdb/conn_pool.go
rahulb-yb: The latest patch addresses this, please do review

installer_scripts/install-yb-voyager
```diff
@@ -299,6 +304,29 @@ update_bashrc() {
...
 
+create_yb_session_vars_file() {
+	vars_file_name="/etc/yb-voyager/ybSessionVariables.sql"
+	if [ -f "$vars_file_name" ]
+	then
+		output "No need to create ybSessionVariables.sql again. Skipping."
+		return
+	fi
+	sudo mkdir -p /etc/yb-voyager
+	sudo touch $vars_file_name
```
amit-yb, line 315: Not needed.

installer_scripts/install-yb-voyager
```diff
@@ -299,6 +304,29 @@ update_bashrc() {
...
+}
+
+create_base_ora2pg_conf_file() {
+	conf_file_name="/etc/yb-voyager/base-ora2pg.conf"
+	if [ -f "$conf_file_name" ]
+	then
+		output "No need to create base-ora2pg.conf again. Skipping."
+		return
+	fi
+	sudo touch $conf_file_name
```
amit-yb, line 326: Not needed.

yb-voyager/cmd/importData.go
```diff
@@ -1010,6 +1007,50 @@ func getProgressAmount(filePath string) int64 {
 	return p
 }
 
+func getYBSessionInitScript() []string {
+	sessionVarsPath := filepath.Join("/", "etc", "yb-voyager", "ybSessionVariables.sql")
```
amit-yb, line 1011: `filepath.Join()` is used when some part of the path is given by runtime variables.
In this particular case all of the path components are known, so you can simply define a string constant.

yb-voyager/cmd/importData.go
```diff
@@ -1010,6 +1007,50 @@ func getProgressAmount(filePath string) int64 {
...
+	defer varsFile.Close()
+	fileScanner := bufio.NewScanner(varsFile)
+
+	var curLine string
+	setVarRegex := regexp.MustCompile(`(?i)^SET `)
+	for fileScanner.Scan() {
+		curLine = strings.TrimSpace(fileScanner.Text())
+		if !setVarRegex.MatchString(curLine) {
+			utils.ErrExit("Only SET statements allowed in ybSessionVariables.sql. Found: %s.", curLine)
+		}
```
amit-yb, line 1040: Let's drop this check. It's an artificial restriction.

amit-yb (approved):
LGTM. Just a few minor nits.

Thanks for patiently accommodating all the review comments. 👍

PR #174, opened 2022-07-05
Disable the use of source and target uri related flags

sanyamsinghal:
This PR removed the use of `--soruce-db-uri` and `--target-db-uri` flags so that to use the separate flags to provide the information.

The reason for moving away from the use of these flags is to avoid ambiguity as for each database type there are many cases where the connection string format followed for corresponding drivers and underlying tools is different.

### DIFF
--- a/yb-voyager/cmd/analyzeSchema.go
+++ b/yb-voyager/cmd/analyzeSchema.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"encoding/xml"
 	"fmt"
+	"path/filepath"
 	"strconv"
 
 	log "github.com/sirupsen/logrus"
@@ -710,16 +711,16 @@ func generateTxtReport(Report utils.Report) string {
 // add info to the 'reportStruct' variable and return
 func analyzeSchemaInternal() utils.Report {
 	reportStruct = utils.Report{}
-	schemaDir := exportDir + "/schema"
+	schemaDir := filepath.Join(exportDir, "schema")
 	sourceObjList = utils.GetSchemaObjectList(source.DBType)
 	initializeSummaryMap()
 	for _, objType := range sourceObjList {
 		var filePath string
 		if objType == "INDEX" {
-			filePath = schemaDir + "/tables/INDEXES_table.sql"
+			filePath = filepath.Join(schemaDir, "tables", "INDEXES_table.sql")
 		} else {
-			filePath = schemaDir + "/" + strings.ToLower(objType) + "s"
-			filePath += "/" + strings.ToLower(objType) + ".sql"
+			filePath = filepath.Join(schemaDir, strings.ToLower(objType)+"s")
+			filePath = filepath.Join(filePath, strings.ToLower(objType)+".sql")
 		}
 
 		if !utils.FileOrFolderExists(filePath) {
@@ -737,7 +738,7 @@ func analyzeSchemaInternal() utils.Report {
 
 func analyzeSchema() {
 	reportFile := "report." + outputFormat
-	reportPath := exportDir + "/reports/" + reportFile
+	reportPath := filepath.Join(exportDir, "reports", reportFile)
 
 	if !schemaIsExported(exportDir) {
 		utils.ErrExit("run export schema before running analyze-schema")
--- a/yb-voyager/cmd/analyzeSchema.go
+++ b/yb-voyager/cmd/analyzeSchema.go
@@ -817,18 +817,13 @@ var analyzeSchemaCmd = &cobra.Command{
 
 		//marking flags as required based on conditions
 		cmd.MarkPersistentFlagRequired("source-db-type")
-		if source.Uri == "" { //if uri is not given
-			cmd.MarkPersistentFlagRequired("source-db-user")
-			cmd.MarkPersistentFlagRequired("source-db-password")
-			if source.DBType != ORACLE {
-				cmd.MarkPersistentFlagRequired("source-db-name")
-			} else if source.DBType == ORACLE {
-				cmd.MarkPersistentFlagRequired("source-db-schema")
-				validateOracleParams()
-			}
-		} else {
-			//check and parse the source
-			source.ParseURI()
+		cmd.MarkPersistentFlagRequired("source-db-user")
+		cmd.MarkPersistentFlagRequired("source-db-password")
+		if source.DBType != ORACLE {
+			cmd.MarkPersistentFlagRequired("source-db-name")
+		} else if source.DBType == ORACLE {
+			cmd.MarkPersistentFlagRequired("source-db-schema")
+			validateOracleParams()
 		}
 
 		if source.TableList != "" {
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -54,10 +54,7 @@ func validateImportFlags(cmd *cobra.Command) {
 	checkExportDirFlag()
 	checkOrSetDefaultTargetSSLMode()
 	validateTargetPortRange()
-	if target.Uri == "" {
-		cmd.MarkFlagRequired("target-db-user")
-		cmd.MarkFlagRequired("target-db-password")
-	}
+
 	if target.TableList != "" {
 		checkTableListFlag(target.TableList)
 	}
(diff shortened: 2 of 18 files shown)


yb-voyager/src/srcdb/postgres.go
```diff
@@ -87,15 +87,15 @@ func (pg *PostgreSQL) GetAllPartitionNames(tableName string) []string {
...
-func (pg *PostgreSQL) getConnectionString() string {
+func (pg *PostgreSQL) GetConnectionUri() string {
 	source := pg.source
-
 	if source.Uri != "" {
 		return source.Uri
 	}
 
-	return fmt.Sprintf("postgresql://%s:%s@%s:%d/%s?%s", source.User, source.Password,
+	pg.source.Uri = fmt.Sprintf("postgresql://%s:%s@%s:%d/%s?%s", source.User, source.Password,
```
rahulb-yb, line 96: Similar to oracle and mysql, reuse `source` definition

rahulb-yb (commented):
Minor changes, but this patch is a good set of changes-I must commend you for especially cutting the code with respect to the `generateSSLStringIfNotExists()` function through the common sourceDB interfaces. After making those smaller changes, LVGTM 😎

yb-voyager/src/srcdb/mysql.go
```diff
@@ -100,9 +104,10 @@ func (ms *MySQL) getConnectionString() string {
 		errMsg := "Incorrect SSL Mode Provided. Please enter a valid sslmode."
 		panic(errMsg)
 	}
-	connStr := fmt.Sprintf("%s:%s@(%s:%d)/%s?%s", source.User, source.Password,
+
+	ms.source.Uri = fmt.Sprintf("%s:%s@(%s:%d)/%s?%s", source.User, source.Password,
```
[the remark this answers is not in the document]
sanyamsinghal: Yeah, since source var is a pointer we can do that, thanks!

yb-voyager/src/srcdb/oracle.go
```diff
@@ -129,7 +132,8 @@ func (ora *Oracle) getConnectionString() string {
 			source.User, source.Password, source.Host, source.Port, source.DBName)
 	}
 
-	return connStr
+	ora.source.Uri = connStr
```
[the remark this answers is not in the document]
sanyamsinghal: done

yb-voyager/cmd/import.go
```diff
@@ -75,16 +72,15 @@ func registerCommonImportFlags(cmd *cobra.Command) {
 
 	cmd.Flags().StringVar(&target.User, "target-db-user", "",
 		"Username with which to connect to the target YugabyteDB server")
+	cmd.MarkFlagRequired("target-db-user")
```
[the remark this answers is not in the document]
sanyamsinghal: > Have you tested whether keeping the flags' checks here will change the functioning of the code? It may be better yet to keep these checks within the `validateImportFlags()` function for uniformity

the motive here is to keep all the statements like `cmd.MarkFlagRequired()` alongside with the flag side wherever possible, as a best practice.

yb-voyager/cmd/import.go
```diff
@@ -165,11 +161,9 @@ func validateTargetPortRange() {
...
-	if target.Uri == "" {
-		if target.SSLMode == "" {
-			target.SSLMode = "prefer"
-		} else if target.SSLMode != "disable" && target.SSLMode != "prefer" && target.SSLMode != "require" && target.SSLMode != "verify-ca" && target.SSLMode != "verify-full" {
-			utils.ErrExit("Invalid sslmode %q. Required one of [disable, allow, prefer, require, verify-ca, verify-full]", target.SSLMode)
-		}
+	if target.SSLMode == "" {
+		target.SSLMode = "prefer"
+	} else if target.SSLMode != "disable" && target.SSLMode != "prefer" && target.SSLMode != "require" && target.SSLMode != "verify-ca" && target.SSLMode != "verify-full" {
+		utils.ErrExit("Invalid sslmode %q. Required one of [disable, allow, prefer, require, verify-ca, verify-full]", target.SSLMode)
```
amit-yb, line 167: Prefer switch case over else-if chain (most of the times)
```
switch target.SSLMode {
case "":
  target.SSLMode = "prefer"
case "disable", "prefer", "require", "verify-ca", "verify-full":
  break
default:
  utils.ErrExit("...")
}
```
In general, prefer "positive" conditions instead of "negations"--the former are easier to decode.

yb-voyager/cmd/importData.go
```diff
@@ -139,7 +139,7 @@ func getYBServers() []*tgtdb.Target {
 		if err != nil {
 			utils.ErrExit("Error resolving host: %v\n", err)
 		}
-		url := getTargetConnectionUri(&target)
+		url := target.GetConnectionUri()
```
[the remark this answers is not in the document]
amit-yb: This is a welcome change. Thanks for this. 👍 

Our eventual goal is to hide the "target connection URI" entirely behind the `TargetDB` object. We should directly ask the `TargetDB` class whatever action we need to perform on the target db; and the `TargetDB` object should internally figure out the details of instantiating and using connections. The general principle behind this idea is called as "Tell, Don't ask!" (https://martinfowler.com/bliki/TellDontAsk.html)

yb-voyager/src/srcdb/srcdb.go
```diff
@@ -9,6 +9,7 @@ import (
 
 type SourceDB interface {
 	Connect() error
+	GetConnectionUri() string
```
amit-yb, line 12: All the callers of the `GetConnectionUri()` are inside the same `srcdb` package. Then why do we need to make it part of a public interface? The "connection management" is an internal implementation detail that should be hidden from the users of `SourceDB`.

yb-voyager/cmd/import.go
sanyamsinghal: yeah, I also prefer the positive condition over negations, this was the previous code.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -710,16 +711,16 @@ func generateTxtReport(Report utils.Report) string {
 // add info to the 'reportStruct' variable and return
 func analyzeSchemaInternal() utils.Report {
 	reportStruct = utils.Report{}
-	schemaDir := exportDir + "/schema"
+	schemaDir := filepath.Join(exportDir, "schema")
```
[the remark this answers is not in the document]
sanyamsinghal: yeah, that's a very nice suggestion, let me create a separate future ticket for this.

yb-voyager/src/srcdb/srcdb.go
sanyamsinghal: Actually with these changes there is a now requirement of Get function for URI for pg_dump.
```
[root@ip-10-9-80-202 yb-voyager]# grep -Rn -A 1 -B 1 "source.sourceDB.GetConnectionUri" *
yb-voyager/src/srcdb/pg_dump_export_data.go-40-	cmd := fmt.Sprintf(`pg_dump "%s" --no-blobs --data-only --compress=0 %s -Fd --file %s --jobs %d`,
yb-voyager/src/srcdb/pg_dump_export_data.go:41:		source.sourceDB.GetConnectionUri(), tableListPatterns, dataDirPath, source.NumConnections)
yb-voyager/src/srcdb/pg_dump_export_data.go-42-
--
yb-voyager/src/srcdb/pg_dump_extract_schema.go-19-	cmd := fmt.Sprintf(`pg_dump "%s" --schema-only --no-owner -f %s`,
yb-voyager/src/srcdb/pg_dump_extract_schema.go:20:		source.sourceDB.GetConnectionUri(), filepath.Join(exportDir, "temp", "schema.sql"))
yb-voyager/src/srcdb/pg_dump_extract_schema.go-21-	log.Infof("Running command: %s", cmd)

```

amit-yb: You can modify `pgDumpExportDataOffline()` to accept a `connectionURI` as an argument. The function has only one caller in the `PostgreSQL.ExportData()`. Make the caller pass the value of `pg.getConnectionString()` to the `pgDumpExportDataOffline()`.

yb-voyager/src/srcdb/pg_dump_extract_schema.go
```diff
@@ -12,12 +12,12 @@ import (
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
 )
 
-func pgdumpExtractSchema(source *Source, exportDir string) {
+func pgdumpExtractSchema(source *Source, connectionUri string, exportDir string) {
```
amit-yb, line 15: `source` is not used anymore.

amit-yb (approved):
LGTM with a minor comment. You can address the comment and merge the PR.
Thanks for the code cleanup. 👍

yb-voyager/src/srcdb/pg_dump_extract_schema.go
sanyamsinghal: done

PR #179, opened 2022-07-08
Specific Schema List support for export from PG

sanyamsinghal:
**Summary:**
1. Refactored code related to flag validation and setting default values for  export
2. --source-db-schema flag is now mandatory for oracle and postgresql
3. **Support for specific schema list option to provide for export from PostgreSQL database**

### DIFF
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -51,13 +51,11 @@ func init() {
 }
 
 func validateImportFlags(cmd *cobra.Command) {
-	checkExportDirFlag()
+	validateExportDirFlag()
 	checkOrSetDefaultTargetSSLMode()
 	validateTargetPortRange()
 
-	if target.TableList != "" {
-		checkTableListFlag(target.TableList)
-	}
+	validateTableListFlag(target.TableList)
 	if target.Schema != YUGABYTEDB_DEFAULT_SCHEMA && ExtractMetaInfo(exportDir).SourceDBType == "postgresql" {
 		utils.ErrExit("ERROR: --target-db-schema flag is only valid for export from 'oracle' and 'mysql' db type")
 	}
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -34,12 +34,13 @@ var importCmd = &cobra.Command{
 	Long:  ``,
 
 	PreRun: func(cmd *cobra.Command, args []string) {
-		validateImportFlags(cmd)
+		validateImportFlags()
+		sourceDBType = ExtractMetaInfo(exportDir).SourceDBType
+		markImportFlagsRequired(cmd)
 	},
 
 	Run: func(cmd *cobra.Command, args []string) {
 		target.ImportMode = true
-		sourceDBType = ExtractMetaInfo(exportDir).SourceDBType
 		importSchema()
 		importData()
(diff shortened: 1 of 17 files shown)


yb-voyager/cmd/import.go
```diff
@@ -34,12 +34,13 @@ var importCmd = &cobra.Command{
 	Long:  ``,
 
 	PreRun: func(cmd *cobra.Command, args []string) {
-		validateImportFlags(cmd)
+		validateImportFlags()
+		sourceDBType = ExtractMetaInfo(exportDir).SourceDBType
```
amit-yb, line 38: This will cause an issue with `import data file`!

yb-voyager/src/srcdb/postgres.go
```diff
@@ -121,7 +124,7 @@ func (pg *PostgreSQL) getConnectionUri() string {
 }
 
 func (pg *PostgreSQL) ExportSchema(exportDir string) {
-	pgdumpExtractSchema(pg.getConnectionUri(), exportDir)
+	pgdumpExtractSchema(pg.source, pg.getConnectionUri(), exportDir)
```
amit-yb, line 127: Why not pass list of schema names as argument instead of sending in the entire `source`?

yb-voyager/src/srcdb/pg_dump_extract_schema.go
```diff
@@ -12,16 +12,16 @@ import (
...
 )
 
-func pgdumpExtractSchema(connectionUri string, exportDir string) {
+func pgdumpExtractSchema(source *Source, connectionUri string, exportDir string) {
 	fmt.Printf("exporting the schema %10s", "")
 	go utils.Wait("done\n", "error\n")
 
-	cmd := fmt.Sprintf(`pg_dump "%s" --schema-only --no-owner -f %s --no-privileges`,
-		connectionUri, filepath.Join(exportDir, "temp", "schema.sql"))
+	cmd := fmt.Sprintf(`pg_dump "%s" --schema-only --schema "%s" --no-owner -f %s --no-privileges`,
```
amit-yb, line 19: Isn't `source.Schema` a pipe separated list?

yb-voyager/cmd/import.go
sanyamsinghal: how? we are not calling some Parent PreRun in `importDataFileCommand`

yb-voyager/src/srcdb/pg_dump_extract_schema.go
sanyamsinghal: yes, pg_dump allows pipe separated list.
Actually `--schema` parameter in `pg_dump` takes in regex pattern for the schema list, if using `|` for joining schema names will serve the purpose for us.

yb-voyager/src/srcdb/postgres.go
sanyamsinghal: hmm, we can do that. Actually i was thinking if something new comes tomorrow we don't have to make function signature complete by passing more and more variables.

yb-voyager/cmd/import.go
amit-yb: Ah! my bad! I thought it is a `PersistentPreRun`.

amit-yb (approved):

PR #180, opened 2022-07-11
[Fixes #108] Add a ConnectionPool.

### LINKED ISSUE #108: Reduce number of connection created during data import using Connection Pool

amit-yb:


yb-voyager/src/tgtdb/conn_pool.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (pool *ConnectionPool) createNewConnection() (conn *pgx.Conn, err error) {
+	idx := pool.getNextUriIndex()
+	n := len(pool.params.ConnUriList)
+	for i := 0; i < n; i++ {
+		uri := pool.params.ConnUriList[(idx+i)%n]
+		conn, err = pgx.Connect(context.Background(), uri)
+		if err != nil {
+			log.Warnf("Failed to connect to %q: %s", uri, err)
+			continue
```
sanyamsinghal, line 80: If we are continuing leaving that server, so i think we should increment this URI function also `getNextUriIndex()`
Because let's say 2 out of 5 target servers are down in that case we will want to distribute equal among the 3 lives ones. So it seems if we don't increment then the live servers(next to dead ones) in your list we will overloaded as compared to others.

Let me know if this makes sense.

sanyamsinghal (commented):
Left some review comments, please respond before merging.
Overall looks great to me, nice approach to implement the connection Pool for our use case.

yb-voyager/src/tgtdb/conn_pool.go
amit-yb: Changed the implementation so that connections get distributed approximately evenly across all alive nodes.

sanyamsinghal (approved):
LGTM, thanks for the quick changes.

PR #183, opened 2022-07-13
Update actual row counts for callhome

### LINKED ISSUE #178: Update Largest Table Size in callhome stats after export data is done
Rightnow the call home stats related to the largest table size is calculated at the start of export on the basis of approx row counts.
Stats should be updated again based on the actual exported row counts at the end of export data command

rahulb-yb:
Correctly updates call-home stats, changed the location of where the relevant statistics (largest and max table rows) are recorded. Fixes #178

### DIFF
--- a/testscript.sh
+++ b/testscript.sh
@@ -1,0 +1,1 @@
+git rev-parse --show-toplevel
--- a/testscript.sh
+++ b/testscript.sh
@@ -1,1 +1,0 @@
-git rev-parse --show-toplevel
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -137,23 +137,14 @@ func getMappingForTableNameVsTableFileName(dataDirPath string) map[string]string
 }
 
 func UpdateTableApproxRowCount(source *srcdb.Source, exportDir string, tablesProgressMetadata map[string]*utils.TableProgressMetadata) {
-	var maxTableLines, totalTableLines int64
-	payload := callhome.GetPayload(exportDir)
 
 	utils.PrintAndLog("calculating approx num of rows to export for each table...")
 	sortedKeys := utils.GetSortedKeys(tablesProgressMetadata)
 	for _, key := range sortedKeys {
 		approxRowCount := source.DB().GetTableApproxRowCount(tablesProgressMetadata[key])
-		if approxRowCount > maxTableLines {
-			maxTableLines = approxRowCount
-		}
-		totalTableLines += approxRowCount
-
 		tablesProgressMetadata[key].CountTotalRows = approxRowCount
 	}
 
-	payload.LargestTableRows = maxTableLines
-	payload.TotalRows = totalTableLines
 	log.Tracef("After updating total approx row count, TablesProgressMetadata: %+v", tablesProgressMetadata)
 }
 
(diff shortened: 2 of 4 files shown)


sanyamsinghal (commented):
@rahulb-yb can we have similar upsert behaviour for the row counts?
1. Initially update the call home stats with whatever approx row counts we got.
2. Later on at the end upsert in same row with exact counts.
I am considering this because export can be a long running process, it might be good to have both information in case complete export fails.

rahulb-yb:
I do not see any real value coming from this, the user will eventually run a successful export, which will provide the relevant statistics-any reason why you would specifically like this? In order to have _both_ the info, I would need to update the call home cluster itself

yb-voyager/cmd/common.go
```diff
@@ -189,11 +181,28 @@ func printExportedRowCount(exportedRowCount map[string]int64) {
 		keys = append(keys, key)
 	}
 	sort.Strings(keys)
+
 	for _, key := range keys {
 		fmt.Printf("|%s|\n", strings.Repeat("-", 65))
 		fmt.Printf("| %30s | %30d |\n", key, exportedRowCount[key])
 	}
 	fmt.Printf("+%s+\n", strings.Repeat("-", 65))
+
```
sanyamsinghal, line 190: nit: remove extra newline

yb-voyager/cmd/common.go
```diff
@@ -180,6 +171,7 @@ func GetTableRowCount(filePath string) map[string]int64 {
 }
 
 func printExportedRowCount(exportedRowCount map[string]int64) {
+
```
sanyamsinghal, line 174: nit: remove extra new line

testscript.sh
```diff
@@ -0,0 +1 @@
+git rev-parse --show-toplevel
```
sanyamsinghal, line 1: why this new file? Looks like accidental commit

yb-voyager/cmd/exportData.go
```diff
@@ -163,7 +163,10 @@ func exportDataOffline() bool {
 	}
 
 	source.DB().ExportDataPostProcessing(exportDir, tablesProgressMetadata)
-	printExportedRowCount(datafile.OpenDescriptor(exportDir).TableRowCount)
+
```
sanyamsinghal, line 166: nit: remove extra new line

yb-voyager/cmd/exportData.go
```diff
@@ -163,7 +163,10 @@ func exportDataOffline() bool {
...
+	tableRowCount := datafile.OpenDescriptor(exportDir).TableRowCount
+	printExportedRowCount(tableRowCount)
+	callhomeActualRowCounts(tableRowCount)
```
sanyamsinghal, line 169: How about making this function a part of callhome package? It will be something like:
```
callhome.TableRowCount(tableRowCount) ...
```

sanyamsinghal (approved):
LGTM! Left some comments, have a look.

testscript.sh
rahulb-yb: Oh no, my bad-was testing if it would be feasible to use the suggestion you had provided, while I was actively using this branch for development

PR #191, opened 2022-07-28
Added support for RHEL

ShivanshGahlot:
Modified the `install-yb-voyager` script as follows:
1) Added a case for 'Red Hat Enterprise Linux' in `main`
2) Added a command to disable the built in PostgresSQL module in `centos_main`
3) Modified the command that installed  `postgresql14-server` to only install `postgresql14` in `centos_main`
4) Added two export statements to resolve locale issues occurring with perl DB driver installation in `install_ora2pg`


installer_scripts/install-yb-voyager
```diff
@@ -258,6 +262,8 @@ install_ora2pg() {
 	ora2pg_license_acknowledgement
 
 	output "Installing perl DB drivers required for ora2pg."
+	export LC_CTYPE=en_US.UTF-8
+	export LC_ALL=en_US.UTF-8
```
amit-yb, line 266: Move these two declarations at the top of the script.

installer_scripts/ora2pg-23.1/INSTALL
```diff
@@ -0,0 +1,3 @@
+
```
amit-yb, line 1: Get rid of the ora2pg-23.1/ directory.

installer_scripts/install-yb-voyager
```diff
@@ -94,7 +97,8 @@ centos_main() {
 	$YUM perl perl-DBI perl-App-cpanminus 1>&2
 	$YUM mysql-devel 1>&2
 	$YUM https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm 1>&2 || true
-	$YUM postgresql14-server 1>&2
+	sudo yum -qy module disable postgresql
```
amit-yb, line 100: Move this line at line 81. We do not want to execute this statement on centos.

Also, will this work on RHEL7?

ShivanshGahlot: This statement will be required on both Alma Linux and RHEL So I have introduced it before running the `centos_main` function in both the cases. 
I'll test the script on RHEL7 and let you know.

ShivanshGahlot: The script ran fine after introducing a new case for 'Red Hat Enterprise Linux Server' and not running the `sudo yum -qy module disable postgresql` command.

installer_scripts/install-yb-voyager
```diff
@@ -75,6 +77,11 @@ main() {
 			centos_main
 			;;
 		"AlmaLinux")
+			sudo yum -qy module disable postgresql
+			centos_main
+			;;
+		"Red Hat Enterprise Linux")
+			sudo yum -qy module disable postgresql
```
amit-yb, line 84: Run this command only if, version is >= 8 . This will ensure that the script also runs on versions 7.*.

amit-yb (commented):
This change is looking very good now. Just one small change and we're ready to merge this.
Do make sure to delete the VMs that you're creating to test this PR. They cost us real money :-)

installer_scripts/install-yb-voyager
```diff
@@ -75,6 +77,19 @@ main() {
...
+			centos_main
+			;;
+		"Red Hat Enterprise Linux")
+			version=$(cat /etc/os-release | grep -w VERSION_ID | awk -F'=' '{print $2}' | tr -d '"' | awk -F'.' '{print $1}')
+			if [ $version -ge 8 ] 
+			then
+				#disable default PostgreSQL repo
+				sudo yum -qy module disable postgresql
+			fi
```
amit-yb, line 89: Define a new function named `disable_default_pg_repo()` and move all this code in that function.
Call `disable_default_pg_repo` for both Alma and RHEL cases.

installer_scripts/install-yb-voyager
ShivanshGahlot: I have introduced the version check in the case of `Red Hat Enterprise Linux`
I am not sure but I don't think version check or disabling the default PG repo will be required in the case of `Red Hat Enterprise Linux Server` as only the rhel7 machine was giving this os name. Let me know if we should include it.

amit-yb: Yes, you're right! We don't need the version check for the "Red Hat Enterprise Linux" case. I missed the change where you added another case for RHEL7. Please add a comment making it clear that "Red Hat Enterprise Linux" means RHEL>=8 and ""Red Hat Enterprise Linux Server" means RHEL7.

installer_scripts/install-yb-voyager
ShivanshGahlot: Do I put this function in the "COMMON" section or the "CENTOS" section

ShivanshGahlot: Guessing from the function prototypes I have put it in the "COMMON" section of the script

amit-yb: You can put it in the centos section.

ShivanshGahlot: Done

installer_scripts/install-yb-voyager
ShivanshGahlot: Okay ill do this

installer_scripts/install-yb-voyager
ShivanshGahlot: I'm sorry I missed your response on the above thread about not requiring the version check :| 
Alma Linux is only version 8 and above. So technically it should not require any version check as well.
And finally do I remove the version check completely from the newly added function and just run the disable command? Let me know.

amit-yb: Yes. Remove the version check. Just run the command. In that case, we don't even need a function. :-)

ShivanshGahlot: Change done and verified on AlmaLinux and RHEL machine as well. Will remove the machines once you review and pass the changes :P

amit-yb (approved):
Well done!!! 👍

PR #193, opened 2022-08-02
Added export data status command

### LINKED ISSUE #119: Use mpb.BarRemoveOnComplete() for export data status progress bars
Progress bar library cannot handle case where count of progress bars exceeds terminal height. So we should remove a progress-bar as soon as it is done.

This may require us implementing export data status command so that user can know the status of data export.

rahulb-yb:
Resolves #119 
-Added export data status command
-Export data now uses mpb.RemoveOnComplete()
-Made some changes to flags(PersistentPreRun->PreRun, similar with flags)in commands, help messages should be more accurate now

Implementation Details:
Due to the lack of data present at this stage of migration, we have opted to only go for table name and its status (MIGRATING, DONE, NOT_STARTED). For postgres, we make use of the toc.txt file, while for oracle and mysql we create the map between in-progress and completed file names at runtime, making use of the "tmp_" prefix. This way, we can find all three statuses for postgres, but only within [MIGRATING, DONE] statuses in the case of oracle and mysql

### DIFF
--- a/yb-voyager/cmd/exportDataStatusCommand.go
+++ b/yb-voyager/cmd/exportDataStatusCommand.go
@@ -15,7 +15,7 @@ import (
 
 var exportDataStatusCmd = &cobra.Command{
 	Use:   "status",
-	Short: "Print status of an ongoing/completed export.",
+	Short: "Print status of an ongoing/completed data export.",
 
 	Run: func(cmd *cobra.Command, args []string) {
 		validateExportDirFlag()
@@ -51,10 +51,11 @@ func runExportDataStatusCmd() error {
 
 	tableMap := make(map[string]string)
 	dataDir := filepath.Join(exportDir, "data")
-	dbTypeFlag := filepath.Join(exportDir, "metainfo", "schema", "source-db-")
-	if utils.FileOrFolderExists(dbTypeFlag + "postgresql") {
+	dbTypeFlag := ExtractMetaInfo(exportDir).SourceDBType
+	source.DBType = dbTypeFlag
+	if dbTypeFlag == "postgresql" {
 		tableMap = GetMappingForTableNameVsTableFileName(dataDir)
-	} else if utils.FileOrFolderExists(dbTypeFlag+"mysql") || utils.FileOrFolderExists(dbTypeFlag+"oracle") {
+	} else if dbTypeFlag == "mysql" || dbTypeFlag == "oracle" {
 		files, err := filepath.Glob(filepath.Join(dataDir, "*_data.sql"))
 		if err != nil {
 			return fmt.Errorf("error while checking data directory for export data status: %v", err)
@@ -64,41 +65,43 @@ func runExportDataStatusCmd() error {
 			fileName = filepath.Base(file)
 			//Sample file name: [tmp_]YB_VOYAGER_TEST_data.sql
 			if strings.HasPrefix(fileName, "tmp_") {
-				tableMap[fileName[4:len(fileName)-9]] = fileName[:len(fileName)-9]
+				tableMap[fileName[4:]] = fileName
 			} else {
-				tableMap[fileName[:len(fileName)-9]] = "tmp_" + fileName[:len(fileName)-9]
+				tableMap[fileName] = "tmp_" + fileName
 			}
 		}
 	} else {
 		return fmt.Errorf("unable to identify source-db-type")
 	}
-
 	if len(tableMap) > 0 {
 		fmt.Printf("%-30s %-12s\n", "TABLE", "STATUS")
 	}
 	var outputRows []*exportTableMigStatusOutputRow
-	var fullTableName string
+	var finalFullTableName string
 	for tableName := range tableMap {
 		//"_" is treated as a wildcard character in regex query for Glob
-		if tableName == "postdata" {
+		if tableName == "tmp_postdata.sql" || tableName == "tmp_data.sql" {
 			continue
 		}
 		if strings.HasPrefix(tableName, "public.") {
-			fullTableName = tableName[7:]
+			finalFullTableName = tableName[7:]
 		} else {
-			fullTableName = tableName
+			finalFullTableName = tableName
 		}
 		var status string
-		if utils.FileOrFolderExists(filepath.Join(dataDir, fullTableName) + "_data.sql") {
+		//postgresql map returns table names, oracle/mysql map contains file names
+		if (source.DBType == POSTGRESQL && utils.FileOrFolderExists(filepath.Join(dataDir, finalFullTableName)+"_data.sql")) || utils.FileOrFolderExists(filepath.Join(dataDir, finalFullTableName)) {
 			status = "DONE"
-		} else if utils.FileOrFolderExists(filepath.Join(dataDir, tableMap[tableName])+"_data.sql") || utils.FileOrFolderExists(filepath.Join(dataDir, tableMap[tableName])) {
+		} else if utils.FileOrFolderExists(filepath.Join(dataDir, tableMap[tableName])) {
 			status = "MIGRATING"
 		} else {
 			status = "NOT_STARTED"
 		}
-
+		if source.DBType == ORACLE || source.DBType == MYSQL {
+			tableName = tableName[:len(tableName)-9]
+		}
 		row := &exportTableMigStatusOutputRow{
-			tableName: tableName,
+			tableName: finalFullTableName,
 			status:    status,
 		}
 		outputRows = append(outputRows, row)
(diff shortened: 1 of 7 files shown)


yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("cannot run `export data status` before schema export is done")
+		}
+		return fmt.Errorf("check if schema export is done: %w", err)
+	}
+
+	tableMap := make(map[string]string)
+	dataDir := filepath.Join(exportDir, "data")
+	dbTypeFlag := filepath.Join(exportDir, "metainfo", "schema", "source-db-")
```
sanyamsinghal, line 54: instead of check for each flag. You can re-use this function ExtractMetaInfo():
```
sourceDBType = ExtractMetaInfo(exportDir).SourceDBType
```

After that you can just use switch case statements based on db types.

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -0,0 +1,123 @@
...
+		files, err := filepath.Glob(filepath.Join(dataDir, "*_data.sql"))
+		if err != nil {
+			return fmt.Errorf("error while checking data directory for export data status: %v", err)
+		}
+		var fileName string
+		for _, file := range files {
+			fileName = filepath.Base(file)
+			//Sample file name: [tmp_]YB_VOYAGER_TEST_data.sql
+			if strings.HasPrefix(fileName, "tmp_") {
+				tableMap[fileName[4:len(fileName)-9]] = fileName[:len(fileName)-9]
```
sanyamsinghal, line 67: why removing last 9 characters `_data.sql`  from filename?
They are the part of filename, right?

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -0,0 +1,123 @@
...
+		if tableName == "postdata" {
+			continue
+		}
+		if strings.HasPrefix(tableName, "public.") {
+			fullTableName = tableName[7:]
+		} else {
+			fullTableName = tableName
+		}
+		var status string
+		if utils.FileOrFolderExists(filepath.Join(dataDir, fullTableName) + "_data.sql") {
```
sanyamsinghal, line 92: here, if you don't exclude `_data.sql` in previous step there would be no need to again add it.

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -0,0 +1,123 @@
...
+			status = "DONE"
+		} else if utils.FileOrFolderExists(filepath.Join(dataDir, tableMap[tableName])+"_data.sql") || utils.FileOrFolderExists(filepath.Join(dataDir, tableMap[tableName])) {
```
sanyamsinghal, line 94: What this second check for - `utils.FileOrFolderExists(filepath.Join(dataDir, tableMap[tableName]))`?
Is it really needed?

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	} else {
+		return fmt.Errorf("unable to identify source-db-type")
+	}
+
+	if len(tableMap) > 0 {
+		fmt.Printf("%-30s %-12s\n", "TABLE", "STATUS")
+	}
+	var outputRows []*exportTableMigStatusOutputRow
+	var fullTableName string
```
sanyamsinghal, line 80: Better to use `finalFullTableName` keyword.
And for the tablemap map, `inProgressTableMap`

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -0,0 +1,123 @@
...
+
+		row := &exportTableMigStatusOutputRow{
+			tableName: tableName,
+			status:    status,
+		}
+		outputRows = append(outputRows, row)
+	}
+
+	// First sort by status and then by table-name.
+	sort.Slice(outputRows, func(i, j int) bool {
```
sanyamsinghal, line 108: I think the same comparator function is used in `importDataStatusCommand`
Can we define this comparator function separately and re-use that at both places.

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -0,0 +1,123 @@
...
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
+)
+
+var exportDataStatusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Print status of an ongoing/completed export.",
```
sanyamsinghal, line 18: nit: "export" -> "data export"

sanyamsinghal (changes requested):
can you try making use of below functions present in `exportData.go` file. There we create a metadata struct which has other extra info also but all these info required for this status command is also there in the `tablesProgressMetadata` struct .

1. initializeExportTableMetadata()
2. initializeExportTablePartitionMetadata()
3. updateFilePaths()

Reason: If possible, re-using those functions will avoid this extra new code and extra work for maintenance/regression-possibilities.

yb-voyager/cmd/exportDataStatusCommand.go
rahulb-yb: That check is for postgres, the first one is for oracle/mysql

yb-voyager/cmd/exportDataStatusCommand.go
rahulb-yb: I will keep it here, and remove it at the time of printing the status of the table

yb-voyager/cmd/exportDataStatusCommand.go
rahulb-yb: Can remove this condition in the next patch, will make use of extracted source db type info

sanyamsinghal: but wouldn't be covered in first check - `utils.FileOrFolderExists(filepath.Join(dataDir, tableMap[tableName])+"_data.sql")`
We name the final file in same manner/syntax for all db types.

rahulb-yb: In the case of postgres, the in-progress file names are ending with `.dat`, and the tableMap provided includes this as well- I can remove the redundancy by renaming values in the map that we make during runtime for export data status command

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -0,0 +1,126 @@
...
+		row := &exportTableMigStatusOutputRow{
+			tableName: finalFullTableName,
+			status:    status,
+		}
+		outputRows = append(outputRows, row)
+	}
+
+	// First sort by status and then by table-name.
+	sort.Slice(outputRows, func(i, j int) bool {
+		ordStates := map[string]int{"MIGRATING": 1, "DONE": 2, "NOT_STARTED": 3}
```
amit-yb, line 112: MIGRATING -> EXPORTING

amit-yb (approved):
LGTM with minor comments.

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -0,0 +1,126 @@
...
+		files, err := filepath.Glob(filepath.Join(dataDir, "*_data.sql"))
+		if err != nil {
+			return fmt.Errorf("error while checking data directory for export data status: %v", err)
+		}
+		var fileName string
+		for _, file := range files {
+			fileName = filepath.Base(file)
+			//Sample file name: [tmp_]YB_VOYAGER_TEST_data.sql
+			if strings.HasPrefix(fileName, "tmp_") {
+				tableMap[fileName[4:]] = fileName
```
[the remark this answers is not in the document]
rahulb-yb: Yes, we simply remove it later on

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+		return fmt.Errorf("check if schema export is done: %w", err)
+	}
+
+	tableMap := make(map[string]string)
+	dataDir := filepath.Join(exportDir, "data")
+	dbTypeFlag := ExtractMetaInfo(exportDir).SourceDBType
+	source.DBType = dbTypeFlag
+	if dbTypeFlag == "postgresql" {
+		tableMap = GetMappingForTableNameVsTableFileName(dataDir)
```
sanyamsinghal, line 57: is `GetMappingForTableNameVsTableFileName()` function really needed to be public?

sanyamsinghal (approved):
LGTM, please take-care/respond to the comment.

yb-voyager/cmd/exportDataStatusCommand.go
rahulb-yb: Ah, yeah I'll make this change back, it doesn't have to be public

PR #196, opened 2022-08-03
Bug fix #195: Now comparing the target.Host IPs with the resolved IPs of the host provided by yb_servers() function

### LINKED ISSUE #195: Bug: False Load Balancer Detection
LB is detected as true by `yb-voyager` in case if the output of `yb_servers()` function contains a DNS name in place of actual IPs
For example:
```
yugabyte=# select * from yb_servers();
    host    | port | num_connections | node_type | cloud  |   region    | zone  | public_ip
------------+------+-----------------+-----------+--------+-------------+-------+------------
 localhost3 | 5433 |               0 | primary   | cloud1 | datacenter1 | rack1 | localhost3
 localhost2 | 5433 |               0 | primary   | cloud1 | datacenter1 | rack1 | localhost2
 localhost  | 5433 |               0 | primary   | cloud1 | datacenter1 | rack1 | localhost
(3 rows)
```

sanyamsinghal:
Fix the bug as mentioned in Issue #195 

Testing/Verification Details:
1. Testing with a local cluster of 3 nodes with localhost, localhost2, localhost3 DNSs defined in `/etc/hosts` file
2. `ns.LookUpIP("172.3.21.10")` function for some given IP Addr returns the same IP Addr i.e. "172.3.21.10"
3. Tested the behaviour with yb-managed cluster where load balancer is actually present.

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"net"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -37,6 +36,7 @@ import (
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/datafile"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/tgtdb"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
+	"golang.org/x/exp/slices"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -160,7 +160,9 @@ func getYBServers() []*tgtdb.Target {
 
 			// check if given host is one of the server in cluster
 			if loadBalancerUsed {
-				tryMatchingWithTargetHost(host, public_ip)
+				if isSeedTargetHost(host, public_ip) {
+					loadBalancerUsed = false
+				}
 			}
 
 			if usePublicIp {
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -164,11 +160,7 @@ func getYBServers() []*tgtdb.Target {
 
 			// check if given host is one of the server in cluster
 			if loadBalancerUsed {
-				for _, ip := range potentialLBIps {
-					if ip.String() == host || ip.String() == public_ip {
-						loadBalancerUsed = false
-					}
-				}
+				tryMatchingWithTargetHost(host, public_ip)
```
amit-yb, line 163: Rename the function as `isSeedHost(host, publicIP)`.
If the function returns true, set `loadBalancerUsed = false`.

yb-voyager/cmd/importData.go
```diff
@@ -226,6 +218,47 @@ func testYbServers(targets []*tgtdb.Target) {
 	log.Infof("all target servers are accessible")
 }
 
+func tryMatchingWithTargetHost(private_ip string, public_ip string) {
```
amit-yb, line 221: You can simplify using https://pkg.go.dev/golang.org/x/exp/slices#Contains

```
func isSeedHost(names ...string) bool {
        seedHostIPList, err := lookupIP(target.Host)
        // Handle err.
        
        var allIPs []string
        for _, name := range names {
                if name != "" {
                        allIPs = append(allIPs, lookupIP(name))
                }
        }
        for _, seedHostIP := range seedHostIPs {
                if slice.Contains(allIPs, seedHostIP) {
                        return true
                }
        }
        return false;
}

func lookupIP(name string) []string {
        var result []string

        ips, err = net.LookupIP(name)
        // Handle err.
        for _, ip := range ips {
                result = append(result, ip.String())
        }
        return result
}
```

sanyamsinghal: @amit-yb that's a nice way to code it. Will do it.
Initially, i thought to `slices.Contains()` directly but was getting an error due to slice elements being net.IP instead of string.

yb-voyager/cmd/importData.go
```diff
@@ -164,11 +160,7 @@ func getYBServers() []*tgtdb.Target {
 
 			// check if given host is one of the server in cluster
 			if loadBalancerUsed {
-				for _, ip := range potentialLBIps {
-					if ip.String() == host || ip.String() == public_ip {
-						loadBalancerUsed = false
-					}
-				}
+				loadBalancerUsed = !isSeedTargetHost(host, public_ip)
```
amit-yb, line 163: Nit: using simple if statement will better capture intent of the statement:
```
if loadBalancerUsed {
  if isSeedTargetHost(host, private_ip) {
    loadBalancerUsed = false
  }
}
```

yb-voyager/cmd/importData.go
```diff
@@ -226,6 +218,24 @@ func testYbServers(targets []*tgtdb.Target) {
 	log.Infof("all target servers are accessible")
 }
 
+func isSeedTargetHost(names ...string) bool {
+	seedHostIPs := utils.LookupIP(target.Host)
```
amit-yb, line 222: Nit: Move this line after line 229.

amit-yb (approved):
LGTM: just a couple of minor nits.

amit-yb (approved):

Issue #209, opened 2022-08-04
Start Clean behaviour in case of import data 


sanyamsinghal:
Currently in the case of `--start-clean` flag given even with `--table-list` flag still it clean all the splits that are not even mentioned in `--table-list`. 
Better functionality will be to just clean and truncate only the concerned tables.

sanyamsinghal:
Fixed in https://github.com/yugabyte/yb-voyager/commit/43f3072c901e10baca49fcb37be585100942c344

PR #210, opened 2022-08-04
Import Data Resume-ability bug fix #208 and start clean behaviour change #209

### LINKED ISSUE #208: Bug: Import Data Resume-ability ability issues in case of retry
There has been cases observed when on retry the `import data` command it is not resuming instead reporting that the import of table has been done.

Fixes have been identified as follows:
1. On retry, ensure to take care of the case when only `*.D` splits are available in `exportDir/metainfo/data` directory and table is still in the interrupted stage, so just need to identify that table as interrupted and continue code to generate new splits and then import them.
2. In some cases, it has been observed that `*.D` are retried so for that to fix just ensure split generation in `splitDataFiles()` function should take into account the already generated `*.D` splits and continue generation of new splits after those.

sanyamsinghal:
Fix/changes are done as mentioned in corresponding tickets #208 and #209.

Test Plan:
Tested by simulating a similar scenario where only *.D splits are present on disruption/exit of yb-voyager import data command and on retry it is generating the remaining splits and continue
Also confirmed that the already Done splits are not regenerated-and-retried(which was happening previously)
For startClean behaviour testing, i just tested  it in combination with --table-list flag, it works fine now as expected.

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -379,8 +379,8 @@ func generateSmallerSplits(taskQueue chan *SplitFileImportTask) {
 		for _, table := range allTables {
 			tableSplitsPatternStr := fmt.Sprintf("%s.%s", table, SPLIT_INFO_PATTERN)
 			filePattern := filepath.Join(exportDir, "metainfo/data", tableSplitsPatternStr)
-			utils.ClearMatchingFiles(filePattern)
 			log.Infof("clearing the generated splits for table %q matching %q pattern", table, filePattern)
+			utils.ClearMatchingFiles(filePattern)
 		}
 		importTables = allTables //since all tables needs to imported now
 	} else {
@@ -513,11 +513,11 @@ func splitDataFiles(importTables []string, taskQueue chan *SplitFileImportTask) 
 
 		doneSplitRegexStr := fmt.Sprintf(".+/%s\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.[D]$", t)
 		doneSplitRegexp := regexp.MustCompile(doneSplitRegexStr)
-		validTableSplitRegexStr := fmt.Sprintf(".+/%s\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.[CPD]$", t)
-		validTableSplitRegex := regexp.MustCompile(validTableSplitRegexStr)
+		splitFileNameRegexStr := fmt.Sprintf(".+/%s\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.[CPD]$", t)
+		splitFileNameRegex := regexp.MustCompile(splitFileNameRegexStr)
 
 		for _, filepath := range matches {
-			submatches := validTableSplitRegex.FindAllStringSubmatch(filepath, -1)
+			submatches := splitFileNameRegex.FindAllStringSubmatch(filepath, -1)
 			for _, match := range submatches {
 				// fmt.Printf("filepath: %s, %v\n", filepath, match)
 				/*
@@ -750,7 +750,10 @@ func getTablesToImport() ([]string, []string, []string, error) {
 		createdMatches, _ := filepath.Glob(createdPattern)
 		lastSplitMatches, _ := filepath.Glob(lastSplitPattern)
 
-		// Not Vaild Now: [Important] This function's return result is based on assumption that the rate of ingestion is slower than splitting
+		if len(lastSplitMatches) > 1 {
+			utils.ErrExit("More than one last split is present, which is not valid scenario. last split names are: %v", lastSplitMatches)
+		}
+
 		if len(lastSplitMatches) > 0 && len(createdMatches) == 0 && len(interruptedMatches) == 0 && len(doneMatches) > 0 {
 			doneTables = append(doneTables, t)
 		} else {
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -375,8 +375,13 @@ func generateSmallerSplits(taskQueue chan *SplitFileImportTask) {
 	if startClean { //start data migraiton from beginning
 		fmt.Printf("Truncating all tables: %v\n", allTables)
 		truncateTables(allTables)
-		log.Infof("cleaning the database and %s/metadata/data directory", exportDir)
-		utils.CleanDir(exportDir + "/metainfo/data")
+
+		for _, table := range allTables {
+			tableSplitsPatternStr := fmt.Sprintf("%s.%s", table, SPLIT_INFO_PATTERN)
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -375,8 +375,13 @@ func generateSmallerSplits(taskQueue chan *SplitFileImportTask) {
...
 		fmt.Printf("Truncating all tables: %v\n", allTables)
 		truncateTables(allTables)
-		log.Infof("cleaning the database and %s/metadata/data directory", exportDir)
-		utils.CleanDir(exportDir + "/metainfo/data")
+
+		for _, table := range allTables {
+			tableSplitsPatternStr := fmt.Sprintf("%s.%s", table, SPLIT_INFO_PATTERN)
+			filePattern := filepath.Join(exportDir, "metainfo/data", tableSplitsPatternStr)
+			utils.ClearMatchingFiles(filePattern)
+			log.Infof("clearing the generated splits for table %q matching %q pattern", table, filePattern)
```
amit-yb, line 383: Move the log line above before line 382.

yb-voyager/cmd/importData.go
```diff
@@ -532,32 +533,9 @@ func splitDataFiles(importTables []string, taskQueue chan *SplitFileImportTask)
...
-				splitNum, _ := strconv.ParseInt(match[1], 10, 64)
-				offsetEnd, _ := strconv.ParseInt(match[2], 10, 64)
-				numLines, _ := strconv.ParseInt(match[3], 10, 64)
-				offsetStart := offsetEnd - numLines
-				if splitNum == LAST_SPLIT_NUM {
-					fileFullySplit = true
-				}
-				if splitNum > largestCreatedSplitSoFar {
-					largestCreatedSplitSoFar = splitNum
+				if !doneSplitRegexp.MatchString(filepath) {
```
amit-yb, line 541: Why not just check if the file-name ends with `.D`?

yb-voyager/cmd/importData.go
```diff
@@ -760,19 +738,23 @@ func getTablesToImport() ([]string, []string, []string, error) {
...
+		lastSplitPattern := fmt.Sprintf("%s/%s.%s.[CPD]", metaInfoDataDir, t, LAST_SPLIT_PATTERN)
 
 		doneMatches, _ := filepath.Glob(donePattern)
 		interruptedMatches, _ := filepath.Glob(interruptedPattern)
 		createdMatches, _ := filepath.Glob(createdPattern)
+		lastSplitMatches, _ := filepath.Glob(lastSplitPattern)
 
-		//[Important] This function's return result is based on assumption that the rate of ingestion is slower than splitting
-		if len(createdMatches) == 0 && len(interruptedMatches) == 0 && len(doneMatches) > 0 {
+		// Not Vaild Now: [Important] This function's return result is based on assumption that the rate of ingestion is slower than splitting
```
amit-yb, line 753: Drop this comment.

yb-voyager/cmd/importData.go
```diff
@@ -760,19 +738,23 @@ func getTablesToImport() ([]string, []string, []string, error) {
 		donePattern := fmt.Sprintf("%s/%s.%s.D", metaInfoDataDir, t, SPLIT_INFO_PATTERN)
 		interruptedPattern := fmt.Sprintf("%s/%s.%s.P", metaInfoDataDir, t, SPLIT_INFO_PATTERN)
 		createdPattern := fmt.Sprintf("%s/%s.%s.C", metaInfoDataDir, t, SPLIT_INFO_PATTERN)
+		lastSplitPattern := fmt.Sprintf("%s/%s.%s.[CPD]", metaInfoDataDir, t, LAST_SPLIT_PATTERN)
 
 		doneMatches, _ := filepath.Glob(donePattern)
 		interruptedMatches, _ := filepath.Glob(interruptedPattern)
 		createdMatches, _ := filepath.Glob(createdPattern)
+		lastSplitMatches, _ := filepath.Glob(lastSplitPattern)
```
amit-yb, line 752: Exit if `len(lastSplitMatches) > 1`. This will let us know if there is any bug in the implementation.

yb-voyager/cmd/importData.go
```diff
@@ -500,19 +499,21 @@ func splitDataFiles(importTables []string, taskQueue chan *SplitFileImportTask)
...
 		fileFullySplit := false
 		pattern := fmt.Sprintf("%s/%s/data/%s.[0-9]*.[0-9]*.[0-9]*.[CPD]", exportDir, metaInfoDir, t)
 		matches, _ := filepath.Glob(pattern)
-		// in progress are interrupted ones
-		interruptedRegexStr := fmt.Sprintf(".+/%s\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.[P]$", t)
-		interruptedRegexp := regexp.MustCompile(interruptedRegexStr)
+
+		doneSplitRegexStr := fmt.Sprintf(".+/%s\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.[D]$", t)
+		doneSplitRegexp := regexp.MustCompile(doneSplitRegexStr)
+		validTableSplitRegexStr := fmt.Sprintf(".+/%s\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.(\\d+)\\.[CPD]$", t)
```
amit-yb, line 516: How about naming it as `regexSplitFileName`?

sanyamsinghal: Can name it as `splitFileNameRegexStr`
currently, all other similar variables names ends with `Regex`, Just following the same pattern here. We can consider renaming to this pattern all variables later.

yb-voyager/cmd/importData.go
sanyamsinghal: done

yb-voyager/cmd/importData.go
sanyamsinghal: done

yb-voyager/cmd/importData.go
sanyamsinghal: Just wanted the check to be strict `*.D`, generated based on the table name.
Your approach should also be fine.

yb-voyager/cmd/importData.go
sanyamsinghal: done

yb-voyager/cmd/importData.go
sanyamsinghal: done

yb-voyager/cmd/importData.go
sanyamsinghal: Is that fine if i name `splitFileNameRegexStr` for now?

amit-yb (approved):

yb-voyager/cmd/importData.go
amit-yb: Yes. You can use the name `splitFileNameRegexStr`.

Issue #138, opened 2022-05-26
Provide an option to exclude specific tables


amit-yb:
If there is an issue with migration because of some particular tables, the user should be able to migrate data excluding the problematic tables.

PR #214, opened 2022-08-08
Added exclude-table-list flag

rahulb-yb:
Added the --exclude-table-list flag (#138) to exclude certain tables during the export data phase
- Works only with export data command
- No-op if `--table-list` is provided
- Removes tables as per the set notation A-B, where A is the set of tables available, and B is the set of tables to be excluded from export

It should be noted that I opted to not use the switches provided by the internal tools for excluding certain tables, but rather removed the tables at the time of populating the table-list. This was mainly done as we populate a table-list to print when export data is initialised, which needs to be updated before printing in the case of excluding tables

### DIFF
--- a/yb-voyager/cmd/export.go
+++ b/yb-voyager/cmd/export.go
@@ -165,7 +165,12 @@ func validateExportFlags() {
 	validatePortRange()
 	validateSSLMode()
 	validateOracleParams()
-	validateTableListFlag(source.TableList)
+
+	if source.TableList != "" && source.ExcludeTableList != "" {
+		utils.ErrExit("Error: Only one of --table-list and --exclude-table-list are allowed")
+	}
+	validateTableListFlag(source.TableList, false)
+	validateTableListFlag(source.ExcludeTableList, true)
 
 	// checking if wrong flag is given used for a db type
 	if source.DBType != ORACLE {
--- a/yb-voyager/cmd/export.go
+++ b/yb-voyager/cmd/export.go
@@ -69,6 +69,9 @@ func init() {
 
 	exportCmd.Flags().BoolVar(&disablePb, "disable-pb", false,
 		"true - to disable progress bar during data export (default false)")
+
+	exportCmd.Flags().StringVar(&source.ExcludeTableList, "exclude-table-list", "",
+		"List of tables to exclude while exporting data (no-op if --table-list is used) (Note: works only for export data command)")
 }
 
 func registerCommonExportFlags(cmd *cobra.Command) {
--- a/yb-voyager/cmd/export.go
+++ b/yb-voyager/cmd/export.go
@@ -169,8 +169,8 @@ func validateExportFlags() {
 	if source.TableList != "" && source.ExcludeTableList != "" {
 		utils.ErrExit("Error: Only one of --table-list and --exclude-table-list are allowed")
 	}
-	validateTableListFlag(source.TableList, false)
-	validateTableListFlag(source.ExcludeTableList, true)
+	validateTableListFlag(source.TableList, "table-list")
+	validateTableListFlag(source.ExcludeTableList, "exclude-table-list")
 
 	// checking if wrong flag is given used for a db type
 	if source.DBType != ORACLE {
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/exp/slices"
 
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/callhome"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/datafile"
@@ -97,20 +98,14 @@ func exportDataOffline() bool {
 	if source.TableList != "" {
 		finalTableList = extractTableListFromString(source.TableList)
 	} else {
-		excludeTable := false
+
 		tableList = source.DB().GetAllTableNames()
 		if len(excludeTableList) > 0 {
 			for _, tableName := range tableList {
-				for _, excludeTableName := range excludeTableList {
-					if excludeTableName == tableName {
-						excludeTable = true
-						break
-					}
+				if slices.Contains(excludeTableList, tableName) {
+					continue
 				}
-				if !excludeTable {
-					finalTableList = append(finalTableList, tableName)
-				}
-				excludeTable = false
+				finalTableList = append(finalTableList, tableName)
 			}
 		} else {
 			finalTableList = tableList
@@ -181,7 +176,7 @@ func exportDataOnline() bool {
 	return false
 }
 
-func validateTableListFlag(tableListString string) {
+func validateTableListFlag(tableListString string, excludeList bool) {
 	if tableListString == "" {
 		return
 	}
@@ -191,7 +186,11 @@ func validateTableListFlag(tableListString string) {
 	tableNameRegex := regexp.MustCompile("[a-zA-Z0-9_.]+")
 	for _, table := range tableList {
 		if !tableNameRegex.MatchString(table) {
-			utils.ErrExit("Error: Invalid table name '%v' provided wtih --table-list flag", table)
+			if !excludeList {
+				utils.ErrExit("Error: Invalid table name '%v' provided wtih --table-list flag", table)
+			} else {
+				utils.ErrExit("Error: Invalid table name '%v' provided wtih --exclude-table-list flag", table)
+			}
 		}
 	}
 }
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -54,6 +54,8 @@ func init() {
 	exportCmd.AddCommand(exportDataCmd)
 	exportDataCmd.Flags().BoolVar(&disablePb, "disable-pb", false,
 		"true - to disable progress bar during data export(default false)")
+	exportDataCmd.Flags().StringVar(&source.ExcludeTableList, "exclude-table-list", "",
+		"List of tables to exclude while exporting data(no-op if --table-list is used)")
 }
 
 func exportData() {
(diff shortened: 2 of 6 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -90,35 +92,33 @@ func exportDataOffline() bool {
 	// defer cancel()
 
 	var tableList []string
+	var finalTableList []string
```
sanyamsinghal, line 95: Instead of introducing a new variable `finalTableList` can we just `tableList` var, and just update/filter the complete `tableList` based on whether the `--table-list` or `--exclude-table-list` is provided.

yb-voyager/cmd/exportData.go
```diff
@@ -90,35 +92,33 @@ func exportDataOffline() bool {
...
-					tableList = append(tableList, table)
-				} else {
-					utils.ErrExit("invalid table name %q in the --table-list flag.", table)
+		finalTableList = extractTableListFromString(source.TableList)
+	} else {
+		excludeTable := false
+		tableList = source.DB().GetAllTableNames()
+		if len(excludeTableList) > 0 {
+			for _, tableName := range tableList {
+				for _, excludeTableName := range excludeTableList {
```
sanyamsinghal, line 104: Use `slices.Contains()` to check this.

yb-voyager/cmd/export.go
```diff
@@ -69,6 +69,9 @@ func init() {
 
 	exportCmd.Flags().BoolVar(&disablePb, "disable-pb", false,
 		"true - to disable progress bar during data export (default false)")
+
+	exportCmd.Flags().StringVar(&source.ExcludeTableList, "exclude-table-list", "",
```
sanyamsinghal, line 73: Instead of `no-op` or silently ignoring. Please add a check in somewhere `PreRun` in case both the table-list flags are provided then we should error out with a message to user, since only of them makes more sense at a time.

yb-voyager/cmd/exportData.go
sanyamsinghal: If you just update the `tableList` var by excluding the ones, will there still be a need to update any code at other places?

sanyamsinghal (changes requested):

yb-voyager/cmd/exportData.go
```diff
@@ -90,35 +93,27 @@ func exportDataOffline() bool {
...
-			for _, table := range userTableList {
-				parts := strings.Split(table, ".")
-				if len(parts) == 1 {
-					tableList = append(tableList, "public."+table)
-				} else if len(parts) == 2 {
-					tableList = append(tableList, table)
-				} else {
-					utils.ErrExit("invalid table name %q in the --table-list flag.", table)
+		tableList = source.DB().GetAllTableNames()
+		if len(excludeTableList) > 0 {
```
amit-yb, line 103: This check is not required. For empty slice or nil, the for loop simply doesn't execute the loop's body.

yb-voyager/cmd/exportData.go
```diff
@@ -181,7 +176,7 @@ func exportDataOnline() bool {
 	return false
 }
 
-func validateTableListFlag(tableListString string) {
+func validateTableListFlag(tableListString string, excludeList bool) {
```
amit-yb, line 179: How about taking `flagName` as argument instead of `excludeList bool`?
That can simplify the error reporting.

yb-voyager/cmd/exportData.go
```diff
@@ -211,6 +210,29 @@ func checkDataDirs() {
 	}
 }
 
+func extractTableListFromString(flagTableList string) []string {
+	var finalTableList []string
+	if flagTableList != "" {
```
amit-yb, line 215: Always arrange if conditions to reduce the required number of indentation levels. For example, in this particular case it is better to check if `flagTableList == ""` and then return empty slice which allows reducing indentation level of the current if block.

```
if flagTableList == "" {
    retur []string{}
}
// Now the remaining code is sure that `flagTableList` is non-empty.
tableList := utils.CsvStringToSlice(...)
if source.DBType != POSTGRESQL {
  return tableList
}
// Remaining code assumes postgresql source.
  ...
}
return finalTableList
```
@see https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_return_early_rather_than_nesting_deeply
@see https://go.dev/doc/effective_go#if

yb-voyager/cmd/importData.go
```diff
@@ -1070,9 +1072,26 @@ func getYBSessionInitScript() []string {
 	return sessionVars
 }
 
+func removeExcludeTables(tableList []string) []string {
```
amit-yb, line 1075: You can take `excludeTableList []string` argument and then you can reuse this function in both export and import code paths.

yb-voyager/cmd/importData.go
```diff
@@ -382,7 +382,9 @@ func generateSmallerSplits(taskQueue chan *SplitFileImportTask) {
 			log.Infof("clearing the generated splits for table %q matching %q pattern", table, filePattern)
 			utils.ClearMatchingFiles(filePattern)
 		}
-		importTables = allTables //since all tables needs to imported now
+
+		//Removing intersection of doneTables and ExcludeTableList from allTables list, just for startClean
+		importTables = removeExcludeTables(allTables) //since all tables needs to imported now
```
sanyamsinghal, line 387: why truncate and clear splits for the tables which are not included?
Instead, just apply this filter beforehand on `allTables` and `importTables` lists

yb-voyager/cmd/importData.go
```diff
@@ -766,7 +768,7 @@ func getTablesToImport() ([]string, []string, []string, error) {
 		}
 	}
 
-	return doneTables, interruptedTables, remainingTables, nil
+	return doneTables, removeExcludeTables(interruptedTables), removeExcludeTables(remainingTables), nil
```
sanyamsinghal, line 771: I think you should this filtering at one place just once only.
Either in this `getTablesToImport` functions or in the `allTables` and `importTables` lists. I prefer the later since all the checks etc since there are kind of final lists after all other flags processing etc.

yb-voyager/cmd/importData.go
sanyamsinghal: Agree to @amit-yb even can make this a general function `filterList` or some better name and put it inside the `utils` function, maybe reusable in future.

yb-voyager/cmd/export.go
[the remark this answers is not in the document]
rahulb-yb: Don't we do an export schema followed by export data when we simply run export? As long as this functionality is still there in voyager, it makes sense to keep this help message handy
I have also ensured to keep two separate help messages within `export.go` and `exportData.go`

yb-voyager/cmd/importData.go
rahulb-yb: I have split the filtering into two separate areas considering the resumability of import data, as well as the possibility of running start clean on an interrupted/fully completed migration. Please review the logic I have used considering all these cases

yb-voyager/cmd/importData.go
rahulb-yb: This specific case applies to when `--start-clean` has been used. The filter has already been applied to `importTables` list by this point. If we wish to truly "start clean", we should empty the all the splits right? That's why I filtered the list after truncating and clearing splits for all tables

yb-voyager/cmd/exportData.go
rahulb-yb: The slices are different right? The for loop uses the tableList which we have to export, while the excludeTableList is the table lists which we have to remove
Still, if we remove the "if" check the end behaviour is the same, so will do it

yb-voyager/cmd/importData.go
sanyamsinghal: That is what the user expectation will be.
Here if i am not wrong, you are applying filter after the startClean process is done, in that case you are going to clear already done tables also.

yb-voyager/cmd/importData.go
sanyamsinghal: Okay so you are filter here mainly, and other place is after the startClean because there you need to consider the tables cleaned. Is that correct?

yb-voyager/cmd/importData.go
sanyamsinghal: What if you just apply the filter at once after line 370 for the `importTables` and `allTables` lists. Once those tables are out/removed there is no need to apply filter again at any place.

sanyamsinghal (changes requested):

yb-voyager/cmd/importData.go
rahulb-yb: Yes, this is correct

yb-voyager/cmd/importData.go
rahulb-yb: I am onboard with this idea, and was actually spending a while trying to understand the "expected" behaviour of start clean- what we can do for now is file a separate ticket introducing the separate mode of start clean as we talked about during today's stand-up (one mode is a full wipe, the other mode is only wipe the tables mentioned through the table list/not the ones mentioned in exclude table list). It will help define behaviour for if/when resumability is introduced in export data as well

yb-voyager/cmd/exportData.go
rahulb-yb: I have consciously tried to keep applying this in my code, but sometimes it just slips! Thanks for the catch again

yb-voyager/cmd/importData.go
sanyamsinghal: Sure, goahead file a ticket for export data. There is a already existing ticket for export data resumability, link/tag that ticket in the new ticket.

sanyamsinghal: > What if you just apply the filter at once after line 370 for the importTables and allTables lists. Once those tables are out/removed there is no need to apply filter again at any place.

I feel, this approach will simplify the code a lot for developers.

rahulb-yb: It may seem simpler, but it will not provide a consistent meaning of what "start clean" means in our code, as of today

This behaviour should then also be reflected on the export side, where `--start-clean` and `--table-list` together should only clean the tables mentioned in the table list, keeping the data for already exported (but not mentioned) tables' data intact

yb-voyager/cmd/export.go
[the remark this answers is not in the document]
amit-yb: Agree! I had forgot the case.

amit-yb (approved):

sanyamsinghal (approved):
LGTM, thanks for taking of each review comment.

PR #221, opened 2022-08-10
Automate labelling of new issues

rahulb-yb:
Added a workflow to automate filtering of tickets to be triaged at a later point in time


amit-yb (approved):
Thanks for this PR!!! 👍

PR #224, opened 2022-08-10
Introduced New Custom Reader for data file splits

sanyamsinghal:
- This will take care of maintaining the lines/rows read from the file split which is to be used in retry if condition

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -1031,23 +1031,22 @@ func getProgressAmount(filePath string) int64 {
 }
 
 func getYBSessionInitScript() []string {
-	var defaultSessionVars []string
-
+	var sessionVars []string
 	if checkSessionVariableSupport(SET_CLIENT_ENCODING_TO_UTF8) {
-		defaultSessionVars = append(defaultSessionVars, SET_CLIENT_ENCODING_TO_UTF8)
+		sessionVars = append(sessionVars, SET_CLIENT_ENCODING_TO_UTF8)
 	}
 	if checkSessionVariableSupport(SET_SESSION_REPLICATE_ROLE_TO_REPLICA) {
-		defaultSessionVars = append(defaultSessionVars, SET_SESSION_REPLICATE_ROLE_TO_REPLICA)
+		sessionVars = append(sessionVars, SET_SESSION_REPLICATE_ROLE_TO_REPLICA)
 	}
 
 	if enableUpsert {
 		// upsert_mode parameters was introduced later than yb_disable_transactional writes in yb releases
 		// hence if upsert_mode is supported then its safe to assume yb_disable_transactional_writes is already there
 		if checkSessionVariableSupport(SET_YB_ENABLE_UPSERT_MODE) {
-			defaultSessionVars = append(defaultSessionVars, SET_YB_ENABLE_UPSERT_MODE)
+			sessionVars = append(sessionVars, SET_YB_ENABLE_UPSERT_MODE)
 			// 	SET_YB_DISABLE_TRANSACTIONAL_WRITES is used only with & if upsert_mode is supported
 			if disableTransactionalWrites && checkSessionVariableSupport(SET_YB_DISABLE_TRANSACTIONAL_WRITES) {
-				defaultSessionVars = append(defaultSessionVars, SET_YB_DISABLE_TRANSACTIONAL_WRITES)
+				sessionVars = append(sessionVars, SET_YB_DISABLE_TRANSACTIONAL_WRITES)
 			}
 		} else {
 			log.Infof("Falling back to transactional inserts of batches during data import")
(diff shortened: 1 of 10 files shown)


yb-voyager/src/utils/recordReader.go
```diff
@@ -0,0 +1,39 @@
...
+type RecordReader struct {
+	file       *os.File
+	NumRecords int64
+	reader     *bufio.Reader
+}
+
+func (rr *RecordReader) Read(p []byte) (n int, err error) {
+	n, err = rr.reader.Read(p)
+
+	for i := 0; i < n; i++ {
```
amit-yb, line 18: If the CSV file is UTF-8 encoded (which is essential), then this code may not be correct. Refer https://golangbyexample.com/iterate-over-a-string-golang/ . Reading each byte separately may lead to incorrectly classifying some bytes as `\n`.

Unfortunately, a for-range loop may not also be enough. Consider a case where only the first byte of a multi-byte character was included in the buffer `p`. In that case, the for loop will fail with "invalid byte in utf-8" error.

Given how low the probability of that happening is, for now, just add a comment stating this problem.

amit-yb (approved):
LGTM with a few minor comments. Please take care of the comments.

yb-voyager/src/utils/recordReader.go
```diff
@@ -0,0 +1,39 @@
...
+		if string(p[i]) == "\n" {
+			rr.NumRecords++
+		}
+	}
+
+	return n, err
+}
+
+func (rr *RecordReader) RestartFromBegin() {
+	rr.file.Seek(0, io.SeekStart)
```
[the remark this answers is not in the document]
sanyamsinghal: done

yb-voyager/src/utils/recordReader.go
```diff
@@ -0,0 +1,39 @@
...
+	rr.reader.Reset(rr.file)
+	rr.NumRecords = 0
+}
+
+func NewRecordReader(file *os.File) *RecordReader {
```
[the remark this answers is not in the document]
sanyamsinghal: done

yb-voyager/src/utils/recordReader.go
sanyamsinghal: Added

yb-voyager/cmd/importData.go
```diff
@@ -595,11 +594,10 @@ func splitFilesForTable(filePath string, t string, taskQueue chan *SplitFileImpo
 			numLinesInThisSplit += 1
 		}
 
-		if linesWrittenToBuffer {
-			line = fmt.Sprintf("\n%s", line)
+		if line != "" {
```
amit-yb, line 597: This condition is not possible because of isDataLine(), is it?

yb-voyager/cmd/importData.go
```diff
@@ -595,11 +594,10 @@ func splitFilesForTable(filePath string, t string, taskQueue chan *SplitFileImpo
...
+			line = fmt.Sprintf("%s\n", line)
```
amit-yb, line 598: This will cause a new memory allocation. Why can't we just write "\n" on its own:
```
length, err := bufferedWriter.WriteString(line)
if err != nil {
  ...
}
_, err = bufferedWriter.WriteString("\n")
if err != nil {
  ...
}
length++ // Account for \n.
```

yb-voyager/cmd/importData.go
```diff
@@ -883,7 +883,12 @@ func doOneImport(task *SplitFileImportTask, connPool *tgtdb.ConnectionPool) {
 
 			log.Infof("%q => %d rows affected", copyCommand, rowsAffected)
 			if copyErr != nil {
-				utils.ErrExit("COPY %q FROM file %q: %s", task.TableName, inProgressFilePath, copyErr)
+				if strings.Contains(copyErr.Error(), "violates unique constraint") {
+					rowsAffected = task.OffsetEnd - task.OffsetStart
```
amit-yb, line 887: If some of the rows from the batch were deleted, then this is going to report "extra" rows imported.

yb-voyager/src/utils/recordReader.go
```diff
@@ -0,0 +1,62 @@
...
+	rr.NumRecords = initializeNumRecords(rr.hasHeader)
+}
+
+func initializeNumRecords(hasHeader bool) int64 {
+	var numRecords int64 = 0
+	if hasHeader {
+		numRecords = -1
+	}
+	return numRecords
+}
```
amit-yb, line 62: This is not worth having a separate function.

yb-voyager/src/utils/recordReader.go
```diff
@@ -0,0 +1,62 @@
...
+	return n, err
+}
+
+func (rr *RecordReader) RestartFromBegin() {
+	_, err := rr.file.Seek(0, io.SeekStart)
+	if err != nil {
+		ErrExit("unable to reset offset to beginning for file %q: %v", rr.file.Name(), err)
+	}
+	rr.reader.Reset(rr.file)
+	rr.NumRecords = initializeNumRecords(rr.hasHeader)
```
amit-yb, line 53:
```
rr.NumRecords = 0
if rr.hasHeader {
  rr.NumRecords--
}
```

yb-voyager/src/utils/recordReader.go
sanyamsinghal: Actually the same code was needed in `ResetToBegin()` function, so created this

yb-voyager/cmd/importData.go
sanyamsinghal: yeah right, it should also be:
```
rowsAffected = reader.NumRecords
```

Good catch

yb-voyager/cmd/importData.go
sanyamsinghal: I think empty `line` is possible in case some `err` because generally when error occurs(ignore `io.EOF` error) the other variable (here `line`) will be empty

yb-voyager/cmd/importData.go
sanyamsinghal: This should be good way. Let me do the changes and test it.

sanyamsinghal: done

yb-voyager/cmd/importData.go
sanyamsinghal: @amit-yb yeah right, Good catch
We can't go with this, because error must interrupted the reader execution:
```
rowsAffect = rr.NumRecords
```

Only approach in that i think will be do something like `wc -l` in this scenario. 
Let me know what you suggest.

yb-voyager/src/utils/recordReader.go
sanyamsinghal: ok let me remove this function.

sanyamsinghal:
Not required anymore. All the requirements are already handled.

Issue #225, opened 2022-08-10
Bug: Import Data hang for importing subset of table using --table-list flag


sanyamsinghal:
If a table is specified with `--table-list` flag but `metainfo/data` directory contains the  `*.C` or `*.P`  splits for some other tables(not mentioned in `--table-list` ) in that case the `checkForDone()` function call will keep waiting for `*.C` and `*.P` to disappear.

Fix: Here check pattern should be based on the table names, not just metainfo/data/*.P pattern

sanyamsinghal:
Fixed and merged into main branch.

PR #226, opened 2022-08-10
BugFix #225: Updated the checkForDone() function code to check based on importTables list

sanyamsinghal:
Fixes the bug #225


amit-yb (approved):
Good catch!!!

PR #234, opened 2022-08-18
Removed logging from status type commands

### LINKED ISSUE #217: Disable Logging for Import & Export Data Status commands
This command can be run multiple times, for example with: `watch` command and the log messages for these are of no use.

That can be achieved easily by setting `log.SetOutput(ioutil.Discard)` in logging initialization for these specific commands.

rahulb-yb:
[Fixes #217]
This PR disables logging for export/import data status, as when used with `watch`, populates the log file until the command is stopped. Note that any errors, if raised will still print to the terminal screen

### DIFF
--- a/yb-voyager/cmd/exportDataStatusCommand.go
+++ b/yb-voyager/cmd/exportDataStatusCommand.go
@@ -8,7 +8,6 @@ import (
 	"sort"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


sanyamsinghal (approved):
Looks good to me. 
Use `disableLogging` instead of `statusCommand` variable name.

PR #237, opened 2022-08-18
Add new issues to GA project

rahulb-yb:
Revamped, more clear version of the functionality provided by PR #227 
Old workflow:
- _Add issue to GA project whenever its corresponding label is assigned_
- _Will only run after the add_triage_label workflow_

New, updated workflow:

- Add a `triage` label to all new issues, and all to GA1.0 project immediately after; new project views easily make items trackable


.github/workflows/add_to_GA.yml
```diff
@@ -0,0 +1,25 @@
+name: Add issues to GA Project
```
amit-yb, line 1: Let's just add the label `triage` and add the issue to the project.

During triage we should decide whether the issue goes to GA1.0 or not.

.github/workflows/add_to_GA.yml
```diff
@@ -0,0 +1,25 @@
...
+on:
+  workflow_run:
+    workflows: ["Label issues"]
+    types:
+      - completed
+  issues:
+    types:
+      - labeled
```
amit-yb, line 9: What does this new `workflow_run` mean?
Don't we run this workflow when a new issue is created or an existing issue reopened?

.github/workflows/add_to_GA.yml
```diff
@@ -0,0 +1,25 @@
...
+    permissions:
+      issues: write
+    steps:
+      - name: Add issue to project
+        uses: actions/add-to-project@RELEASE_VERSION
+        with:
+        #Project 10 is yb-voyager 1.0 GA
+          project-url: https://github.com/orgs/yugabyte/projects/10
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
+          labeled: GA1.0
```
amit-yb, line 23: As suggested, add only `triage` label.

amit-yb (commented):
A few minor comments and questions.

.github/workflows/add_to_GA.yml
rahulb-yb: Okay, looking at the new planned workflow, I will have to make some changes, please ignore this PR- will ping you when I'm ready with the corresponding changes

.github/workflows/add_to_GA.yml
rahulb-yb: This solves the concurrency issue of "which workflow will run first"
I also kept the files separate for this implementation as there was no way I found where I could ignore the "if an issue is labelled" case in the "assign triage label" workflow

rahulb-yb: This implies (as far as my understanding goes) that this specific workflow would run only after the triage label is assigned/whenever a issue is labelled-note that there is also a label filter later in the steps
P.S. this workflow will now be changed to handle the new way we triage tickets, and is outdated

.github/workflows/organize_issues.yml
```diff
@@ -24,3 +24,10 @@ jobs:
         with:
           add-labels: "triage"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Add issue to project
+        uses: actions/add-to-project@RELEASE_VERSION
+        with:
+        #Project 10 is yb-voyager 1.0 GA
```
amit-yb, line 30: Nit: Either you can drop this comment or specify the updated project name `yb-voyager`.
Nit: Space after `#`.

amit-yb (approved):
Thanks a ton for this PR. 👍

.github/workflows/organize_issues.yml
rahulb-yb: Will drop the comment, not much value addition now that I think of it

PR #257, opened 2022-08-24
Multiple import data COPY command error handling improvements.

sanyamsinghal:
- Ignore Unique constraint violation in case of transactional mode
- Add ROWS_PER_TRANSACTION in generated COPY command
- Added extra layer of check before enabling upsert and transaction mode
  to findout if it is supported on that server version.
- Improved retrying logic to retry for based on retryable vs
  non-retryable errors
- Improved logging for automation tests in case of failure


sanyamsinghal:
Closing this as cherry-picked is done already via multiple single commit PRs

PR #289, opened 2022-09-01
Fixed comment DDL export

### LINKED ISSUE #246: [MySQL] Column Comments
Hi, While migrating the table with column comments from MySQL to YBDB, there are column comments written in MySQL that are missing in the Table.SQL file exported from Voyager and we don't see any column comments for those columns after schema migration in YBDB.

Example 

in MYSQL

 `id` int NOT NULL AUTO_INCREMENT COMMENT 'Auto generated Identifier. This will be the primary key of the table.',

In YBDB these comments are missing?

Can this be fixed?

rahulb-yb:
[Fixes #246]
yb-voyager was unable to export comment DDLs from MySQL or Oracle as a source. This patch fixes this behaviour.

Oddly enough it was a comment that stopped us from exporting comments...

### DIFF
--- a/yb-voyager/src/srcdb/data/sample-ora2pg.conf
+++ b/yb-voyager/src/srcdb/data/sample-ora2pg.conf
@@ -169,7 +169,7 @@ TYPE		TABLE
 
 # Set this to 1 if you don't want to export comments associated to tables and
 # column definitions. Default is enabled.
-DISABLE_COMMENT         0   #Important
+DISABLE_COMMENT         0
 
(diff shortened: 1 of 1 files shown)


sanyamsinghal (approved):
LGTM.
Although the reason/fix is still not  make sense. 
Is it possible for you to verify what was happening with the comment:
1. You can modify the ora2pg code for printing the value of this flag on the console to check what was happening.
2. Or can raise a small ticket on the ora2pg repo asking the same question

It will be better to know the ora2pg is working in case of comments like this, can help to avoid again any future bug

PR #290, opened 2022-09-01
Double quote reserved words in PG

### LINKED ISSUE #248: [MySQL] reserved words in Table Name
In MySQL, we have a table CREATE TABLE `user` after conversion through Voyager CREATE TABLE user and its thrown as error 
ERROR:  syntax error at or near "user"
LINE 1: CREATE TABLE  user (

it expects CREATE TABLE  "user" 

This needs to apply for ALTER TABLE, CREATE INDEX and wherever it is referred in its child tables.

This is applicable to other reserve words as well "CLUSTER" 

Kindly fix it

rahulb-yb:
[Fixes #248]
yb-voyager now exports reserved words from MySQL and Oracle surrounded by double quotation marks.


amit-yb (approved):
LGTM. Thanks for the change.

PR #296, opened 2022-09-05
Added check for binutils > 2.25 for CentOS/RHEL/Ubuntu

ShivanshGahlot:


installer_scripts/install-yb-voyager
```diff
@@ -165,6 +167,27 @@ macos_main() {
...
+	
+	if [ "$os" == "centos" ]
+	then
+		version=$(ld -v | grep -o 'version [0-9.]\+' | awk -F' ' '{print $2}')
+	elif [ "$os" == "ubuntu" ]
+	then
+		version=$(ld -v | grep -o '[0-9.]\+')
+	fi
+
+	 if [ "$(echo "$version $required_version" | awk '{print ($1 > $2)}')" == "0" ]
```
amit-yb, line 183:
```
[amit.jambure@dev-server-ajambure film]$ min_required_version=2.25
[amit.jambure@dev-server-ajambure film]$ version=2.26
[amit.jambure@dev-server-ajambure film]$ if [[ "$version" > "$min_required_version" ]]; then echo true; else echo false; fi
true
[amit.jambure@dev-server-ajambure film]$ version=2.19
[amit.jambure@dev-server-ajambure film]$ if [[ "$version" > "$min_required_version" ]]; then echo true; else echo false; fi
false
```

amit-yb (commented):
Just a few minor comments.

installer_scripts/install-yb-voyager
```diff
@@ -165,6 +167,35 @@ macos_main() {
...
+			# Example output of "ld -v" on Ubuntu:
+			# GNU ld (GNU Binutils for Ubuntu) 2.38
+			version=$(ld -v | grep -o '[0-9.]\+')
+			;;
+		*)
+			echo "ERROR: unsupported linux distro"
+			exit 1
+			;;
+	esac
+
```
amit-yb, line 191: With the following approach, we can get rid of this case statement as well as distinction between centos/ubuntu:
```
[amit.jambure@dev-server-ajambure ~]$ ld -v | awk '{print $NF}'
2.27-44.base.el7_9.1
[amit.jambure@dev-server-ajambure ~]$ ld -v | awk '{print $NF}' | awk -F '-' '{print $1}'
2.27
[amit.jambure@dev-server-ajambure ~]$ echo -n "GNU ld (GNU Binutils for Ubuntu) 2.38" | awk '{print $NF}' | awk -F '-' '{print $1}'
2.38
```
`NF` is an awk built in variable which means number of "fields" in the current row. So `print $NF` outputs last word of a line. We then pass the last word again to awk--this time with `-` as a field separator and then take the first "field".

ShivanshGahlot: Got that!

amit-yb (approved):

installer_scripts/install-yb-voyager
amit-yb: Out of curiosity, found even shorter method:
```
[centos@ip-10-9-12-152 ~]$ ld -v | awk '{split($NF, a, "-"); print a[1]}'
2.27
[centos@ip-10-9-12-152 ~]$ echo -n "GNU ld (GNU Binutils for Ubuntu) 2.38" | awk '{split($NF, a, "-"); print a[1]}'
2.38
```
No need to make this change, though.

PR #306, opened 2022-09-07
Changed default batch size to 20k for better working with transactional mode in import data

sanyamsinghal:


amit-yb (approved):
Thanks for quickly raising this PR.

PR #312, opened 2022-09-09
Added whitespaces where necessary in regex checks

### LINKED ISSUE #249: [MySQL] Issue in analyse schema reports needs more details
Hi when we look at the analyse schema output in report.txt file

Here the below issue is misleading that ALTER TABLE IS NOT SUPPORTED, here alter table is supported by CLUSTER is the reserved word. when we manually correct it and then it's working.

Issues:
Error in Object TABLE:
-Object Name:
-Reason: ALTER TABLE CLUSTER not supported yet.
-SQL Statement: ALTER TABLE cluster_abc ADD CONSTRAINT fk_cluster_abc_cluster FOREIGN KEY (cluster_id) REFERENCES cluster(id) MATCH SIMPLE ON DELETE NO ACTION ON UPDATE NO ACTION;
-File Path: /home/centos/mysqltest/exportdir/schema/tables/table.sql
-Github Issue Link: https://github.com/YugaByte/yugabyte-db/issues/1124
----------------------------

Kindly fix it.

rahulb-yb:
[Fixes #249]
Some regexes did not handle the border cases of certain reserved words being a prefix to object names. This particular issue (#249) arose as the regex simply checked for the keyword "cluster". We needed to ensure this keyword was being used as a stand-alone term rather than a prefix (regex form should be `(?i)ALTER TABLE CLUSTER `. Note the additional whitespace after the keyword cluster)

All similar instances of regexes have been corrected


sanyamsinghal (approved):
LGTM, thanks for the quick fix @rahulb-yb

PR #323, opened 2022-09-15
Added a check in --ignore-exist for some primary key cases

### LINKED ISSUE #293: Ignore Exist functionality in case of same PRIMARY KEY has already been created in last run
If i re-run the import schema command with `--ignore-exist` flag it avoids creation/check for other objects but in the case of the primary key we get this error:
```
    ERROR: multiple primary keys for table "actor" are not allowed (SQLSTATE 42P16)
    STATEMENT: ALTER TABLE ONLY public.actor
    ADD CONSTRAINT actor_pkey PRIMARY KEY (actor_id);

```

The `--ignore-exist` check can be smart enough to determine and decide if the primary key tried to create is the one that already exists.

rahulb-yb:
[Fixes #293]
If a primary key DDL is provided as an alter table statement rather than a direct declaration in the schema DDL (eg. `ALTER TABLE ONLY public.actor ADD CONSTRAINT actor_pkey PRIMARY KEY (actor_id);`), --ignore-exist flag will incorrectly flag an error stating `multiple primary keys are not allowed`. This patch adds a check for this statement, and fixes said issue.

### DIFF
--- a/yb-voyager/cmd/importSchemaYugabyteDB.go
+++ b/yb-voyager/cmd/importSchemaYugabyteDB.go
@@ -91,17 +91,21 @@ func YugabyteDBImportSchema(target *tgtdb.Target, exportDir string) {
 			_, err := conn.Exec(context.Background(), sqlInfo.stmt)
 			if err != nil {
 				log.Errorf("Previous SQL statement failed with error: %s", err)
-				if strings.Contains(err.Error(), "already exists") && !target.IgnoreIfExists && !reCreateSchema.MatchString(sqlInfo.formattedStmtStr) {
-					fmt.Printf("\b \n    %s\n", err.Error())
-					fmt.Printf("    STATEMENT: %s\n", sqlInfo.formattedStmtStr)
-					if !target.ContinueOnError {
-						os.Exit(1)
+				if strings.Contains(err.Error(), "already exists") {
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -99,6 +99,14 @@ func YugabyteDBImportSchema(target *tgtdb.Target, exportDir string) {
 							os.Exit(1)
 						}
 					}
+				} else if strings.Contains(err.Error(), "multiple primary keys") {
```
amit-yb, line 102: Shouldn't this issue be flagged by `analyze-schema` and corrected by the user. We can't just ignore this error.

rahulb-yb: I believe this is PR is to help with the case of the user entering `--ignore-exist`. The DDL is sound as it introduces a PK constraint through ALTER TABLE, but this particular case is skipped with regards to the usage of this flag

sanyamsinghal, line 102: Yeah, I agree with @amit-yb this might be a genuine error in some corner case maybe user adds an extra ALTER command to create one more PK. 
A better and strict check would have been to check for string + make system catalog query to find if already a same-name/same-column index exists on this table.

amit-yb (approved):
We can merge this change.

Issue #319, opened 2022-09-14
Import data file needs to verify if mentioned export-dir has data dir set up within it


rahulb-yb:
If the _complete_ project structure is not available within the export dir during import data file, if the user attempts to simulate the usual project structure for their data dir (For eg. `--export-dir=/voyager_export_dir/ --data-dir=/voyager_export_dir/data/`), yb-voyager currently erases this data directory while creating a new project structure within this export-dir; the contents of the new data dir are broken symlink(s) which point to files that no longer exist.
The user needs to be warned of this operation before proceeding/a new option has to be provided to incorporate a data-dir within the export-dir (similar to the usual migration cycle).

Issue #314, opened 2022-09-09
Documentation: Not use the same export-dir for both command - import data and import data file


sanyamsinghal:
We have to document it properly that for every unique/new migration or different command like `import data` and `import data file` user should use separate export directory otherwise the metainfo stored in the directory for the ongoing migration might cause issues with the new migration started.

sanyamsinghal:
Instead of documentation, will be covered as a part of check in the `import data file` command

PR #324, opened 2022-09-16
Added check for valid path for data-dir wrt export-dir flag

sanyamsinghal:
It will fix the issues: #319 #314

### DIFF
--- a/yb-voyager/cmd/importDataFileCommand.go
+++ b/yb-voyager/cmd/importDataFileCommand.go
@@ -203,16 +203,26 @@ func checkFileFormat() {
 
 func checkDataDirFlag() {
 	if dataDir == "" {
-		fmt.Fprintln(os.Stderr, `ERROR: required flag "data-dir" not set`)
-		os.Exit(1)
+		utils.ErrExit(`ERROR: required flag "data-dir" not set`)
 	}
 	if !utils.FileOrFolderExists(dataDir) {
-		fmt.Fprintf(os.Stderr, "Directory: %s doesn't exists!!\n", dataDir)
-		os.Exit(1)
-	} else if dataDir == "." {
+		utils.ErrExit("data-dir: %s doesn't exists!!", dataDir)
+	}
+	dataDirAbs, err := filepath.Abs(dataDir)
+	if err != nil {
+		utils.ErrExit("unable to resolve absolute path for data-dir(%q): %v", dataDir, err)
+	}
+
+	exportDirAbs, err := filepath.Abs(exportDir)
+	if err != nil {
+		utils.ErrExit("unable to resolve absolute path for export-dir(%q): %v", exportDir, err)
+	}
+
+	if strings.HasPrefix(dataDirAbs, exportDirAbs) {
+		utils.ErrExit("ERROR: data-dir must be outside the export-dir")
+	}
+	if dataDir == "." {
 		fmt.Println("Note: Using current working directory as data directory")
-	} else {
-		dataDir = strings.TrimRight(dataDir, "/")
 	}
 }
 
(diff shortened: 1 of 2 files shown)


priyanshi-yb (approved):

yb-voyager/cmd/importDataFileCommand.go
```diff
@@ -203,16 +203,18 @@ func checkFileFormat() {
...
 	}
 	if !utils.FileOrFolderExists(dataDir) {
-		fmt.Fprintf(os.Stderr, "Directory: %s doesn't exists!!\n", dataDir)
-		os.Exit(1)
-	} else if dataDir == "." {
+		utils.ErrExit("Directory: %s doesn't exists!!", dataDir)
+	}
+	dataDir = filepath.Clean(dataDir)
+	if filepath.Dir(dataDir) == exportDir || dataDir == filepath.Clean(exportDir) {
+		utils.ErrExit("ERROR: Either --export-dir or --data-dir flag is invalid, both needs to be at different location/directory")
```
rahulb-yb, line 213: if the data directory is anywhere within the export directory, the error will arise, please look at the details of the issue I had raised (#319) for an example of this. It does not necessarily only mention the case where data-dir and export-dir are the same

rahulb-yb (dismissed):

yb-voyager/cmd/importDataFileCommand.go
priyanshi-yb: I think `filePath.Dir()` returns get the parent directory of `dataDir` and this condition checks it if parent directory of `dataDir` is `exportDir` or not, I think this will handle that case where `dataDir` is within `exportDir` . @sanyamsinghal let me know if this understanding of mine is wrong?

sanyamsinghal: yes @priyanshi-yb 
@rahulb-yb I am handling both the cases if they are the same or dataDir is within exportDir. So i think this should the case mentioned in your ticket. let me know if i am missing something.

amit-yb: Can't we just check if export dir is prefix of data dir?
```
if strings.HasPrefix(filepath.Abs(dataDir), filepath.Abs(exportDir)) {
  // ERROR: data dir must be outside of export dir
}
```

sanyamsinghal: yes @amit-yb we can do that. But I think this code is more readable and meaningful

amit-yb: From the code it looks like you're only checking two cases (ignoring path cleaning):
- export dir == data dir
- data dir is an immediate child of export dir

As rahul is suggesting, the issue can happen whenever data dir is anywhere in the export-dir tree.

sanyamsinghal: I understand the prefix logic but do we really need to create an error for all those cases - what if data dir is in `temp` or `reports` directory inside the `export-dir`, technically that's not a problem.

amit-yb: Okay. Let's go with your approach.

yb-voyager/cmd/importDataFileCommand.go
```diff
@@ -203,16 +203,18 @@ func checkFileFormat() {
...
+		utils.ErrExit(`ERROR: required flag "data-dir" not set`)
 	}
 	if !utils.FileOrFolderExists(dataDir) {
-		fmt.Fprintf(os.Stderr, "Directory: %s doesn't exists!!\n", dataDir)
-		os.Exit(1)
-	} else if dataDir == "." {
+		utils.ErrExit("Directory: %s doesn't exists!!", dataDir)
+	}
+	dataDir = filepath.Clean(dataDir)
+	if filepath.Dir(dataDir) == exportDir || dataDir == filepath.Clean(exportDir) {
```
amit-yb, line 212: What if exportDir is provided using relative path e.g. `./` and dataDir is provided using abs path `$PWD` . Will this code work?

amit-yb (approved):

Issue #311, opened 2022-09-07
[MYSQL] Interrupted 'import-data' restarted from beginning for table with Primary Key


shubham-yb:
After I interrupted the `import-data` command in between for around 200000 rows on MYSQL, instead of resuming from the last point, it restarted from the beginning after throwing this statement.

`truncate table 'MOCK_DATA' with NO Primary Key for import of data to restart from beginning...`

Even though my table had a Primary Key. Branch used : v1.0.0-xbeta

This is the schema.

```
create table MOCK_DATA (
	id int primary key auto_increment,
	first_name VARCHAR(50),
	last_name VARCHAR(50),
	email VARCHAR(50),
	gender VARCHAR(50),
	ip_address VARCHAR(20)
);
```

This is the description on the source.

![Image](https://user-images.githubusercontent.com/65844208/189043629-b0adf766-9cb8-4942-89d6-31537b083b7d.png)

This is the description on the target.

![Image](https://user-images.githubusercontent.com/65844208/189043679-26159fd6-94ec-4f8f-90db-b67fa55d8ea7.png)

sanyamsinghal:
Thanks @shubham-yb 

The issue is in the code where we check the primary key presence. 
The table names in MySQL are Case sensitive by default, so it was like:
MOCK_DATA -> MySQL
mock_data -> YugabyteDB
but the query was checking for MOCK_DATA in YugabyteDB, which is even not present in the target.

Fix:
1. In import data, if the table names in unquoted then convert to the small case before comprising it in system catalog tables, where if quoted just use it as it.
2. Error out in case the exact table is not present in target while checking for PK presence.

priyanshi-yb:
resolved by #332

PR #332, opened 2022-09-20
Modified checkPrimaryKey function and resolves the issue #311

priyanshi-yb:
This PR resolves this issue - #311

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -458,15 +458,15 @@ func checkPrimaryKey(tableName string) bool {
 
 	rows, err := conn.Query(context.Background(), checkTableSql)
 	if err != nil {
-		utils.ErrExit("error in querying to check table %q is present: %s", table, err)
-	}
-	
-	if(rows.Next()){
-		log.Infof("table %s is present in DB", table)
+		if strings.Contains(err.Error(), "does not exist"){
+			utils.ErrExit("table %q doesn't exist in target DB", table)
+		} else {
+			utils.ErrExit("error in querying to check table %q is present: %v", table, err)
+		}
 	} else {
-		utils.ErrExit("table %q doesn't exist in target DB", table)
+		log.Infof("table %s is present in DB", table)
 	}
-
+		
     rows.Close()
 	
 	/* currently object names for yugabytedb is implemented as case-sensitive i.e. lower-case
@@ -475,12 +475,11 @@ func checkPrimaryKey(tableName string) bool {
 
 	checkPKSql := fmt.Sprintf(`SELECT * FROM information_schema.table_constraints
 	WHERE constraint_type = 'PRIMARY KEY' AND table_name = '%s' AND table_schema = '%s';`, table, schema)
-	// fmt.Println(checkPKSql)
 
 	log.Infof("Running query on target DB: %s", checkPKSql)
 	rows, err = conn.Query(context.Background(), checkPKSql)
 	if err != nil {
-		utils.ErrExit("error in querying to check PK on table %q: %s", table, err)
+		utils.ErrExit("error in querying to check PK on table %q: %v", table, err)
 	}
 	defer rows.Close()
 
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -436,7 +436,7 @@ func checkPrimaryKey(tableName string) bool {
 	}
 	defer conn.Close(context.Background())
 
-	var table, schema string
+	var table, schema, originalTableName string
 	if len(strings.Split(tableName, ".")) == 2 {
 		schema = strings.Split(tableName, ".")[0]
 		table = strings.Split(tableName, ".")[1]
@@ -444,12 +444,18 @@ func checkPrimaryKey(tableName string) bool {
 		schema = target.Schema
 		table = strings.Split(tableName, ".")[0]
 	}
+	originalTableName = table
 
-	if sourceDBType == ORACLE || !utils.IsQuotedString(table) {
+
+	if !utils.IsQuotedString(table) {
 		table = strings.ToLower(table)
+	} else {
+		var unQuotedTableName string
+		unQuotedTableName = table[1 : len(table)-1]
+		table = unQuotedTableName
 	}
 
-	checkTableSql := fmt.Sprintf(`SELECT '%s.%s'::regclass;`, schema, table)
+	checkTableSql := fmt.Sprintf(`SELECT '%s.%s'::regclass;`, schema, originalTableName)
 	log.Infof("Running query on target DB: %s", checkTableSql)
 
 	rowTable, e := conn.Query(context.Background(), checkTableSql)
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -447,30 +447,28 @@ func checkPrimaryKey(tableName string) bool {
 	originalTableName = table
 
 
-	if !utils.IsQuotedString(table) {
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -436,22 +436,46 @@ func checkPrimaryKey(tableName string) bool {
...
+	originalTableName = table
 
-	if sourceDBType == ORACLE && !utils.IsQuotedString(table) {
+
+	if !utils.IsQuotedString(table) {
 		table = strings.ToLower(table)
+	} else {
+		var unQuotedTableName string
+		unQuotedTableName = table[1 : len(table)-1]
+		table = unQuotedTableName
```
amit-yb, line 455: Simplify:
```
if utils.IsQuotedString(table) {
  table = strings.Trim(table, `"`)
} else {
  table = strings.ToLower(table)
}
```

yb-voyager/cmd/importData.go
```diff
@@ -436,22 +436,46 @@ func checkPrimaryKey(tableName string) bool {
...
 	}
 
+	checkTableSql := fmt.Sprintf(`SELECT '%s.%s'::regclass;`, schema, originalTableName)
+	log.Infof("Running query on target DB: %s", checkTableSql)
+
+	rowTable, e := conn.Query(context.Background(), checkTableSql)
```
amit-yb, line 461: It is customary to name error variable as `err`.
Rename `rowTable` -> `rows`

yb-voyager/cmd/importData.go
```diff
@@ -436,22 +436,46 @@ func checkPrimaryKey(tableName string) bool {
...
+	}
+	
+	if(rowTable.Next()){
+		log.Infof("table %s is present in DB", table)
+	} else {
+		log.Infof("table %s is not present in DB", table)
+		return false
+	}
+
+    rowTable.Close()
```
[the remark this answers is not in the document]
priyanshi-yb: No, actually that is not enough as if in a scenario table doesn't exists, that query on 480 also returns the 0 rows in the result, due to which the error logged will be `table does not has a PK` but that should be like `table doesn't exists` or something. So with only that query we won't be able to check whether table is there or not.

yb-voyager/cmd/importData.go
priyanshi-yb: Thanks!

yb-voyager/cmd/importData.go
priyanshi-yb: Done.

yb-voyager/cmd/importData.go
```diff
@@ -436,28 +436,50 @@ func checkPrimaryKey(tableName string) bool {
...
+	rows, err := conn.Query(context.Background(), checkTableSql)
+	if err != nil {
+		utils.ErrExit("error in querying to check table %q is present: %s", table, err)
+	}
+	
+	if(rows.Next()){
+		log.Infof("table %s is present in DB", table)
+	} else {
+		log.Infof("table %s is not present in DB", table)
+		return false
```
amit-yb, line 468: Looks like you will have to exit yb-voyager. If `import data` is called for a table, then the table MUST be present on the target.
If we return `false` here, we gain nothing with this additional check--anyway the earlier code was also returning `false` for non-existent tables.

priyanshi-yb: Yes, I think in this case when table doesn't exist, we should exit with throw error like this, right? 
```
	if(rows.Next()){
		log.Infof("table %s is present in DB", table)
	} else {
		utils.ErrExit("table %q doesn't exist in target DB", table)
	}
```

priyanshi-yb: @amit-yb , I think this else case is not even required because if the query to check table is already ran fine without error then this else is not requried because empty rows can not be returned because that query selects the table name in the schema and if error came in this query we are already exiting in line 461. I will remove this else block.

amit-yb (approved):

yb-voyager/cmd/importData.go
```diff
@@ -436,28 +436,49 @@ func checkPrimaryKey(tableName string) bool {
...
 
+    rows.Close()
+	
 	/* currently object names for yugabytedb is implemented as case-sensitive i.e. lower-case
 	but in case of oracle exported data files(which we use for to extract tablename)
 	so eg. file EMPLOYEE_data.sql -> table EMPLOYEE which needs to converted for using further */
+
 	checkPKSql := fmt.Sprintf(`SELECT * FROM information_schema.table_constraints
 	WHERE constraint_type = 'PRIMARY KEY' AND table_name = '%s' AND table_schema = '%s';`, table, schema)
 	// fmt.Println(checkPKSql)
```
sanyamsinghal, line 478: nit: remove this commented print stat

yb-voyager/cmd/importData.go
```diff
@@ -436,28 +436,49 @@ func checkPrimaryKey(tableName string) bool {
...
+	  } else {
 		table = strings.ToLower(table)
+	  }
+
+	checkTableSql := fmt.Sprintf(`SELECT '%s.%s'::regclass;`, schema, originalTableName)
+	log.Infof("Running query on target DB: %s", checkTableSql)
+
+	rows, err := conn.Query(context.Background(), checkTableSql)
+	if err != nil {
+		utils.ErrExit("error in querying to check table %q is present: %s", table, err)
```
sanyamsinghal, line 461: err variable is not of type string.
either use `%v` with `err` or `err.String()` with `%s`

yb-voyager/cmd/importData.go
```diff
@@ -436,28 +436,49 @@ func checkPrimaryKey(tableName string) bool {
...
-	var table, schema string
+	var table, schema, originalTableName string
 	if len(strings.Split(tableName, ".")) == 2 {
 		schema = strings.Split(tableName, ".")[0]
 		table = strings.Split(tableName, ".")[1]
 	} else {
 		schema = target.Schema
 		table = strings.Split(tableName, ".")[0]
 	}
+	originalTableName = table
```
sanyamsinghal, line 447: why is this extra variable;e required? can't you just use the `table` variable as its place.

sanyamsinghal (approved):
LGTM, thanks for the fix.
Provided some minor comments.

yb-voyager/cmd/importData.go
```diff
@@ -436,28 +436,49 @@ func checkPrimaryKey(tableName string) bool {
...
+
+	checkTableSql := fmt.Sprintf(`SELECT '%s.%s'::regclass;`, schema, originalTableName)
+	log.Infof("Running query on target DB: %s", checkTableSql)
+
+	rows, err := conn.Query(context.Background(), checkTableSql)
+	if err != nil {
+		utils.ErrExit("error in querying to check table %q is present: %s", table, err)
+	}
+	
+	if(rows.Next()){
```
amit-yb, line 464: No need of parenthesis around expression. Setup your IDE/editor to automatically format the code using `go fmt`.

yb-voyager/cmd/importData.go
amit-yb: In that case, you don't need this entire `if...else` block. Just successfully running the `SELECT 'foo.bar'::regclass;` query is enough.
Make sure the output proper message when the `err.Error()` contains `"does not exist"`.

yb-voyager/cmd/importData.go
priyanshi-yb: This variable is required for the query to check table as if there are quotes in table and we are removing them, so quoted table name is required to check if that table exist or not. It stores the original table name with quotes as well if present.

PR #340, opened 2022-09-22
Import index, FK, and triggers in a separate phase after import data

amit-yb:
This PR introduces `--post-import-data` flag to the `import schema` command. The user is expected, to run `import schema` with the flag, AFTER data import is complete.
With this change, we assume that no FK, index, or triggers are present when `import data` is running. That allows us to get rid of the `CASCADE` clause from the `TRUNCATE TABLE`.

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -996,10 +996,11 @@ func executeSqlStmtWithRetries(conn **pgx.Conn, sqlInfo sqlInfo, objType string)
 			time.Sleep(time.Second * 5)
 			log.Infof("RETRYING DDL: %q", sqlInfo.stmt)
 			continue
-		} else if strings.Contains(err.Error(), "already exists") &&
-			(target.IgnoreIfExists || strings.EqualFold(sqlInfo.stmt, "CREATE SCHEMA public;")) {
-
-			err = nil
+		} else if strings.Contains(err.Error(), "already exists") {
+			fmt.Printf("XXX already exists error: %q\n", sqlInfo.stmt)
+			if target.IgnoreIfExists || strings.EqualFold(strings.Trim(sqlInfo.stmt, " \n"), "CREATE SCHEMA public;") {
+				err = nil
+			}
 		}
 		break // no more iteration in case of non retriable error
 	}
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -446,19 +446,18 @@ func checkPrimaryKey(tableName string) bool {
 	}
 	originalTableName = table
 
-
 	if utils.IsQuotedString(table) {
 		table = strings.Trim(table, `"`)
-	  } else {
+	} else {
 		table = strings.ToLower(table)
-	  }
+	}
 
 	checkTableSql := fmt.Sprintf(`SELECT '%s.%s'::regclass;`, schema, originalTableName)
 	log.Infof("Running query on target DB: %s", checkTableSql)
 
 	rows, err := conn.Query(context.Background(), checkTableSql)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -982,8 +976,11 @@ func executeSqlStmtWithRetries(conn **pgx.Conn, sqlInfo sqlInfo, objType string)
 			time.Sleep(time.Second * 5)
 			log.Infof("RETRYING DDL: %q", sqlInfo.stmt)
 			continue
-		} else if strings.Contains(err.Error(), "already exists") && target.IgnoreIfExists {
-			err = nil
+		} else if strings.Contains(err.Error(), "already exists") {
+			fmt.Printf("XXX already exists error: %q\n", sqlInfo.stmt)
+			if target.IgnoreIfExists || strings.EqualFold(strings.Trim(sqlInfo.stmt, " \n"), "CREATE SCHEMA public;") {
```
sanyamsinghal, line 981: can we comment here why/when `CREATE SCHEMA public` comes and required to be removed.

yb-voyager/cmd/importData.go
```diff
@@ -941,27 +927,35 @@ func executeSqlFile(file string, objType string) {
...
 	}()
 
-	var errOccured = 0
 	sqlInfoArr := createSqlStrInfoArray(file, objType)
 	for _, sqlInfo := range sqlInfoArr {
 		if conn == nil {
 			conn = newTargetConn()
 		}
+
+		setOrSelectStmt := strings.HasPrefix(sqlInfo.stmt, "SET ") || strings.HasPrefix(sqlInfo.stmt, "SELECT ")
```
sanyamsinghal, line 936: can we make the string comparison case insensitive here also as asked at other place.

amit-yb: Done.

yb-voyager/cmd/importData.go
amit-yb: Done.

yb-voyager/cmd/importSchema.go
```diff
@@ -51,22 +51,49 @@ func init() {
...
 
+	if !flagPostImportData {
+		createTargetSchemas(conn)
+	}
+	var importObjectList []string
+	var skipFn func(objType, stmt string) bool
+	isCreateFKStmt := func(objType, stmt string) bool {
+		return objType == "TABLE" && strings.HasPrefix(stmt, "ALTER TABLE") &&
+			strings.Contains(stmt, "ADD CONSTRAINT") &&
+			strings.Contains(stmt, "FOREIGN KEY")
```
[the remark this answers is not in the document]
amit-yb: Done.

yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -85,8 +83,14 @@ func YugabyteDBImportSchema(target *tgtdb.Target, exportDir string) {
...
 		sqlInfoArr := createSqlStrInfoArray(importObjectFilePath, importObjectType)
-		errOccured := 0
 		for _, sqlInfo := range sqlInfoArr {
+			if !(strings.HasPrefix(sqlInfo.stmt, "SET ") || strings.HasPrefix(sqlInfo.stmt, "SELECT ")) {
+				if len(sqlInfo.stmt) < 80 {
+					fmt.Printf("%s\n", sqlInfo.stmt)
+				} else {
+					fmt.Printf("%s ...\n", sqlInfo.stmt[:80])
+				}
+			}
```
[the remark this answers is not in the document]
amit-yb: Because `math.Min()` takes `float64` as argument, using `Min()` would require multiple (total 3) type-casts. That's why went with a simple `if...else` block.

sanyamsinghal (approved):

Issue #336, opened 2022-09-21
[MYSQL] Lower case schema gets dropped on the target while importing in upper case


shubham-yb:
If we have two schemas on the target with the same name but opposite cases for example, xyz and XYZ, if we import into the upper case schema, the import goes through but the lower case one gets dropped from the target.

Steps to reproduce:

1. Create two schemas on the target with same names but opposite case ex: xyz and XYZ.
2. Do a `\dn` to list the schemas.
3. Populate the `TARGET_DB_SCHEMA` variable with the upper case schema.
4. Run import schema.
5. Go to the target and do `\dn` again. You'll find the lower case one missing.

Before:

![Image](https://user-images.githubusercontent.com/65844208/191453305-4f2f1566-7dec-4297-820c-b6fd29905f3e.png)

After:

![Image](https://user-images.githubusercontent.com/65844208/191452683-b7c891b1-6a54-4e55-8fb0-c47010e39105.png)

Issue #335, opened 2022-09-21
[MYSQL] Importing into a non existing schema (involving upper-case chars) fails but the schema gets created on the target


shubham-yb:
If we try to import into a schema which does not exist, i.e. we populate the `TARGET_DB_SCHEMA` env with a non-existing schema name and then use the `--target-db-schema` flag, the schema gets created on the target side but the table fails to import, throwing an error.

Edit:
Occurs only for schema names with mixed or upper case characters.

**List of target schemas before:**

![Image](https://user-images.githubusercontent.com/65844208/191438766-537acae1-7711-4a1d-85ac-22d3fb6a17c9.png)

**Populating the env variable with a non-existing schema:**

![Image](https://user-images.githubusercontent.com/65844208/191439011-9ee9895b-0146-4128-99da-8d41ed2b067f.png)

**Running the import:**

![Image](https://user-images.githubusercontent.com/65844208/191439291-e164b4cc-2ad2-4f0c-b0f5-8d3188aee05e.png)

**List of target schemas after:**

![Image](https://user-images.githubusercontent.com/65844208/191439558-ad266de7-dbf4-4eac-8a81-71116ebf8c94.png)

sanyamsinghal:
@shubham-yb can you please attach the log file and steps to repro this issue.

shubham-yb:
Sure @sanyamsinghal 

I've made a small edit in the ticket, it repros only for schema names with Upper Case characters.
So it might be a subset of [#334](https://github.com/yugabyte/yb-voyager/issues/334#issue-1380428939)


Here are the steps to reproduce the issue:

1. Populate the `TARGET_DB_SCHEMA` env variable with a **non-existing schema** name having  upper case characters ex: XYZ, Xyz.
2. Run `import schema` using the `--target-db-schema` flag.
3. The command fails.
4. Do a `\dn` on the target. You will find the schema there.


Here are the log files for this:

```
2022-09-21 08:30:31 INFO exportData.go:70 Export of data completed.
2022-09-21 08:30:44 INFO logging.go:65 Logging initialised.
2022-09-21 08:30:44 INFO logging.go:66 Args: [yb-voyager import schema --export-dir /home/centos/export-dir --target-db-host 172.151.41.221 --target-db-user ybvoyager --target-db-schema ABCD --target-db-password password --target-db-name sch_check --start-clean --yes]
[...]

PR #357, opened 2022-09-27
Convert schema names to case-insensitive during import schema

### LINKED ISSUE #334: [MYSQL] --target-db-schema flag fails for case sensitive schema names
If we create a schema in the YB side with a case sensitive name, and pass the env variable with the same, the `import schema` fails with error.

Steps to reproduce:

1. Create a schema on the target with case sensitive name example: XYZ, Xyz.
2. Pass the same in the env variable of `TARGET_DB_SCHEMA`.
3. Try importing into that schema using the `--target-db-schema` flag.

Corresponding error:

Image

rahulb-yb:
[Fixes #334, #335, #336]
Voyager now converts any case sensitive names for target schema to lowercase. The user can convert the schema names to case-sensitive names post import using alter statements, if needed

### DIFF
--- a/yb-voyager/src/utils/utils.go
+++ b/yb-voyager/src/utils/utils.go
@@ -304,3 +304,13 @@ func InsensitiveSliceContains(slice []string, s string) bool {
 	log.Infof("string s=%q did not match with any string in %v", s, slice)
 	return false
 }
+
+func ToInsensitiveSlice(slice []string) []string {
+	for i, object := range slice {
+		if IsQuotedString(object) {
+			object = strings.Trim(object, "\"")
+		}
+		slice[i] = strings.ToLower(object)
+	}
+	return slice
+}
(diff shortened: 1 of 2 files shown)


yb-voyager/src/utils/utils.go
```diff
@@ -304,3 +304,13 @@ func InsensitiveSliceContains(slice []string, s string) bool {
 	log.Infof("string s=%q did not match with any string in %v", s, slice)
 	return false
 }
+
+func ToInsensitiveSlice(slice []string) []string {
```
amit-yb, line 308: Nit: `ToCaseInsensitiveNames(names []string)`.

yb-voyager/src/utils/utils.go
```diff
@@ -304,3 +304,13 @@ func InsensitiveSliceContains(slice []string, s string) bool {
...
+	for i, object := range slice {
+		if IsQuotedString(object) {
```
amit-yb, line 310: Nit: This check is not required. For unquoted string, the `Trim()` call will be no-op.

amit-yb (approved):

PR #359, opened 2022-09-27
Change GitHub workflow to use latest YugabyteDB docker images

rahulb-yb:


.github/workflows/mysql-migtests.yml
```diff
@@ -7,8 +7,10 @@ on:
     branches: [ main, v1.0.0-xbeta ]
 
 jobs:
-
   run-mysql-migration-tests:
+    strategy:
+      matrix:
+        version: [2.14.2.0-b25, 2.15.2.0-b87]
```
sanyamsinghal, line 13: Is it possible to do something like: For odd versions can we fetch the latest yb version available and use that.

sanyamsinghal (approved):
LGTM
Left a minor comment.

.github/workflows/mysql-migtests.yml
rahulb-yb: we can use the "latest" tag, but there is no guarantee of it being the latest odd version, so decided to explicitly use latest two versions as of now

sanyamsinghal: ok

PR #372, opened 2022-09-30
Added a case for TABLE ATTACH

ShivanshGahlot:


sanyamsinghal (approved):
LGTM, thanks for finding out and fixing it

PR #381, opened 2022-09-30
Modified checkPrimaryKey function and resolves the issue #311

priyanshi-yb:


amit-yb (commented):
The PR message must remain exactly the same as the commit being cherry-picked (no need to mention that it is a cherry-pick).

PR #398, opened 2022-10-04
Fixed the issue of the password containing special characters for Oracle, refer -…

priyanshi-yb:
… https://pkg.go.dev/github.com/godror/godror#readme-connection

Modified the connection String passed to `godror` for Oracle DB by passing the `user`, `password` and `connectString` params in the string.


sanyamsinghal (approved):

amit-yb (approved):
Nit: Fix the commit message to move the `refer ...` part on a separate line.

PR #403, opened 2022-10-04
Added flags '--only-pg-support' and '-v' or '--version' to the installer script

ShivanshGahlot:


installer_scripts/install-yb-voyager
```diff
@@ -6,19 +6,37 @@ export LC_ALL=en_US.UTF-8
...
-		# Install from locally checked out branch.
-		VERSION=local
-	else
-		echo "ERROR: Invalid argument: '$1'"
-		exit 1
-	fi
+OPTS=$(getopt -o v: --long version:,only-pg-support -n 'install-yb-voyager' -- "$@")
+
+if [ $? != 0 ]; then 
+	echo "Unable to process the options. Exiting..." >&2;
```
amit-yb, line 12: Does the user get the exact error why the script couldn't process options?

installer_scripts/install-yb-voyager
```diff
@@ -6,19 +6,37 @@ export LC_ALL=en_US.UTF-8
...
+		-v | --version ) 
+			if [ "$2" == local ]; 
+			then 
+				# Install from locally checked out branch.
+				VERSION="$2"; shift 2;
+			else
+				echo "ERROR: Invalid version"
+				exit 1
+			fi 
+			;;
```
amit-yb, line 32: Let's have a single boolean flag named `--install-from-local-source`. If passed on CLI, set the `VERSION` to `local`.

installer_scripts/install-yb-voyager
```diff
@@ -6,19 +6,37 @@ export LC_ALL=en_US.UTF-8
...
+		--only-pg-support ) ONLY_PG="true"; shift ;;
```
amit-yb, line 33: Move the option processing into separate lines:
```
  --only-pg-support )
    ONLY_PG="true"
    shift
    ;;
```

installer_scripts/install-yb-voyager
```diff
@@ -6,19 +6,37 @@ export LC_ALL=en_US.UTF-8
...
+		-- ) shift; break ;;
+		* ) break ;;
```
amit-yb, line 35: Just for my understanding, why is this required?

installer_scripts/install-yb-voyager
```diff
@@ -108,19 +126,23 @@ centos_main() {
...
+		centos_install_oracle_instant_clients
+		update_yb_voyager_bashrc
+		install_ora2pg
+	fi
 	install_yb_voyager
 	create_yb_session_vars_file
-	create_base_ora2pg_conf_file
+	if [ $ONLY_PG == "false" ] ; then
+		create_base_ora2pg_conf_file
+	fi
```
amit-yb, line 145: Merge this if block into the earlier if block by moving line 144 after line 139.

installer_scripts/install-yb-voyager
```diff
@@ -132,20 +154,24 @@ ubuntu_main() {
...
+		ubuntu_install_oracle_instant_clients
+		update_yb_voyager_bashrc
+		install_ora2pg
+	fi
 	install_yb_voyager
 	create_yb_session_vars_file
-	create_base_ora2pg_conf_file
+	if [ $ONLY_PG == "false" ] ; then
+		create_base_ora2pg_conf_file
+	fi
```
amit-yb, line 174: Same as above.

installer_scripts/install-yb-voyager
ShivanshGahlot: Done

installer_scripts/install-yb-voyager
ShivanshGahlot: Done

installer_scripts/install-yb-voyager
ShivanshGahlot: Actually I did this only earlier. But the `/etc/yb-voyager` is created in the `create_yb_session_vars_file`. I have now added a` mkdir -p /etc/yb-voyager` statement in `create_base_ora2pg_conf_file` too. Since in case of macos we only run `create_yb_session_vars_file`, I can't remove the `mkdir -p /etc/yb-voyager` from that function.

installer_scripts/install-yb-voyager
ShivanshGahlot: The user gets an error if a wrong option is passed. I think this statement is unnecessary. I found it in a template on stackoverflow. I have removed it for now.
Also a small finding. getopt doesn't match the whole string. For example, the result would be the same if either `--only-pg-support` or `--only `is passed.

installer_scripts/install-yb-voyager
ShivanshGahlot: This is also unnecessary. I have removed it for now. I tried running without this and it worked fine.

amit-yb (approved):
Good work! 👍

PR #407, opened 2022-10-06
Modified extracting the table name from the table-list flag according to the default cases of Source DBs if not quoted

priyanshi-yb:
This PR includes modifying the table-names from the `--table-list` flag into default cases of the Source DBs if not quoted -
1. For Oracle, convert the name to Upper if not quoted else pass as is without quotes.
2. For PostgreSQL, convert the name to Lower if not quoted else pass as is with quotes.
3. For MySQL, pass the table list as is because names are case Sensitive in MySQL.

### DIFF
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -213,16 +213,32 @@ func extractTableListFromString(flagTableList string) []string {
 		return []string{}
 	}
 	tableList := utils.CsvStringToSlice(flagTableList)
-	if source.DBType != POSTGRESQL {
+	if source.DBType == MYSQL {
 		return tableList
 	}
 	// in postgres format should be schema.table, public is default and other parts of code assume schema.table format
 	for _, table := range tableList {
 		parts := strings.Split(table, ".")
 		if len(parts) == 1 {
-			finalTableList = append(finalTableList, "public."+table)
+			if source.DBType == ORACLE {
+				if utils.IsQuotedString(parts[0]) {
+					parts[0] = strings.Trim(parts[0],`"`)
+				} else {
+					parts[0] = strings.ToUpper(parts[0])
+				}
+			} else if  source.DBType == POSTGRESQL {
+				finalTableList = append(finalTableList, "public."+table)
+			}
 		} else if len(parts) == 2 {
-			finalTableList = append(finalTableList, table)
+			if source.DBType == ORACLE {
+				if utils.IsQuotedString(parts[1]) {
+					parts[1] = strings.Trim(parts[1],`"`)
+				} else {
+					parts[1] = strings.ToUpper(parts[1])
+				}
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -213,16 +213,44 @@ func extractTableListFromString(flagTableList string) []string {
...
+				finalTableList = append(finalTableList, "public."+table)
+			}
 		} else if len(parts) == 2 {
-			finalTableList = append(finalTableList, table)
+			if source.DBType == ORACLE {
+				if utils.IsQuotedString(parts[1]) {
+					parts[1] = strings.Trim(parts[1],`"`)
+				} else {
+					parts[1] = strings.ToUpper(parts[1])
+				}
```
amit-yb, line 244: This implementation is too complicated. Can we simplify in the following lines:
```
		parts := strings.Split(table, ".")
		if len(parts) > 2 {
			utils.ErrExit("invalid table name: %q", table)
		}
		tableName := parts[0]
		if utils.IsQuotedString(tableName) {
			tableName = strings.Trim(tableName,`"`)
		} else {
			switch source.DBType {
			case ORACLE:
				tableName = strings.ToUpper(tableName)
			case POSTGRESQL:
				tableName = strings.ToLower(tableName)
			case MYSQL:
				// No conversion needed for MySQL.
			}
		}

		if len(parts) == 2 {
			tableName = fmt.Sprintf("%s.%s", parts[1], tableName)
		} else if source.DBType == POSTGRESQL {
			tableName = fmt.Sprintf("public.%s", tableName)
		}
		finalTableList = append(finalTableList, tableName)
	}
```

yb-voyager/cmd/exportData.go
```diff
@@ -213,19 +213,42 @@ func extractTableListFromString(flagTableList string) []string {
...
-		} else if len(parts) == 2 {
-			finalTableList = append(finalTableList, table)
+			tableName = parts[0]
+		} else {
+			tableName = parts[1]
+		}
+		if utils.IsQuotedString(tableName) {
+			if  source.DBType == ORACLE {
+				tableName = strings.Trim(tableName,`"`)
+			}
```
amit-yb, line 234: For PG and MySQL, won't you need to trim the double quotes?

priyanshi-yb: No, for PG we need not to remove the double quotes, the queries work with that as is. And for MySQL as the table-name are case-Sensitive the quotes are not even there and also I had tried creating a table with quotes but it seems like that it doesn't happen in MySQL - 
```
MySQL [pk_missing]> create table "Test"(id int);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '"Test"(id int)' at line 1
```
so we need not to handle anything for MySQL

amit-yb (approved):
LGTM.

PR #458, opened 2022-10-13
Schemas are imported by default when source is PG

### LINKED ISSUE #453: [POSTGRES] The "schema" object type must be imported as default
For postgres, the `schema` object type must not be chosen by the user and be imported as default while using `--object-list` or `--exclude-object-list` as the other objects require the schema to be imported into. 

If the `schema` object type is not in `--object-list`, it throws an error:

`ERROR: schema "public" does not exist (SQLSTATE 3F000) `

rahulb-yb:
[Fixes #453]
If the `schema` object type were excluded using either `--object-list` or `--exclude-object-list` when source is PG, migration would fail at the import schema stage as there would be no expected schema to import into; this patch fixes this behaviour.

### DIFF
--- a/yb-voyager/cmd/importSchemaYugabyteDB.go
+++ b/yb-voyager/cmd/importSchemaYugabyteDB.go
@@ -92,5 +92,8 @@ func applySchemaObjectFilterFlags(importObjectOrderList []string) []string {
 	} else {
 		finalImportObjectList = utils.SetDifference(importObjectOrderList, excludeObjectList)
 	}
+	if sourceDBType == "postgresql" && !slices.Contains(finalImportObjectList, "SCHEMA") { // Schema should be migrated by default.
+		finalImportObjectList = append([]string{"SCHEMA"}, finalImportObjectList...)
+	}
 	return finalImportObjectList
 }


yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -92,5 +92,8 @@ func applySchemaObjectFilterFlags(importObjectOrderList []string) []string {
 	} else {
 		finalImportObjectList = utils.SetDifference(importObjectOrderList, excludeObjectList)
 	}
+	if sourceDBType == "postgresql" && !slices.Contains(finalImportObjectList, "SCHEMA") { // Schema should be migrated by default.
```
kneeraj, line 95: @rahulb-yb if the source schema is passed in the arg then this should be fine no? We can assume that the user is looking for that schema and just check if it exists or not in the target, and  create it if it doesn't. We know about the schema from the exported sql files? Does it make sense?

rahulb-yb: For import schema, we simply replay the .sql files that have been exported, this is based on object type rather than one (or more) specific schema to import into. When the user does not include `schema` in the object list, we do not make any schemas available for importing the remainder of the schema to be consolidated into

This is the simplest approach I could think of which considers all scenarios, simply append the `schema` object type if it already doesn't exist, to the list of object types we are to import on the target

kneeraj: ok. sounds good.

kneeraj (approved):

PR #470, opened 2022-10-17
Added skip functions to skip alter table statements in the first call to importSchemaInternal

ShivanshGahlot:
and then import only alter table statements in the second call.

### DIFF
--- a/yb-voyager/cmd/importSchema.go
+++ b/yb-voyager/cmd/importSchema.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/srcdb"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
+	"golang.org/x/exp/slices"
 
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/callhome"
 )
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/importSchema.go
```diff
@@ -81,7 +82,18 @@ func importSchema() {
 	}
 	objectList = applySchemaObjectFilterFlags(objectList)
 	log.Infof("List of schema objects to import: %v", objectList)
-	importSchemaInternal(&target, exportDir, objectList, nil)
+
+	skipFn := func(objType, stmt string) bool {
+		stmt = strings.ToUpper(stmt)
+		return strings.HasPrefix(stmt, "ALTER TABLE")
+	}
```
amit-yb, line 89: Name this function as `isAlterTableStmt`.

    return objType == "SEQUENCE" && Strings.HasPrefix(...)

We need to skip the alter table statements only from the sequences.sql . Your current implementation skips them from all the files. So the extra condition on `objType` is required.

yb-voyager/cmd/importSchema.go
```diff
@@ -81,7 +82,18 @@ func importSchema() {
...
+	importSchemaInternal(&target, exportDir, objectList, skipFn)
+
+	skipFn = func(objType, stmt string) bool {
+		stmt = strings.ToUpper(stmt)
+		return !strings.HasPrefix(stmt, "ALTER TABLE")
+	}
```
amit-yb, line 95: You can frame the implementation as a complement of the previous function e.g.

    skipFn = func(objType, stmt string) bool {
        return !isAlterTableStmt(objType, stmt)
    }

This implementation declares the "intent" of the code. i.e. we're importing the statements skipped by the previous invocation.

yb-voyager/cmd/importSchema.go
```diff
@@ -81,7 +82,18 @@ func importSchema() {
 	}
 	objectList = applySchemaObjectFilterFlags(objectList)
 	log.Infof("List of schema objects to import: %v", objectList)
-	importSchemaInternal(&target, exportDir, objectList, nil)
+
```
amit-yb, line 85: Add a comment stating: Import `ALTER TABLE` statements from the sequences.sql only after importing everything else.

amit-yb (approved):
LGTM.

PR #477, opened 2022-10-17
[Fixes #475] Rename ybvoyager schema to ybvoyager_metadata .

### LINKED ISSUE #475: Tables get imported in the "user" schema as default
If there exists a schema with the name same as the `user` ex: `ybvoyager` then it is considered as default and not the `public` schema.

Consequently the tables get imported in that schema instead of `public` and it comes up empty.

Steps to reproduce:

1. Create a schema on the target which has the same name as the user. ex: ybvoyager
2. Run `import-schema`
3. The tables get imported in `ybvoyager` schema instead of `public`.

amit-yb:
Voyeger's target user creation guide recommends user name `ybvoyager`. If there is a schema with the same name, then it will be treated as default schema instead of the usual "public".

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -309,8 +309,8 @@ func importData() {
 
 func createVoyagerSchemaOnTarget(connPool *tgtdb.ConnectionPool) error {
 	cmds := []string{
-		"CREATE SCHEMA IF NOT EXISTS ybvoyager",
-		`CREATE TABLE IF NOT EXISTS ybvoyager.batches (
+		"CREATE SCHEMA IF NOT EXISTS ybvoyager_metadata",
+		`CREATE TABLE IF NOT EXISTS ybvoyager_metadata.batches (
 			schema_name VARCHAR(250),
 			file_name VARCHAR(250),
 			rows_imported BIGINT,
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -309,8 +309,8 @@ func importData() {
 
 func createVoyagerSchemaOnTarget(connPool *tgtdb.ConnectionPool) error {
 	cmds := []string{
-		"CREATE SCHEMA IF NOT EXISTS ybvoyager",
-		`CREATE TABLE IF NOT EXISTS ybvoyager.batches (
+		"CREATE SCHEMA IF NOT EXISTS ybvoyager_metadata",
```
sanyamsinghal, line 312: nit: Consider using a `const` variable for storing the metainfo schema name

sanyamsinghal (approved):
LGTM, left a minor commetn

yb-voyager/cmd/importData.go
```diff
@@ -309,8 +309,8 @@ func importData() {
...
+		`CREATE TABLE IF NOT EXISTS ybvoyager_metadata.batches (
 			schema_name VARCHAR(250),
 			file_name VARCHAR(250),
 			rows_imported BIGINT,
```
ShivanshGahlot, line 316: Nit:
The error message can also be edited on line number `327` to:
`return fmt.Errorf("create ybvoyager_metadata schema on target: %w", err)`

ShivanshGahlot (approved):

PR #483, opened 2022-10-18
Added case in ignore-exists flag for existing partition

### LINKED ISSUE #479: Add "is already a partition" case to ignore exist flag
This is the error and corresponding SQL statement which also needs to be covered as a part of ignore-exist flag
```
    ERROR: "nist_hashes_1" is already a partition (SQLSTATE 42809)
    STATEMENT: ALTER TABLE ONLY public.nist_hashes ATTACH PARTITION public.nist_hashes_1 FOR VALUES IN (1);
```

rahulb-yb:
[Fixes #479]
Added a case for when a partition already exists on the target DB in the case of an import schema retry, to be ignored when using `ignore-exists`

### DIFF
--- a/yb-voyager/cmd/importSchema.go
+++ b/yb-voyager/cmd/importSchema.go
@@ -182,11 +182,11 @@ func checkIfTargetSchemaExists(conn *pgx.Conn, targetSchema string) bool {
 	return fetchedSchema == targetSchema
 }
 
-func isDuplicateDDL(errString string) bool {
-	duplicateDDLErrors := []string{"already exists",
+func isAlreadyExists(errString string) bool {
+	alreadyExistsErrors := []string{"already exists",
 		"multiple primary keys",
 		"already a partition"}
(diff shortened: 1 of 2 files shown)


rahulb-yb:
P.S. changes to go.mod and go.sum were made automatically after running `go mod tidy`

yb-voyager/cmd/importSchema.go
```diff
@@ -181,3 +181,15 @@ func checkIfTargetSchemaExists(conn *pgx.Conn, targetSchema string) bool {
 
 	return fetchedSchema == targetSchema
 }
+
+func isDuplicateDDL(errString string) bool {
+	duplicateDDLErrors := []string{"already exists",
```
sanyamsinghal, line 186: `duplicateDDLErrors` --> `alreadyExistsErrors`

sanyamsinghal (approved):
LGTM, left some minor comments have a look at it.

PR #490, opened 2022-10-19
Add test result validation framework.

amit-yb:
This PR adds a way for automation tests to run validations against the target DB.
The test driver expects an executable named `validate` in the test-dir. If present, the test-driver invokes the script towards the end of the migration.
The `validate` can be an executable Python script. Such a Python script can use a utility class `PostgresDB` from the `yb` module (`migtests/lib/yb.py`).


kneeraj (approved):
lgtm

sanyamsinghal (approved):

PR #505, opened 2022-10-21
Enhance pg-dvdrental automation test with custom collation DDL

rahulb-yb:
With the addition of custom collation support in PG[#465], we need to enhance our automation tests with the same. This patch addresses this issue, by enhancing the `pg-dvdrental` automation test.

### DIFF
--- a/migtests/tests/pg-dvdrental/init-db
+++ b/migtests/tests/pg-dvdrental/init-db
@@ -19,5 +19,8 @@ run_psql postgres "CREATE DATABASE ${SOURCE_DB_NAME};"
 echo "Initialising source database."
 run_pg_restore ${SOURCE_DB_NAME} dvdrental.tar
 
+echo "Adding custom collation."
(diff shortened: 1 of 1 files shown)


migtests/tests/pg-dvdrental/init-db
```diff
@@ -19,5 +19,8 @@ run_psql postgres "CREATE DATABASE ${SOURCE_DB_NAME};"
 echo "Initialising source database."
 run_pg_restore ${SOURCE_DB_NAME} dvdrental.tar
 
+echo "Adding custom collation."
+run_psql ${SOURCE_DB_NAME} "CREATE COLLATION fulah (provider= icu, locale='ff-x-icu');"
```
rahulb-yb, line 23: Note that I have chosen a collation which is not provided by `libc`, YB only supports C for the libc provider, which is the default collation used

amit-yb (approved):
Thanks for this PR. 👍

PR #510, opened 2022-10-26
Determine the number of parallel jobs to use by default during import data

### LINKED ISSUE #468: import data: determine default value of --parallel-jobs based on num cores in target cluster
Currently `import data` command sets the default value of `--parallel-jobs` to the number of nodes in the target cluster. But the default value is too low. voyager should set the value to half the number of cores in the target cluster. e.g. if the target cluster has 48 cores, it should use 24 as default value.

rahulb-yb:
[Fixes #468]
For PG 9.3+, superusers are allowed to fire (and use the returned output of) shell commands from within the PG session. We make use of this functionality to gather the number of cores for each node in the cluster, based on which the default value of `--parallel-jobs` is decided. The current workflow is like so:
- If we fail to query even a single node in the cluster, default value is `len(nodes) * 2`
- [Edited] If we query all nodes in the cluster, default value is `(total_cores/2)`.
- [Edited] The older version of this patch extrapolated partially failed cases over the number of nodes in the cluster, this is no longer the case.

P.S. review does not have to be done commit-by-commit, some implementations were incorrect in the previous versions of this patch

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -208,6 +208,7 @@ func fetchDefaultParllelJobs(targets []*tgtdb.Target) int {
 	totalCores := 0
 	targetCores := 0
 	for _, target := range targets {
+		log.Infof("Determining CPU core count on: %s", utils.GetRedactedURLs([]string{target.Uri})[0])
 		conn, err := pgx.Connect(context.Background(), target.Uri)
 		if err != nil {
 			log.Warnf("Unable to reach target while querying cores: %v", err)
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -205,30 +205,34 @@ func getYBServers() []*tgtdb.Target {
 }
 
 func fetchDefaultParllelJobs(targets []*tgtdb.Target) int {
-	coreQueries := []string{"CREATE TEMP TABLE yb_voyager_cores(num_cores int);",
-		"COPY yb_voyager_cores(num_cores) FROM PROGRAM 'grep processor /proc/cpuinfo|wc -l';",
-		"SELECT num_cores FROM yb_voyager_cores;"}
 	totalCores := 0
 	targetCores := 0
 	for _, target := range targets {
 		conn, err := pgx.Connect(context.Background(), target.Uri)
 		if err != nil {
-			utils.ErrExit("Unable to reach target while querying cores: %v", err)
-		}
-		defer conn.Close(context.Background())
-		_, err = conn.Exec(context.Background(), coreQueries[0])
-		if err != nil {
-			utils.ErrExit("Unable to create tables on target DB: %v", err)
-		}
-		_, err = conn.Exec(context.Background(), coreQueries[1])
-		if err != nil {
-			// There may occur scenarios where we cannot query the cores, store in log as a warning, continue with nodeCount * 2.
-			log.Warnf("Error while running query %s on host %s: %v", coreQueries[1], utils.GetRedactedURLs([]string{target.Uri}), err)
+			log.Warnf("Unable to reach target while querying cores: %v", err)
 			return len(targets) * 2
 		}
-		rows := conn.QueryRow(context.Background(), coreQueries[2])
-		if err = rows.Scan(&targetCores); err != nil {
-			utils.ErrExit("Error while running query %s: %v", coreQueries[2], err)
+		defer conn.Close(context.Background())
+
+		cmd := "CREATE TEMP TABLE yb_voyager_cores(num_cores int);"
+		_, err = conn.Exec(context.Background(), cmd)
+		if err != nil {
+			log.Warnf("Unable to create tables on target DB: %v", err)
+			return len(targets) * 2
+		}
+
+		cmd = "COPY yb_voyager_cores(num_cores) FROM PROGRAM 'grep processor /proc/cpuinfo|wc -l';"
+		_, err = conn.Exec(context.Background(), cmd)
+		if err != nil {
+			log.Warnf("Error while running query %s on host %s: %v", cmd, utils.GetRedactedURLs([]string{target.Uri}), err)
+			return len(targets) * 2
+		}
+
+		cmd = "SELECT num_cores FROM yb_voyager_cores;"
+		if err = conn.QueryRow(context.Background(), cmd).Scan(&targetCores); err != nil {
+			log.Warnf("Error while running query %s: %v", cmd, err)
+			return len(targets) * 2
 		}
 		totalCores += targetCores
 	}
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -194,7 +194,8 @@ func getYBServers() []*tgtdb.Target {
 	}
 
 	if parallelImportJobs == -1 {
-		parallelImportJobs = len(targets)
+		parallelImportJobs = fetchParllelJobs(targets)
+		fmt.Printf("Uhhh you have %d jobs", parallelImportJobs)
 	}
 
 	if loadBalancerUsed { // if load balancer is used no need to check direct connectivity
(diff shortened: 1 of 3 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -206,6 +207,43 @@ func getYBServers() []*tgtdb.Target {
...
 
+func fetchParllelJobs(targets []*tgtdb.Target) int {
+	totalCores := 0
+	targetCores := 0
+	queriedTargets := 0
+	for _, target := range targets {
+		conn, err := pgx.Connect(context.Background(), target.Uri)
+		if err != nil {
+			utils.ErrExit("Unable to reach target while querying cores: %v", err)
+		}
```
amit-yb, line 218: defer conn.Close()

yb-voyager/cmd/importData.go
```diff
@@ -194,7 +194,8 @@ func getYBServers() []*tgtdb.Target {
 	}
 
 	if parallelImportJobs == -1 {
-		parallelImportJobs = len(targets)
+		parallelImportJobs = fetchParllelJobs(targets)
+		utils.PrintAndLog("Using %d parallel jobs by default. Use --parallel-jobs to specify a custom value", parallelImportJobs)
 	}
```
amit-yb, line 199: `getYBServers()` is doing two jobs:
- get the list of yb servers
- find out default value of `parallelImportJobs`

Let's move the code associated with the second aspect to a separate function `getDefaultImportJobs()`. And move the call to the new function inside the `importData()`.

yb-voyager/cmd/importData.go
```diff
@@ -206,6 +207,43 @@ func getYBServers() []*tgtdb.Target {
 	return targets
 }
 
+func fetchParllelJobs(targets []*tgtdb.Target) int {
+	totalCores := 0
+	targetCores := 0
+	queriedTargets := 0
+	for _, target := range targets {
+		conn, err := pgx.Connect(context.Background(), target.Uri)
```
amit-yb, line 215: With LB, all `target.Uri` are going to point to the same IP. In that case, we don't know to which instance the request will hit.

yb-voyager/cmd/importData.go
```diff
@@ -206,6 +207,43 @@ func getYBServers() []*tgtdb.Target {
...
+		if err != nil {
+			utils.ErrExit("Unable to reach target while querying cores: %v", err)
+		}
+		_, err = conn.Exec(context.Background(), "CREATE TEMP TABLE yb_voyager_cores(num_cores int);")
+		if err != nil {
+			utils.ErrExit("Unable to create tables on target DB: %v", err)
+		}
+		_, err = conn.Exec(context.Background(), "COPY yb_voyager_cores(num_cores) FROM PROGRAM 'grep processor /proc/cpuinfo|wc -l';")
+		if err != nil {
+			continue
```
amit-yb, line 225: Log the redacted URI, query, and the error message.

yb-voyager/cmd/importData.go
```diff
@@ -206,6 +207,43 @@ func getYBServers() []*tgtdb.Target {
...
+		}
+		rows, err := conn.Query(context.Background(), "SELECT num_cores FROM yb_voyager_cores;")
+		if err != nil {
+			continue
```
amit-yb, line 229: Log the query and the error message.

yb-voyager/cmd/importData.go
```diff
@@ -206,6 +207,43 @@ func getYBServers() []*tgtdb.Target {
...
+		}
+		_, err = conn.Exec(context.Background(), "CREATE TEMP TABLE yb_voyager_cores(num_cores int);")
+		if err != nil {
+			utils.ErrExit("Unable to create tables on target DB: %v", err)
+		}
+		_, err = conn.Exec(context.Background(), "COPY yb_voyager_cores(num_cores) FROM PROGRAM 'grep processor /proc/cpuinfo|wc -l';")
+		if err != nil {
+			continue
+		}
+		rows, err := conn.Query(context.Background(), "SELECT num_cores FROM yb_voyager_cores;")
```
amit-yb, line 227: `QueryRow()` is easier while fetching only one row from a table.

yb-voyager/cmd/importData.go
```diff
@@ -206,6 +207,43 @@ func getYBServers() []*tgtdb.Target {
...
+		}
+		queriedTargets++
+		totalCores += targetCores
+	}
+	// Return twice the number of nodes if querying was not possible.
+	if queriedTargets == 0 {
+		return len(targets) * 2
+	}
+	// Return the extrapolated value of sum(cores)/2 in the case of partial or full querying.
+	return totalCores * len(targets) / (2 * queriedTargets)
```
amit-yb, line 244: Simplify the calculation:
(Rename `totalCores` as `coreCount`.)
```
coresPerNode := float32(coreCount) / queriedTargets
totalCores := int(coresPerNode * len(targets))
return totalCores/2
```

yb-voyager/cmd/importData.go
rahulb-yb: Is there a scenario where, on YBM, there is a difference in the number of cores for any pair of nodes?

yb-voyager/cmd/importData.go
```diff
@@ -193,19 +193,48 @@ func getYBServers() []*tgtdb.Target {
...
+func fetchDefaultParllelJobs(targets []*tgtdb.Target) int {
+	coreQueries := []string{"CREATE TEMP TABLE yb_voyager_cores(num_cores int);",
+		"COPY yb_voyager_cores(num_cores) FROM PROGRAM 'grep processor /proc/cpuinfo|wc -l';",
+		"SELECT num_cores FROM yb_voyager_cores;"}
+	totalCores := 0
+	targetCores := 0
+	for _, target := range targets {
+		conn, err := pgx.Connect(context.Background(), target.Uri)
+		if err != nil {
+			utils.ErrExit("Unable to reach target while querying cores: %v", err)
```
amit-yb, line 216: We must not fail+exit data import when we cannot determine core count. Instead, just log a WARNING msg and use the default `len(targets) * 2`.

yb-voyager/cmd/importData.go
```diff
@@ -193,19 +193,48 @@ func getYBServers() []*tgtdb.Target {
...
 	} else {
 		testYbServers(targets)
 	}
 	return targets
 }
 
+func fetchDefaultParllelJobs(targets []*tgtdb.Target) int {
+	coreQueries := []string{"CREATE TEMP TABLE yb_voyager_cores(num_cores int);",
+		"COPY yb_voyager_cores(num_cores) FROM PROGRAM 'grep processor /proc/cpuinfo|wc -l';",
+		"SELECT num_cores FROM yb_voyager_cores;"}
```
amit-yb, line 210: I would suggest to just create a simple `cmd` variable near the query:
```
cmd := "CREATE TEMP TABLE ..."
_, err = conn.Exec(context.Background(), cmd)
if err != nil {
  ...
}
cmd = "COPY ..."
_, err = conn.Exec(context.Background(), cmd)
if err != nil {
  ...
}
```

yb-voyager/cmd/importData.go
```diff
@@ -193,19 +193,48 @@ func getYBServers() []*tgtdb.Target {
...
+			utils.ErrExit("Unable to create tables on target DB: %v", err)
+		}
+		_, err = conn.Exec(context.Background(), coreQueries[1])
+		if err != nil {
+			// There may occur scenarios where we cannot query the cores, store in log as a warning, continue with nodeCount * 2.
+			log.Warnf("Error while running query %s on host %s: %v", coreQueries[1], utils.GetRedactedURLs([]string{target.Uri}), err)
+			return len(targets) * 2
+		}
+		rows := conn.QueryRow(context.Background(), coreQueries[2])
+		if err = rows.Scan(&targetCores); err != nil {
```
amit-yb, line 230: Nit: it is idiomatic to chain the `Scan()` call:
```
err = conn.QueryRow(...).Scan(&varName)
```

yb-voyager/cmd/importData.go
```diff
@@ -276,6 +305,12 @@ func importData() {
 		targetUriList = append(targetUriList, t.Uri)
 	}
 	log.Infof("targetUriList: %s", utils.GetRedactedURLs(targetUriList))
+
+	if parallelImportJobs == -1 {
+		parallelImportJobs = fetchDefaultParllelJobs(targets)
+		utils.PrintAndLog("Using %d parallel jobs by default. Use --parallel-jobs to specify a custom value", parallelImportJobs)
```
amit-yb, line 311: Take this call out of the `if` block. That will ensure that the message is show even for LB cases.

yb-voyager/cmd/importData.go
amit-yb: There can be e.g. user deployed a cluster with N cores per node. After few days, the user added a new node to the cluster with M cores.
But we don't need to be precise in this particular case. Aprox core count is also good enough for us.

yb-voyager/cmd/importData.go
```diff
@@ -193,19 +193,48 @@ func getYBServers() []*tgtdb.Target {
...
 	}
 
-	if parallelImportJobs == -1 {
-		parallelImportJobs = len(targets)
-	}
-
 	if loadBalancerUsed { // if load balancer is used no need to check direct connectivity
 		utils.PrintAndLog(LB_WARN_MSG)
+		parallelImportJobs = 2 * len(targets)
+		utils.PrintAndLog("Using %d parallel jobs by default. Use --parallel-jobs to specify a custom value", parallelImportJobs)
```
amit-yb, line 199: After you address the comment for line 311, you won't need this additional `PrintAndLog()`.

yb-voyager/cmd/importData.go
rahulb-yb: If you are talking about the call to the print message, I only show this if parallel-jobs is not provided explicitly, which is why I have added two separate instances of print messages for the LB case and the other generic case

yb-voyager/cmd/importData.go
```diff
@@ -193,19 +193,52 @@ func getYBServers() []*tgtdb.Target {
...
 		testYbServers(targets)
 	}
 	return targets
 }
 
+func fetchDefaultParllelJobs(targets []*tgtdb.Target) int {
+	totalCores := 0
+	targetCores := 0
+	for _, target := range targets {
+		conn, err := pgx.Connect(context.Background(), target.Uri)
```
amit-yb, line 211: Nit: Consider adding an Info level log stating `"Determining CPU core count on: %s", redactedURI`

amit-yb (approved):
Thanks for accommodating suggested changes. 👍

PR #513, opened 2022-10-26
Enabled export of pg-extension fixes #512

### LINKED ISSUE #512: [PostgreSQL] Extension not getting exported
Bug: pg-extensions are not getting exported as part of `export schema` command.
Fix: Previously they used to be exported but after the schema list flag feature it is required to enable extensions export in pg_dump. Add `--extension "*"` to `pg_dump` args, this will export all the extensions from postgres database.

sanyamsinghal:
- Previously they used to be exported but after the schema list flag feature it is required to enable extensions export in pg_dump


amit-yb (approved):
LGTM.

Issue #531, opened 2022-11-02
[PostgreSQL] Exported Schema with comments present in SQL statement


sanyamsinghal:
Voyager fails to export the function having code block with comments

Example:
```

CREATE FUNCTION public.init_export_preset() RETURNS void
    AS $$
BEGIN
    SELECT * from t1;

--------------------------------------------------
-- Comment 
--------------------------------------------------
    INSERT INTO t1

END;
$$;
    

```

PR #532, opened 2022-11-02
[#531] Bug fix for PostgreSQL Export Schema having SQL statements with comments in it

sanyamsinghal:


amit-yb (approved):
LGTM. Thanks for accommodating such a large change request.

PR #533, opened 2022-11-02
Update mysql_views_automation.sql

shubham-yb:
Commented out the view DDL which wasn't exporting as expected and throwing syntax error. Will reinstate once the fix has landed.


rahulb-yb (approved):
Please file a corresponding ticket on our GH to allow us to keep track of this, in addition to the one you will be filing on ora2pg GH

PR #545, opened 2022-11-04
Regression in 938d5f3d0d91e1763b12694cb3763242f3363492: Not exporting the last SQL statement from temp/schema.sql file

sanyamsinghal:


amit-yb (approved):
Good catch!

PR #580, opened 2022-11-10
Added few more automation tests with validations for MySQL and PG

priyanshi-yb:
1. MySQL - triggers and procedures
2. PG - views, Types, DOMAIN, case-sensitive table-name.

### DIFF
--- a/.github/workflows/pg-migtests.yml
+++ b/.github/workflows/pg-migtests.yml
@@ -80,7 +80,7 @@ jobs:
     - name: "TEST: pg-views"
       run: migtests/scripts/run-test.sh pg-tests/pg-views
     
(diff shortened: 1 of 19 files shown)


.github/workflows/pg-migtests.yml
```diff
@@ -76,6 +76,12 @@ jobs:
 
     - name: "TEST: pg-sequences"
       run: migtests/scripts/run-test.sh pg-tests/pg-sequences
+    
+    - name: "TEST: pg-views"
+      run: migtests/scripts/run-test.sh pg-tests/pg-views
+    
+    - name: "TEST: pg-misc-objects-1 (Enum, case-sensitive-table-name, Domain)"
```
shubham-yb, line 83:
```suggestion
    - name: "TEST: pg-misc-objects-1 (Types, case-sensitive-table-name, Domain)"
```

PR #591, opened 2022-11-16
Shivansh/change tar options installer script

ShivanshGahlot:
Changed options of tar to xzf in the installer-script


sanyamsinghal (approved):
LGTM, @ShivanshGahlot  please update the final commit message before merging. Maybe something like this(or whatever is more meaningful):
```
Changed options of tar to 'xzf' from 'xf' in the installer-script to decompress gzip files
```

PR #592, opened 2022-11-16
Use Parent Table for loading data of table partitions during import data

sanyamsinghal:
1. Enabled --load-via-partition-root flag in pg_dump to use parent table for table partitions
2. Updated extracting copy command for tables in postgres from toc.txt file
    - Need to query pg catalog tables to findout the parentTable for using in copyCommandRegex

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -1177,6 +1177,7 @@ func extractCopyStmtForTable(table string, fileToSearchIn string) {
 		defer conn.Close(context.Background())
 
 		parentTable := getParentTable(table, conn)
+		log.Infof("parentTable=%s for table=%s", parentTable, table)
 		if parentTable != "" { // this is partitioned table
 			// add schema name to parentTable
 			splits := strings.Split(table, ".")
@@ -1187,7 +1188,7 @@ func extractCopyStmtForTable(table string, fileToSearchIn string) {
 			}
 			copyCommandRegex = regexp.MustCompile(fmt.Sprintf(`(?i)COPY %s[\s]+\(.*\) FROM STDIN`, parentTable))
 		} else {
-			// if no schema then add public in tableName as it is there in postgres' toc file
+			// if no schema is provided use public with tableName as it is there in postgres' toc file
 			if len(strings.Split(table, ".")) == 1 {
 				copyCommandRegex = regexp.MustCompile(fmt.Sprintf(`(?i)COPY public.%s[\s]+\(.*\) FROM STDIN`, table))
 			} else {
@@ -1228,7 +1229,7 @@ func extractCopyStmtForTable(table string, fileToSearchIn string) {
 func getParentTable(table string, conn *pgx.Conn) string {
 	var parentTable string
 	query := fmt.Sprintf(`SELECT inhparent::pg_catalog.regclass
-	FROM pg_catalog.pg_class c JOIN pg_catalog.pg_inherits i ON c.oid = inhrelid
+	FROM pg_catalog.pg_class c JOIN pg_catalog.pg_inherits ON c.oid = inhrelid
 	WHERE c.oid = '%s'::regclass::oid`, table)
 
 	var currentParentTable string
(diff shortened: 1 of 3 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -1206,6 +1223,33 @@ func extractCopyStmtForTable(table string, fileToSearchIn string) {
...
+	if len(currentParentTable) == 0 {
+		return ""
+	} else {
+		parentTable = getParentTable(currentParentTable, conn)
+		if len(parentTable) == 0 {
+			parentTable = currentParentTable
+		}
+	}
+
+	log.Infof("parentTable=%s for table=%s", parentTable, table)
```
rahulb-yb, line 1249: We can probably put this log statement outside the recursion of the function, it could otherwise lead to a lot of unnecessary cluttering in the log file; log the "oldest" parent of the table we wish to query

yb-voyager/cmd/importData.go
```diff
@@ -1206,6 +1223,33 @@ func extractCopyStmtForTable(table string, fileToSearchIn string) {
...
+	if err != pgx.ErrNoRows && err != nil {
+		utils.ErrExit("Error in querying parent tablename for table=%s: %v", table, err)
+	}
+
+	if len(currentParentTable) == 0 {
+		return ""
+	} else {
+		parentTable = getParentTable(currentParentTable, conn)
+		if len(parentTable) == 0 {
+			parentTable = currentParentTable
```
rahulb-yb, line 1245: Hmm, this recursion can be cut down by adding only one check of the type `len(table name)==0`, try exiting early by checking the length of the `table` argument as soon as you enter the function. 

P.S. You may need to alter the function return type and assign the "parent table" to a pointer declared outside the function, haven't given it much thought, but speaking from past experiences 🙂

yb-voyager/cmd/importData.go
```diff
@@ -1206,6 +1223,33 @@ func extractCopyStmtForTable(table string, fileToSearchIn string) {
 	}
 }
 
+// get the parent table if its a partitioned table in yugabytedb
+// also covers the case when the partitions are further subpartitioned to multiple levels
+func getParentTable(table string, conn *pgx.Conn) string {
+	var parentTable string
+	query := fmt.Sprintf(`SELECT inhparent::pg_catalog.regclass
+	FROM pg_catalog.pg_class c JOIN pg_catalog.pg_inherits i ON c.oid = inhrelid
```
rahulb-yb, line 1231: Any reason for the alias `pg_catalog.pg_inherits i`?...I don't see it being used anywhere..

yb-voyager/cmd/importData.go
```diff
@@ -1168,14 +1168,31 @@ func extractCopyStmtForTable(table string, fileToSearchIn string) {
...
+			// add schema name to parentTable
+			splits := strings.Split(table, ".")
+			if len(splits) > 1 {
+				parentTable = splits[0] + "." + parentTable
+			} else {
+				parentTable = "public." + parentTable
+			}
+			copyCommandRegex = regexp.MustCompile(fmt.Sprintf(`(?i)COPY %s[\s]+\(.*\) FROM STDIN`, parentTable))
+		} else {
+			// if no schema then add public in tableName as it is there in postgres' toc file
```
rahulb-yb, line 1190: This comment probably does not belong here..

rahulb-yb (commented):
Left some comments, please address/respond to them

yb-voyager/cmd/importData.go
sanyamsinghal: done

yb-voyager/cmd/importData.go
sanyamsinghal: Didn't use it with `inhrelid` column. Removing it as not needed.

yb-voyager/cmd/importData.go
sanyamsinghal: why? It specific to the next line

yb-voyager/cmd/importData.go
sanyamsinghal: current function code is also similar. Just the base condition is at different point, not passing the empty tableName to the function, stopping it before that.
Its better not to use some global or outside variable and just make the function as separated out as possible.

rahulb-yb (approved):

PR #599, opened 2022-11-17
Added a flag --straight-order which when specified imports the schema objects in the order specified

ShivanshGahlot:
Added a flag --straight-order which when specified imports the scheme objects in the order specified by the user. Also added the required code.


rahulb-yb (approved):
LGTM, thanks for change

Issue #598, opened 2022-11-17
support the schema names in the target-db-schema with any case


priyanshi-yb:
This issue is to support the schema names to be passed with any case in `target-db-schema`.

Jira ticket for more info - [DB-4219](https://yugabyte.atlassian.net/browse/DB-4219)

PR #600, opened 2022-11-17
fixed GH #598, converted the schema name for target to lower and adde…

priyanshi-yb:
…d check to determine schema exist or not (#598)

### DIFF
--- a/migtests/tests/pg-tests/pg-views/env.sh
+++ b/migtests/tests/pg-tests/pg-views/env.sh
@@ -1,3 +1,4 @@
 export SOURCE_DB_TYPE="postgresql"
 export SOURCE_DB_NAME=${SOURCE_DB_NAME:-"pg_views"}
-export SOURCE_DB_SCHEMA="public"
+export SOURCE_DB_SCHEMA="test_case"
+export TARGET_DB_SCHEMA="TEST_Case"
--- a/migtests/tests/pg-tests/pg-views/env.sh
+++ b/migtests/tests/pg-tests/pg-views/env.sh
@@ -1,4 +1,4 @@
 export SOURCE_DB_TYPE="postgresql"
 export SOURCE_DB_NAME=${SOURCE_DB_NAME:-"pg_views"}
-export SOURCE_DB_SCHEMA="test_case"
-export TARGET_DB_SCHEMA="TEST_Case"
+export SOURCE_DB_SCHEMA="test_views"
+export TARGET_DB_SCHEMA="TEST_Views"
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -1005,26 +1005,18 @@ func setTargetSchema(conn *pgx.Conn) {
 		// No need to set schema if importing in the default schema.
 		return
 	}
-	target.Schema = strings.ToLower(target.Schema)
 	checkSchemaExistsQuery := fmt.Sprintf("SELECT count(schema_name) FROM information_schema.schemata WHERE schema_name = '%s'", target.Schema)
-	rows, err := conn.Query(context.Background(), checkSchemaExistsQuery)
-	if err != nil {
+	var cntSchemaName int
+	
+	if err := conn.QueryRow(context.Background(), checkSchemaExistsQuery).Scan(&cntSchemaName); err != nil {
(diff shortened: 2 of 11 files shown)


rahulb-yb (approved):
LGTM

migtests/tests/pg-tests/pg-views/env.sh
```diff
@@ -1,3 +1,4 @@
 export SOURCE_DB_TYPE="postgresql"
 export SOURCE_DB_NAME=${SOURCE_DB_NAME:-"pg_views"}
-export SOURCE_DB_SCHEMA="public"
+export SOURCE_DB_SCHEMA="test_case"
+export TARGET_DB_SCHEMA="TEST_Case"
```
sanyamsinghal, line 4: this is a special case, don't remove `public` schema case. We couple of cases for tests.
1. public schema(i.e. default)
2. non-public schema name
3. case sensitive non-public schema name

yb-voyager/cmd/importData.go
```diff
@@ -1005,9 +1005,26 @@ func setTargetSchema(conn *pgx.Conn) {
 		// No need to set schema if importing in the default schema.
 		return
 	}
+	target.Schema = strings.ToLower(target.Schema)
+	checkSchemaExistsQuery := fmt.Sprintf("SELECT count(schema_name) FROM information_schema.schemata WHERE schema_name = '%s'", target.Schema)
+	rows, err := conn.Query(context.Background(), checkSchemaExistsQuery)
```
sanyamsinghal, line 1010: try using `QueryRow()` function instead of `Query()`. Already used at multiple places in code
it will reduce the code

yb-voyager/cmd/importData.go
```diff
@@ -1005,9 +1005,26 @@ func setTargetSchema(conn *pgx.Conn) {
 		// No need to set schema if importing in the default schema.
 		return
 	}
+	target.Schema = strings.ToLower(target.Schema)
```
sanyamsinghal, line 1008: Is it possible to do this conversion somewhere else in the code. Where it is more intuitive, at some upper layer of the code.
This function will be called multiple times and everytime this conversion will be done.

rahulb-yb (commented):
Please change the name of the schema to something more relevant to the test itself (views in this case)

yb-voyager/cmd/importData.go
sanyamsinghal: Maybe somewhere in `init()` of the commands

migtests/tests/pg-tests/pg-views/env.sh
sanyamsinghal: yes for target-db-schema

yb-voyager/cmd/importData.go
```diff
@@ -1328,4 +1320,5 @@ func init() {
 	importCmd.AddCommand(importDataCmd)
 	registerCommonImportFlags(importDataCmd)
 	registerImportDataFlags(importDataCmd)
+	target.Schema = strings.ToLower(target.Schema)
```
sanyamsinghal, line 1323: please verify if it is required for `import data file` command also

sanyamsinghal (approved):
LGTM

migtests/tests/pg-tests/pg-views/env.sh
priyanshi-yb: Okay, added a case for 2nd one as 1st and 3rd are already added.

ChrisnaMishra (approved):
_i don't understandanything_

yb-voyager/cmd/importSchema.go
```diff
@@ -51,6 +51,7 @@ func init() {
 	importCmd.AddCommand(importSchemaCmd)
 	registerCommonImportFlags(importSchemaCmd)
 	registerImportSchemaFlags(importSchemaCmd)
+	target.Schema = strings.ToLower(target.Schema)
```
sanyamsinghal, line 54: @priyanshi-yb have you verify whether adding it here is working for `import schema` and `import data`
Because this is a parent command which we don't support now.

priyanshi-yb: I have added in specifically `importSchema` and `importData` init() functions and not in `import`'s init(). So, I believe it is fine and it working as well.

sanyamsinghal: ok

PR #608, opened 2022-11-21
Ask for password if not already provided as an argument

### LINKED ISSUE #597: yb-voyager shows password in list of processes (ps -aef | grep yb-voyager)
[Jira Link for additional details](https://yugabyte.atlassian.net/browse/DB-4221)

rahulb-yb:
[Fixes #597]
This patch removes the need to provide a source or target DB password as a command-line argument. If not provided, the user is prompted for the password before beginning a migration. This hides the password from the command `ps -aef | grep voyager`

P.S. Adding an automation test case is not applicable here, in order to make sure the user inputs do not echo to the terminal session, I used a `term` package which additionally ensures that the input coming in must be from a terminal instance; buffered input errors out with this error: `inappropriate ioctl for device`
[Additional details](https://unix.stackexchange.com/questions/382747/should-i-try-to-get-rid-of-inappropriate-ioctl-for-device-in-strace-output-for)

### DIFF
--- a/yb-voyager/cmd/export.go
+++ b/yb-voyager/cmd/export.go
@@ -143,7 +143,7 @@ func setDefaultSSLMode() {
 	}
 }
 
-func validateExportFlags() {
+func validateExportFlags(cmd *cobra.Command) {
 	validateExportDirFlag()
 	validateSourceDBType()
 	validateSourceSchema()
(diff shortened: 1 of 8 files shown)


yb-voyager/cmd/export.go
```diff
@@ -238,11 +241,24 @@ func validateOracleParams() {
...
+	fmt.Print("Password to connect to source:")
+	bytePassword, err := term.ReadPassword(int(syscall.Stdin))
+	if err != nil {
+		utils.ErrExit("read password: %v", err)
+		return
+	}
+	fmt.Print("\n")
+	source.Password = string(bytePassword)
+}
+
```
priyanshi-yb, line 257: Is this also tested with special characters in password while entering through input?

priyanshi-yb (approved):

PR #611, opened 2022-11-22
Bug #610: Fetched Parent Table already has schema name in it if its non-public, add public if there is no schema name

### LINKED ISSUE #610: [PostgreSQL] Import Data fail when Parent table and Its partition are in different schema
Schema:

```
CREATE TABLE p1.sales_region (id int, amount int, branch text, region text) PARTITION BY LIST (region);
CREATE TABLE p2.London PARTITION OF p1.sales_region FOR VALUES IN ('London');
CREATE TABLE p2.Sydney PARTITION OF p1.sales_region FOR VALUES IN ('Sydney');
CREATE TABLE p2.Boston PARTITION OF p1.sales_region FOR VALUES IN ('Boston');
```

In this case import data fails for sales_region tables and its partition.

Reason: extracting copy command with schema name is broken for this scenario.

sanyamsinghal:
[Fixes #610]
This change fixes the scenario when the table and its partitions are in a different schema by correctly fetching the parent table name and copy command for it. 
For more details about the refer to the filed bug

Tests: 
1. Updated the test `migtests/tests/pg-partitions` in the automation framework
2. Also added validations for the `pg-partitions` test case.

### DIFF
--- a/migtests/tests/pg-partitions/env.sh
+++ b/migtests/tests/pg-partitions/env.sh
@@ -1,3 +1,3 @@
 export SOURCE_DB_TYPE="postgresql"
 export SOURCE_DB_NAME=${SOURCE_DB_NAME:-"partitions"}
-export SOURCE_DB_SCHEMA="public"
+export SOURCE_DB_SCHEMA="public,p1,p2"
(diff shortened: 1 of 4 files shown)


priyanshi-yb (approved):
LGTM

PR #616, opened 2022-11-22
fixed #564 to execute all the UNIQUE INDEX DDLs before FK creation

### LINKED ISSUE #564: Import Schema dependency issue of Foreign Key on Unique Index
(This is observed during the migration from the source as PostgreSQL)

We get the following error while creating an FK during the import schema phase
```
 ERROR: there is no unique constraint matching given keys for referenced table "xyz" (SQLSTATE 42830)
```

SQL caused the above error:
```
    STATEMENT: ALTER TABLE ONLY public.abc ADD CONSTRAINT abc_fk1 FOREIGN KEY (c1) REFERENCES public.xyz(c1) ON DELETE CASCADE;
```

Error says there is no unique constraint on the`c1` column of table `xyz` so it can create a FK referencing that column.

But that UNIQUE constraint on that column is provided by the CREATE UNIQUE INDEX DDL:
```
CREATE UNIQUE INDEX xyz_ix2 ON public.xyz USING btree (c1);
```

Since we now postImportData phase, there is no way to create this index beforehand and on the other hand, all the FK are created before the postImportData phase. So FK creation is dependent on INDEX creation now, which voyager can’t handle today.

priyanshi-yb:
1. fixed #564 
2. Added a case in `migtests/tests/pg-tests/pg-constraints` for the same with validation.

### DIFF
--- a/migtests/tests/pg-tests/pg-constraints/validate
+++ b/migtests/tests/pg-tests/pg-constraints/validate
@@ -31,7 +31,7 @@ QUERIES_CHECK = {
 		'query': "insert into check_test (id, first_name, last_name, age) values (7, 'Tom', 'Stewan', 15);",
 		'code': 23514
 	},
-	'FORIEGN_CHECK': {
+	'FOREIGN_CHECK': {
 		'query': "insert into foreign_test values (5,1,7);",
 		'code': 23503
 	},
@@ -39,7 +39,7 @@ QUERIES_CHECK = {
 		'query': "insert into xyz values(3);",
 		'code': 23505
 	},
-	'FORIEGN_CHECK_WITH_UNIQUE_INDEX': {
+	'FOREIGN_CHECK_WITH_UNIQUE_INDEX': {
 		'query': "insert into abc values ('dsfs', 4);",
 		'code': 23503
 	},
--- a/migtests/tests/pg-tests/pg-constraints/validate
+++ b/migtests/tests/pg-tests/pg-constraints/validate
@@ -14,7 +14,9 @@ EXPECTED_ROW_COUNT = {
 	'foreign_test': 4,         
 	'not_null_check': 6,
 	'primary_test': 6,  
-	'unique_test': 5
+	'unique_test': 5,
+	'xyz':3,
+	'abc':3
 }
 QUERIES_CHECK = {
 	'NULL_CHECK': {
@@ -32,13 +34,22 @@ QUERIES_CHECK = {
 	'FORIEGN_CHECK': {
 		'query': "insert into foreign_test values (5,1,7);",
 		'code': 23503
-	}
+	},
+	'UNIQUE_INDEX': {
+		'query': "insert into xyz values(3);",
+		'code': 23505
+	},
+	'FORIEGN_CHECK_WITH_UNIQUE_INDEX': {
+		'query': "insert into abc values ('dsfs', 4);",
+		'code': 23503
+	},
+
 }	
 
 def migration_completed_checks(tgt):
 	table_list = tgt.get_table_names("public")
 	print("table_list:", table_list)
-	assert len(table_list) == 6
+	assert len(table_list) == 8
 
 
 	got_row_count = tgt.row_count_of_all_tables("public")
--- a/yb-voyager/cmd/importSchema.go
+++ b/yb-voyager/cmd/importSchema.go
@@ -90,30 +90,38 @@ func importSchema() {
 	// Import ALTER TABLE statements from sequence.sql only after importing everything else
 	isAlterStatement := func(objType, stmt string) bool {
 		stmt = strings.ToUpper(strings.TrimSpace(stmt))
-		return (objType == "SEQUENCE" || objType == "TABLE") && strings.HasPrefix(stmt, "ALTER TABLE")
+		if objType == "SEQUENCE" && strings.HasPrefix(stmt, "ALTER TABLE"){
+			return true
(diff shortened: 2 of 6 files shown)


yb-voyager/cmd/importSchema.go
```diff
@@ -90,23 +90,40 @@ func importSchema() {
 	// Import ALTER TABLE statements from sequence.sql only after importing everything else
 	isAlterStatement := func(objType, stmt string) bool {
```
priyanshi-yb, line 91: I think we can change this `isAlterStatement` to `isSkipStatement` to make it general. Will do.

migtests/tests/pg-tests/pg-constraints/validate
```diff
@@ -32,13 +34,22 @@ QUERIES_CHECK = {
 	'FORIEGN_CHECK': {
 		'query': "insert into foreign_test values (5,1,7);",
 		'code': 23503
-	}
+	},
+	'UNIQUE_INDEX': {
+		'query': "insert into xyz values(3);",
+		'code': 23505
+	},
+	'FORIEGN_CHECK_WITH_UNIQUE_INDEX': {
```
sanyamsinghal, line 42: nit: `FOREIGN`

yb-voyager/cmd/importSchema.go
```diff
@@ -90,23 +90,40 @@ func importSchema() {
...
 		stmt = strings.ToUpper(strings.TrimSpace(stmt))
-		return objType == "SEQUENCE" && strings.HasPrefix(stmt, "ALTER TABLE")
+		if objType == "SEQUENCE" && strings.HasPrefix(stmt, "ALTER TABLE"){
+			return true
+		} else if objType == "TABLE" && strings.Contains(stmt, "ALTER TABLE") && strings.Contains(stmt, "FOREIGN KEY"){
+			return true
+		} else if objType == "UNIQUE INDEX" && !strings.Contains(stmt, objType){
```
sanyamsinghal, line 97: are both of these if conditions at line - 97 and 99 required?
Or these can be merged into one?

yb-voyager/cmd/importSchema.go
sanyamsinghal: or we can break it into multiple specific ones like `isALTERStatemet` , `createUniqueIndex` etc...

yb-voyager/cmd/importSchema.go
priyanshi-yb: I am not sure if these can be merged, as one is to only run UNIQUE INDEX DDLs for objType UNIQUE INDEX and the other one is to skip only UNIQUE INDEX DDLs at the time of INDEX objType.

yb-voyager/cmd/importSchema.go
priyanshi-yb: That is something I was also thinking but then as we pass only one skipFn to importSchemaInternal, we have to pass multiple functions to it I guess.

yb-voyager/cmd/importSchema.go
```diff
@@ -90,21 +90,30 @@ func importSchema() {
...
+		return (objType == "SEQUENCE" || objType == "TABLE") && strings.HasPrefix(stmt, "ALTER TABLE")
 	}
 
 	skipFn := isAlterStatement
 	importSchemaInternal(exportDir, objectList, skipFn)
 
-	// Import the skipped ALTER TABLE statements from sequence.sql if it exists
-	if slices.Contains(objectList, "SEQUENCE") {
+	// Import the skipped ALTER TABLE statements from sequence.sql and table.sql if it exists
+	if slices.Contains(objectList, "SEQUENCE") || slices.Contains(objectList, "TABLE") {
```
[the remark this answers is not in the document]
priyanshi-yb: Done already

yb-voyager/cmd/importSchema.go
sanyamsinghal: yeah, ok, maybe you can multiple conditions named separately and then add and pass them in one skipFn

yb-voyager/cmd/importSchema.go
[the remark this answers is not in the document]
sanyamsinghal: yeah, it was an yesterday's comment which wasn't pushed

yb-voyager/cmd/importSchema.go
priyanshi-yb: Yes, that can be done.

yb-voyager/cmd/importSchema.go
```diff
@@ -88,25 +88,45 @@ func importSchema() {
...
+		return objType == "UNIQUE INDEX" && !strings.Contains(stmt, objType)
+	}
+	isUniqueIndexInIndex := func(objType, stmt string) bool{
+		return objType == "INDEX" && strings.Contains(stmt, "UNIQUE INDEX")
+	}
+	isSkipStatement := func(objType, stmt string) bool {
+		stmt = strings.ToUpper(strings.TrimSpace(stmt))
+		return isAlterTableInSequence(objType, stmt) || isAlterTableFKInTable(objType, stmt) || 
+			isNotUniqueIndexInUniqueIndex(objType, stmt) || isUniqueIndexInIndex(objType, stmt) 
+	}
```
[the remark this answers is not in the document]
priyanshi-yb: Okay, will make the change.

amit-yb (approved):
LGTM.

sanyamsinghal (approved):

PR #631, opened 2022-11-29
Fixed #618 to check all the schemas passed in the --source-db-schema …

### LINKED ISSUE #618: No check for schema list provided in `source-db-schema` flag during export schema
During `export data` it throws an error saying: `Following schemas are not present in source database [p1 p2], please provide a valid schema list.`

But during the `export schema` phase there is no check, it is just a no-op and only the valid schemas are exported.

priyanshi-yb:
…exists

Issue - #618 
Fix - we already have the check while Exporting data to check all the schemas exists or not, refactored the code to make a function for that part of code as `checkSchemasExists()` and called it in both export data and the export schema commands.
Test case -  N/A

### DIFF
--- a/yb-voyager/src/srcdb/postgres.go
+++ b/yb-voyager/src/srcdb/postgres.go
@@ -80,7 +80,8 @@ func (pg *PostgreSQL) GetVersion() string {
 	}
 	return version
 }
-func checkSchemasExists(pg *PostgreSQL) []string{
+
+func (pg *PostgreSQL) checkSchemasExists() []string{
 	list := strings.Split(pg.source.Schema, "|")
 	var trimmedList []string
 	for _, schema := range list {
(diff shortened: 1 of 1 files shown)


yb-voyager/src/srcdb/postgres.go
```diff
@@ -80,8 +80,7 @@ func (pg *PostgreSQL) GetVersion() string {
 	}
 	return version
 }
-
-func (pg *PostgreSQL) GetAllTableNames() []string {
+func checkSchemasExists(pg *PostgreSQL) []string{
```
amit-yb, line 83: Can't we make this as a method of `PostgreSQL` type?

Leave one empty line above this function definition.

amit-yb (approved):
LGTM with a minor comment.

PR #633, opened 2022-11-29
Refactor deferred statements to a separate function

rahulb-yb:
This patch refactors the deferral of execution of sql files relating to corner cases into a separate function to allow for read-friendly accumulation of such cases in the future.


yb-voyager/cmd/importSchema.go
```diff
@@ -90,26 +95,10 @@ func importSchema() {
...
-	// Import the skipped ALTER TABLE statements from sequence.sql if it exists
-	if slices.Contains(objectList, "SEQUENCE") {
-		skipFn = func(objType, stmt string) bool {
-			return !isAlterStatement(objType, stmt)
-		}
-		sequenceFilePath := utils.GetObjectFilePath(filepath.Join(exportDir, "schema"), "SEQUENCE")
-		if utils.FileOrFolderExists(sequenceFilePath) {
-			fmt.Printf("\nImporting ALTER TABLE DDLs from %q\n\n", sequenceFilePath)
-			executeSqlFile(sequenceFilePath, "SEQUENCE", skipFn)
-		}
```
amit-yb, line 110: This is the problematic part. We should use the `importSchemaInternal()` instead of copying its behaviour here.

yb-voyager/cmd/importSchema.go
```diff
@@ -92,43 +91,34 @@ func importSchema() {
...
+		importSchemaInternal(exportDir, []string{"SEQUENCE"}, skipFn)
 	}
 	if slices.Contains(objectList, "TABLE") {
-		tableFilePath := utils.GetObjectFilePath(filepath.Join(exportDir, "schema"), "TABLE")
-		if utils.FileOrFolderExists(tableFilePath) {
-			fmt.Printf("\nImporting FOREIGN KEY DDLs from %q\n\n", tableFilePath)
-			executeSqlFile(tableFilePath, "TABLE", skipFn)
-		}
+		importSchemaInternal(exportDir, []string{"TABLE"}, skipFn)
```
amit-yb, line 121: What about the INDEX and UNIQUE INDEX statements that we skipped in the first pass?

amit-yb (approved):
LGTM.

yb-voyager/cmd/importSchema.go
priyanshi-yb: INDEX statements are already being executed in PostImportData phase, and the UNIQUE INDEX which we skip while importing Indexes are already imported before.

Issue #632, opened 2022-11-29
uncategorized.sql should be merged in schema/sequence


tichimura:
# Problem Description

Schema file `uncategorized.sql` which `export schema` generated should be merged in schema/sequence

# Issues

```
tichimura@ ~ % yb-voyager version
VERSION=0.81
tichimura@ ~ % yb-voyager export schema --export-dir ${EXPORT_DIR} \
        --source-db-type ${SOURCE_DB_TYPE} \
        --source-db-host ${SOURCE_DB_HOST} \
        --source-db-name voyager \
        --source-db-port 5432 \
        --source-db-user tichimura \
        --source-db-password '' --source-db-schema public --start-clean
export of schema for source type as 'postgresql'
postgresql version: 11.17
exporting the schema          |
Some uncategorized sql statements are present in "/Users/tichimura/Downloads/20221129_YB/schema/uncategorized.sql", Needs to review and import them manually!!

Exported schema files created under directory: /Users/tichimura/Downloads/20221129_YB/schema
```

The sample outputs of `uncategorized.sql` is like below, which includes 50 of `ALTER SEQUENCE` lines

```
-- setting variables for current session
SET statement_timeout = 0;
SET lock_timeout = 0;
SET idle_in_transaction_session_timeout = 0;
SET client_encoding = 'UTF8';
SET standard_conforming_strings = on;
SELECT pg_catalog.set_config('search_path', '', false);
SET check_function_bodies = false;
SET xmloption = content;
SET client_min_messages = warning;
SET row_security = off;


ALTER SEQUENCE public.xxxx_id_seq OWNED BY public.xxx_id;

<snip> 

```

# Expectation

without generating `uncategorized.sql` the `ALTER SEQUENCE` should be included in `schema/sequences/sequence.sql`

# Workaround

manually add `uncategorized.sql` via ysqlsh

PR #638, opened 2022-12-01
Bug #632: Storing uncategorized ALTER SEQUENCE DDLs in schema/sequence.sql file

sanyamsinghal:
For example: 
```
ALTER SEQUENCE public.users_user_id_seq OWNED BY public.users.user_id;
```

These DDLs should be a part of `sequence.sql` ideally, but they should run after the TABLE creation.
So, now we will run `ALTER SEQUENCE` stmts at the end of the import schema after everything is done along with `ALTER TABLE` statements

Testing: Added a commit, to check the fetched owner with the expected owner of the sequences


priyanshi-yb (approved):
LGTM

PR #642, opened 2022-12-01
Fixes #640 Upper/Mixed case issue with target-db-schema

### LINKED ISSUE #640: Mixed/Upper case values in target-db-schema behave different for import schema and data
When we use the flag `--target-db-schema` while import using Mixed/Upper case values, it's behaviour is different for `import-schema ` and different for `import-data`.

This was tested as a part of 0.82 verification.

If we pass some mixed case values in import schema, it considers it as lower case.
If we pass some mixed case values in import data, it matches the case provided in the command.

example:

Running import schema with `--target-db-schema XYZ` creates a schema `xyz` on the target

```
[centos@ip-10-9-6-84 ~]$ yb-voyager import schema --export-dir ${EXPORT_DIR}         --target-db-host ${TARGET_DB_HOST}         --target-db-user ${TARGET_DB_USER}                 --target-db-name ${TARGET_DB_NAME} --start-clean --yes --target-db-schema XYZ

Password to connect to target:

YugabyteDB version: 11.2-YB-2.15.3.0-b0
schemas to be present in target database "automation": [xyz]
dropping schema 'xyz' in target database
creating schema 'xyz' in target database...

```

Then running import data with the same `--target-db-schema XYZ` throws error saying that schema `XYZ` doesn't exist

```
[centos@ip-10-9-6-84 ~]$ yb-voyager import data --export-dir ${EXPORT_DIR}         --target-db-host ${TARGET_DB_HOST}         --target-db-user ${TARGET_DB_USER}                 --target-db-name ${TARGET_DB_NAME} --start-clean --yes --target-db-schema XYZ

Password to connect to target:

import of data in "automation" database started
Target YugabyteDB version: 11.2-YB-2.15.3.0-b0
Using 3 parallel jobs by default. Use --parallel-jobs to specify a custom value
schema 'XYZ' does not exist in target

```

priyanshi-yb:
Fixes - #640 
Issue - When schema name in target-db-schema is passed in Mixed/Upper case throwing error schema not exist, bcz schema was not being converted into lower.

Fix - changed the line of converting the target-db-schema in lower to main functions of commands, which executed first when command start its working.

Test cases - Added cases in MySQL for Mixed and Upper both.

### DIFF
--- a/migtests/tests/mysql-tests/mysql-constraints/env.sh
+++ b/migtests/tests/mysql-tests/mysql-constraints/env.sh
@@ -1,2 +1,3 @@
 export SOURCE_DB_TYPE="mysql"
 export SOURCE_DB_NAME=${SOURCE_DB_NAME:-"mysql_constraints"}
+export TARGET_DB_SCHEMA="Test_Mysql_Constraints"
--- a/migtests/tests/mysql-tests/mysql-constraints/validate
+++ b/migtests/tests/mysql-tests/mysql-constraints/validate
@@ -18,30 +18,30 @@ EXPECTED_ROW_COUNT = {
 }
 QUERIES_CHECK = {
 	'NULL_CHECK': {
-		'query': "insert into not_null_check(id, first_name, last_name, email, gender, ip_address) values (7, NULL, 'something', 'msomething@java.com', 'Female', '230.188.87.42');",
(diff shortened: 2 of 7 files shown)


amit-yb (approved):
LGTM.

PR #659, opened 2022-12-06
Import data file tests for csv and tsv delimited files

chetank-yb:


migtests/tests/import-txt-file-test/run-import-file-test
```diff
@@ -0,0 +1,74 @@
...
+	step "Create target database."
+	run_ysql yugabyte "DROP DATABASE IF EXISTS ${TARGET_DB_NAME};"
+	run_ysql yugabyte "CREATE DATABASE ${TARGET_DB_NAME}"
+
+	step "Create target table."
+	run_ysql ${TARGET_DB_NAME} "${CMD}"
+
+	step "Import data file"
+	import_data_file --data-dir ${TEST_DIR} --format text --delimiter '\t' \
+		--file-table-map "${DATA_FILE_NAME}:${TARGET_TABLE_NAME}"
```
amit-yb, line 68: Can you consider using as many options of `import data file` as possible?

migtests/tests/import-csv-file-test/run-import-file-test
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
```
amit-yb, line 1: Can't we just add multiple invocations of `import data file` in the existing test case itself?

migtests/tests/import-txt-file-test/SMSA.txt
```diff
@@ -0,0 +1,60 @@
+Akron	OH	27	71	59	36	921.87	11.4	3243	8.8	42.6	660328	3.34	29560	21	15	59	15
```
amit-yb, line 1: Consider using delimiter other than tab (Something like a `|` character).
Add some rows where the delimiter itself appears as a part of a string and thus you will have to escape it.

PR #664, opened 2022-12-08
[PostgreSQL] Moved import of RULE after TABLE and VIEW

sanyamsinghal:
- RULE can be defined on tables and views only, so needs to be imported later.
- Added automation tests for rules in 'pg-tests/pg-views-and-rules' test case


amit-yb (approved):
LGTM.

PR #667, opened 2022-12-09
Added test cases to automation for import data file command

sanyamsinghal:
- test1 is importing multiple table data files and test2 is text file format with --has-header flag


amit-yb (approved):
LGTM 👍 . Thank you for adding this in the automation.

Issue #585, opened 2022-11-11
Import Data failed if LB is present and cluster servers host is not resolvable


sanyamsinghal:
If there is a Load Balancer is present and yb servers IP is not resolvable from voyager machine, then import data command is erroring out. 
Ideally it should use the load balancer for importing the data.

More details of the issue mentioned in this ticket - https://github.com/yugabyte/yb-voyager/issues/553

PR #675, opened 2022-12-13
Fixed Bug #585 and enhances the behaviour to use reachable yb servers

sanyamsinghal:
First commit: fixes bug #585
Second commit: change the import data behavior to use the reachable servers instead of using all and failing, if not all.

Testing:
1. Tested a scenario where the output of the yb_servers() gives the DNS of servers which is not resolvable from the client machine, so the provided `target-db-host` is just used for importData
2. Cluster on YB Cloud where LB is present
3. Normal cluster without a load balancer

### DIFF
--- a/yb-voyager/src/utils/utils.go
+++ b/yb-voyager/src/utils/utils.go
@@ -290,7 +290,8 @@ func LookupIP(name string) []string {
 
 	ips, err := net.LookupIP(name)
 	if err != nil {
-		ErrExit("Error Resolving name=%s: %v", name, err)
+		log.Infof("Error Resolving name=%s: %v", name, err)
+		return []string{name}
 	}
 
 	for _, ip := range ips {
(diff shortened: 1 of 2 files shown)


rahulb-yb (approved):
lgtm

yb-voyager/src/utils/utils.go
```diff
@@ -290,7 +290,8 @@ func LookupIP(name string) []string {
 
 	ips, err := net.LookupIP(name)
 	if err != nil {
-		ErrExit("Error Resolving name=%s: %v", name, err)
+		log.Infof("Error Resolving name=%s: %v", name, err)
+		return []string{name}
```
amit-yb, line 294: This is sneaky! Return an empty slice if the name can't be resolved. But do not play with the semantics of a function. In this case, the callers (or at least someone who is reading the code) would expect that the `LookupIP()` is going to return a list of IP addresses. Please, fix this before merging.

sanyamsinghal: ok, will fix it before merging

sanyamsinghal: done

PR #677, opened 2022-12-14
Updated the ora2pg hash for dummy run

shubham-yb:
Created a dummy pull request to run our automation tests against the latest ora2pg hash.


shubham-yb:
We have already updated our Ora2pg hash in the main. So there's no need for this pull request.

Issue #673, opened 2022-12-12
[ORACLE] Error in CREATE VIEW DDL in synonym.sql, includes full name of view/table with schema same as in source


priyanshi-yb:
While exporting SYNONYMs from Oracle, it exports the `CREATE OR REPLACE VIEW` DDLs with full name of view with schema_name included and that schema will be same as the schema in which the synonym is present in Oracle sourceDB, but the schema with that name is not present in target and hence import schema fails with error of `does not exist`.

Example- 
```
CREATE OR REPLACE VIEW sakila_demo.offices AS SELECT * FROM sakila_demo.locations;
```
Error in import schema - 
```
ERROR: relation "sakila_demo.locations" does not exist (SQLSTATE 42P01)
```

as `locations` table is imported in public schema.

priyanshi-yb:
There are these two options to decide between on one for the solutions - 

- Adding this case in analyze-schema and leaving it on user to do the needful manually like removing the schema name from the DDL or create the schema/required objects in the respective schema in the target.
- Voyager explicitly remove the schema name from the DDLs at the end of export-schema phase.

priyanshi-yb:
fixed using option-2 in this PR #701

PR #682, opened 2022-12-16
Added the workarounds for Synonym export from Oracle in analyze-schema in Details section

priyanshi-yb:
Added more details for #673 in analyze-schema in Details section of `SYNONYM` object. Here is the screenshot for that in the analyze report - 

Screenshot 2022-12-16 at 4 00 10 PM

### DIFF
--- a/yb-voyager/cmd/analyzeSchema.go
+++ b/yb-voyager/cmd/analyzeSchema.go
@@ -672,7 +672,7 @@ func initializeSummaryMap() {
 		if objType == "PACKAGE" {
 			summaryMap[objType].details["Packages in oracle are exported as schema, please review and edit them(if needed) to match your requirements"] = true
 		} else if objType == "SYNONYM" {
-			summaryMap[objType].details["Synonyms in oracle are exported as view, please review and edit them(if needed) to match your requirements"] = true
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/analyzeSchema.go
```diff
@@ -672,7 +672,7 @@ func initializeSummaryMap() {
 		if objType == "PACKAGE" {
 			summaryMap[objType].details["Packages in oracle are exported as schema, please review and edit them(if needed) to match your requirements"] = true
 		} else if objType == "SYNONYM" {
-			summaryMap[objType].details["Synonyms in oracle are exported as view, please review and edit them(if needed) to match your requirements"] = true
+			summaryMap[objType].details["\nSynonyms in oracle are exported as view with full classified name of the object which may have source schema name with object name, which will not be present by default and required to be present in the target database. \nHere are two workarounds, modify the DDLs as per requirement using any of these: \n- Create the schema mentioned in object names and the objects required in that Schema used in those DDLs on the target database.\n- Remove the schema name from all the object names from the DDLs and then it will execute that DDL in the respective target-db-schema passed."] = true
```
sanyamsinghal, line 675: summaryMap[objType].details["\nSynonyms in oracle are exported as views with fully qualified name of the object which may have source schema name with object name, which will not be present by default and required to be created in the target database. \nHere are two workarounds, modify the DDLs as per requirement using any of these: \n- Create the schema mentioned in object names and the objects required in that Schema used in those DDLs on the target database.\n- Remove the schema name from all the object names from the DDLs and then it will execute that DDL in the provided target-db-schema(default public)."] = true

sanyamsinghal (approved):
```

```

priyanshi-yb:
This PR is no more required as it is fixed in this PR #701

PR #693, opened 2022-12-20
Accept file names containing whitespaces within import directives

rahulb-yb:
When import directives (`\i 'filename.sql'`) are such that the filename contains whitespace characters, we would only take the first token, separated by whitespaces, as the entire file name while parsing it internally. This patch now regards the entire contents of the string between the two quotation marks `'` ahead of the directive `\i` as the file name to be considered.

This patch also adds a case to our automation test framework for the same.

### DIFF
--- a/migtests/scripts/run-test.sh
+++ b/migtests/scripts/run-test.sh
@@ -48,7 +48,7 @@ main() {
 
 	step "Export schema."
 	export_schema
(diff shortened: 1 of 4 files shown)


amit-yb (approved):
LGTM. Thanks for this quick fix!

PR #695, opened 2022-12-21
Added the Dockerfile to create the ubuntu image with yb-voyager installed and the yb-voyager-docker shell script to run the yb-voyager commands inside docker.

ShivanshGahlot:
The files have been added in a new folder '`docker`' in the yb-voyager repo


docker/yb-voyager-docker
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
```
amit-yb, line 1: We can simplify it a bit:
```
#!/bin/sh

export_dir=""
args=( "$@" )
i=0
while [ $i -lt $# ]
do
        w=${args[$i]}
        if [ "${w}" == '--export-dir' ]
        then
                j=$(( $i + 1))
                export_dir=${args[$j]}
                args[$j]='/export-dir'
        fi
        i=$(( $i + 1))
done
if [ -z ${export_dir} ]
then
        echo "error: required argument 'export-dir' not set"
        exit 1
fi
cmd="docker run -v ${export_dir}:/export-dir -it --rm --privileged --name voyager yb-voyager:latest yb-voyager ${args[*]}"
echo "$cmd"
```

docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM amd64/ubuntu:22.04
+
+RUN apt-get update && apt-get upgrade -y && apt-get upgrade -y binutils && (1 | apt-get upgrade -y tzdata) && apt-get install -y wget && wget https://s3.us-west-2.amazonaws.com/downloads.yugabyte.com/repos/reporpms/yb-apt-repo_1.0.0_all.deb && apt-get install -y ./yb-apt-repo_1.0.0_all.deb && apt-get clean && apt-get update && apt-get install -y yb-voyager && rm -rf /yb-apt-repo_1.0.0_all.deb && apt-get remove -y wget && apt-get clean -y
```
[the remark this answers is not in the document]
ShivanshGahlot: Removed the upgrade command

`apt` tries to upgrade` tzdata `while installing `yb-voyager`. At that time it asks for prompts to select the timezone. Due to this the build fails. That's why I explicitly upgrade` tzdata `and feed it 1 to select the timezone.

apt-get update needs to be the first step as well. Otherwise it wont be able to find the wget package.

docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
...
+
+CMD ["yb-voyager", "version"]
```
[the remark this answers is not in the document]
ShivanshGahlot: Done

docker/yb-voyager-docker
ShivanshGahlot: This will give error even while running `--help` for commands since we don't specify `export-dir` for them.

docker/yb-voyager-docker
```diff
@@ -0,0 +1,21 @@
...
+        args[$j]='/export-dir'
+    fi
+    i=$(( $i + 1))
+done
+
+dockerCmd="docker run ${vol} -it --network=host --rm --privileged localhost/yb-voyager:0.82 yb-voyager ${cmdToRun}"
+
+echo $dockerCmd
+
+$dockerCmd
```
ShivanshGahlot, line 21: Simplified the script using the code @amit-yb suggested

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM amd64/ubuntu:22.04
+
+RUN apt-get update && \
+    apt-get install -y wget && \
+    wget https://s3.us-west-2.amazonaws.com/downloads.yugabyte.com/repos/reporpms/yb-apt-repo_1.0.0_all.deb && \
+    apt-get install -y ./yb-apt-repo_1.0.0_all.deb && \ 
+    apt-get clean && \
```
amit-yb, line 7: Can we get rid of this `apt-get clean` instance? We already have another `apt-get clean` towards the end.

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM amd64/ubuntu:22.04
+
+RUN apt-get update && \
+    apt-get install -y wget && \
```
amit-yb, line 4: Doesn't the base ubuntu image already contain `wget`?

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    wget https://s3.us-west-2.amazonaws.com/downloads.yugabyte.com/repos/reporpms/yb-apt-repo_1.0.0_all.deb && \
+    apt-get install -y ./yb-apt-repo_1.0.0_all.deb && \ 
+    apt-get clean && \
+    apt-get update && \
+    apt-get upgrade -y binutils && \
+    (1 | apt-get upgrade -y tzdata) && \
```
amit-yb, line 10: Why do we need `tzdata`?

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    apt-get install -y yb-voyager && \
+    rm -rf /yb-apt-repo_1.0.0_all.deb && \
+    apt-get remove -y wget && \
```
amit-yb, line 13: Leave the `wget` packaged in the container. I don't think it will take a lot of space.

docker/yb-voyager-docker
```diff
@@ -0,0 +1,21 @@
...
+    if [[ ${argv[${i}]} == "--export-dir" ]]
+    then
+        j=$(( $i + 1))
+        vol="-v ${argv[$j]}:/export-dir"
+        argv[$j]='/export-dir'
+    fi
+    i=$(( $i + 1))
+done
+
+dockerCmd="docker run ${vol} -it --network=host --rm --privileged localhost/yb-voyager:0.82 yb-voyager ${argv[*]}"
```
amit-yb, line 17: Use `latest` tag instead of `0.82`. We don't want to update this script after every release.
Why do we need the `--network=host`?

docker/yb-voyager-docker
```diff
@@ -0,0 +1,21 @@
...
+
+echo $dockerCmd
```
amit-yb, line 19: Comment out this line. It is okay for debugging. But we don't want to show it as part of normal output.

docker/yb-voyager-docker
ShivanshGahlot: Used the latest tag

amit-yb (approved):
Excellent! Good work! 👍

PR #715, opened 2023-01-02
[Fixes #654] Enable running multiple ora2pg instances concurrently.

### LINKED ISSUE #654: [Oracle] Concurrent export fails
When we try parallel export in Oracle, the export which finishes first, throws the error:

```
Data export failed: exit status 2

WARNING: an error occurs during data export. Please check what's happen.
```

The second export goes through successfully.

Note: `export data status` says done for both and the files with the row counts match.

This error comes up **using just Ora2pg** as well.

Steps to repro:

1. Create an Oracle schema with >=2 tables large enough for parallel export (you may have same number of rows so that comparison is easier)
2. Create two different export-dir 
3. Run export data in parallel for the same schema using different export dir and different tables using table-list

PS. All this was from a 2 parallel export perspective.

amit-yb:
The shared `/tmp` dir was not allowing multiple instances to run concurrently.

This patch creates a `export-dir/temp/ora2pg_temp_dir` and passes it to the ora2pg using its `-T` option. This makes each ora2pg to have its own temp dir.

### DIFF
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -201,7 +201,11 @@ func printExportedRowCount(exportedRowCount map[string]int64) {
 // setup a project having subdirs for various database objects IF NOT EXISTS
 func CreateMigrationProjectIfNotExists(dbType string, exportDir string) {
 	// TODO: add a check/prompt if any directories apart from required ones are present in export-dir
-	var projectSubdirs = []string{"schema", "data", "reports", "metainfo", "metainfo/data", "metainfo/schema", "metainfo/flags", "temp"}
+	var projectSubdirs = []string{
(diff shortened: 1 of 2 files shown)


amit-yb:
Here is how the ora2pg invocation looks like:
```
2023-01-02 17:20:56 INFO ora2pg_export_data.go:75 Executing command: ora2pg -q -t COPY -P 4 -o data.sql -b /home/centos/tmp/export-dir-2/data -c /home/centos/tmp/export-dir-2/temp/.ora2pg.conf --no_header -T /home/centos/tmp/export-dir-2/temp/ora2pg_temp_dir
```

rahulb-yb (approved):
LGTM 👍🏽

PR #733, opened 2023-01-16
Added default value in cases where we silently error out while obtain…

rahulb-yb:
…ing core count

When the target is running on a MacOS machine, we are unable to determine to core count for default parallel jobs during import data. This patch adds another failsafe for such scenarios.

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -239,6 +239,9 @@ func fetchDefaultParllelJobs(targets []*tgtdb.Target) int {
 		}
 		totalCores += targetCores
 	}
+	if totalCores == 0 { //if target is running on MacOS, we are unable to determine totalCores
+		return 3
+	}
 	return totalCores / 2
 }
 


priyanshi-yb (approved):
LGTM

yb-voyager/cmd/importData.go
```diff
@@ -239,6 +239,9 @@ func fetchDefaultParllelJobs(targets []*tgtdb.Target) int {
 		}
 		totalCores += targetCores
 	}
+	if totalCores == 0 { //if target is running on MacOS, we are unable to determine totalCores
+		return 3
```
sanyamsinghal, line 243: Instead of a fixed constant `3`, can we have a better default like `2 * number of nodes` ?

rahulb-yb: I believe we had decided during the standup meeting to keep it as a constant, and to keep it as 3- this case only arises when the target is running on MacOS (which is only possible when the user is running a local installation of YugabyteDB)

Issue #740, opened 2023-01-18
[ORACLE] Nested materialized view data not exported


chetank-yb:
**Steps** : Create a nested materialized view on the source DB and perform the export & import or schema & data.
**Expected** : Nested materialized view data should be migrated.
**Actual** : Nested materialized view data not migrated. 

**Schema used for testing** : https://github.com/yugabyte/yb-voyager/blob/oracle_mview_auto/migtests/tests/oracle-tests/ora-mview-tests/mv_fast_refreshable_schema_data.sql

Actual number of rows in mv on source : 3105
Number of rows in mv on target : 0

Issue discussed with @sanyamsinghal

PR #741, opened 2023-01-19
Added a note in post-import-data and refresh-mviews phase for nested mviews which are not refreshed

priyanshi-yb:
Issue - #740 
This PR is for adding a note by detecting the mviews which are not refreshed for users to refresh them manually. 
This issue will not be there with PostgreSQL as pg-dump dumps the mviews in correct order of their dependency. Added a test case for that.


amit-yb (approved):
LGTM.

sanyamsinghal (approved):

PR #745, opened 2023-01-19
Fixed parallel-jobs for loadBalancer case to take the one passed in the command

### LINKED ISSUE #744: [Load Balancer] --parallel-jobs is broken
Even if we provide a different value for the flag `--parallel-jobs` in case of load balancer, the default value of 6 is used.

Image

Image

priyanshi-yb:
Fixes #744

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -197,8 +197,10 @@ func getYBServers() []*tgtdb.Target {
 
 	if loadBalancerUsed { // if load balancer is used no need to check direct connectivity
 		utils.PrintAndLog(LB_WARN_MSG)
-		parallelImportJobs = 2 * len(targets)
-		utils.PrintAndLog("Using %d parallel jobs by default. Use --parallel-jobs to specify a custom value", parallelImportJobs)
+		if parallelImportJobs == -1 {
+			parallelImportJobs = 2 * len(targets)
+			utils.PrintAndLog("Using %d parallel jobs by default. Use --parallel-jobs to specify a custom value", parallelImportJobs)
+		}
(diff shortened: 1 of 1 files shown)


amit-yb (approved):
LGTM. Please test the fix before merging it.

yb-voyager/cmd/importData.go
```diff
@@ -197,8 +197,10 @@ func getYBServers() []*tgtdb.Target {
 
 	if loadBalancerUsed { // if load balancer is used no need to check direct connectivity
 		utils.PrintAndLog(LB_WARN_MSG)
-		parallelImportJobs = 2 * len(targets)
-		utils.PrintAndLog("Using %d parallel jobs by default. Use --parallel-jobs to specify a custom value", parallelImportJobs)
+		if parallelImportJobs == -1 {
+			parallelImportJobs = 2 * len(targets)
+			utils.PrintAndLog("Using %d parallel jobs by default. Use --parallel-jobs to specify a custom value", parallelImportJobs)
```
sanyamsinghal, line 202: Nit: Remove this print line from here
Also remove similar line from line 328
Add this line just once at line 330

sanyamsinghal (approved):
LGTM, provided a minor comment

yb-voyager/cmd/importData.go
priyanshi-yb: this print should be only in the case of the default parallel-jobs where user is not using --parallel-jobs which is this condition - (parallelImportJobs == -1), how can we remove it from both the places add at some other?

priyanshi-yb: resolved, over discussion.

PR #752, opened 2023-01-23
Reusing the same memory allocated for reading each csv record by setting reader.ReuseRecord = true

sanyamsinghal:


amit-yb (approved):
LGTM. Thanks for this change.

PR #756, opened 2023-01-25
Simplify ora2pg conf generation by using templates.

amit-yb:


sanyamsinghal (commented):
I have one concern
Flags like: FILE_PER_TABLE, PREFIX_PARTITION are the kinds of flags that are required for the successful execution of migration. 
Previously they were hardcoded, i.e. even the changes in Base Conf file will not have any effect on them but now they can be changed via Base Conf file

sanyamsinghal (approved):
LGTM, thanks for the change

PR #758, opened 2023-01-30
Added conditional statements to download the ora2pg conf file either from the latest release commit hash or from the current yb-voyager depending on version to install

ShivanshGahlot:
If the version is 'latest' then the conf file from the latest release hash will be downloaded.
If the version is 'local' then the conf file from the current yb-voyager directory will be copied to the required location.

### DIFF
--- a/installer_scripts/install-yb-voyager
+++ b/installer_scripts/install-yb-voyager
@@ -389,7 +389,14 @@ create_base_ora2pg_conf_file() {
 		output "Overwriting ${conf_file_name}. Existing contents are backed up in ${backup_file_name}"
 	fi
 	output "Installing the latest base-ora2pg.conf"
-	sudo wget -nv -O $conf_file_name https://github.com/yugabyte/yb-voyager/raw/main/yb-voyager/src/srcdb/data/sample-ora2pg.conf
+
+	if [ "${VERSION}" == "latest" ]
+	then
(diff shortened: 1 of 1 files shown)


installer_scripts/install-yb-voyager
```diff
@@ -389,7 +389,13 @@ create_base_ora2pg_conf_file() {
 		output "Overwriting ${conf_file_name}. Existing contents are backed up in ${backup_file_name}"
 	fi
 	output "Installing the latest base-ora2pg.conf"
-	sudo wget -nv -O $conf_file_name https://github.com/yugabyte/yb-voyager/raw/main/yb-voyager/src/srcdb/data/sample-ora2pg.conf
+
+	if [ "${VERSION}" == "latest" ]
+	then
+		sudo wget -nv -O $conf_file_name https://github.com/yugabyte/yb-voyager/raw/$YB_VOYAGER_GIT_HASH/yb-voyager/src/srcdb/data/sample-ora2pg.conf
+	else
+		sudo cp ../yb-voyager/src/srcdb/data/sample-ora2pg.conf $conf_file_name
```
amit-yb, line 397: You should not assume that the script is being run from the `installer_scripts/` directory.

There is already an example where we first find the root of the git repo and switch to it:
```
        pushd `git rev-parse --show-toplevel` > /dev/null
        pushd yb-voyager > /dev/null
        $GO install
        popd > /dev/null
        popd > /dev/null
```
While using the above example for reference, you won't have to `cd` into the `yb-voyager` directory.

amit-yb (approved):
LGTM. Thanks for this quick change. 👍

PR #759, opened 2023-02-01
Define and use a class to represent table names

amit-yb:
This patch introduces a struct type SourceName and
TargetName to encapsulate all variations of
table names -- quoted, unquoted, and quoted-iff-required.

This ensures that code referring a table by name is
explicit about what form of table name is in use.


priyanshi-yb (approved):
LGTM 👍

yb-voyager/src/utils/sqlname/sqlname.go
```diff
@@ -0,0 +1,230 @@
...
+			return `"` + objectName + `"`
+		}
+	default:
+		panic("invalid source db type")
+	}
+}
+
+func isAllUppercase(s string) bool {
+	for _, c := range s {
+		if c >= 'a' && c <= 'z' {
```
sanyamsinghal, line 216: what if there are special characters in the name, this `if` will be fail/invalid
Since we don't support special chars in objName, maybe you add a TODO comment here for future

yb-voyager/src/utils/sqlname/sqlname.go
```diff
@@ -0,0 +1,230 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+func isAllLowercase(s string) bool {
+	for _, c := range s {
+		if c >= 'A' && c <= 'Z' {
```
sanyamsinghal, line 225: same as above

yb-voyager/src/utils/sqlname/sqlname.go
amit-yb: In fact, `isAllUppercase()` characters returns `false` only if it sees any lower-case character. Else it returns `true` (which includes names containing all capital letters or special characters).

The same applies to the `isAllLowercase()` implementation.

sanyamsinghal: yeah, misunderstood the code. Thanks

sanyamsinghal (approved):

Issue #765, opened 2023-02-02
Voyager doesn't recognize Hasura's syntax for stored procedures


dmagda:
I was migrating the following eCommerce application from PostgreSQL: 
https://github.com/YugabyteDB-Samples/hasura-ecommerce

The application has a public schema with the app-specific data and `hdb_catalog` schema with Hasura metadata. The following command was used during the `schema export` step:
```shell
yb-voyager export schema --export-dir $VOYAGER_EXPORT_DIR \
    --source-db-type postgresql \
    --source-db-host 34.23.188.24 \
    --source-db-user postgres \
    --source-db-password 'password' \
    --source-db-name postgres \
    --source-db-schema 'public,hdb_catalog'
```

The tool exports the following stored procedure of Hasura to the `voyager_export_dir\schema\functions\functions.sql` file:
```sql
CREATE FUNCTION hdb_catalog.gen_hasura_uuid() RETURNS uuid
    LANGUAGE sql
    AS $$select gen_random_uuid()$$; 
```

The `analyze-schema` and `import schema` commands complete succesfully and don't report any issues. But the stored procedure above (and all the others that are defined below it) are not imported. The import fails silently.

As a workaround, the stored procedure's format needs to be changed to the following:
```sql
CREATE FUNCTION hdb_catalog.gen_hasura_uuid() RETURNS uuid
LANGUAGE sql
AS $$
        select gen_random_uuid()
$$;
```

sanyamsinghal:
fix merged to `main`

PR #768, opened 2023-02-03
bug #765: Fixed function parsing issue in case function's complete body is in one single line

sanyamsinghal:


migtests/lib/yb.py
```diff
@@ -176,7 +176,17 @@ def fetch_all_function_names(self, schema_name="public") -> List[str]:
 		cur.execute(f"SELECT routine_name FROM information_schema.routines WHERE routine_schema = '{schema_name}'")
 		return [function[0] for function in cur.fetchall()]
 
-	def fetch_all_table_rows(self, table_name, schema_name="public") -> set[str]:
+	def fetch_all_table_rows(self, table_name, schema_name="public") -> any:
```
amit-yb, line 179: Specifying `any` is as good as specifying nothing about the return value.
What made you replace the `set[str]` to `any`?

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -676,12 +676,12 @@ func createSqlStrInfoArray(path string, objType string) []sqlInfo {
 
 		// Assuming that both the dollar quote strings will not be in same line
 		if dollarQuoteFlag == 0 {
-			if matches := dollarQuoteRegex.FindStringSubmatch(currLine); matches != nil {
-				dollarQuoteFlag = 1 //denotes start of the code/body part
-				codeBlockDelimiter = matches[0]
-			} else if strings.Contains(currLine, ";") { // in case, there is no body part
+			if strings.Contains(currLine, ";") { // in case, there is no body part or body part is in single line
```
amit-yb, line 679: What if there is a string literal containing `;` in it? Will it trigger false detection of end of function? How about checking if last non-white-space character is semi-colon?

sanyamsinghal: yeah, nice catch. Will make that change.

sanyamsinghal: but what if there is a sql comment after the semi-colon

amit-yb (approved):
LGTM.

PR #773, opened 2023-02-23
Fix the issue of export schema hanging in between when some error is coming from ora2pg for export of an object

priyanshi-yb:


yb-voyager/src/srcdb/ora2pg_extract_schema.go
```diff
@@ -83,13 +54,36 @@ func ora2pgExtractSchema(source *Source, exportDir string) {
 		}
 
 		err = exportSchemaObjectCommand.Wait()
+		var errMsgs []string
```
amit-yb, line 57: Let's simplify even more. Just like the `export data` code path, log the STDOUT and STDERR at this point.
In the `else` block on line 62, all you need to do is check if `errbuf.String()` and `outbuf.String()` contain `"error"` (that you can check using `strings.Contains()`). If there is an `"error"`, just write 1 to the channel. No need of iterating it line-by-line.

sanyamsinghal (approved):

yb-voyager/src/srcdb/ora2pg_extract_schema.go
```diff
@@ -83,13 +54,24 @@ func ora2pgExtractSchema(source *Source, exportDir string) {
...
 		err = exportSchemaObjectCommand.Wait()
+		if outbuf.String() != "" {
+			log.Infof(`ora2pg STDOUT: "%s"`, outbuf.String())
+		}
+		if errbuf.String() != "" {
+			log.Errorf(`ora2pg STDERR in export of %s : "%s"`, exportObject, errbuf.String())
+		}
 		if err != nil {
 			utils.PrintAndLog("Error while waiting for export command exit: %v", err)
 			exportSchemaObjectCommand.Process.Kill()
```
amit-yb, line 65: We don't need to explicitly call this `Process.Kill()`. So remove all such occurrences.

yb-voyager/src/srcdb/ora2pg_extract_schema.go
```diff
@@ -83,13 +54,24 @@ func ora2pgExtractSchema(source *Source, exportDir string) {
...
 			continue
```
amit-yb, line 66: Because of this `continue`, I suspect the "spinner" will continue spinning. After the line 28 (`go utils.Wait(...)`) it is a MUST to either send 1 or 0 on the `utils.WaitChannel`. Can you ensure that the channel is informed on all paths leading out of the iteration? (maybe using a proper `defer` stmt?)

amit-yb (approved):
LGTM. Good work!!! 👍

PR #788, opened 2023-03-09
Bug fix in Installer scripts for version comparisions.

makalaaneesh:
String comparison of versions is not correct in bash. 
For example, 
```
~/Documents/test » cat compver.sh                                                                 
min_ver="$1"
actual_ver="$2"

if [[ "$actual_ver" > "$min_ver" ]]
then
   echo "requirements met"
else
   echo "requirements NOT met"
fi

-----------------------------------------------------------------------------------------------------------------------
~/Documents/test » ./compver.sh 2.25 2.26                                                         
requirements met
-----------------------------------------------------------------------------------------------------------------------
~/Documents/test » ./compver.sh 2.25 2.100                                                        
requirements NOT met
```

Therefore, incorporating a proper solution for comparing version numbers - https://stackoverflow.com/a/4025065 that splits by DOTs and does an integer comparison for each split.


installer_scripts/install-yb-voyager-debezium
```diff
@@ -65,6 +65,48 @@ main() {
...
+        if ((10#${ver1[i]} < 10#${ver2[i]}))
+        then
+            echo 2; return;
+        fi
+    done
+    echo 0; return;
+}
+
+# usage: version_satisfied <minimum_version_required> <actual_version>
+version_satisfied() {
```
ShivanshGahlot, line 101: Hey @amit-yb do we need this abstraction in `version_satisfied` or can we just directly use the output of `vercomp`. We might encounter some cases in the future where we want the versions to be exactly equal and that cannot be checked using just the basic `-eq` for example `1.02` and `1.2`. 
In such a case we will anyway have to use `vercomp` directly.

installer_scripts/install-yb-voyager-debezium
```diff
@@ -65,6 +65,48 @@ main() {
 # COMMON
 #=============================================================================
 
+# https://stackoverflow.com/a/4025065
```
ShivanshGahlot, line 68: Can you add a comment stating `vercomp` returns 0 for equal 1 for var1>var2 etc.

amit-yb (approved):
LGTM.

installer_scripts/install-yb-voyager-debezium
makalaaneesh: done

installer_scripts/install-yb-voyager-debezium
makalaaneesh: @ShivanshGahlot merging this for now. Don't see a strong reason for choosing either approach, so leaving it as is.

PR #790, opened 2023-03-10
Added Blob/Clob automation

shubham-yb:
- Added Blob/Clob automation for Oracle and Mysql
- Fixed Oracle Datatypes automation test


priyanshi-yb (approved):
LGTM

PR #791, opened 2023-03-13
Updated YB versions for automation tests

shubham-yb:
Updated the docker version tags to newer releases of DB.
Added 2.17 to test the fix of regression https://github.com/yugabyte/yugabyte-db/issues/14529


amit-yb (approved):
LGTM. Thanks for updating this.

PR #792, opened 2023-03-13
Added a function to drop orafce in yb.py and used it before running validations of oracle/datatypes

ShivanshGahlot:
While running the `validations` for `datatypes` test in `oracle`, we need to fetch the columns of each table with their column type etc to validate from the `public` schema of yb. If` orafce` extension is left on, some of the tables created by `orafce` were also getting fetched. That is why I added a function to drop `orafce` in `yb.py`. This can be invoked before the tests that require `orafce` to be turned off. 
`Orafce` creates many functions and this can pose a problem in the future while validating.


shubham-yb (approved):
LGTM

priyanshi-yb (approved):

PR #803, opened 2023-03-28
Bug fix: Invalid config value for database.server.id, only numeric values are allowed

sanyamsinghal:
- using random number generator to generate and assign a unique value for the migration

Not using a constant fixed numeric value since multiple migrations might be running at source MySQL using debezium/yb-voyager

doc reference: [link](https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-property-database-server-id)

### DIFF
--- a/yb-voyager/src/dbzm/config.go
+++ b/yb-voyager/src/dbzm/config.go
@@ -2,6 +2,7 @@ package dbzm
 
 import (
 	"fmt"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"strings"
@@ -77,7 +78,7 @@ var mysqlSrcConfigTemplate = baseSrcConfigTemplate + `
 debezium.source.connector.class=io.debezium.connector.mysql.MySqlConnector
 
 debezium.source.database.include.list=%s
-debezium.source.database.server.id=ybvoyager-dbzm
+debezium.source.database.server.id=%d
 
 
 
@@ -123,6 +124,7 @@ func (c *Config) String() string {
 			c.Host, c.Port, c.Username, c.Password,
 			strings.Join(c.TableList, ","),
 			c.DatabaseName,
+			rand.Int(),
 			filepath.Join(c.ExportDir, "data", "schema_history.json"))
 	default:
 		panic(fmt.Sprintf("unknown source db type %s", c.SourceDBType))
--- a/yb-voyager/src/dbzm/config.go
+++ b/yb-voyager/src/dbzm/config.go
@@ -2,9 +2,12 @@ package dbzm
 
 import (
 	"fmt"
+	"log"
 	"math/rand"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


yb-voyager/src/dbzm/config.go
```diff
@@ -139,3 +142,38 @@ func (c *Config) WriteToFile(filePath string) error {
 	}
 	return nil
 }
+
+// read config file DEBEZIUM_CONF_FILEPATH into a string
+func readConfigFile() (string, error) {
+	configFile, err := os.ReadFile(DEBEZIUM_CONF_FILEPATH)
```
amit-yb, line 148: `configFile` -> `config`
Fix all such occurrences.

yb-voyager/src/dbzm/config.go
```diff
@@ -139,3 +142,38 @@ func (c *Config) WriteToFile(filePath string) error {
...
+		matches := re.FindStringSubmatch(configFile)
+		if len(matches) == 2 {
+			databaseServerId, err = strconv.Atoi(matches[1])
+			if err != nil {
+				log.Fatalf("failed to convert database server id to int: %v", err)
+				return databaseServerId
+			}
+		}
+	}
+
```
amit-yb, line 177: Add a log message that prints the server id selected.

yb-voyager/src/dbzm/config.go
```diff
@@ -139,3 +142,38 @@ func (c *Config) WriteToFile(filePath string) error {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to read config file %s: %w", DEBEZIUM_CONF_FILEPATH, err)
+	}
+
+	return string(configFile), nil
+}
+
+// generate/fetch the value for 'debezium.source.database.server.id' property for MySQL
+func getDatabaseServerID() int {
+	databaseServerId := rand.Intn(100000)
```
amit-yb, line 158: Ensure that the id is greater than 10000. Just to avoid any conflict with existing IDs.

yb-voyager/src/dbzm/config.go
```diff
@@ -139,3 +142,38 @@ func (c *Config) WriteToFile(filePath string) error {
...
+	configFile, err := readConfigFile()
+	if err != nil {
+		log.Fatalf("failed to read config file: %v", err)
+		return databaseServerId
```
amit-yb, line 162: Add a comment stating it is unreachable stmt. Consider returning `-1`. Same applies another similar error handler.

amit-yb, line 162: Add a comment stating it is unreachable stmt. Consider returning `-1`. Same applies another similar error handler.

sanyamsinghal: changed `Fatalf` to `Errorf`, i think we should move on to creating a new conf file in case of any error encountered here.

yb-voyager/src/dbzm/config.go
amit-yb: Nit: log message?

amit-yb (approved):
LGTM with a minor comment.

PR #811, opened 2023-04-11
Added error context to failed ddl file

rahulb-yb:
This patch helps voyager append the error message of failed DDL statements to the respective DDL statements which failed within the failed.sql file


yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -60,6 +60,7 @@ func importDefferedStatements() {
 	defer func() { conn.Close(context.Background()) }()
 
 	var err error
+	var errString string
```
sanyamsinghal, line 63: nit: no need to separately declare the variable
use `:=` at line 77

sanyamsinghal (approved):
LGTM

PR #821, opened 2023-04-26
Added separate workflow for import data file and analyse schema

shubham-yb:
- Import data file and Analyse schema tests used to run under PG tests. Added a separate workflow.
- Updated Docker image versions


shubham-yb:
Closing the PR since there won't be a major improvement over the current implementation.

PR #825, opened 2023-04-27
Sanyam/fix partitions debezium

sanyamsinghal:


yb-voyager/cmd/exportData.go
```diff
@@ -262,10 +265,25 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName) er
 	return nil
 }
 
+func filterTablePartitions(tableList []*sqlname.SourceName) []*sqlname.SourceName {
+	if source.DBType != POSTGRESQL || source.TableList != "" {
```
makalaaneesh, line 269: Don't MySQL and Oracle also have partitioned tables?

sanyamsinghal: Yes they do, but the query used in `GetAllTableNames()` does return only parent table in case of oracle and mysql.
But in case of postgresql, table partitions are considered as normal tables and hence they are also returned in the list and hence we only need to filter them out for the postgresql

sanyamsinghal:
Oracle Automation test pipelines results: 
1. https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/306/

yb-voyager/cmd/exportData.go
makalaaneesh: makes sense

yb-voyager/src/srcdb/postgres.go
```diff
@@ -272,3 +272,17 @@ func (pg *PostgreSQL) FilterEmptyTables(tableList []*sqlname.SourceName) ([]*sql
...
 	return nonEmptyTableList, emptyTableList
 }
+
+func (pg *PostgreSQL) IsTablePartition(table *sqlname.SourceName) bool {
+	var parentTable *sqlname.SourceName
+	query := fmt.Sprintf(`SELECT inhparent::pg_catalog.regclass
+	FROM pg_catalog.pg_class c JOIN pg_catalog.pg_inherits ON c.oid = inhrelid
+	WHERE c.oid = '%s'::regclass::oid`, table.Qualified.Unquoted)
+
+	err := pg.db.QueryRow(context.Background(), query).Scan(&parentTable)
```
amit-yb, line 282: I don't think you will be able to extract the parent table name in a variable of type `*sqlname.SourceName`. You will need a plain string.

amit-yb (approved):
LGTM with an important comment. Please fix it and then proceed to merge.

PR #830, opened 2023-05-02
[Oracle] Bug fix: Qualified table name is required whenever a different(non-schema) user is specified as source-db-user

sanyamsinghal:
For example: Let's say `source-db-user` is `admin`, but the `source-db-schema` to export is `sakila`. In these kind of cases, it is mandatory to use qualified table names in `Select *` kind of queries, since it is not know which schema to look for. But if `source-db-user` is also `sakila` then it is not mandatory


shubham-yb (approved):

sanyamsinghal:
Oracle automation tests results: https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/263/console

PR #837, opened 2023-05-09
[DB-6100] & [DB-6012]: Fixed Debezium Data Export bugs in case of case-sensitive tables(PostgreSQL) and reserved keywords tables(Oracle/MySQL)

sanyamsinghal:


amit-yb (approved):
LGTM. Good work!

PR #854, opened 2023-05-18
Fix progress bar issue on Ctrl-C

makalaaneesh:
- Newer version 8 has a shutdown method available on the progress struct. 
- Added exit hook for ProgressTracker which basically shutsdown entire progressbar.
- Reordered progressTracker struct creation above Debezium Start, because atexit handlers are run in the same order which they were registered in.


amit-yb (approved):
LGTM.

sanyamsinghal (approved):
LGTM. We need to manually test this update for every possible to check that it doesn't break any functionality or behaviour

PR #858, opened 2023-05-22
[DB-6633] added java check before running java

makalaaneesh:
https://yugabyte.atlassian.net/browse/DB-6633


makalaaneesh:
Decided to add a check in debezium's run.sh script instead. Closing this

PR #861, opened 2023-05-22
Filtering tables with empty supported column list

priyanshi-yb:
Issue - https://yugabyte.atlassian.net/browse/DB-6635
Fix - Skipping the table with all unsupported columns

Oracle pipeline run for this branch - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/483/

### DIFF
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -136,8 +136,8 @@ func exportDataOffline() bool {
 			"\nDo you want to ignore just these columns' data and continue with export") {
 			utils.ErrExit("Exiting at user's request. Use `--exclude-table-list` flag to continue without these tables")
 		}
+		finalTableList = filterTableWithEmptySupportedColumnList(finalTableList, tablesColumnList)
 	}
-	finalTableList = filterTableWithEmptySupportedColumnList(finalTableList, tablesColumnList)
 
 	if len(finalTableList) == 0 {
 		fmt.Println("no tables present to export, exiting...")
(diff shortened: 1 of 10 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -148,6 +148,7 @@ func exportDataOffline() bool {
 			utils.ErrExit("Exiting at user's request. Use `--exclude-table-list` flag to continue without these tables")
 		}
 	}
+	finalTableList = filterTableWithEmptySupportedColumnList(finalTableList, tablesColumnList)
```
makalaaneesh, line 151: Slightly unrelated question: What happens if as a result of all the filtering that we do (empty tables, unsupported columns, etc), we end up with no tables. Do we throw an error?

amit-yb (approved):
LGTM.

yb-voyager/cmd/exportData.go
```diff
@@ -305,6 +306,20 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName, ta
 	return nil
 }
 
+func filterTableWithEmptySupportedColumnList(finalTableList []*sqlname.SourceName, tablesColumnList map[*sqlname.SourceName][]string) []*sqlname.SourceName {
+	if tablesColumnList == nil {
```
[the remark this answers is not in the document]
priyanshi-yb: In case of PG, we are not even preparing this list as there are no unsupported types for the PG.

yb-voyager/cmd/exportData.go
priyanshi-yb: Oh yes, thanks we do that after filtering of tablelist for other cases few lines above, but I missed that here. Will now do that after this case of filtering. Thanks

priyanshi-yb: yes, we return from there by saying no tables to be exported and hence, export is completed.

yb-voyager/cmd/exportData.go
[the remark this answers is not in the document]
makalaaneesh: Cool, got it! 
nit: In which case, I think you can just call this function inside the `if len(unsupportedColumnNames) > 0` block. Then you wouldn't need this check

makalaaneesh (approved):
LGTM with a small comment

yb-voyager/cmd/exportData.go
makalaaneesh: Got it, thanks!

PR #865, opened 2023-05-22
Exiting with exit code=1 in export data if success==false, currently we are exiting normally i.e. with exit code=0

sanyamsinghal:
- for example: automation's run-tests.sh is moving to next step assuming that export data completed fine in case of data export error

### DIFF
--- a/migtests/scripts/run-test.sh
+++ b/migtests/scripts/run-test.sh
@@ -65,9 +65,8 @@ main() {
 	tail -20 ${EXPORT_DIR}/reports/report.txt
 
 	step "Export data."
-	{ #TRY
-		export_data
-	}|| { #CATCH
+	# false if exit code of export_data is non-zero
+	export_data || { 
 		# Print debezium.log file if present and exportDataDone flag file is not present
 		if [ -f "${EXPORT_DIR}/logs/debezium.log" ] && [ ! -f "${EXPORT_DIR}/metainfo/flags/exportDataDone" ]
 		then
--- a/migtests/scripts/run-test.sh
+++ b/migtests/scripts/run-test.sh
@@ -66,8 +66,7 @@ main() {
 
 	step "Export data."
 	{ #TRY
-		export_data 
-		export_data_exit_code=$?
+		export_data
 	}|| { #CATCH
 		# Print debezium.log file if present and exportDataDone flag file is not present
 		if [ -f "${EXPORT_DIR}/logs/debezium.log" ] && [ ! -f "${EXPORT_DIR}/metainfo/flags/exportDataDone" ]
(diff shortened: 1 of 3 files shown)


sanyamsinghal:
Oracle Automation Test Results: https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/484/
Status: Pending

migtests/scripts/run-test.sh
```diff
@@ -65,17 +65,22 @@ main() {
 	tail -20 ${EXPORT_DIR}/reports/report.txt
 
 	step "Export data."
-	export_data
-	ls -l ${EXPORT_DIR}/data
+	{ #TRY
+		export_data 
+		export_data_exit_code=$?
```
makalaaneesh, line 70: Doesn't seem like we are using this variable anywhere?

migtests/scripts/run-test.sh
```diff
@@ -65,17 +65,22 @@ main() {
...
+			echo "Printing debezium.log file"
+			tail -n 100 ${EXPORT_DIR}/logs/debezium.log
+		else
+			echo "No debezium.log found."
+		fi	
+		exit 1 #since the TRY block failed
+	}
 
-	# Print debezium.log file if present and exportDataDone flag file is not present
-	if [ -f "${EXPORT_DIR}/logs/debezium.log" ] && [ ! -f "${EXPORT_DIR}/metainfo/flags/exportDataDone" ]
```
makalaaneesh, line 72: Is this condition not working currently? Even if exit code is 0, this condition should be met, right?

makalaaneesh: I'm assuming this works currently. this PR is not about fixing the print of log file, but more about making run_tests.sh exit when export data fails. 

Correct?

migtests/scripts/run-test.sh
sanyamsinghal: yes removed it.

migtests/scripts/run-test.sh
sanyamsinghal: currently if actually export data fails with exit code=1, then this won't be printed, because the `run-tests.sh` script will exit because of `set -e`

makalaaneesh (approved):
LGTM. Thanks for fixing this! 
Still have no idea why the setup-java step runs here, but nevertheless, this should be good to merge

migtests/scripts/run-test.sh
```diff
@@ -65,17 +65,21 @@ main() {
 	tail -20 ${EXPORT_DIR}/reports/report.txt
 
 	step "Export data."
-	export_data
-	ls -l ${EXPORT_DIR}/data
+	{ #TRY
+		export_data
+	}|| { #CATCH
```
amit-yb, line 70: No need of curly braces for just a single command.
```
export_data || {
  ...
  ...
}
```

amit-yb (approved):
LGTM with a minor comment.

migtests/scripts/run-test.sh
sanyamsinghal: done

PR #869, opened 2023-05-23
Bug fix for analyze-schema reporting the supported DDL in unsupported case

priyanshi-yb:
Fixes - Bug reported by a user in the community slack, where analyze-schema was detecting few supported queries as unsupported, for ex - 
```
Alter table only parent_tbl add constraint party_profile_pk primary key (party_profile_id);
```
this is detected for the case of ALTER TABLE xyz OF case, where OF is part of the identifier name, and previous regex for this case was not considering a space after OF and due to which this query was being matched with the regex and hence reported. Fixed the regex.


amit-yb (approved):
LGTM.

PR #876, opened 2023-05-25
Run all tests only on main branch, latest release on PRs

rahulb-yb:
This patch resolves [DB-6689](https://yugabyte.atlassian.net/browse/DB-6689)
Disables automation tests from running on all versions of YugabyteDB during a PR, run only on latest available LTS version; maintain runs on all versions when merging into main branch.

### DIFF
--- a/.github/workflows/pg-migtests.yml
+++ b/.github/workflows/pg-migtests.yml
@@ -14,7 +14,7 @@ jobs:
         EXPERIMENTAL_FAST_EXPORT: [0, 1]
     env:
       EXPERIMENTAL_FAST_EXPORT: ${{ matrix.EXPERIMENTAL_FAST_EXPORT }}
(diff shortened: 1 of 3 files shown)


.github/workflows/pg-migtests.yml
```diff
@@ -14,6 +14,7 @@ jobs:
         EXPERIMENTAL_FAST_EXPORT: [0, 1]
     env:
       EXPERIMENTAL_FAST_EXPORT: ${{ matrix.EXPERIMENTAL_FAST_EXPORT }}
+    if: (github.event_name != 'pull_request') || (github.event_name== 'pull_request' && matrix.version == '2.18.0.0-b65')
```
amit-yb, line 17: Can we introduce a variable like `LATEST_LTS_VERSION` and use it while comparing `matrix.version`?

If the `||` is short-circuit operator, then there is no need of the `github.event_name=='pull_request'` condition.

amit-yb:
Closing this PR for now. We will revisit this later.

PR #872, opened 2023-05-24
Added support for characters such as backspace(\b) in quote and escape char with --file-opts in import data file

priyanshi-yb:
1. Added support for `backspace(\b)` type of characters in the import data file. 
2. added automation tests for the same with different combinations.

Oracle pipeline run on this branch - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/508/ [Passed]

PR #881, opened 2023-05-26
Added support for characters such as backspace(\b) in quote and escape char with --file-opts in import data file (#872)

priyanshi-yb:
* Added support for characters such as backspace(\b) in quote and escape character with --file-opts in import data file

* added automation tests for backspace char


amit-yb (approved):
LGTM.

PR #908, opened 2023-06-07
Changed java check and jdk version in dockerfile

ShivanshGahlot:


docker/Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM amd64/ubuntu:22.04
 ARG DEBIAN_FRONTEND=noninteractive
 ENV TZ=UTC
 RUN apt-get update && \
-    apt-get install -y wget openjdk-17-jre && \
```
amit-yb, line 8: Let's keep jdk 17 in our docker image. The docker env is ours! We can use any version we want ;-)

amit-yb (approved):
LGTM with one comment.

PR #909, opened 2023-06-12
Made DEBEZIUM_RELEASE_TAG, DEBEZIUM_VERSION and ORA2PG_VERSION variables overridable in the installer script.

ShivanshGahlot:
1) `DEBEZIUM_RELEASE_TAG`, `DEBEZIUM_VERSION`: You can export these variables before running the script to specify what version of debezium you want to install.
2) `ORA2PG_VERSION`: You can export this variable before running the script to install a particular hash of ora2pg. 

Both of these additions are for testing purposes in the Jenkins pipeline.

### DIFF
--- a/installer_scripts/install-yb-voyager
+++ b/installer_scripts/install-yb-voyager
@@ -233,7 +233,7 @@ get_passed_options() {
 				shift
 				;;
 			-d | --install-debezium-server-latest )
-        		DEBEZIUM_VERSION=latest
+				DEBEZIUM_VERSION=latest
 				DEBEZIUM_RELEASE_TAG=voyager-debezium
 				shift
 				;;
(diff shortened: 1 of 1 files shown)


makalaaneesh (commented):
@ShivanshGahlot Do we want to give users the option of installing latest debezium, or is it just for internal dev workflows? 
If it's just for internal purposes, how about we do without adding a new flag. 
We can achieve the same by making `DEBEZIUM_RELEASE_TAG` and `DEBEZIUM_VERSION` env vars overridable. 

What do you think?

ShivanshGahlot:
@makalaaneesh Made the change you suggested. It will make things easier in the jenkins pipeline too.

makalaaneesh (approved):
LGTM

PR #914, opened 2023-06-16
SSL support with Debezium - Oracle

makalaaneesh:
Related PR: https://github.com/yugabyte/debezium/pull/60/files
ref: https://yugabyte.atlassian.net/browse/DB-6534


yb-voyager/src/dbzm/dbzm.go
```diff
@@ -65,6 +65,11 @@ func (d *Debezium) Start() error {
 
 	log.Infof("starting debezium...")
 	d.cmd = exec.Command(filepath.Join(DEBEZIUM_DIST_DIR, "run.sh"), DEBEZIUM_CONF_FILEPATH)
+	d.cmd.Env = os.Environ()
+	// setting ORACLE_HOME so that wallet location and tns_admin are correctly set as
+	// $ORACLE_HOME/network/admin in debezium, which will enable using TNS alias
+	// as well as wallets for SSL.
+	d.cmd.Env = append(d.cmd.Env, fmt.Sprintf("ORACLE_HOME=%s", d.Config.OracleHome))
```
makalaaneesh, line 72: Context: https://github.com/yugabyte/debezium/pull/60/files

yb-voyager/cmd/exportData.go
```diff
@@ -340,6 +360,54 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName, ta
 	return nil
 }
 
+func getConnectionUriForDebezium(s srcdb.Source) (string, error) {
+	if s.DBType == "oracle" {
+		connectionStringRegex := regexp.MustCompile(`.*connectString="(?P<connectString>.*)".*`)
```
sanyamsinghal, line 365: please add a comment explaining that what and where are we extracting from.

yb-voyager/cmd/exportData.go
```diff
@@ -340,6 +360,54 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName, ta
...
+		return false, nil
+	}
+	tnsAdmin, err := getTNSAdmin(s)
+	if err != nil {
+		return false, fmt.Errorf("failed to get tns admin")
+	}
+	ojdbcPropertiesFilePath := filepath.Join(tnsAdmin, "ojdbc.properties")
+	if _, err := os.Stat(ojdbcPropertiesFilePath); errors.Is(err, os.ErrNotExist) {
+		// file does not exist
+		return false, nil
```
sanyamsinghal, line 389: why returning `nil` here? so that it picks up the default?

yb-voyager/cmd/exportData.go
```diff
@@ -340,6 +360,54 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName, ta
...
+		return connectionString, nil
+	}
+	return s.Uri, nil
+}
+
+// oracle wallet location can be optionally set in $TNS_ADMIN/ojdbc.properties as
+// oracle.net.wallet_location=<>
+func isOracleJDBCWalletLocationSet(s srcdb.Source) (bool, error) {
+	if s.DBType != "oracle" {
+		return false, nil
```
sanyamsinghal, line 380: instead of nil, not supported source db type?

migtests/tests/oracle/ssl/env.sh
```diff
@@ -0,0 +1,5 @@
+export SOURCE_DB_TYPE="oracle"
```
sanyamsinghal, line 1: We can test for ssl using some of the existing tests to run with ssl connections instead of creation some new extra tests, if possible.

yb-voyager/cmd/exportData.go
```diff
@@ -340,6 +360,54 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName, ta
...
+	}
+	ojdbcProperties := properties.MustLoadFile(ojdbcPropertiesFilePath, properties.UTF8)
+	walletLocationKey := "oracle.net.wallet_location"
+	_, present := ojdbcProperties.Get(walletLocationKey)
+	return present, nil
+}
+
+// https://www.orafaq.com/wiki/TNS_ADMIN
+// default is $ORACLE_HOME/network/admin
+func getTNSAdmin(s srcdb.Source) (string, error) {
```
sanyamsinghal, line 399: maybe we don't even need a error as a return type for this function?

yb-voyager/cmd/exportData.go
```diff
@@ -340,6 +360,54 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName, ta
...
+	if s.DBType != "oracle" {
+		return "", nil
```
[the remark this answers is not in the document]
makalaaneesh: Hmm, I thought we could have the caller just call this function irrespective of the db type, but I also see your point. it seems wrong to return nil error if db type is not oracle. caller should ensure that this is only called for oracle 👍

yb-voyager/src/srcdb/source.go
```diff
@@ -56,6 +56,14 @@ func (s *Source) DB() SourceDB {
 	return s.sourceDB
 }
 
+func (s *Source) GetOracleHome() string {
```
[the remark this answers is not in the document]
makalaaneesh: My concern is: We don't use the env var $ORACLE_HOME for ora2pg or while making a connection from voyager. So it'd be inconsistent to use it for debezium. 

Whatever the behaviour is, I think we should be consistent

yb-voyager/cmd/exportData.go
makalaaneesh: Yes

migtests/tests/oracle/ssl/env.sh
makalaaneesh: Thought about this, but i wanted to test multiple combinations - for example, with tnsadmin != oraclehome/network/admin, with a different wallet location, etc, and so I thought it'd be better to just have a separate test that was as simple as possible.

yb-voyager/src/srcdb/source.go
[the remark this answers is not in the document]
sanyamsinghal: But i am thinking shouldn't we do the same for ora2pg also? Like looking for the env var before assuming the default path?

migtests/tests/oracle/ssl/env.sh
sanyamsinghal: ok, makes sense

yb-voyager/cmd/exportData.go
```diff
@@ -396,15 +391,15 @@ func isOracleJDBCWalletLocationSet(s srcdb.Source) (bool, error) {
 
 // https://www.orafaq.com/wiki/TNS_ADMIN
 // default is $ORACLE_HOME/network/admin
-func getTNSAdmin(s srcdb.Source) (string, error) {
+func getTNSAdmin(s srcdb.Source) string {
 	if s.DBType != "oracle" {
-		return "", nil
+		return ""
```
sanyamsinghal: what will happen later on, when this is passed as empty to debezium conf?

yb-voyager/cmd/exportData.go
```diff
@@ -264,21 +264,29 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName, ta
...
+			return fmt.Errorf("failed to generate uri connection string: %v", err)
+		}
+		tnsAdmin, err = getTNSAdmin(source)
+		if err != nil {
+			return fmt.Errorf("failed to get tns admin: %w", err)
+		}
+		oracleJDBCWalletLocationIsSet, err = isOracleJDBCWalletLocationSet(source)
+		if err != nil {
+			return fmt.Errorf("failed to determine if Oracle JDBC wallet location is set: %v", err)
+		}
```
sanyamsinghal, line 287: can you move this to `PrepareSSLParamsForDebezium` ?

yb-voyager/cmd/exportData.go
```diff
@@ -274,11 +295,14 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName, ta
...
-		ColumnList:        dbzmColumnList,
-		ColumnSequenceMap: columnSequenceMap,
+		DatabaseName:                source.DBName,
+		SchemaNames:                 source.Schema,
+		TableList:                   dbzmTableList,
+		ColumnList:                  dbzmColumnList,
+		ColumnSequenceMap:           columnSequenceMap,
+		Uri:                         uri,
+		TNSAdmin:                    tnsAdmin,
+		OracleJDBCWalletLocationSet: oracleJDBCWalletLocationIsSet,
```
sanyamsinghal, line 305: thinking shall we set these props only for oracle db type? Currently looks like common props

yb-voyager/cmd/exportData.go
makalaaneesh: Hey, this is outdated. raising an error if dbtype!=oracle

sanyamsinghal: ok

sanyamsinghal (approved):

PR #917, opened 2023-06-19
[DB-6897] Do not quote identifiers coming from Oracle.

amit-yb:
This is a regression caused by my previous commit. We do not honour case-sensitivity for Oracle identifiers (like ora2pg). But the refactoring missed this part.


yb-voyager/cmd/exportData.go
```diff
@@ -379,12 +379,8 @@ func writeDataFileDescriptor(exportDir string, status *dbzm.ExportStatus) error
 	dataFileList := make([]*datafile.FileEntry, 0)
 	for _, table := range status.Tables {
 		// TODO: TableName and FilePath must be quoted by debezium plugin.
-		tableName := table.TableName
-		if (sqlname.IsCaseSensitive(tableName, source.DBType) || sqlname.IsReservedKeyword(tableName)) &&
```
sanyamsinghal, line 383: why have we removed the check `IsCaseSensitive()` for non postgres source db type?

PR #920, opened 2023-06-21
[DB-6923] Throttle COPY commands to match behaviour with old implementation

amit-yb:
The earlier implementation used an atomic int counter (that acted like a semaphore) to limit the number of batches submitted to target. Unlike a semaphore, voyager checked the counter only once every 2 seconds. That made voyager wake up every 2 second and ensure that there are `parallelism` number of batches in flight. During these 2 seconds, even if some batches finish there import, voyager will not submit new batches.

This behaviour provided some form of consistent throttling.

The refactoring changed this behaviour. Now that the code started using synchronisation primitives (e.g. channels and Pool), as soon as import of a batch finishes, voyager immediately sends the next batch. This caused too much load on the target thereby resulting in increased number of COPY command timeouts.

This PR tweaks the ConnPool implementation so that voyager behaves just like the old implementation.

### DIFF
--- a/yb-voyager/src/tgtdb/conn_pool.go
+++ b/yb-voyager/src/tgtdb/conn_pool.go
@@ -50,6 +50,8 @@ func (pool *ConnectionPool) WithConn(fn func(*pgx.Conn) (bool, error)) error {
 	for retry {
 		conn, gotIt := <-pool.conns
 		if !gotIt {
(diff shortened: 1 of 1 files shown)


sanyamsinghal (approved):
LGTM

sanyamsinghal (approved):

PR #921, opened 2023-06-21
[DB-6958] Handle relative paths specified in the --data-dir option.

amit-yb:
`import data file` command concats value of --data-dir and file-name from the --file-table-mapping to get the full path of the data file. The full path is expected to be absolute.

If a relative path is passed to the --data-dir, the computed full-path is also relative. Which eventually breaks the `import data status` command.

This patch converts the --data-dir to absolute path before it is used any further.


rahulb-yb (approved):
LGTM 👍🏽

PR #922, opened 2023-06-22
[DB-6975] Count the empty lines in the CSV files as well.

amit-yb:
Because, the new csv.Reader was ignoring empty new lines, the CsvDataFile was not getting any chance to account bytes of new-line. This caused mismatch in totalByteCount and importedByteCount during `import data status`.

This patch fixes the issue by reporting the number of bytes it skipped before returning the current line. This gives the caller a chance to account for the skipped bytes.

Updated an existing test-case to import CSV file containing empty lines.


priyanshi-yb (approved):
LGTM

PR #923, opened 2023-06-22
[DB-6976] Fix the Glob() implementation of S3DataStore.

amit-yb:
With the earlier Glob() implementation, a pattern of "test.txt" and bucket name "s3://foo", would result in listing objects in the internal "directories" as well e.g. "s3://foo/test.txt", "s3://foo/dir1/test.txt", etc.

This patch makes the regex matcher more specific so that it will list objects only from the given directory.


rahulb-yb (approved):
LGTM, nice catch!

PR #924, opened 2023-06-22
[DB-6975] From the descriptor, treat s3, gs, azure paths as absolute.

amit-yb:
In the dataFileDescriptor.json, if a data file path starts with s3://, gs://, or https:// then treat them as absolute path.

voyager was treating these paths as relative (as they don't start with the root dir (`/`)) and hence prefixing them with the export-dir.


priyanshi-yb (approved):
LGTM

PR #926, opened 2023-06-26
Added a line to print yb-voyager log file if export data fails

ShivanshGahlot:


migtests/scripts/functions.sh
```diff
@@ -268,3 +268,15 @@ import_data_file() {
 			exit 1
 		}
 }
+
+print_log_file() {
+	log_file_name=$1
+	if [ -f "${EXPORT_DIR}/logs/${log_file_name}" ] && [ ! -f "${EXPORT_DIR}/metainfo/flags/exportDataDone" ]
```
amit-yb, line 274: Take the `exportDataDone` check outside of this function (consider a case where we want to output voyager log when `export schema` command fails--there won't be this file).

migtests/scripts/functions.sh
```diff
@@ -268,3 +268,15 @@ import_data_file() {
 			exit 1
 		}
 }
+
+print_log_file() {
```
amit-yb, line 272: `tail_log_file()`.

migtests/scripts/functions.sh
```diff
@@ -268,3 +268,15 @@ import_data_file() {
...
+	log_file_name=$1
+	if [ -f "${EXPORT_DIR}/logs/${log_file_name}" ] && [ ! -f "${EXPORT_DIR}/metainfo/flags/exportDataDone" ]
+	then
+		echo "Printing ${log_file_name} file"
+		tail -n 100 "${EXPORT_DIR}/logs/${log_file_name}"
+	else
+		echo "No ${log_file_name} found."
+	fi	
+	exit 1 #since the TRY block failed
```
amit-yb, line 281: Take this statement outside of this function.

migtests/scripts/run-test.sh
```diff
@@ -67,15 +67,10 @@ main() {
...
 		# Print debezium.log file if present and exportDataDone flag file is not present
-		if [ -f "${EXPORT_DIR}/logs/debezium.log" ] && [ ! -f "${EXPORT_DIR}/metainfo/flags/exportDataDone" ]
-		then
-			echo "Printing debezium.log file"
-			tail -n 100 ${EXPORT_DIR}/logs/debezium.log
-		else
-			echo "No debezium.log found."
-		fi	
-		exit 1 #since the TRY block failed
+		print_log_file "debezium.log"
```
amit-yb, line 73: With the current implementation of `print_log_file()`, this line will never get executed--because the earlier print_log_file will call `exit 1`.

amit-yb (commented):
Just a few nits. Thanks for the PR 👍

migtests/scripts/functions.sh
ShivanshGahlot: Fixed. Btw we havent added an option to tail log files if export schema fails. Should I add?

migtests/scripts/run-test.sh
```diff
@@ -67,15 +67,14 @@ main() {
...
-		if [ -f "${EXPORT_DIR}/logs/debezium.log" ] && [ ! -f "${EXPORT_DIR}/metainfo/flags/exportDataDone" ]
+		if [ ! -f "${EXPORT_DIR}/metainfo/flags/exportDataDone" ]
 		then
-			echo "Printing debezium.log file"
-			tail -n 100 ${EXPORT_DIR}/logs/debezium.log
-		else
-			echo "No debezium.log found."
-		fi	
-		exit 1 #since the TRY block failed
+			# Print yb-voyager.log file if present and exportDataDone flag file is not present
```
amit-yb, line 72: These two comments do not add much value. You can safely remove them.

migtests/scripts/run-test.sh
```diff
@@ -67,15 +67,14 @@ main() {
 	step "Export data."
 	# false if exit code of export_data is non-zero
 	export_data || { 
-		# Print debezium.log file if present and exportDataDone flag file is not present
-		if [ -f "${EXPORT_DIR}/logs/debezium.log" ] && [ ! -f "${EXPORT_DIR}/metainfo/flags/exportDataDone" ]
+		if [ ! -f "${EXPORT_DIR}/metainfo/flags/exportDataDone" ]
```
amit-yb, line 70: I think we can get rid of this check as well. I guess, earlier this was added to detect export data failure. With the way we arranged `export_data() || {`, when the control flow reaches line 70, we "know" that export data filed).

amit-yb (approved):
LGTM with couple of minor nits.

migtests/scripts/run-test.sh
ShivanshGahlot: Yeah I was thinking the same

PR #930, opened 2023-06-27
[DB-7015]Added the optimisation to import data file from S3 to fetch objects only for given prefix in --data-dir

priyanshi-yb:
[DB-7015](https://yugabyte.atlassian.net/browse/DB-7015)


amit-yb (approved):
LGTM. Thanks for making these changes on priority. 👍

PR #931, opened 2023-06-27
Make `import data status` independent of location of export-dir.

amit-yb:
`export data` command outputs its files in the $EXPORT_DIR/data directory. A hash of the absolute path of the data file is included in the name of the import data state directory of the file. If, after the partial/full import of the data, the user moves the export-dir to some other location, the hashes in the import data state directory names will become invalid ( because the hashes were calculated on the abs path of the data files which includes the old export-dir path as well).

This PR fixes the issue by not including the `export-dir` path while calculating the path hashes.


sanyamsinghal (approved):
LGTM

priyanshi-yb (approved):

PR #933, opened 2023-06-28
Rename ybvoyager_import_data_batches_metainfo table as ybvoyager_impo…

amit-yb:
…rt_data_batches_metainfo_v2

The rename is required so that voyager 1.4 doesn't try to use the old table created by 1.3 and lower. The 1.4 version is incompatible with the import data state created by the earlier versions.


sanyamsinghal (approved):
LGTM

PR #943, opened 2023-07-04
Implemented backdoor way to provide pg_dump args for PostgreSQL migrations

sanyamsinghal:
This PR introduces a new `pg_dump-args.ini` file under `/etc/yb-voyager` directory which is created as a part of voyager installation.

If this file is present, voyager takes the args from it and uses them. 
If not, then internally the args are prepared and used as per the user's input.

```
[centos@ip-10-9-2-160 ~]$ cat /etc/yb-voyager/pg_dump-args.ini
[schema]
--schema-only={{or .SchemaOnly true}}
--schema="{{ .Schema }}"
--extension={{ .ExtensionPattern }}
--file={{ .SchemaTempFilePath }}
--no-comments={{or .NoComments true}}
--no-owner={{or .NoOwner true}}
--no-privileges={{or .NoPrivileges true}}
--no-tablespaces={{or .NoTablespaces true}}
--load-via-partition-root={{or .LoadViaPartitionRoot true}}

[data]
--data-only={{or .DataOnly true}}
--no-blobs={{or .NoBlobs true}}
--compress=0
--table={{ .TablesListPattern }}
--format={{or .DataFormat "directory"}}
--file={{ .DataDirPath }}
--jobs={{ .ParallelJobs }}
--no-owner={{or .NoOwner true}}
--no-privileges={{or .NoPrivileges true}}
--no-tablespaces={{or .NoTablespaces true}}
--load-via-partition-root={{or .LoadViaPartitionRoot true}}
``` 

TODO: Update `yum/apt/brew/docker` installation packages


sanyamsinghal:
Oracle Jenkins Pipeline Test Results: 
1. https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/695/
2. https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/696/

yb-voyager/src/srcdb/pg_dump_export_data.go
```diff
@@ -33,22 +36,25 @@ import (
...
 
-	pgDumpArgs := fmt.Sprintf(`--no-blobs --data-only --no-owner --compress=0 %s -Fd --file %s --jobs %d --no-privileges --no-tablespaces --load-via-partition-root`,
-		tableListPatterns, dataDirPath, source.NumConnections)
+	pgDumpArgs.DataDirPath = filepath.Join(exportDir, "data")
+	pgDumpArgs.TablesListPattern = createTableListPatterns(tableList)
+	pgDumpArgs.ParallelJobs = strconv.Itoa(source.NumConnections)
+	pgDumpArgs.DataFormat = "directory"
+
+	args := source.getPgDumpDataArgsFromFile()
+	if args == "" {
```
makalaaneesh, line 50: Under what circumstances would be get empty args? 
Even if the file is not present, we can have a default conf file embedded into go, and parse that, no?

That would be more consistent than having a different set of default args in code. Basically, one single source of args is what I'm getting at.

yb-voyager/src/srcdb/pg_dump_extract_schema.go
```diff
@@ -39,14 +42,23 @@ func pgdumpExtractSchema(source *Source, connectionUri string, exportDir string)
...
+	pgDumpArgs.Schema = source.Schema
+	pgDumpArgs.SchemaTempFilePath = filepath.Join(exportDir, "temp", "schema.sql")
+	pgDumpArgs.NoComments = strconv.FormatBool(!source.CommentsOnObjects)
+	pgDumpArgs.ExtensionPattern = `"*"`
+
+	args := source.getPgDumpSchemaArgsFromFile()
+	if args == "" {
+		args = fmt.Sprintf(`--schema-only --schema "%s" --no-owner -f %s --no-privileges --no-tablespaces --extension "*" --load-via-partition-root`,
+			source.Schema, filepath.Join(exportDir, "temp", "schema.sql"))
+		if !source.CommentsOnObjects {
```
makalaaneesh, line 54: Why special handling for this arg? Won't it be read from the file anyway?

yb-voyager/src/srcdb/source.go
```diff
@@ -202,6 +203,26 @@ type Ora2pgConfig struct {
 	ModifyStruct     string
 }
 
+type PgDumpArgs struct {
```
makalaaneesh, line 206: For the boolean ones, should be boolean type no?

yb-voyager/src/srcdb/source.go
```diff
@@ -315,6 +336,93 @@ func (source *Source) extrapolateDSNfromSSLParams(DSN string) string {
...
+	section := iniData.Section("schema")
+	args := ""
+	for _, key := range section.Keys() {
+		if key.Value() == "false" {
+			continue
+		}
+		if key.Value() == "true" {
+			args += fmt.Sprintf(" %s", key.Name())
+		} else {
+			if key.Name() == "--schema" {
```
makalaaneesh, line 374: Don't see a difference in the if-else blocks 🤔 What is the intention here?

yb-voyager/src/srcdb/source.go
```diff
@@ -315,6 +336,93 @@ func (source *Source) extrapolateDSNfromSSLParams(DSN string) string {
...
+	}
+
+	iniData, err := ini.Load(output.Bytes())
+	if err != nil {
+		utils.ErrExit("Error while ini loading pg_dump arguments file: %v", err)
+	}
+	section := iniData.Section("data")
+	args := ""
+	for _, key := range section.Keys() {
+		if key.Value() == "false" {
```
makalaaneesh, line 413:
```
if key.Value() == "false" {
    ...
} else if key.Value() == "true" {
    ...
} else {
    ...
}
```

is more readable IMO.

yb-voyager/src/srcdb/source.go
```diff
@@ -315,6 +336,93 @@ func (source *Source) extrapolateDSNfromSSLParams(DSN string) string {
 	return DSN
 }
 
+func (source *Source) getPgDumpSchemaArgsFromFile() string {
```
makalaaneesh, line 339: Try and merge `getPgDumpSchemaArgsFromFile` and `getPgDumpDataArgsFromFile` into a single function, and parameterise the section?

makalaaneesh, line 339: Don't think this method should be on the source struct. (Remember we were discussing this?). It doesn't use the source struct's fields/methods. Moreover, this is pg specific functionality that shouldn't be on the abstract `source` struct. Move it to another file `pg_dump.go` if it's common to both pg_dump schema and data. 

Speaking of, I made some related refactoring in https://github.com/yugabyte/yb-voyager/pull/929 😄  Do take a look !

yb-voyager/src/srcdb/pg_dump_extract_schema.go
sanyamsinghal: yeah it is the case when the file is not present. I agree with you idea of having an internal pg-dump-args file also. So this will be gone then

yb-voyager/src/srcdb/pg_dump_export_data.go
sanyamsinghal: yeah agreed, will make the change

yb-voyager/src/srcdb/source.go
sanyamsinghal: the schema names are comma separate multiple names. Need to provided within quotes. 
I tried other ways like putting the quotes inside `.ini` file or passing it as a value but they got considered as a part of string.

yb-voyager/src/srcdb/source.go
sanyamsinghal: `ini` parser read it in form of strings only. We will need to parse it into respective datatype first, but in the end we just want pg_dump args in the form of string.

amit-yb (approved):
LGTM.

makalaaneesh (approved):
LGTM. Pls make sure that the other installation methods also install the pg_dump_args file

PR #946, opened 2023-07-07
Added datatype transformations in import data

priyanshi-yb:
This PR introduces a `valueConverter` interface which requires two functions `ConvertRow()` for transforming a row from the datafile, and `ConvertEvent()` which is used to transform the values in the event. 
Currently, `DebeziumValueConverter` is implementing the `valueConverter` for the data coming from debezium as after [this](https://github.com/yugabyte/debezium/pull/62) change in debezium, transformations are not performed in the debezium plugin. It implements both the functions and use the `valueConverterSuite` as per target DB.
Oracle tests - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/815/


yb-voyager/cmd/exportData.go
```diff
@@ -541,6 +540,18 @@ func renameDbzmExportedDataFiles() error {
 		if err != nil {
 			return fmt.Errorf("failed to rename dbzm exported data file: %w", err)
 		}
+
+		//rename table schema file as well
```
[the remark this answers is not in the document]
priyanshi-yb: I kept it similar to datafiles, as it will either rename it or change the table name while reading the file for the case-sensitive/reserved words. So the logic to rename all case-sensitive/reserved word table files was already here, I applied the same to schema files as well.

yb-voyager/cmd/live_migration.go
```diff
@@ -57,8 +59,40 @@ func streamChanges() error {
...
-	// TODO: Convert values in the event to make it suitable for target DB.
+	tableName := event.TableName
+	if sourceDBType == "postgresql" && event.SchemaName != "public" {
+		tableName = event.SchemaName + "." + event.TableName
+	}
+	// preparing value converters for the streaming mode
+	tableSchema := dbzm.GetTableSchema(tableName, exportDir)
+	columnToConverterFn := make(map[string]func(string) (string, error))
+	for _, column := range tableSchema.Columns {
+		logicalType := column.ColDbzSchema.Name
```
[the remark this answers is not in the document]
priyanshi-yb: common function approach is fine, but what is the point to have a separate file for just this `convertValue` function?
I think keeping this function in `schemaRegistry.go` should also be fine right?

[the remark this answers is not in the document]
makalaaneesh: This wouldn't really have anything to do with schemaRegistry, right? It's more about converting the values, so I thought new file would be fine :) If it's a small enough function, I'm okay with keeping it along with the schemaRegistry, but then rename the file to `schema.go`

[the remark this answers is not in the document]
priyanshi-yb: Hmm..Okay, let's separate out the transforming value and schema part.

yb-voyager/cmd/importData.go
```diff
@@ -661,6 +676,48 @@ func getTargetSchemaName(tableName string) string {
 	return tconf.Schema // default set to "public"
 }
 
+func prepareValueConverters(tasks []*ImportFileTask) {
```
makalaaneesh, line 679: nit: `prepareValueConvertersPerTable`

yb-voyager/cmd/live_migration.go
```diff
@@ -57,8 +59,30 @@ func streamChanges() error {
 
 func handleEvent(event *tgtdb.Event) error {
 	log.Debugf("Handling event: %v", event)
-
-	// TODO: Convert values in the event to make it suitable for target DB.
+	tableName := event.TableName
```
makalaaneesh, line 62: nit: move this to a separate func `transformEventFields`

yb-voyager/cmd/live_migration.go
```diff
@@ -57,8 +59,30 @@ func streamChanges() error {
...
+	if sourceDBType == "postgresql" && event.SchemaName != "public" {
+		tableName = event.SchemaName + "." + event.TableName
+	}
+	// preparing value converters for the streaming mode
+	tableSchema := dbzm.GetTableSchema(tableName, exportDir)
+	columnToSchema := make(map[string]dbzm.Schema)
+	for _, column := range tableSchema.Columns {
+		columnToSchema[column.ColName] = column
+	}
+	for column, value := range event.Fields {
```
makalaaneesh, line 72: We might have to do this for event.Key as well, I guess? (what if the primary key is one of the types that needs transformation)

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -774,3 +777,133 @@ func (yb *TargetYugabyteDB) recordEntryInDB(tx pgx.Tx, batch Batch, rowsAffected
 	}
 	return nil
 }
+
+func (yb *TargetYugabyteDB) GetDebeziumValueConverterSuite(isStreamingMode bool) map[string]func(string) (string, error) {
+	converterValueSuite := make(map[string]func(string) (string, error))
+	converterValueSuite["io.debezium.time.Date"] = func(columnValue string) (string, error) {
+		epochDays, err := strconv.ParseUint(columnValue, 10, 64)
+		fmt.Printf("epochDays: %v", epochDays)
```
makalaaneesh, line 785: nit: remove printf

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -774,3 +777,133 @@ func (yb *TargetYugabyteDB) recordEntryInDB(tx pgx.Tx, batch Batch, rowsAffected
...
+		if err != nil {
+			return columnValue, fmt.Errorf("parsing epoch seconds: %v", err)
+		}
+		epochSecs := epochDays * 24 * 60 * 60
+		return time.Unix(int64(epochSecs), 0).Local().Format(time.DateOnly), nil
+	}
+	converterValueSuite["io.debezium.time.Timestamp"] = func(columnValue string) (string, error) {
+		epochMilliSecs, err := strconv.ParseInt(columnValue, 10, 64)
```
makalaaneesh, line 793: so the int 64 bit size is equivalent to long right?

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -774,3 +777,133 @@ func (yb *TargetYugabyteDB) recordEntryInDB(tx pgx.Tx, batch Batch, rowsAffected
...
+		return time.Unix(epochSecs, 0).Local().Format(time.DateTime), nil
+	}
+	converterValueSuite["io.debezium.time.MicroTimestamp"] = func(columnValue string) (string, error) {
+		epochMicroSecs, err := strconv.ParseInt(columnValue, 10, 64)
+		if err != nil {
+			return columnValue, fmt.Errorf("parsing epoch microseconds: %v", err)
+		}
+		epochSeconds := epochMicroSecs / 1000000
+		epochNanos := (epochMicroSecs % 1000000) * 1000
+		microTimeStamp, err := time.Parse(time.RFC3339Nano, time.Unix(epochSeconds, epochNanos).Local().Format(time.RFC3339Nano))
```
makalaaneesh, line 807: I don't understand the RFC3339Nano but this is micro timestamp, so maybe it should be RFC3339Micro?

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -774,3 +777,133 @@ func (yb *TargetYugabyteDB) recordEntryInDB(tx pgx.Tx, batch Batch, rowsAffected
...
+		return strings.TrimSuffix(microTimeStamp.String(), " +0000 UTC"), nil
+	}
+	converterValueSuite["io.debezium.time.NanoTimestamp"] = func(columnValue string) (string, error) {
+		epochNanoSecs, err := strconv.ParseInt(columnValue, 10, 64)
+		if err != nil {
+			return columnValue, fmt.Errorf("parsing epoch nanoseconds: %v", err)
+		}
+		epochSeconds := epochNanoSecs / 1000000000
+		epochNanos := epochNanoSecs % 1000000000
+		nanoTimeStamp, err := time.Parse(time.RFC3339Nano, time.Unix(epochSeconds, epochNanos).Local().Format(time.RFC3339Nano))
```
makalaaneesh, line 820: In all the time formating, pls make sure it's UTC.

makalaaneesh (approved):
Overall LGTM @priyanshi-yb  . Nice work on the converter suite! 
Pls check one important comment about also transforming the key of the event(in addition to the fields). 
Rest of the comments about datatype conversion, you can add TODOs and move on to more pressing tasks.

yb-voyager/src/tgtdb/yugabytedb.go
priyanshi-yb: this `RFC3339Nano` format is used to format the time with Nano seconds in timestamp, but there is not anything as such `RFC3339Micro `, so I had to use the nano one available.

yb-voyager/src/tgtdb/yugabytedb.go
priyanshi-yb: Yes

yb-voyager/src/tgtdb/yugabytedb.go
priyanshi-yb: I have used a Micro secs time format in MicroTime one, but at that time when I tried that similar format here in timestamp it didn't. I will add a todo and check this later.

yb-voyager/src/tgtdb/yugabytedb.go
priyanshi-yb: this `time` package, by default, it gives UTC format only unless we specify. So all the time formatting here is in UTC.

yb-voyager/cmd/live_migration.go
priyanshi-yb: yes, completely missed it. I remember it in the starting but while working missed that. Thanks for catching that. I will add it.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,139 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/tgtdb"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
+)
+
+type ValueConverter interface {
```
makalaaneesh, line 26: oo I like this design :D

priyanshi-yb: Thanks to @amit-yb ! :)

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/tgtdb"
+)
+
+type DbzSchema struct {
+	Type string `json:"type"`
+	Name string `json:"name"`
+	// Not decoding the rest of the fields for now.
+}
+
+type Schema struct {
```
amit-yb, line 33: Change the type name to `Column`.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+	ColName      string    `json:"name"`
```
amit-yb, line 34: Change the variable name as just `Name`.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+	Index        int64     `json:"index"`
+	ColDbzSchema DbzSchema `json:"schema"`
```
amit-yb, line 36: ColDbzSchema -> Schema

In general, we can use shorter name by using the context in which an identifier is used.
Also, use abbreviations judiciously. Excessive use can slow down a reader.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/tgtdb"
+)
+
+type DbzSchema struct {
```
amit-yb, line 27: Change type name to `ColumnSchema`.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+	ColName      string    `json:"name"`
+	Index        int64     `json:"index"`
+	ColDbzSchema DbzSchema `json:"schema"`
+}
+
+type TableSchema struct {
+	Columns []Schema `json:"columns"`
+}
+
+func GetTableSchema(tableName string, exportDir string) (TableSchema, error) {
```
amit-yb, line 43: Make it a private method of `SchemaRegistry`.

Almost always: return pointer types instead of value types. In this case, return `*TableSchema`. That will allow you to return `nil` TableSchema in the presence of an error. Fix all such occurrences.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+	schemaFileName := fmt.Sprintf("%s_schema.json", tableName)
+	schemaFilePath := filepath.Join(exportDir, "data", "schemas", schemaFileName)
+	schemaFile, err := os.Open(schemaFilePath)
+	if err != nil {
+		return TableSchema{}, fmt.Errorf("failed to open schema file %s: %v", schemaFilePath, err)
```
amit-yb, line 48: `return nil, fmt.Errorf(...)`

Fix all such occurrences in this function and other functions as well.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return TableSchema{}, fmt.Errorf("failed to decode schema file %s: %v", schemaFilePath, err)
+	}
+	return tableSchema, nil
+
+}
+
+func (ts *TableSchema) GetColumnSchema(columnName string) Schema {
+	for _, colSchema := range ts.Columns {
+		if colSchema.ColName == columnName {
```
amit-yb, line 62: Strip quotes around `columnName` if they exist.
I think we should use case-insensitive comparison to avoid all the issues related to case sensitivity.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+	var tableSchema TableSchema
+	err = json.NewDecoder(schemaFile).Decode(&tableSchema)
+	if err != nil {
+		return TableSchema{}, fmt.Errorf("failed to decode schema file %s: %v", schemaFilePath, err)
+	}
+	return tableSchema, nil
+
+}
+
+func (ts *TableSchema) GetColumnSchema(columnName string) Schema {
```
amit-yb, line 60: `GetColumn(...) *Column`

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, colSchema := range ts.Columns {
+		if colSchema.ColName == columnName {
+			return colSchema
+		}
+	}
+	return Schema{}
```
amit-yb, line 66: `return nil`.

Log a warning message before returning. The log message must dump the `columnName` we are looking up and the entire `TableSchema`.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+
+type SchemaRegistry struct {
+	exportDir string
+}
+
+func NewSchemaRegistry(exportDir string) *SchemaRegistry {
+	return &SchemaRegistry{exportDir: exportDir}
+}
+
+func (sreg *SchemaRegistry) GetColumnSchemas(tableName string, columnNames []string) ([]Schema, error) {
```
amit-yb, line 90: I would suggest returning just the type names from this functions (and change the function name to `GetColumnTypes()`).

As a general principle, a class should hide its internal implementation detail. The API of the class should fulfil some use-case of the caller.

For column type, mostly, you will need to return `ColumnSchema.Type`. But if its value is `"STRUCT"`, return `ColumnSchema.Name` instead.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+	tableSchema, err := GetTableSchema(tableName, sreg.exportDir)
+	if err != nil {
+		return nil, err
+	}
+	columnToSchema := make([]Schema, len(columnNames))
+	for idx, columnName := range columnNames {
+		colSchema := tableSchema.GetColumnSchema(columnName)
```
amit-yb, line 97: Probably you can replace `GetColumnSchema` with `GetColumnType(columnName) string`.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,114 @@
...
+		columnToSchema[idx] = colSchema
+	}
+	return columnToSchema, nil
+}
+
+func (sreg *SchemaRegistry) GetColumnSchema(tableName, columnName string) (Schema, error) {
```
amit-yb, line 103: As suggested above make this function directly return `typeName` instead of asking users the call some method on the `Schema`.
The point is the caller need not know that `SchemaRegistry` read some json file or contacted some external service to get the details. Just give the table and column name and ask the `SchemaRegistry` for the typeName.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,142 @@
...
+	result := conv.converterFnCache[tableName]
+	if result == nil {
+		colSchemas, err := conv.schemaRegistry.GetColumnSchemas(tableName, columnNames)
+		if err != nil {
+			return nil, err
+		}
+		result = make([]tgtdb.ConverterFn, len(columnNames))
+		for i := range columnNames {
+			schema := colSchemas[i]
+			result[i] = schema.ColDbzSchema.GetConverterFn(conv.valueConverterSuite)
```
amit-yb, line 104: Get rid of the `GetConverterFn()` -- that's too much engineering ;-) . Directly lookup the type name in the converter suite.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+		conv.converterFnCache[tableName] = result
+	}
+	return result, nil
+}
+
+func (conv *DebeziumValueConverter) ConvertEvent(ev *tgtdb.Event, table string) error {
+	err := conv.convertMap(table, ev.Key)
+	if err != nil {
+		return fmt.Errorf("convert event fields: %v", err)
```
amit-yb, line 114: Always use `%w` format specifier for `err`.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	err = conv.convertMap(table, ev.Fields)
+	if err != nil {
+		return fmt.Errorf("convert event keys: %v", err)
+	}
+	return nil
+}
+
+func (conv *DebeziumValueConverter) convertMap(tableName string, m map[string]interface{}) error {
```
amit-yb, line 123: The `Key` and `Fields` MUST be of type `map[string]string`! I don't want json conversion to mess up values exported by DB.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,142 @@
...
+	for column, value := range m {
+		if value == nil {
+			m[column] = "NULL"
+			continue
+		}
+		columnValue := fmt.Sprintf("%v", value)
+		colSchema, err := conv.schemaRegistry.GetColumnSchema(tableName, column)
```
amit-yb, line 130: 🤯  For each event, `GetColumnSchema()` will open the table schema file, parse it, and then look for the type info!!

Define `schemaRegistry.Init() error` method that will open all the schema files, parse them, and store it in a map inside `SchemaRegistry`.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,142 @@
...
+		if err != nil {
+			return fmt.Errorf("fetch column schema: %v",err)
+		}
+		converterFn := colSchema.ColDbzSchema.GetConverterFn(conv.valueConverterSuite)
+		columnValue, err = converterFn(columnValue)
+		if err != nil {
+			return fmt.Errorf("converting value: %v",err)
```
amit-yb, line 137: Provide enough information in the error context so that one can start debugging e.g.
```
return fmt.Errorf("error while converting %s.%s of type %s in event[%d]: %w", tableName, columnName, typeName, event.SequenceID, err)
```

yb-voyager/src/tgtdb/event.go
```diff
@@ -21,11 +21,11 @@ import (
...
-	Op         string            `json:"op"`
-	SchemaName string            `json:"schema_name"`
-	TableName  string            `json:"table_name"`
-	Key        map[string]string `json:"key"`
-	Fields     map[string]string `json:"fields"`
+	Op         string                 `json:"op"`
+	SchemaName string                 `json:"schema_name"`
+	TableName  string                 `json:"table_name"`
+	Key        map[string]interface{} `json:"key"`
+	Fields     map[string]interface{} `json:"fields"`
```
amit-yb, line 28: Strictly no!! Keep the values coming from debezium as `string`. Only the value converter functions should interpret the values. We don't want golang type semantics, json type semantics to come into play. e.g. consider a high precision float number or decimal number.

yb-voyager/src/tgtdb/event.go
```diff
@@ -69,12 +73,12 @@ func (event *Event) getUpdateStmt(targetSchema string) string {
 	}
 	var setClauses []string
 	for column, value := range event.Fields {
-		setClauses = append(setClauses, fmt.Sprintf("%s = %s", column, value))
+		setClauses = append(setClauses, fmt.Sprintf("%s = '%s'", column, value))
```
amit-yb, line 76: Not all values should be quoted! The `ValueConverter.ConvertEvent()` must put quotes only if required by the value's type (e.g. string).

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -33,8 +33,12 @@ type TargetDB interface {
 	ImportBatch(batch Batch, args *ImportBatchArgs) (int64, error)
 	IfRequiredQuoteColumnNames(tableName string, columns []string) ([]string, error)
 	ExecuteBatch(batch []*Event) error
+	GetDebeziumValueConverterSuite(isStreamingMode bool) map[string]ConverterFn
 }
 
+//value converter Function type 
+type ConverterFn func(v string) (string, error)
```
amit-yb, line 40: `ConverterFn` should take one more boolean argument `quoteIfRequired`. If the flag is true, the return value should be such that it can be directly passed in a SQL statement (INSERT, UPDATE, etc.). If the flag is false, it should return values suitable to be written in a batch file.

amit-yb (commented):
Overall on a right track. Not completely reviewed the PR. Posting the comments accumulated until now.

yb-voyager/src/tgtdb/event.go
priyanshi-yb: @makalaaneesh, can you please make that change again?

yb-voyager/cmd/exportData.go
```diff
@@ -432,6 +427,10 @@ func checkAndHandleSnapshotComplete(status *dbzm.ExportStatus, progressTracker *
 		return false, fmt.Errorf("failed to write data file descriptor: %w", err)
 	}
 	log.Infof("snapshot export is complete.")
+	err = renameDbzmExportedDataFiles()
```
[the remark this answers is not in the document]
priyanshi-yb: this function is called from debeziumExportData() only, this is only called when debezium is in use for exporting data.

yb-voyager/src/dbzm/schemaRegistry.go
priyanshi-yb: here, columnName(header) and colSchema.Name(schema file) both column names are coming from debezium and they both will be with case-sensitivity. I don't think there will be any case-sensitivity issue here as such.

yb-voyager/src/dbzm/valueConverter.go
priyanshi-yb: Is there any SequenceId for each event?

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,150 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("fetching column schemas: %w", err)
+		}
+		result = make([]tgtdb.ConverterFn, len(columnNames))
+		for i := range columnNames {
+			colType := colTypes[i]
+			if conv.valueConverterSuite[colType] != nil {
+				result[i] = conv.valueConverterSuite[colType]
+			} else {
+				result[i] = func(v string, formatIfRequired bool) (string, error) { return v, nil }
```
amit-yb, line 108: This may become the most frequently called function! Consider leaving `result[i]` as `nil`. On line 83, skip conversion if `converterFn[i]` is `nil`.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		conv.converterFnCache[tableName] = result
+	}
+	return result, nil
+}
+
+func (conv *DebeziumValueConverter) ConvertEvent(ev *tgtdb.Event, table string) error {
+	err := conv.convertMap(table, ev.Key)
+	if err != nil {
+		return fmt.Errorf("convert event fields: %w", err)
```
amit-yb, line 119: On this line and line 123 swap `"fields"` and `"keys"` from the error messages.

Include event's sequenceID in the message. This comment applies to all such similar occurrences.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return fmt.Errorf("convert event keys: %w", err)
+	}
+	return nil
+}
+
+func (conv *DebeziumValueConverter) convertMap(tableName string, m map[string]string) error {
+	for column, value := range m {
+		if value == "" {
+			m[column] = "NULL"
```
amit-yb, line 131: Nope!! Consider an update event that cleared value of a string column by setting it to `""`. This line will result in writing `"NULL"` in the target table.

Debezium plugin must make the distinction of whether to use empty string or `"NULL"`. Voyager should not handle this case.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,150 @@
...
+			continue
+		}
+		columnValue := fmt.Sprintf("%v", value)
```
amit-yb, line 134: `value` is already a string!

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,150 @@
...
+		colType, err := conv.schemaRegistry.GetColumnType(tableName, column)
+		if err != nil {
+			return fmt.Errorf("fetch column schema: %w", err)
+		}
+		converterFn := conv.valueConverterSuite[*colType]
+		if converterFn == nil {
+			converterFn = func(v string, formatIfRequired bool) (string, error) { return v, nil }
+		}
+		columnValue, err = converterFn(columnValue, true)
```
amit-yb, line 143: Invert the condition:
```
if converterFn != nil {
  columnValue, err = converterFn(columnValue, true)
}
```

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -774,3 +777,180 @@ func (yb *TargetYugabyteDB) recordEntryInDB(tx pgx.Tx, batch Batch, rowsAffected
 	}
 	return nil
 }
+
+func (yb *TargetYugabyteDB) GetDebeziumValueConverterSuite(snapshotMode bool) map[string]ConverterFn {
+	converterValueSuite := make(map[string]ConverterFn)
```
amit-yb, line 782: Declare and populate `ybValueConverterSuite` at the top of this file. Only return the reference to `ybValueConverterSuite` map from the `GetDebeziumValueConverterSuite()`. That way, no matter how many times the function is called, only one map is created.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -774,3 +777,180 @@ func (yb *TargetYugabyteDB) recordEntryInDB(tx pgx.Tx, batch Batch, rowsAffected
...
+	converterValueSuite["io.debezium.time.Date"] = func(columnValue string, formatIfRequired bool) (string, error) {
```
amit-yb, line 783: I will skip reviewing individual converter functions and trust Aneesh's opinion on this.

yb-voyager/cmd/importData.go
```diff
@@ -393,6 +404,13 @@ func splitFilesForTable(state *ImportDataState, filePath string, t string,
 			// handling possible case: last dataline(i.e. EOF) but no newline char at the end
 			numLinesTaken += 1
 		}
+		if line != "" && isDebeziumExport {
+			table := batchWriter.tableName
+			line, err = valueConverter.ConvertRow(table, TableToColumnNames[table], line)
```
amit-yb, line 409: We don't need the `isDebeziumExport` flag, the `TableToColumnNames` map, and the `prepareTableToColumns()` function.

Just take the column names from the `ImportBatchArgs`. Also, it is okay to call the `ConvertRow()` for non-dbzm exports as well--because in that case it will be a no-op converter.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("initializing schema registry: %w", err)
+	}
+	tdbValueConverterSuite := tdb.GetDebeziumValueConverterSuite(snapshotMode)
+
+	return &DebeziumValueConverter{
+		schemaRegistry:      schemaRegistry,
+		valueConverterSuite: tdbValueConverterSuite,
+		converterFnCache:    map[string][]tgtdb.ConverterFn{},
+		snapshotMode:        snapshotMode,
```
amit-yb, line 71: `snapshotMode` is not used anymore. Get rid of it.

yb-voyager/cmd/live_migration.go
```diff
@@ -55,6 +56,10 @@ func streamChangesFromSegment(segment *EventQueueSegment) error {
 	}
 	defer segment.Close()
 	log.Infof("streaming changes for segment %s", segment.FilePath)
+	valueConverter, err = dbzm.NewValueConverter(exportDir, tdb, false) //streaming valueConverter
```
amit-yb, line 59: Once you drop the `snapshotMode`, you won't need to re-instantiate a new converter.

amit-yb (commented):
Second batch of comments. We're in the last mile! :-)

yb-voyager/src/tgtdb/event.go
```diff
@@ -54,7 +54,7 @@ func (event *Event) getInsertStmt(targetSchema string) string {
 	valueList := make([]string, 0, len(event.Fields))
 	for column, value := range event.Fields {
 		columnList = append(columnList, column)
-		valueList = append(valueList, value)
+		valueList = append(valueList, fmt.Sprintf("%v", value))
```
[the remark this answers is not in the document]
priyanshi-yb: Oh yes, missed that. now we are getting string so no need for this conversion! thanks!

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,150 @@
...
+//============================================================================
+
+type DebeziumValueConverter struct {
+	schemaRegistry      *SchemaRegistry
+	valueConverterSuite map[string]tgtdb.ConverterFn
+	converterFnCache    map[string][]tgtdb.ConverterFn
+	snapshotMode        bool
+}
+
+func NewDebeziumValueConverter(exportDir string, tdb tgtdb.TargetDB, snapshotMode bool) (*DebeziumValueConverter, error) {
```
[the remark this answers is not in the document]
priyanshi-yb: Actually no need for this. I missed removing this. now we have modified ConverterFn for this purpose. will remove it.

yb-voyager/cmd/live_migration.go
priyanshi-yb: Oh yeah!

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,150 @@
...
+	columnValues := strings.Split(row, "\t")
+	for i, columnValue := range columnValues {
+		if columnValue != "\\N" {
+			transformedValue, err := converterFns[i](columnValue, false)
+			if err != nil {
+				return "", fmt.Errorf("converting value: %w", err)
+			}
+			columnValues[i] = transformedValue
+		} else {
+			columnValues[i] = columnValue
```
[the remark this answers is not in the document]
priyanshi-yb: added a TODO.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -0,0 +1,111 @@
...
+
+func NewSchemaRegistry(exportDir string) *SchemaRegistry {
+	return &SchemaRegistry{
+		exportDir:     exportDir,
+		tableNameToSchema: make(map[string]*TableSchema),
+	}
+}
+
+func (sreg *SchemaRegistry) GetColumnTypes(tableName string, columnNames []string) ([]string, error) {
+	tableSchema := sreg.tableNameToSchema[tableName]
```
amit-yb, line 72: Return an error if the `tableName` is not present in the map.

yb-voyager/cmd/importData.go
```diff
@@ -393,6 +381,13 @@ func splitFilesForTable(state *ImportDataState, filePath string, t string,
 			// handling possible case: last dataline(i.e. EOF) but no newline char at the end
 			numLinesTaken += 1
 		}
+		if line != "" {
+			table := batchWriter.tableName
+			line, err = valueConverter.ConvertRow(table, TableToColumnNames[table], line)
+			if err != nil {
+				utils.ErrExit("transforming line for table %q: %s", t, err)
```
amit-yb, line 388: Add the file name and line number in the log message.

yb-voyager/cmd/importData.go
```diff
@@ -393,6 +381,13 @@ func splitFilesForTable(state *ImportDataState, filePath string, t string,
 			// handling possible case: last dataline(i.e. EOF) but no newline char at the end
 			numLinesTaken += 1
 		}
+		if line != "" {
+			table := batchWriter.tableName
+			line, err = valueConverter.ConvertRow(table, TableToColumnNames[table], line)
```
amit-yb, line 386: Add a comment stating why we can't use `importBatchArgsProto.Columns`.

yb-voyager/cmd/live_migration.go
```diff
@@ -55,6 +55,9 @@ func streamChangesFromSegment(segment *EventQueueSegment) error {
 	}
 	defer segment.Close()
 	log.Infof("streaming changes for segment %s", segment.FilePath)
+	if err != nil {
+		return fmt.Errorf("error creating value converter: %v", err)
+	}
```
amit-yb, line 60: Yikes!!! 😬

amit-yb (commented):
One more small batch of comments.

yb-voyager/src/tgtdb/event.go
```diff
@@ -16,16 +16,17 @@ limitations under the License.
...
-	Op         string            `json:"op"`
-	SchemaName string            `json:"schema_name"`
-	TableName  string            `json:"table_name"`
-	Key        map[string]string `json:"key"`
-	Fields     map[string]string `json:"fields"`
+	Op         string                    `json:"op"`
+	SchemaName string                    `json:"schema_name"`
+	TableName  string                    `json:"table_name"`
+	Key        map[string]sql.NullString `json:"key"`
+	Fields     map[string]sql.NullString `json:"fields"`
```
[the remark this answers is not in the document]
priyanshi-yb: yes, debezium does that,  it comes as `"column_name": null`.

yb-voyager/cmd/live_migration.go
priyanshi-yb: Oops. thanks!

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (conv *DebeziumValueConverter) ConvertRow(tableName string, columnNames []string, row string) (string, error) {
+	converterFns, err := conv.getConverterFns(tableName, columnNames)
+	if err != nil {
+		return "", fmt.Errorf("fetching converter functions: %w", err)
+	}
+	columnValues := strings.Split(row, "\t")
+	for i, columnValue := range columnValues {
+		if columnValue == "\\N" || converterFns[i] == nil { // TODO: make "\\N" condition Target specific tdb.NullString()
+			columnValues[i] = columnValue
```
amit-yb, line 81: Nit: This is effectively a no-op.

amit-yb (approved):
LGTM. Good work!! 👏

sanyamsinghal (approved):
LGTM. Nice work

PR #947, opened 2023-07-10
fixed a bug in import data file with CSV format, header and delimiter other than ','

priyanshi-yb:
Bug - In import data file, CSV datafile with header and delimiter other than a comma (,) will return a syntax error in copy as when a datafile with header has delimiter other than a comma, the columns in the header will be separated with delimiter and hence not with a comma and we are using the same as is in the COPY command as columns assuming that they will be separated with a comma. 
Fix - create comma-separated columns explicitly and pass those in COPY command.


priyanshi-yb:
Resolved in #959

PR #950, opened 2023-07-10
Fixed Oracle test

shubham-yb:
Due to a trigger placed on the table, the inserts were only allowed on certain times. Due to this the test was failing since the Jenkins job ran at night. Removing the check for now since we already have a test which validates it.


makalaaneesh (approved):
LGTM

PR #951, opened 2023-07-11
Error out in case Voyager 1.4 is used to continue the migration started by lower versions (<=1.3)

priyanshi-yb:


amit-yb (approved):
LGTM.

PR #953, opened 2023-07-13
Added a gh actions file to build binaries for voyager and made relevant changes in the installer script

ShivanshGahlot:


installer_scripts/install-yb-voyager
```diff
@@ -314,11 +315,26 @@ update_yb_voyager_bashrc() {
 
 
 install_yb_voyager() {
+	os=$1
 	output "Installing yb-voyager:${VERSION}."
 	if [ "${VERSION}" == "latest" ]
 	then
-		$GO install github.com/yugabyte/yb-voyager/yb-voyager@${YB_VOYAGER_GIT_HASH}
```
amit-yb, line 320: After this do we still need `install_golang`?

installer_scripts/install-yb-voyager
```diff
@@ -314,11 +315,26 @@ update_yb_voyager_bashrc() {
...
+			arch="$(uname -m)"
+			if [ "${arch}" == "arm64" ]
+			then
+				binary_name="voyager-darwin-arm64"
+			else
+				binary_name="voyager-darwin-amd64"
+			fi
+		fi
+
+		wget https://github.com/yugabyte/yb-voyager/releases/download/yb-voyager%2Fv${VOYAGER_RELEASE_TAG}/${binary_name} 1>&2
```
amit-yb, line 336: A voyager "release" must be a TAR file. And we should be able to bundle as many files in it as we want (e.g. base_ora2pg.conf).

installer_scripts/install-yb-voyager
ShivanshGahlot: We should not I guess. We will be just running the binary.

PR #961, opened 2023-07-19
Oracle CDB support for Debezium Live Migration

makalaaneesh:
This adds new params to exportData
- oracle-cdb-name
- oracle-cdb-tns-alias
- oracle-cdb-sid

which allow specifying the cdb identifiers needed to perform live migration with debezium. 
One constraint is:
if specifying cdb identifiers, then the only way you can specify the pdb is via --source-db-name (because that's what debezium allows)

### DIFF
--- a/yb-voyager/cmd/export.go
+++ b/yb-voyager/cmd/export.go
@@ -79,8 +79,11 @@ func registerCommonExportFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&source.CDBName, "oracle-cdb-name", "",
 		"[For Oracle Only] Oracle Container Database Name in case you are using a multitenant container database.")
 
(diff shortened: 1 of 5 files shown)


yb-voyager/cmd/export.go
```diff
@@ -229,6 +238,23 @@ func validateOracleParams() {
 	} else if source.DBSid != "" {
 		utils.PrintAndLog("Using SID for export.")
 	}
+	if source.IsOracleCDBSetup() {
```
amit-yb, line 241: How to convey that we don't need these CLI arguments for offline migration.

amit-yb (approved):
LGTM. 👍

yb-voyager/cmd/export.go
makalaaneesh: Can add it to the command help doc. Additionally we can also add it to our voyager documentation on the web.

PR #970, opened 2023-07-31
[export data] Support in exporter for YugabyteDB to stream changes

priyanshi-yb:
[DB-7364](https://yugabyte.atlassian.net/browse/DB-7364) 

For stream id generation, created a yb-client-wrapper java jar to create stream id using yb-client APIs and packaged it with debezium-server (PR - https://github.com/yugabyte/debezium/pull/72).


yb-voyager/src/srcdb/yugabytedb.go
makalaaneesh, line 1: I assume this is simply a copy of postgres? lmk if there were any changes you made!

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -64,12 +76,67 @@ func findDebeziumDistribution() error {
 	return nil
 }
 
+func GetYugabyteDBStreamID(config Config) (string, error) {
```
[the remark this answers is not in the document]
makalaaneesh: Alternatively, name this func `generateAndStoreYbStreamID` ! :)

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -64,12 +76,67 @@ func findDebeziumDistribution() error {
...
+		if outbuf.String() != "" {
+			log.Infof("Output of the command %s: %s", command, outbuf.String())
+		}
+		log.Infof("Failed to wait for command: %s , error: %s", command, err)
+		return "", err
+	}
+	//output of yb-admin command - CDC Stream ID: <stream_id>
+	rgx := regexp.MustCompile(`CDC Stream ID: ([a-zA-Z0-9]+)`)
+	matches := rgx.FindStringSubmatch(outbuf.String())
+	if len(matches) != 2 {
```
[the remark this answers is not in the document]
priyanshi-yb: first one is the complete string matched with the regex string (e.g. "CDC Stream ID: <stream_id>") and the second one will be the match of the expression matching for the stream id (e.g. "<stream_id>").

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -180,3 +247,7 @@ func (d *Debezium) Stop() error {
 	}
 	return nil
 }
+
```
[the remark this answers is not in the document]
priyanshi-yb: Oops! I missed removing this, was trying out something. Thanks for pointing!

yb-voyager/src/dbzm/config.go
```diff
@@ -173,6 +174,31 @@ debezium.source.database.ssl.truststore=%s
...
+debezium.source.database.dbname=%s
+debezium.source.database.streamid=%s
+debezium.source.database.master.addresses=%s
+debezium.source.schema.include.list=%s
+debezium.source.hstore.handling.mode=map
+debezium.source.converters=postgres_source_converter
+debezium.source.postgres_source_converter.type=io.debezium.server.ybexporter.PostgresToYbValueConverter
+debezium.source.transforms=unwrap
+debezium.source.transforms.unwrap.type=io.debezium.connector.yugabytedb.transforms.PGCompatible
+debezium.source.transforms.extract.drop.tombstones=false
```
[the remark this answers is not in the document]
priyanshi-yb: yes, actually this has to be `unwrap.drop.tombstones` but deletes are working fine without this as well so not required.

yb-voyager/src/srcdb/postgres.go
```diff
@@ -459,3 +459,7 @@ func generateSSLQueryStringIfNotExists(s *Source) string {
 		return ""
 	}
 }
+
+func (pg *PostgreSQL) GetServers() string {
```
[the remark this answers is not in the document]
priyanshi-yb: Hmm, makes sense!

yb-voyager/src/srcdb/yugabytedb.go
priyanshi-yb: Yes it's mostly the same as PG, but the difference is I have removed a few functions that are not required for export data for yugabytedb. Let me think on this and try if I can just use PG one for source-type yb.

makalaaneesh: To clarify, what I was thinking is: 
```
type YugabyteDB struct{
    PostgresDB
}

func () GetServers (){

}
```
That way you get to re-use all the methods of postgresDB struct, (maintaining the TargetDB interface) and just add whatever is YB-specific. 

I'm not sure though, because if you've made other significant changes (maybe connection related), or some deletions, then you may be better off just copy pasting for now (we can take it up later)

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -0,0 +1,409 @@
...
+
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/datafile"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils/sqlname"
+)
+
+type YugabyteDB struct {
+	source *Source
+
+	db *pgx.Conn
```
amit-yb, line 38: Rename `db` to `conn`.

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -0,0 +1,409 @@
...
+
+func (yb *YugabyteDB) Disconnect() {
+	if yb.db == nil {
+		log.Infof("No connection to the source database to close")
+		return
+	}
+
+	err := yb.db.Close(context.Background())
+	if err != nil {
+		log.Infof("Failed to close connection to the source database: %s", err)
```
amit-yb, line 59: `log.Errorf`

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+}
+
+func (yb *YugabyteDB) CheckRequiredToolsAreInstalled() {
+	checkTools("strings")
+}
+
+func (yb *YugabyteDB) GetTableRowCount(tableName string) int64 {
```
amit-yb, line 67: This and many other methods are applicable only for offline migration. Do we need to implement them for YB?

yb-voyager/cmd/exportData.go
```diff
@@ -312,6 +311,20 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName, ta
 		if err != nil {
 			return fmt.Errorf("failed to determine if Oracle JDBC wallet location is set: %v", err)
 		}
+	} else if source.DBType == "yugabytedb" {
+		config.YBServers = source.DB().GetServers()
+		if startClean {
```
amit-yb, line 316: What about starting exporter for the first time (when --start-clean flag is not provided)?

This code should follow code structure as below:
```
if migrationType == "CHANGES_ONLY" {
  config.YBStreamID, err = dbzm.ReadYBStreamID(exportDir)
  if err is stream id not found {
    config.YBStreamID, err = dbzm.GenerateAndStoreYBStreamID(*config)
    if err != nil {
      ...
    }
    utils.PrintAndLog("Generated YugabyteDB CDC stream-id: %s", ...)
  } else if err != nil {
    return fmt.Errorf(...)
  }

}
```

yb-voyager/cmd/exportData.go
```diff
@@ -312,6 +311,20 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName, ta
...
+			config.YBStreamID, err = dbzm.GenerateAndStoreYBStreamID(*config)
```
amit-yb, line 317: You should change the `GenerateAndStoreYBStreamID()` to accept pointer to `Config`. That way you won't need to send config by-value here.

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -64,12 +76,80 @@ func findDebeziumDistribution() error {
...
+func ReadYBStreamID(exportDir string) (string, error) {
+	streamIDFilePath := filepath.Join(exportDir, "metainfo", "yb_cdc_stream_id.txt")
+	if utils.FileOrFolderExists(streamIDFilePath) {
+		file, err := os.ReadFile(streamIDFilePath)
+		if err != nil {
+			return "", fmt.Errorf("failed to read stream id file: %w", err)
+		}
+		return string(file), nil
+	}
+	return "", fmt.Errorf("yugabytedb cdc stream id not found at %s. Please use --start-clean to start the streaming", streamIDFilePath)
```
amit-yb, line 143: If stream-id is not found, the code should generate one. No need of passing `--start-clean`.

If `--start-clean` is passed and there is already existing stream-id, we should see if we can "delete" the stream-id from the server itself.

amit-yb (commented):
Just a few minor comments. Overall very good work! We can reuse some of this work if we need to support YB as source DB.

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -64,12 +76,80 @@ func findDebeziumDistribution() error {
...
 }
 
+func GenerateAndStoreYBStreamID(config Config) (string, error) {
+	err := findDebeziumDistribution(config.SourceDBType)
+	if err != nil {
+		return "", fmt.Errorf("error in finding debezium distribution: %s", err)
+	}
+	YB_CLIENT_WRAPPER_JAR := fmt.Sprintf("%s/yb-client-cdc-stream-wrapper.jar", DEBEZIUM_DIST_DIR) //$DEBEZIUM_DIST/yb-client-cdc-stream-wrapper.jar
+	tableName := strings.Split(config.TableList[0], ".")[1]                                        //any table name in the database is required by yb-client createCDCStream(...) API
+	command := fmt.Sprintf("java -jar %s -master_addresses %s -table_name %s -db_name %s ", YB_CLIENT_WRAPPER_JAR, config.YBServers, tableName, config.DatabaseName)
```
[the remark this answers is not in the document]
priyanshi-yb: This table name is just one table in that database which is required by yb-client's `createCDCStream()` API to route the RPI call from a table on the tserver and has no effect on the stream id, it is always per database we are passing which means all the tables in that database. I have confirmed this from CDC team.

yb-voyager/src/dbzm/config.go
```diff
@@ -210,7 +234,27 @@ func (c *Config) String() string {
...
+			c.Username,
+			c.Password,
+			"never",
+			offsetFile,
+			strings.Join(c.TableList, ","),
 
+			c.Host, c.Port,
+			c.DatabaseName,
+			c.YBStreamID,
+			c.YBServers,
```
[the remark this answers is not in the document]
priyanshi-yb: master nodes

yb-voyager/src/srcdb/yugabytedb.go
priyanshi-yb: No, should I remove them or leave it for now?

yb-voyager/src/srcdb/yugabytedb.go
priyanshi-yb: I think, we can take this up later.

yb-voyager/src/dbzm/dbzm.go
priyanshi-yb: added delete option for start-clean

yb-voyager/src/dbzm/config.go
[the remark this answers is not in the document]
amit-yb: The `c.YBServers` is set to ALL nodes of the cluster. `SourceDB.GetServers()` is returning all cluster nodes.

yb-voyager/src/dbzm/config.go
```diff
@@ -55,8 +55,9 @@ type Config struct {
 	SSLKeyStorePassword   string
 	SSLTrustStore         string
 	SSLTrustStorePassword string
-
-	SnapshotMode string
+	YBStreamID            string
+	YBServers             string
```
amit-yb, line 59: If you really want this variable to represent master nodes only then choose the name to reflect it e.g. YBMasterNodes

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -42,14 +45,22 @@ type Debezium struct {
...
+func findDebeziumDistribution(sourceDBType string) error {
 	if distDir := os.Getenv("DEBEZIUM_DIST_DIR"); distDir != "" {
 		DEBEZIUM_DIST_DIR = distDir
 	} else {
-		possiblePaths := []string{
-			"/opt/homebrew/Cellar/debezium@" + DEBEZIUM_VERSION + "/" + DEBEZIUM_VERSION + "/debezium-server",
-			"/usr/local/Cellar/debezium@" + DEBEZIUM_VERSION + "/" + DEBEZIUM_VERSION + "/debezium-server",
-			"/opt/yb-voyager/debezium-server"}
+		var possiblePaths []string
+		if sourceDBType == "yugabytedb" {
```
amit-yb, line 53: Nit: You can simplify the code as
```
pathSuffix := ""
if sourceDBType == YUGABYTEDB {
  pathSuffix = "/debezium-server-fall-forward"
}
possiblePaths = []string{
  fmt.Sprintf(".../debezium@%s/%s/debezium-server/debezium-server%s", DEBEZIUM_VERSION, DEBEZIUM_VERSION, pathSuffix),
  ...
  ...
}
```

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -64,12 +75,111 @@ func findDebeziumDistribution() error {
 	return nil
 }
 
+func GenerateAndStoreYBStreamID(config *Config) (string, error) {
+	err := findDebeziumDistribution(config.SourceDBType)
+	if err != nil {
+		return "", fmt.Errorf("error in finding debezium distribution: %s", err)
+	}
+	YB_CLIENT_WRAPPER_JAR := fmt.Sprintf("%s/yb-client-cdc-stream-wrapper.jar", DEBEZIUM_DIST_DIR) //$DEBEZIUM_DIST/yb-client-cdc-stream-wrapper.jar
```
amit-yb, line 83: Nit: the comment is not adding any value. Drop it.

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -64,12 +75,111 @@ func findDebeziumDistribution() error {
...
+	tableName := strings.Split(config.TableList[0], ".")[1]                                        //any table name in the database is required by yb-client createCDCStream(...) API
```
amit-yb, line 84: Are we sure that the elements in the `config.TableList` are always qualified by the schema name?

How would the server find the table if it is in non-public schema and the table-name is not qualified?

Nit: avoid having arbitrary long lines. You can always put a comment just above the line it is describing. e.g.
```
// Any table name in the database is required by yb-client createCDCStream(...) API.
tableName := strings.Split(config.TableList[0], ".")[1]
```

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -64,12 +75,111 @@ func findDebeziumDistribution() error {
...
+	command := fmt.Sprintf("java -jar %s -create -master_addresses %s -table_name %s -db_name %s ", YB_CLIENT_WRAPPER_JAR, config.YBServers, tableName, config.DatabaseName)
```
amit-yb, line 85: Try limiting line length to below 100 columns.

You can add a line break after the format string and pass the arguments on the second line.

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -64,12 +75,111 @@ func findDebeziumDistribution() error {
...
+	var errbuf bytes.Buffer
+	cmd.Stdout = &outbuf
+	cmd.Stderr = &errbuf
+	err = cmd.Start()
+	if err != nil {
+		if outbuf.String() != "" {
+			log.Infof("Output of the command %s: %s", command, outbuf.String())
+		}
+		log.Infof("Failed to start command: %s, error: %s", command, err)
+		return "", err
```
amit-yb, line 102: Wrap the error telling which command failed.

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -64,12 +75,111 @@ func findDebeziumDistribution() error {
...
+	var outbuf bytes.Buffer
+	var errbuf bytes.Buffer
+	cmd.Stdout = &outbuf
+	cmd.Stderr = &errbuf
+	err = cmd.Start()
+	if err != nil {
+		if outbuf.String() != "" {
+			log.Infof("Output of the command %s: %s", command, outbuf.String())
+		}
+		log.Infof("Failed to start command: %s, error: %s", command, err)
```
amit-yb, line 101: `log.Errorf()`. Fix all such occurrences.

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -64,12 +75,111 @@ func findDebeziumDistribution() error {
...
+		return "", err
+	}
+	//output of yb-admin command - CDC Stream ID: <stream_id>
+	output := outbuf.String()
+	streamID := strings.Trim(strings.Split(output, ":")[1], " \n")
+	//save streamID in a file
+	streamIDFile := filepath.Join(config.ExportDir, "metainfo", "yb_cdc_stream_id.txt")
+	file, err := os.Create(streamIDFile)
+	if err != nil {
+		return "", fmt.Errorf(" creating file: %s, error: %s", streamIDFile, err)
```
amit-yb, line 119: Avoid a white space after the opening double-quote `" creating..."`. Fix all such cases.

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -64,12 +75,111 @@ func findDebeziumDistribution() error {
 	return nil
 }
 
+func GenerateAndStoreYBStreamID(config *Config) (string, error) {
```
amit-yb, line 78: If possible consider defining a new class in a new file `cdc-client.go`
```
type YBCDCClient struct {

}

func NewYBCDCClient(exportDir, masterNodes, dbName, tableName string) *YBCDCClient {

}

func (YBCDCClient *c) Init() error {
  // Locate the distribution and client wrapper jar. Store the paths as
  // private fields of YBCDCClient.
} 
// If existing stream ID is not found, generate one. Else, return the existing stream ID.
func (YBCDCClient *c) GetStreamID() (string, error)
func (YBCDCClient *c) DeleteStreamID(streamID string) error

// A private method to invoke the wrapper jar. `c.runCommand()` will be called
// from both `GetStreamID()` and `DeleteStreamID()`.
func (YBCDCClient *c) runCommand(args string) (stdout string, err error)
```

amit-yb (commented):
Few more nits.

yb-voyager/src/dbzm/config.go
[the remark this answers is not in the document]
priyanshi-yb: yes, we are sending all nodes here, as I couldn't find a way to just get the master nodes. So, thought it's better to pass all the nodes. 
Though there is one API/command with yb-client/yb-admin to list all masters but that needs master addresses.

yb-voyager/src/dbzm/dbzm.go
priyanshi-yb: > How would the server find the table if it is in non-public schema and the table-name is not qualified?

For this, on the yb-client wrapper code side what is happening is we are doing a rest call to the server to get the list of tables and then we are the fetching the one with this tableName to get the object of some YBTable struct and only that object can be passed to createCDCStream() API as the table. ([refer](https://github.com/yugabyte/debezium/pull/72/files#diff-7bd741feeba6787d8e87e9380af921ff8a3e4ccc3d4841fde2080bd93bc817f5)) 
But I see a point to specify the schema name as well to jar and match the qualified name. Will make that change.

amit-yb (approved):
LGTM. Just take care of sending correct master node list to the CDC wrapper command.

yb-voyager/cmd/exportData.go
```diff
@@ -317,6 +316,32 @@ func debeziumExportData(ctx context.Context, tableList []*sqlname.SourceName, ta
 		if err != nil {
 			return fmt.Errorf("failed to determine if Oracle JDBC wallet location is set: %v", err)
 		}
+	} else if source.DBType == "yugabytedb" {
+		if liveMigration { //TODO: for migration type CHANGES_ONLY
+			ybServers := source.DB().GetServers()
+			ybCDCClient := dbzm.NewYugabyteDBCDCClient(exportDir, ybServers, config.SSLRootCert, config.DatabaseName, config.TableList[0])
+			ybCDCClient.Init()
```
amit-yb, line 323: Handle error from the `Init()`.

yb-voyager/src/dbzm/yb_cdc_client.go
```diff
@@ -0,0 +1,176 @@
...
+	_, err = ybc.runCommand(args)
+	if err != nil {
+		return fmt.Errorf("running command with args: %s, error: %s", args, err)
+	}
+	utils.PrintAndLog("Deleted YugabyteDB CDC stream-id: %s", streamID)
+	return nil
+}
+
+func (ybc *YugabyteDBCDCClient) ListMastersNodes() (string, error) {
+	args := fmt.Sprintf("-list_masters -master_addresses %s ", ybc.masterAddreses)
```
amit-yb, line 135: Is it okay to pass both master and non-master nodes to the `-master_addresses` argument?

priyanshi-yb: There is no problem as such in this func as we are just listing the master nodes and it is working fine I tried so seems to be no problem there, but I think that to be on the safer side I should pass only master nodes to create and delete APIs after listing them so that there can never be a problem in future with those two APIs as they are crucial for streaming. Makes sense right?

PR #972, opened 2023-08-01
parallel streaming import

makalaaneesh:


yb-voyager/cmd/live_migration.go
```diff
@@ -18,20 +18,43 @@ package cmd
...
+var END_OF_SOURCE_QUEUE_SEGMENT_EVENT = &tgtdb.Event{Op: "end_of_source_queue_segment"}
+
 func streamChanges() error {
-	eventQueue := NewEventQueue(exportDir)
-	log.Infof("streaming changes from %s", eventQueue.QueueDirPath)
+	sourceEventQueue := NewSourceEventQueue(exportDir)
+	// setup target channels
+	var targetEventChans []chan *tgtdb.Event
+
+	// targetEventChans = make([]chan *tgtdb.Event, NUM_TARGET_EVENT_CHANNELS)
```
amit-yb, line 43: Delete.

yb-voyager/cmd/live_migration.go
```diff
@@ -18,20 +18,43 @@ package cmd
...
+	var eventProcessingDoneChans []chan bool
+	for i := 0; i < NUM_TARGET_EVENT_CHANNELS; i++ {
+		targetEventChans = append(targetEventChans, make(chan *tgtdb.Event, TARGET_EVENT_CHANNEL_SIZE))
+		eventProcessingDoneChans = append(eventProcessingDoneChans, make(chan bool))
+	}
+	// eventProcessingDoneChans = make([]chan bool, NUM_TARGET_EVENT_CHANNELS)
```
amit-yb, line 49: Delete.

yb-voyager/cmd/live_migration.go
```diff
@@ -18,20 +18,43 @@ package cmd
...
+	// start target event channel processors
+	for i := 0; i < NUM_TARGET_EVENT_CHANNELS; i++ {
+		go process_events_from_target_channel(i, targetEventChans[i], eventProcessingDoneChans[i])
```
amit-yb, line 52: Use camelCasing for identifiers.

Just `processEvents` should be enough.

yb-voyager/cmd/live_migration.go
```diff
@@ -88,10 +116,60 @@ func handleEvent(event *tgtdb.Event) error {
...
+func insertIntoTargetEventChan(e *tgtdb.Event, targetEventChans []chan *tgtdb.Event) {
+	// hash event
+	delimiter := "-"
+	stringToBeHashed := e.SchemaName + delimiter + e.TableName
+	for _, value := range e.Key {
+		stringToBeHashed += delimiter + *value
+	}
+	hash := fnv.New64a()
+	hash.Write([]byte(stringToBeHashed))
+	hashValue := hash.Sum64()
```
amit-yb, line 132: Too many memory allocations!! Instead of building one giant string to be hashed, can we write each individual string into the `hash` object?
No need of delimiter.
Ordering of keys inside the `e.Key` matters.
I assume the hash function is not too compute intensive.

yb-voyager/cmd/live_migration.go
```diff
@@ -88,10 +116,60 @@ func handleEvent(event *tgtdb.Event) error {
...
+				if event == END_OF_SOURCE_QUEUE_SEGMENT_EVENT {
+					endOfProcessing = true
+					batchReady = true
+				}
+				batch = append(batch, event)
+				if len(batch) >= MAX_EVENTS_PER_BATCH {
+					// utils.PrintAndLog("reached max events per batch")
+					batchReady = true
+				}
+			case <-time.After(time.Duration(MAX_TIME_PER_BATCH) * time.Millisecond):
```
amit-yb, line 159: We need one more enhancement. The current design sends the batch if no event is received on the channel for 2 seconds (even if the batch is not full). We want to send a batch in two cases:
- batch is full
- every 5 seconds (irrespective of when the last event was received on the channel).

sanyamsinghal (commented):
Some review comments based on my understanding of design and implementation.

yb-voyager/cmd/live_migration.go
makalaaneesh: Makes sense. With current implementation, we could end up in a situation where we get an event in the channel every 1 second, but will only get executed once we reach MAX_EVENTS_PER_BATCH.

yb-voyager/cmd/eventQueue.go
```diff
@@ -34,33 +34,33 @@ const (
 	QUEUE_SEGMENT_FILE_EXTENSION = "ndjson"
 )
 
-type EventQueue struct {
+type SourceEventQueue struct {
```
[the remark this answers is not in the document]
makalaaneesh: Hmm, I added it becausee I think it still adds value. "EventQueue" just by itself isn't clear which queue it is. We have a queue of events coming from the source, and a queue of events (split into channels) that we want to apply on the target. :) 

What do you think?

yb-voyager/cmd/live_migration.go
```diff
@@ -88,10 +116,60 @@ func handleEvent(event *tgtdb.Event) error {
...
+		batch := []*tgtdb.Event{}
+		batchReady := false
+		for !batchReady {
+			// read from channel until MAX_EVENTS_PER_BATCH or MAX_TIME_PER_BATCH
+			select {
+			case event := <-targetEventChan:
+				// utils.PrintAndLog("read event %v from channel", event)
+				if event == END_OF_SOURCE_QUEUE_SEGMENT_EVENT {
+					endOfProcessing = true
+					batchReady = true
```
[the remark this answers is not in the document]
makalaaneesh: the break only breaks out of the select apparently! - https://stackoverflow.com/a/11105482/4434664

yb-voyager/cmd/live_migration.go
```diff
@@ -88,10 +116,60 @@ func handleEvent(event *tgtdb.Event) error {
...
-	batch := []*tgtdb.Event{event}
-	err = tdb.ExecuteBatch(batch)
-	if err != nil {
-		return fmt.Errorf("error executing batch: %v", err)
-	}
+	insertIntoTargetEventChan(event, targetEventChans)
 	return nil
 }
+
+func insertIntoTargetEventChan(e *tgtdb.Event, targetEventChans []chan *tgtdb.Event) {
```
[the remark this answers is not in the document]
makalaaneesh: Thought about this, but this is not a generic hash of the event. This is a hash (only of the primary key values essentially) specifically to insert into a targetEventChan, so I thought it'd better be a method in the live_migration.go, as opposed to be on the Event struct.

yb-voyager/cmd/live_migration.go
makalaaneesh: didn't do too much research into hash functions, but the fnv was designed to be fast to compute. Didn't stand out in the profile either. 

https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#Non-cryptographic_hash

yb-voyager/cmd/live_migration.go
```diff
@@ -88,10 +116,60 @@ func handleEvent(event *tgtdb.Event) error {
...
+	// hash event
+	delimiter := "-"
+	stringToBeHashed := e.SchemaName + delimiter + e.TableName
```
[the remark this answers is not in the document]
makalaaneesh: Nice!. In this case though, I am going to use multiple hash.Writes (instead of concating a string) which is essentially just writing byte buffers, so I think this should be convered

yb-voyager/cmd/live_migration.go
```diff
@@ -18,20 +18,43 @@ package cmd
...
 	"time"
 
 	log "github.com/sirupsen/logrus"
 
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/tgtdb"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
 )
 
+var NUM_TARGET_EVENT_CHANNELS = 64
+var TARGET_EVENT_CHANNEL_SIZE = 2000 // has to be > MAX_EVENTS_PER_BATCH
```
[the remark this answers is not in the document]
makalaaneesh: Yeah, my thought process was - 
there's enough buffer such that if execute of batch is complete, immediately, you can create and execute the next batch.

yb-voyager/cmd/live_migration.go
```diff
@@ -18,20 +18,43 @@ package cmd
...
+var MAX_EVENTS_PER_BATCH = 100
```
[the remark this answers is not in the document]
makalaaneesh: Good point, let me allow env var override

yb-voyager/cmd/eventQueue.go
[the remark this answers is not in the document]
amit-yb: The `EventQueue` can have events coming from source-db, target-db, and in future maybe from fall-forward-db as well. In that case, the SourceEventQueue is misleading.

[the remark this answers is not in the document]
makalaaneesh: Ah, good point, I see what you mean. `ExportedEventQueue` or `GoldenReplicaEventQueue` is probably a more technically correct name :D . For now, will revert.

yb-voyager/cmd/live_migration.go
```diff
@@ -18,17 +18,41 @@ package cmd
...
+var NUM_TARGET_EVENT_CHANNELS int
+var TARGET_EVENT_CHANNEL_SIZE int // has to be > MAX_EVENTS_PER_BATCH
+var MAX_EVENTS_PER_BATCH int
+var MAX_TIME_PER_BATCH int //ms
+var END_OF_SOURCE_QUEUE_SEGMENT_EVENT = &tgtdb.Event{Op: "end_of_source_queue_segment"}
+
+func init() {
+	NUM_TARGET_EVENT_CHANNELS = utils.GetEnvAsInt("NUM_TARGET_EVENT_CHANNELS", 512)
+	TARGET_EVENT_CHANNEL_SIZE = utils.GetEnvAsInt("TARGET_EVENT_CHANNEL_SIZE", 2000)
+	MAX_EVENTS_PER_BATCH = utils.GetEnvAsInt("MAX_EVENTS_PER_BATCH", 100)
```
amit-yb, line 42: Let's set this to 2000.

yb-voyager/cmd/live_migration.go
```diff
@@ -65,19 +95,26 @@ func streamChangesFromSegment(segment *EventQueueSegment) error {
...
-		err = handleEvent(event)
+		err = handleEvent(event, evChans)
 		if err != nil {
 			return fmt.Errorf("error handling event: %v", err)
 		}
 	}
 
-	log.Infof("finished streaming changes from segment %s", segment.FilePath)
+	for i := 0; i < NUM_TARGET_EVENT_CHANNELS; i++ {
+		evChans[i] <- END_OF_SOURCE_QUEUE_SEGMENT_EVENT
```
amit-yb, line 105: END_OF_QUEUE_SEGMENT ?

yb-voyager/cmd/live_migration.go
```diff
@@ -88,10 +125,61 @@ func handleEvent(event *tgtdb.Event) error {
...
+		hash.Write([]byte(*e.Key[k]))
+	}
+	return int(hash.Sum64() % (uint64(NUM_TARGET_EVENT_CHANNELS)))
+}
+
+func processEvents(chanNo int, evChan chan *tgtdb.Event, done chan bool) {
+	endOfProcessing := false
+	for !endOfProcessing {
+		batch := []*tgtdb.Event{}
+		timeThreshold := time.After(time.Duration(MAX_TIME_PER_BATCH) * time.Millisecond)
```
amit-yb, line 156: From the documentation:
__The underlying Timer is not recovered by the garbage collector until the timer fires.__

Consider using `timer := NewTimer(<same-arg-as-time-after>)`; then in the select statement read the `timer.C`; and after the `Batching for` loop call `timer.Stop()`.

yb-voyager/cmd/live_migration.go
```diff
@@ -18,17 +18,41 @@ package cmd
...
+var TARGET_EVENT_CHANNEL_SIZE int // has to be > MAX_EVENTS_PER_BATCH
+var MAX_EVENTS_PER_BATCH int
+var MAX_TIME_PER_BATCH int //ms
+var END_OF_SOURCE_QUEUE_SEGMENT_EVENT = &tgtdb.Event{Op: "end_of_source_queue_segment"}
+
+func init() {
+	NUM_TARGET_EVENT_CHANNELS = utils.GetEnvAsInt("NUM_TARGET_EVENT_CHANNELS", 512)
+	TARGET_EVENT_CHANNEL_SIZE = utils.GetEnvAsInt("TARGET_EVENT_CHANNEL_SIZE", 2000)
+	MAX_EVENTS_PER_BATCH = utils.GetEnvAsInt("MAX_EVENTS_PER_BATCH", 100)
+	MAX_TIME_PER_BATCH = utils.GetEnvAsInt("MAX_TIME_PER_BATCH", 2000)
```
amit-yb, line 43: MAX_INTERVAL_BETWEEN_BATCHES ??

yb-voyager/cmd/live_migration.go
```diff
@@ -18,17 +18,41 @@ package cmd
...
 )
 
+var NUM_TARGET_EVENT_CHANNELS int
+var TARGET_EVENT_CHANNEL_SIZE int // has to be > MAX_EVENTS_PER_BATCH
+var MAX_EVENTS_PER_BATCH int
+var MAX_TIME_PER_BATCH int //ms
+var END_OF_SOURCE_QUEUE_SEGMENT_EVENT = &tgtdb.Event{Op: "end_of_source_queue_segment"}
+
+func init() {
+	NUM_TARGET_EVENT_CHANNELS = utils.GetEnvAsInt("NUM_TARGET_EVENT_CHANNELS", 512)
```
amit-yb, line 40: Do we still need `"TARGET"` in the names?

amit-yb (approved):
LGTM. Just a few minor nits. Good work! 👍

makalaaneesh:
Passing build for oracle tests (incl live migration) - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/850/

yb-voyager/cmd/live_migration.go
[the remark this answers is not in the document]
sanyamsinghal: yeah, makes sense

PR #983, opened 2023-08-08
init sqlite db for storing meta info

makalaaneesh:


yb-voyager/cmd/common.go
```diff
@@ -308,6 +318,73 @@ func retrieveMigrationUUID(exportDir string) error {
...
+	return nil
+}
+
+func initMetaDB(path string) error {
+	conn, err := sql.Open("sqlite3", path)
+	if err != nil {
+		return fmt.Errorf("error while opening meta db :%w", err)
+	}
+	cmds := []string{
+		fmt.Sprintf(`CREATE TABLE %s (segment_no INTEGER, size_committed INTEGER);`, QUEUE_SEGMENT_META_TABLE_NAME),
```
amit-yb, line 360: Do we need `IF NOT EXISTS` clause?

yb-voyager/cmd/common.go
```diff
@@ -308,6 +318,73 @@ func retrieveMigrationUUID(exportDir string) error {
...
+		log.Infof("Executed query on meta db - %s", cmd)
+	}
+	return nil
+}
+
+func truncateTablesInMetaDb(exportDir string, tableNames []string) error {
+	conn, err := sql.Open("sqlite3", getMetaDBPath(exportDir))
+	if err != nil {
+		return fmt.Errorf("error while opening meta db :%w", err)
+	}
```
amit-yb, line 376: `conn.Close()`??

yb-voyager/cmd/common.go
```diff
@@ -308,6 +318,73 @@ func retrieveMigrationUUID(exportDir string) error {
 	return nil
 }
 
+func getMetaDBPath(exportDir string) string {
```
amit-yb, line 321: How about creating a new file `metadata-db.go` for all functions related to the meta db?

amit-yb (approved):
LGTM.

yb-voyager/cmd/common.go
amit-yb: Add a column named `file_path`. It should be segment file path relative to export-dir.

makalaaneesh: We only run the DDLs when we init the database for the first time, so `IF NOT EXISTS` is not needed! 

Added `file_path` but how would it help 🤔 queue segment no should be enough, no?

PR #989, opened 2023-08-11
Added datatypes transformations for Oracle TargetDB

priyanshi-yb:
Jenkins pipeline - 
1. Without debezium - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/1013
2. debezium - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/1014


yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -40,10 +40,14 @@ type TableSchema struct {
 	Columns []Column `json:"columns"`
 }
 
-func (ts *TableSchema) getColumnType(columnName string) (string, error) {
+func (ts *TableSchema) getColumnType(columnName string, targetDBType string) (string, error) {
 	for _, colSchema := range ts.Columns {
```
makalaaneesh, line 44: Todo for later: we should probably store this as a map instead of a list, so that it's faster to lookup column from columnName. I suppose this code is in the critical path.

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -40,10 +40,14 @@ type TableSchema struct {
...
 		if colSchema.Name == columnName {
-			if colSchema.Schema.Name != "" { // in case Kafka speicfic type which start with io.debezium...
+			if dbColType, ok := colSchema.Schema.Parameters["__debezium.source.column.type"]; ok && targetDBType == "oracle" &&
+				(strings.Contains(dbColType, "DATE") || strings.Contains(dbColType, "INTERVAL")) {
```
makalaaneesh, line 47: What about for CHAR, TIMESTAMP, LONG? You've also included those types int he debezium config.

yb-voyager/src/tgtdb/suites/oracleSuite.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+		var transformedMapValue string
+		for key, value := range mapValue {
+			transformedMapValue = transformedMapValue + fmt.Sprintf("\"%s\"=>\"%s\",", key, value)
+		}
+		return fmt.Sprintf("'%s'", transformedMapValue[:len(transformedMapValue)-1]), nil //remove last comma and add quotes
+	},
+	"STRING": func(columnValue string, formatIfRequired bool) (string, error) {
+		if formatIfRequired {
+			formattedColumnValue := strings.Replace(columnValue, "'", "''", -1)
```
makalaaneesh, line 140: I think we got rid of this? Just be careful when merging

yb-voyager/src/tgtdb/suites/oracleSuite.go
```diff
@@ -0,0 +1,206 @@
...
+			return fmt.Sprintf("'%s'", formattedColumnValue), nil
+		} else {
+			return columnValue, nil
+		}
+	},
+	"INTERVAL YEAR TO MONTH": func(columnValue string, formatIfRequired bool) (string, error) {
```
makalaaneesh, line 146: uff, the interval handling looks like a lot of manually written logic 😅  
Have you checked interval.handling.mode in debezium? THere is a numeric option. Maybe that might be ingestible to oracle directly?

yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -0,0 +1,208 @@
+/*
```
makalaaneesh, line 1: Assuming this is unchanged, you've just moved to a diff file.

yb-voyager/src/tgtdb/oracle.go
```diff
@@ -619,3 +634,24 @@ func (tdb *TargetOracleDB) getConnectionString(tconf *TargetConf) string {
 func (tdb *TargetOracleDB) MaxBatchSizeInBytes() int64 {
 	return 2 * 1024 * 1024 * 1024 // 2GB
 }
+
+func (tdb *TargetOracleDB) getColumnDataTypes(tableName string) (map[string]string, error) {
```
makalaaneesh, line 638: do we still need this given that we're getting the source data types from debezium.

yb-voyager/src/tgtdb/oracle.go
```diff
@@ -587,7 +589,18 @@ func (tdb *TargetOracleDB) InitConnPool() error {
 	return nil
 }
 
-func (tdb *TargetOracleDB) GetDebeziumValueConverterSuite() map[string]ConverterFn {
+func (tdb *TargetOracleDB) GetDebeziumValueConverterSuite() map[string]tgtdbsuite.ConverterFn {
+	oraValueConverterSuite := tgtdbsuite.OraValueConverterSuite
+	for _, i := range []int{1, 2, 3, 4, 5, 6, 7, 8, 9} {
```
makalaaneesh, line 594: is 1-9 the precision of the milliseconds? Can you just add documentation in comments?

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -141,3 +148,16 @@ func (conv *DebeziumValueConverter) convertMap(tableName string, m map[string]*s
 	}
 	return nil
 }
+
+func (conv *DebeziumValueConverter) GetTableNameToSchema() map[string]map[string]map[string]string {
+	//need to create explicit map with required details only as can't use TableSchema directly in import area because of cyclic dependency
```
makalaaneesh, line 153: Let's discuss this once on call? I want to understand the cyclic dependency once.

makalaaneesh (commented):
Added some comments. Let me know after you address them & also merge main.

yb-voyager/src/dbzm/valueConverter.go
priyanshi-yb: yes sure

yb-voyager/src/tgtdb/oracle.go
priyanshi-yb: yes, its precision, will add comment

yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
priyanshi-yb: yes same

yb-voyager/src/tgtdb/suites/oracleSuite.go
priyanshi-yb: I can check that. Currently, the way debezium dumps interval data is easily ingested using COPY as is to YB but with SQlldr we need some transformation.

yb-voyager/src/tgtdb/suites/oracleSuite.go
priyanshi-yb: no, we are still formatting the values. It's just for prepared statements I am not formatting these values by unsettling the `formatIfRequired`.

yb-voyager/src/dbzm/schemaRegistry.go
priyanshi-yb: there are different usage of all of them. See below else-if conditions. for example for CHAR types I am using length to set in the CTL file.

yb-voyager/src/dbzm/schemaRegistry.go
priyanshi-yb: Yes agreed!

yb-voyager/src/dbzm/schemaRegistry.go
makalaaneesh: ok, just wanted to ensure you didn't miss it

yb-voyager/src/tgtdb/oracle.go
makalaaneesh: oops my bad.

yb-voyager/src/tgtdb/suites/oracleSuite.go
makalaaneesh: Ah, right. forgot that we also need to worry about YB 😅 
Ok, just see if the numeric option makes it easy. Otherwise, if you've tested this logic, leave it for now, and we can revisit as this is not a very highly used data type?

priyanshi-yb: I checked the numeric type it gives values in microseconds e.g. -
```
"1","-44706600000000","-1560000000"
```
 COPY can't work with this. I haven't tested sqlldr yet but I think it will also not be able to handle this and need to be manipulated to some understandable format. 
I think the current string format is fine. I have also tested my logic and it works fine.

yb-voyager/src/tgtdb/suites/oracleSuite.go
```diff
@@ -0,0 +1,206 @@
...
+		epochSecs := epochMilliSecs / 1000
+		timestamp := time.Unix(epochSecs, 0).Local()
+		oracleTimestampFormat := "02-01-2006 03.04.05.000 PM" //format: DD-MM-YY HH.MI.SS.FFF PM
+		formattedTimestamp := timestamp.Format(oracleTimestampFormat)
+		if formatIfRequired {
+			formattedTimestamp = fmt.Sprintf("'%s'", formattedTimestamp)
+		}
+		return formattedTimestamp, nil
+	},
+	"io.debezium.time.MicroTimestamp": func(columnValue string, formatIfRequired bool) (string, error) {
```
makalaaneesh, line 61: @priyanshi-yb pls take into consideration here the fixes i made in https://github.com/yugabyte/yb-voyager/pull/1023

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -39,10 +40,14 @@ type TableSchema struct {
 	Columns []Column `json:"columns"`
 }
 
-func (ts *TableSchema) getColumnType(columnName string) (string, error) {
-	for _, colSchema := range ts.Columns {
-		if colSchema.Name == columnName {
-			if colSchema.Schema.Name != "" { // in case Kafka speicfic type which start with io.debezium...
+func (ts *TableSchema) getColumnType(columnName string, targetDBType string) (string, error) {
```
makalaaneesh, line 43: Leave it to you whether you want to do it now, or add a TODO for later:
it would be better to add a param that is at the same abstraction level as that of the schemaRegistry. 

so, instead of taking the targetDBType as input, how about you take a boolean flag `returnSourceDbType` which the caller will set to true. 

basically a similar API as that of `debezium.source.datatype.propagate.source.type`

Basically, the main concern being - SchemaRegistry and TableSchema should have no awareness of what the targetDBTYpe is

yb-voyager/src/dbzm/schemaRegistry.go
```diff
@@ -56,36 +61,36 @@ func (ts *TableSchema) getColumnType(columnName string) (string, error) {
...
 	return &SchemaRegistry{
-		exportDir:         exportDir,
+		exportDir:             exportDir,
 		exporterRole:      exporterRole,
 		tableNameToSchema: make(map[string]*TableSchema),
 	}
 }
 
-func (sreg *SchemaRegistry) GetColumnTypes(tableName string, columnNames []string) ([]string, error) {
+func (sreg *SchemaRegistry) GetColumnTypes(targetDBType, tableName string, columnNames []string) ([]string, error) {
```
makalaaneesh, line 77: consider setting targetDBType at the schemaRegistry level, so you won't have to change these   APIs.

yb-voyager/src/dbzm/valueConverter.go
makalaaneesh: pls add a TODO here to fix

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -120,15 +118,25 @@ func (args *ImportBatchArgs) GetYBCopyStatement() string {
...
 
-func (args *ImportBatchArgs) GetSqlLdrControlFile(schema string) string {
+func (args *ImportBatchArgs) GetSqlLdrControlFile(schema string, tableSchema map[string]map[string]string) string {
 	var columns string
 	if len(args.Columns) > 0 {
 		columnsSlice := make([]string, 0, len(args.Columns))
 		for _, col := range args.Columns {
 			// Add the column name and the NULLIF clause after it
-			columnsSlice = append(columnsSlice, fmt.Sprintf(`%s NULLIF %s='\\N'`, col, col))
+			dataType := tableSchema[col]["__debezium.source.column.type"]
```
makalaaneesh, line 127: Add a TODO comment here as well. 
It seems mixed up for the importBatch layer to have awareness of debezium. Given that we absolutely need it - we should call it something else - like `source_db_datatype`

yb-voyager/src/tgtdb/suites/oracleSuite.go
makalaaneesh: Also, test timestamp fields while streaming.

yb-voyager/cmd/importData.go
```diff
@@ -173,6 +174,11 @@ func applyTableListFilter(importFileTasks []*ImportFileTask) []*ImportFileTask {
 }
 
 func importData(importFileTasks []*ImportFileTask) {
+	if tconf.TargetDBType == YUGABYTEDB {
```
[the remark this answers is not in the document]
makalaaneesh: I'm somehow still not clear on the case sensitivity handling here. I thought we didn't have to deal with quotedStrings? Let's discuss this once.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -172,3 +182,18 @@ func (conv *DebeziumValueConverter) convertMap(tableName string, m map[string]*s
 	}
 	return nil
 }
+
+func (conv *DebeziumValueConverter) GetTableNameToSchema() map[string]map[string]map[string]string {
```
makalaaneesh, line 186: (again for the future :D ) A probably nice way to deal with this might be: 
- instead of returning just a single columnType string for each column, return a column struct (that has all the information pertaining to that column - the name, the source type, the debezium type). 

Then,
- the valueConverter can work with that field struct (as opposed to a single keyword right now). 
- we can embed each column's schema in the data file descriptor (similar to how we add the delimiter/quote char/etc info in it. Then, we won't have to pass the schema directly to importBatch. 

Makes sense? Just add a TODO with some notes on how we can make this better later.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -172,3 +182,18 @@ func (conv *DebeziumValueConverter) convertMap(tableName string, m map[string]*s
...
+
+	//need to create explicit map with required details only as can't use TableSchema directly in import area because of cyclic dependency
+	//TODO: fix this cyclic dependency
```
makalaaneesh, line 189: Let's remember to clean up this dependency mess immediately post the release.

yb-voyager/src/tgtdb/oracle.go
```diff
@@ -31,27 +31,21 @@ import (
...
-}
-
 type TargetOracleDB struct {
 	sync.Mutex
 	tconf *TargetConf
 	oraDB *sql.DB
 	conn  *sql.Conn
 }
 
+var TableToColumnDataTypesCache = make(map[string]map[string]string)
```
makalaaneesh, line 47: Doesn't seem like you're using this?

yb-voyager/src/tgtdb/oracle.go
```diff
@@ -658,6 +653,77 @@ func (tdb *TargetOracleDB) setTargetSchema(conn *sql.Conn) {
 }
 
 func (tdb *TargetOracleDB) IfRequiredQuoteColumnNames(tableName string, columns []string) ([]string, error) {
+	result := make([]string, len(columns))
```
makalaaneesh, line 656: Let's discuss this on call :) I'll need context

yb-voyager/src/tgtdb/suites/oracleSuite.go
```diff
@@ -0,0 +1,210 @@
...
+		parsedTime, err := time.Parse(debeziumFormat, columnValue)
+		if err != nil {
+			return "", fmt.Errorf("parsing timestamp: %v", err)
+		}
+
+		oracleFormat := "06-01-02 3:04:05.000000000 PM -07:00"
+		format := "RR-MM-DD HH:MI:SS.FF9 AM TZR"
+		timstampFn := "TO_TIMESTAMP_TZ"
+		if parsedTime.Location().String() == "UTC" { //LOCAL TIMEZONE Case
+			oracleFormat = "06-01-02 3:04:05.000000000 PM" // TODO: for timezone ones sqlldr is inserting as GMT though GMT/UTC is similar
```
makalaaneesh, line 103: Hmm, I'm a little unsure about this. What is this scenario where we don't have to include timezone information while inserting? It's a TZ included field, so I think we should always make the timezone explicit.

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -120,15 +118,35 @@ func (args *ImportBatchArgs) GetYBCopyStatement() string {
...
 
-func (args *ImportBatchArgs) GetSqlLdrControlFile(schema string) string {
+func (args *ImportBatchArgs) GetSqlLdrControlFile(schema string, tableSchema map[string]map[string]string) string {
 	var columns string
 	if len(args.Columns) > 0 {
 		var columnsList []string
 		for _, column := range args.Columns {
 			//setting the null string for each column
-			columnsList = append(columnsList, fmt.Sprintf("%s NULLIF %s='%s'", column, column, args.NullString))
+			dataType := tableSchema[column]["__debezium.source.column.type"]  //TODO: rename this to some thing like source-db-datatype
```
makalaaneesh, line 127: how come this field is always present? I would have thought you should check for existence first?

yb-voyager/src/tgtdb/oracle.go
makalaaneesh: to discuss: are we making a call to TgtDB in the critical path?

yb-voyager/src/tgtdb/oracle.go
```diff
@@ -658,6 +653,77 @@ func (tdb *TargetOracleDB) setTargetSchema(conn *sql.Conn) {
 }
 
 func (tdb *TargetOracleDB) IfRequiredQuoteColumnNames(tableName string, columns []string) ([]string, error) {
```
makalaaneesh, line 655: Important: test streaming phase for these column types.

yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+		return columnValue, nil
+	},
+	"io.debezium.time.Time": func(columnValue string, formatIfRequired bool) (string, error) {
+		epochMilliSecs, err := strconv.ParseInt(columnValue, 10, 64)
+		if err != nil {
+			return columnValue, fmt.Errorf("parsing epoch milliseconds: %v", err)
+		}
+		epochSecs := epochMilliSecs / 1000
+		timeValue := time.Unix(epochSecs, 0).Local().Format(time.TimeOnly)
```
makalaaneesh, line 98: change to UTC()

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -120,15 +118,35 @@ func (args *ImportBatchArgs) GetYBCopyStatement() string {
...
+			charLength := tableSchema[column]["__debezium.source.column.length"]
+			if  strings.Contains(dataType, "INTERVAL") {
```
makalaaneesh, line 129: let's use a switch case instead here.

makalaaneesh (approved):
LGTM

PR #990, opened 2023-08-11
Added a test case for import data into oracle

ShivanshGahlot:
- Added a test case under `migtests/tests/oracle/import-test`
- Added a validation script `oracle.py`


yb-voyager/src/tgtdb/oracle.go
```diff
@@ -166,80 +166,6 @@ func (tdb *TargetOracleDB) GetVersion() string {
 }
 
 func (tdb *TargetOracleDB) CreateVoyagerSchema() error {
-	createUserQuery := fmt.Sprintf(`BEGIN
```
priyanshi-yb, line 169: Why have we removed this from the import data?
AFAIR, we decided to keep it here as well, for the case where the user is using admin user for migrating in such a scenario it should work fine. 
You can add a check for specific error msg while executing these to check for `insufficient privileges..` and error out with these DDLs to the user for manually creating these or give sufficient permissions to the user.

PR #997, opened 2023-08-21
fixed the unquoting issue with strings for insert/delete event with YugabyteDB target

priyanshi-yb:


priyanshi-yb:
closing this as raised new PR #999

PR #1009, opened 2023-08-24
import data stats - remaining events and estimated time to catch up

priyanshi-yb:


yb-voyager/cmd/live_migration.go
```diff
@@ -225,3 +226,16 @@ func processEvents(chanNo int, evChan chan *tgtdb.Event, lastAppliedVsn int64, d
 	}
 	done <- true
 }
+
+func updateExportedEventsStats(statsReporter *reporter.StreamImportStatsReporter) {
```
amit-yb, line 230: Instead of having a separate go-routine, can't we fetch the export counts in the `ReportStats()` for-loop itself (i.e. line 78) just the way we slide the window we can also make another call to update the export stats.

amit-yb (approved):
LGTM with a suggestion to simplify the code structure.

PR #1010, opened 2023-08-24
separating out the log files per command

priyanshi-yb:


yb-voyager/cmd/analyzeSchema.go
```diff
@@ -1062,7 +1062,8 @@ var analyzeSchemaCmd = &cobra.Command{
 	Short: "Analyze source database schema and generate report about YB incompatible constructs",
 	Long:  ``,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
-		cmd.Parent().PersistentPreRun(cmd.Parent(), args)
+		lockExportDir(cmd)
```
amit-yb, line 1065: Why?

amit-yb (approved):
LGTM.

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: `analyze-schema`' Parent cmd is `yb-voyager` and hence it was getting added with the name in the log file. 
But I think this `PersistentPreRun` is not even required for this command explicitly in the subcommand and can be replaced with `PreRun` (and even maintains uniformity across commands as we are using `PreRun` for other commands) which solves the problem of fetching cmdName directly in root cmd.

PR #1013, opened 2023-08-28
added an analyze-schema case for JSONARRAYAGG() in any DDL

priyanshi-yb:


yb-voyager/cmd/analyzeSchema.go
```diff
@@ -186,6 +186,7 @@ var (
 	unsupportedCommentRegex2   = re("--", anything, "please edit to match PostgreSQL syntax")
 	typeUnsupportedRegex       = re("Inherited types are not supported", anything, "replacing with inherited table")
 	bulkCollectRegex           = re("BULK COLLECT") // ora2pg unable to convert this oracle feature into a PostgreSQL compatible syntax
+	viewJsonFuncRegex          = re("CREATE", opt("OR REPLACE"), "VIEW", capture(ident), anything, "JSON_ARRAYAGG")
```
shubham-yb, line 189: Will it check for this function only in VIEW DDLs?
What happens in case if this is used in other objects like functions/procedures/triggers etc?

priyanshi-yb: Yes fixed that, I was thinking of this but was facing some issues with regex so I dropped that. But added it now. Thanks!

shubham-yb (approved):
LGTM

PR #1016, opened 2023-08-28
Changes to support exported datafile format as CSV 

priyanshi-yb:


yb-voyager/src/tgtdb/oracle.go
```diff
@@ -562,7 +562,7 @@ func (tdb *TargetOracleDB) importBatch(conn *sql.Conn, batch Batch, args *Import
 	password := tdb.tconf.Password
 	connectString := tdb.getConnectionString(tdb.tconf)
 	oracleConnectionString := fmt.Sprintf("%s@\"%s\"", user, connectString)
-	sqlldrArgs := fmt.Sprintf("userid=%s control=%s log=%s DIRECT=TRUE NO_INDEX_ERRORS=TRUE", oracleConnectionString, sqlldrControlFilePath, sqlldrLogFilePath)
+	sqlldrArgs := fmt.Sprintf("userid=%s control=%s log=%s DIRECT=TRUE NO_INDEX_ERRORS=TRUE SKIP=1 ERRORS=0", oracleConnectionString, sqlldrControlFilePath, sqlldrLogFilePath)
```
makalaaneesh, line 565: what do these options mean? can you pls comment here, or link to doc

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -134,7 +134,8 @@ INFILE '%s'
 APPEND
 INTO TABLE %s
 REENABLE DISABLED_CONSTRAINTS
-FIELDS TERMINATED BY '%s'
+FIELDS CSV WITH EMBEDDED 
+TRAILING NULLCOLS
```
makalaaneesh, line 138: pls link to doc

makalaaneesh (commented):
Looks  good from my side. @amit-yb can you help with the "import data file" scenario review

yb-voyager/src/tgtdb/oracle.go
priyanshi-yb: ERRORS=0, set the error limit to 0 which means that any single error in the datafile can't be skipped. `sqlldr` should error out at that time itself. 
SKIP=1 is to skip the first row which is the header.
will add a comment.

yb-voyager/cmd/importDataState.go
```diff
@@ -394,7 +394,7 @@ func (bw *BatchWriter) WriteRecord(record string) error {
 		return nil
 	}
 	var err error
-	if bw.flagFirstRecordWritten {
+	if bw.flagFirstRecordWritten && !strings.HasSuffix(record, "\n") { //TODO: check if this is better way to check if \n is already added or not
```
makalaaneesh, line 397: I assume this is because the csvWriter is appending a \n at the end. 
How about preemptively removing that in ConvertRow with strings.TrimSuffix ? it

.github/workflows/mysql-migtests.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         yes | ./installer_scripts/install-yb-voyager --install-from-local-source
       env:
         ON_INSTALLER_ERROR_OUTPUT_LOG: Y
-        DEBEZIUM_VERSION: latest
+        DEBEZIUM_VERSION: priyanshi_csv-format
```
amit-yb, line 37: You must fix these before you merge the PR.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -86,7 +92,12 @@ func (conv *DebeziumValueConverter) ConvertRow(tableName string, columnNames []s
 		}
 		columnValues[i] = transformedValue
 	}
-	return strings.Join(columnValues, "\t"), nil
+	var buffer bytes.Buffer
```
amit-yb, line 95: Consider having a `bytes.Buffer` instance as a member of `DebeziumValueConverter`.

Reuse the same buffer (after calling `conv.buf.Reset()`) across multiple runs of the `ConvertRow()`.

yb-voyager/src/utils/commonVariables.go
```diff
@@ -26,6 +26,7 @@ const (
 	TABLE_MIGRATION_IN_PROGRESS
 	TABLE_MIGRATION_DONE
 	TABLE_MIGRATION_COMPLETED
+	YB_VOYAGER_NULL_STRING = "__$#__YBVOYAGER_NULL__$#__"
```
amit-yb, line 29: I understand why you might want to have a string like that. But this can (in certain cases) blow up the snapshot file size.

A similar but smaller alternative is `__YBV_NULL__`. This won't cause any conflict even if there is a string field containing this "keyword" -- because the string field will be quoted whereas the `__YBV_NULL__` will appear without any quotes.

amit-yb (approved):
Overall looks good! Good work 👍

.github/workflows/mysql-migtests.yml
priyanshi-yb: yes will do.

PR #1012, opened 2023-08-27
Enabling ff workflow

makalaaneesh:

PR #1017, opened 2023-08-29
Fixed stream Id generation with start-clean and error logging in yb_cdc_client

priyanshi-yb:


yb-voyager/src/dbzm/yb_cdc_client.go
```diff
@@ -115,7 +115,9 @@ func (ybc *YugabyteDBCDCClient) readYBStreamID() (string, error) {
 
 func (ybc *YugabyteDBCDCClient) DeleteStreamID() error {
 	streamID, err := ybc.readYBStreamID()
-	if err != nil {
+	if err != nil && strings.Contains(err.Error(), "stream id not found") {
```
makalaaneesh, line 118: Why ignore this error?

makalaaneesh (commented):
@priyanshi-yb I already made the dist dir changes as part of https://github.com/yugabyte/yb-voyager/pull/1012. Changed the logic a bit to make sure that we always add the suffix so that the high level debezium dir is always the one specified.

I don't fully understand the change you made in the yb_cdc_client.

yb-voyager/src/dbzm/yb_cdc_client.go
priyanshi-yb: Actually with `--start-clean`, we first delete the existing stream-id for which we first read the file `yb_cdc_stream_id.txt` and fetch the stream-id from it and then delete it but when `--start-clean` is used for the for first this file will not be present in the export-dir and hence it was error out saying `stream ID not found`. but that is not an ideal case. if stream-id is not present then we should generate one in spite of erroring out. This was one of [Amit's comments](https://github.com/yugabyte/yb-voyager/pull/970#discussion_r1286861690) which I think I messed up while changing the code structure.

makalaaneesh: Got it, makes sense! Can you edit this PR to only retain this change as the other changes are taken care in the other PR?

amit-yb (approved):
LGTM.

PR #1019, opened 2023-08-29
Record ExportType and FallForwardDBExists in metaDB

ShivanshGahlot:


yb-voyager/cmd/metadataDB.go
```diff
@@ -136,7 +142,7 @@ type MetaDB struct {
 }
 
 func NewMetaDB(exportDir string) (*MetaDB, error) {
-	db, err := sql.Open("sqlite3", getMetaDBPath(exportDir))
+	db, err := sql.Open("sqlite3", fmt.Sprintf("%s?_txlock=immediate&_timeout=30000", getMetaDBPath(exportDir)))
```
amit-yb, line 145: Who duplicates sqlite options like this!!! 🙄 
Define a const `SQLITE_OPTIONS` and then use it to construct the connection string.

yb-voyager/cmd/export.go
```diff
@@ -316,3 +316,12 @@ func validateExportTypeFlag() {
 		utils.ErrExit("Error: Invalid export-type: %q. Supported export types are: %s", exportType, validExportTypes)
 	}
 }
+
+func updateExportTypeInMetaDB() {
```
amit-yb, line 320: `saveExportTypeInMetaDB()`.

yb-voyager/cmd/importData.go
```diff
@@ -214,6 +213,18 @@ func importData(importFileTasks []*ImportFileTask) {
 		utils.ErrExit("Failed to initialize meta db: %s", err)
 	}
 
+	if importDestinationType == TARGET_DB {
+		record, err := GetMigrationStatusRecord()
```
amit-yb, line 217: We will have to ensure that the `import data` code path continues to work with `import data file` use-case.

Here, it is possible that the MigrationStatusRecord is not inserted in the meta-db (and hence the `record` will be `nil`).

We should insert a `MigrationStatusRecord`, with initial default values, at the time of project creation (where we initialise meta-db as well).

ShivanshGahlot: Done

yb-voyager/cmd/migrationStatus.go
```diff
@@ -0,0 +1,37 @@
...
+		return nil, nil
+	}
+	return record, nil
+}
+
+func InitMigrationStatusRecord(migUUID string) error {
+	return UpdateMigrationStatusRecord(func(record *MigrationStatusRecord) {
+		record.MigrationUUID = migUUID
+		record.ExportType = SNAPSHOT_ONLY
+	})
```
amit-yb, line 36: BUG: The current implementation is not idempotent. `CreateMigrationProjectIfNotExists()` is called from every command invocation. So `InitMigrationStatusRecord()` will be called once for every command.

The current implementation will correctly initialise `migUUID` and `ExportType` the first time it runs. On the next run, it will overwrite `migUUID` with empty UUID and reset `ExportType` back to `SNAPSHOT_ONLY`.

One approach to make this function idempotent is to immediately return if `record.MigrationUUID` is not "nil uuid" (there is a const to represent a nil uuid).

yb-voyager/cmd/migrationStatus.go
```diff
@@ -30,6 +30,14 @@ func GetMigrationStatusRecord() (*MigrationStatusRecord, error) {
...
 
 func InitMigrationStatusRecord(migUUID string) error {
+	record, err := GetMigrationStatusRecord()
+	if err != nil {
+		return err
+	}
+	if record.MigrationUUID != "" {
+		return nil
+	}
+
```
amit-yb, line 40: No need of explicitly reading the record here. The `UpdateMigrationStatusRecord()` internally reads the latest record and passes it to the updateFn (the anonymous func). So all you have to do is:
```
	return UpdateMigrationStatusRecord(func(record *MigrationStatusRecord) {
        if record.MigrationUUID != uuid.Nil {
            log.Info("MigrationStatusRecord is already initialised")
            return
        }
		record.MigrationUUID = migUUID
		record.ExportType = SNAPSHOT_ONLY
    })
```

amit-yb (approved):
LGTM with one comment. Address it, test the changes, and proceed to merging the PR. Good work :-)

PR #1022, opened 2023-08-30
exporter stats refresh every 10s

makalaaneesh:


amit-yb (approved):
That was a very small change. :D LVGTM.

PR #1023, opened 2023-08-30
Data transformation fix: Local->UTC

makalaaneesh:
Timestamp fields without timezone are represented by debezium as microseconds/nanoseconds since Unix Epoch (which is in UTC). Therefore, we need to tie the time calculation to UTC timezone.

### DIFF
--- a/yb-voyager/src/tgtdb/yugabytedb.go
+++ b/yb-voyager/src/tgtdb/yugabytedb.go
@@ -80,11 +80,12 @@ var ybValueConverterSuite = map[string]ConverterFn{
 		}
 		epochSeconds := epochMicroSecs / 1000000
 		epochNanos := (epochMicroSecs % 1000000) * 1000
-		microTimeStamp, err := time.Parse(time.RFC3339Nano, time.Unix(epochSeconds, epochNanos).UTC().Format(time.RFC3339Nano)) //TODO: check if proper format for Micro can work
-		if err != nil {
-			return columnValue, err
-		}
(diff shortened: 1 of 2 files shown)


yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -80,11 +80,7 @@ var ybValueConverterSuite = map[string]ConverterFn{
 		}
 		epochSeconds := epochMicroSecs / 1000000
 		epochNanos := (epochMicroSecs % 1000000) * 1000
-		microTimeStamp, err := time.Parse(time.RFC3339Nano, time.Unix(epochSeconds, epochNanos).Local().Format(time.RFC3339Nano)) //TODO: check if proper format for Micro can work
-		if err != nil {
-			return columnValue, err
-		}
-		timestamp := strings.TrimSuffix(microTimeStamp.String(), " +0000 UTC")
```
priyanshi-yb, line 87: Is trimming the suffix not required anymore with `UTC()`?

priyanshi-yb (approved):
LGTM!

yb-voyager/src/tgtdb/yugabytedb.go
makalaaneesh: Ah, no. 
trimming has become unnecessary because now I've got rid of RFC3339Nano and replaced it with a custom format - for micro and nano (i just removed the timezone part of the RFC3339Nano so now we don't have to remove the UTC from the string)

priyanshi-yb: Oh okay. Yes. Thanks!

PR #1027, opened 2023-08-31
Enabling ff integration

makalaaneesh:


yb-voyager/cmd/root.go
```diff
@@ -46,7 +46,7 @@ Refer to docs (https://docs.yugabyte.com/preview/migrate/) for more details like
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		if exportDir != "" && utils.FileOrFolderExists(exportDir) {
 			// TODO: avoid lock only for fall-forward setup
-			if cmd.Use != "version" && cmd.Use != "status" && cmd.Use != "initiate" && !(cmd.Use == "data" && cmd.Parent().Use == "import"){
+			if cmd.Use != "version" && cmd.Use != "status" && cmd.Use != "initiate" && !(cmd.Use == "setup" && cmd.Parent().Use == "fall-forward") {
```
sanyamsinghal, line 49: nit: maybe a separate function and array to store these commands names can be introduced here.

yb-voyager/cmd/root.go
```diff
@@ -65,7 +65,7 @@ Refer to docs (https://docs.yugabyte.com/preview/migrate/) for more details like
 	},
 
 	PersistentPostRun: func(cmd *cobra.Command, args []string) {
-		if exportDir != "" && utils.FileOrFolderExists(exportDir) && cmd.Use != "version" && cmd.Use != "status" && cmd.Use != "initiate" && !(cmd.Use == "data" && cmd.Parent().Use == "import"){
+		if exportDir != "" && utils.FileOrFolderExists(exportDir) && cmd.Use != "version" && cmd.Use != "status" && cmd.Use != "initiate" && !(cmd.Use == "setup" && cmd.Parent().Use == "fall-forward") {
```
sanyamsinghal, line 68: same comment as above

yb-voyager/cmd/root.go
sanyamsinghal: this condition can also be modified: `cmd.Use != "initiate"`
to
`!(cmd.Use != "initiate" && cmd.Use!="cutover")

I skipped it since cutover is only command command having `initiate` 
But lets do it for both then

sanyamsinghal (approved):
LGTM
Assuming you have tested the workflow end-to-end

PR #1031, opened 2023-09-01
Added sqlite3 in installer script

ShivanshGahlot:


installer_scripts/install-yb-voyager
```diff
@@ -179,6 +181,7 @@ macos_main() {
 	macos_install_pg_dump
 	create_pg_dump_args_file
 	brew install wget
+	brew install sqlite
```
amit-yb, line 184: Don't we need to specify `sqlite` version in case of `brew`?

amit-yb (approved):
LGTM.

PR #1032, opened 2023-09-01
More changes for ff integration

makalaaneesh:


yb-voyager/cmd/root.go
```diff
@@ -65,12 +64,27 @@ Refer to docs (https://docs.yugabyte.com/preview/migrate/) for more details like
...
+	}
+	if cmd.Use == "initiate" && cmd.Parent().Use == "cutover" {
+		return false
+	}
+	if cmd.Parent().Use == "fall-forward" {
+		if cmd.Use == "setup" || cmd.Use == "switchover" {
+			return false
+		}
+	}
+	return false
```
sanyamsinghal, line 85: should it be `return true` for other cases?

yb-voyager/cmd/root.go
```diff
@@ -65,12 +64,27 @@ Refer to docs (https://docs.yugabyte.com/preview/migrate/) for more details like
...
 }
 
+func shouldLock(cmd *cobra.Command) bool {
+	if cmd.Use == "version" || cmd.Use == "status" {
+		return false
+	}
+	if cmd.Use == "initiate" && cmd.Parent().Use == "cutover" {
+		return false
+	}
+	if cmd.Parent().Use == "fall-forward" {
```
sanyamsinghal, line 80: also there are chances that there is no parent command, in that case this might result in `nil pointer exception` here

sanyamsinghal (commented):
LGTM.
Couple of important comments on `shouldLock()` function

sanyamsinghal (approved):
LGTM, nice work

PR #1035, opened 2023-09-01
Added possible paths for pg_dump-args.ini

ShivanshGahlot:
In case of brew, we cannot install files directly into the `/etc/yb-voyager` folder due to permission issues. Hence, I install the `pg_dump-args.ini` file in `${HOMEBREW_PREFIX}/etc/yb-voyager`

### DIFF
--- a/yb-voyager/src/srcdb/pg_dump.go
+++ b/yb-voyager/src/srcdb/pg_dump.go
@@ -40,8 +40,23 @@ var pgDumpArgs PgDumpArgs
 var pgDumpArgsFile string
 
 func getPgDumpArgsFromFile(sectionToRead string) string {
-	basePgDumpArgsFilePath := filepath.Join("/", "etc", "yb-voyager", "pg_dump-args.ini")
-	if utils.FileOrFolderExists(basePgDumpArgsFilePath) {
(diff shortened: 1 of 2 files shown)


yb-voyager/src/srcdb/pg_dump.go
```diff
@@ -40,8 +40,23 @@ var pgDumpArgs PgDumpArgs
...
+		filepath.Join("/", "usr", "local", "Cellar", homebrewVoyagerDir, utils.YB_VOYAGER_VERSION, "etc", "yb-voyager", "pg_dump-args.ini"),
+	}
+
+	for _, path := range possiblePaths {
+		if utils.FileOrFolderExists(path) {
+			basePgDumpArgsFilePath = path
+			break
+		}
+	}
+	if basePgDumpArgsFilePath == "" {
```
sanyamsinghal, line 57: you can remove this check. In case if it is not found it will be generated and saved there(just verfiy once).
Because of this code:
```
//go:embed data/pg_dump-args.ini
var pgDumpArgsFile string
```

sanyamsinghal (approved):

amit-yb (approved):
LGTM. take care of the comment from sanyam.

PR #1040, opened 2023-09-03
Bug import stats: finalize the updation of all stats once cutover completes

priyanshi-yb:


yb-voyager/cmd/live_migration.go
```diff
@@ -57,14 +58,14 @@ func streamChanges() error {
...
 		return fmt.Errorf("failed to fetch event channel meta info from target : %w", err)
 	}
-	statsReporter := reporter.NewStreamImportStatsReporter()
+	statsReporter = reporter.NewStreamImportStatsReporter()
 	err = statsReporter.Init(tdb, migrationUUID)
 	if err != nil {
 		return fmt.Errorf("failed to initialize stats reporter: %w", err)
 	}
-	defer statsReporter.Finalize()
+	defer finalizeStats()
```
sanyamsinghal, line 66: can we make - `finalizeStatus` and `updateExportedEventsStats` methods of the `statsReporter` struct

yb-voyager/cmd/live_migration.go
```diff
@@ -243,15 +244,24 @@ func processEvents(chanNo int, evChan chan *tgtdb.Event, lastAppliedVsn int64, d
...
-		if err != nil {
-			utils.ErrExit("failed to fetch exported events stats from meta db: %v", err)
-		}
-		statsReporter.UpdateRemainingEvents(totalExportedEvents)
+		fetchAndUpdateEventsStats()
+	}
+}
+
+func fetchAndUpdateEventsStats() {
+	totalExportedEvents, _, err := metaDB.GetTotalExportedEvents(time.Now().String())
```
sanyamsinghal, line 257: how about having a retry logic, maybe 5 times before exiting the process here?

priyanshi-yb: How will that help? like in which case retry can succeed?

yb-voyager/cmd/live_migration.go
priyanshi-yb: No, we can't there is some cyclic dependency issue in having a separate connection to MetaDB in the `streamImportReporter` class. Need to do refactoring of metaDB to some other location like `utils`. to be able to use it everywhere. That's a future task. See this comment for more information https://github.com/yugabyte/yb-voyager/pull/1009#discussion_r1304251485

yb-voyager/cmd/live_migration.go
sanyamsinghal: I am not 100% sure, but there is one error we throw based on this condition:
```
if !strings.Contains(err.Error(), "converting NULL to int64 is unsupported")
```
when will that occur?

can this error occur in case the stats table is not populated by the debezium at initial moments?

priyanshi-yb: this error is handled inside the `GetTotalExportedEvents()` function in `metadataDB.go`. I don't it will occur here.

sanyamsinghal: I know that. I am asking when will db throw that error?
Maybe can this error occur in case the stats table is not populated by the debezium at initial moments?

sanyamsinghal (approved):

PR #1041, opened 2023-09-03
miscellaneous changes - separating debezium logs, skip import table list with streaming, skip CLOB NCLOB types from oracle

priyanshi-yb:


yb-voyager/cmd/importData.go
```diff
@@ -92,7 +92,11 @@ func importDataCommandFn(cmd *cobra.Command, args []string) {
 	// TODO: handle case-sensitive in table names with oracle ff-db
 	// quoteTableNameIfRequired()
 	importFileTasks := discoverFilesToImport()
-	importFileTasks = applyTableListFilter(importFileTasks)
+	if changeStreamingIsEnabled(importType) && (tconf.TableList != "" || tconf.ExcludeTableList != "") {
```
makalaaneesh, line 95: I'm wondering - should we just exit and have the user re-run without the table-list flags. 
I think a log is not sufficient

sanyamsinghal: agree with Aneesh, for now we should error out saying that these flags are not supported for live migration. 
Because in future we might start supporting them.

PR #1043, opened 2023-09-05
set empty password in conf file instead of env variable for debezium

priyanshi-yb:
Issue - https://yugabyte.atlassian.net/browse/DB-7533

### DIFF
--- a/yb-voyager/src/dbzm/config.go
+++ b/yb-voyager/src/dbzm/config.go
@@ -355,6 +355,9 @@ func (c *Config) String() string {
 
 func (c *Config) WriteToFile(filePath string) error {
 	config := c.String()
+	if c.Password == "" { //empty password have issues with Env variable https://yugabyte.atlassian.net/browse/DB-7533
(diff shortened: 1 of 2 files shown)


yb-voyager/src/dbzm/config.go
```diff
@@ -355,6 +355,9 @@ func (c *Config) String() string {
 
 func (c *Config) WriteToFile(filePath string) error {
 	config := c.String()
+	if c.Password == "" { //empty password have issues with Env variable https://yugabyte.atlassian.net/browse/DB-7533
+		config += fmt.Sprintf("\ndebezium.source.database.password=%s", c.Password)
```
amit-yb, line 359: Just asking: do you need quotes around the value to represent empty string?

priyanshi-yb: No, I don't think it is required. Earlier when env wasn't introduced and in the conf, we were setting like only. cc: @makalaaneesh

makalaaneesh (approved):

PR #1046, opened 2023-09-05
Fixed a bug in import data file where default batchSize 20K not being set

priyanshi-yb:
1. Fixed a bug in the `import data file` where the progress bar was kind of broken for the import data file, because the default batch size of 20K was not being set for the import data file command, and hence the splits were generating of 200MB size limit
2. Introduced a new `importerRole` as `import_file`.

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -358,7 +358,9 @@ func importData(importFileTasks []*ImportFileTask) {
 	callhome.PackAndSendPayload(exportDir)
 	if !dbzm.IsDebeziumForDataExport(exportDir) {
 		executePostImportDataSqls()
-		displayImportedRowCountSnapshot(importFileTasks)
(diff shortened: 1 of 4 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -358,7 +358,9 @@ func importData(importFileTasks []*ImportFileTask) {
 	callhome.PackAndSendPayload(exportDir)
 	if !dbzm.IsDebeziumForDataExport(exportDir) {
 		executePostImportDataSqls()
-		displayImportedRowCountSnapshot(importFileTasks)
+		if !reportProgressInBytes { //shouldn't report in case import data file
+			displayImportedRowCountSnapshot(importFileTasks)
```
amit-yb, line 362: Print the report even in case of `import data file`.

amit-yb (approved):
LGTM with a minor comment.

PR #1048, opened 2023-09-05
Backdoor option for master and tserver port using environment variable

priyanshi-yb:
`yb_cdc_client_wrapper` jar change PR on debezium - https://github.com/yugabyte/debezium/pull/89

### DIFF
--- a/yb-voyager/src/srcdb/yugabytedb.go
+++ b/yb-voyager/src/srcdb/yugabytedb.go
@@ -20,6 +20,7 @@ import (
 	"database/sql"
 	"fmt"
 	"net/url"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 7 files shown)


yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -391,7 +392,10 @@ func (yb *YugabyteDB) GetColumnToSequenceMap(tableList []*sqlname.SourceName) ma
 
 func (yb *YugabyteDB) GetServers() string {
```
amit-yb, line 393: This code has become confusing. It returns all nodes of the cluster -- but with master port! The port will not be valid for all nodes.

It will be better to return a list of ipOrHostNames of all the servers from this method (that will also justify the name). The caller is then free to append each ipOrHostName with any desired port before using it.

amit-yb (approved):
LGTM with a suggestion.

PR #1054, opened 2023-09-06
COALESCE while summing up stats in final export report. 

makalaaneesh:
Not required on import side because we init the metadatatable with 0 values for all tables, SUM will always return a 0. We don't do the same on the import side. We only insert a row when we first encounter an event for that table.

### DIFF
--- a/yb-voyager/cmd/metadataDB.go
+++ b/yb-voyager/cmd/metadataDB.go
@@ -340,7 +340,9 @@ func (m *MetaDB) GetExportedEventsStatsForTable(schemaName string, tableName str
 	var updates int64
 	var deletes int64
 	// Using SUM + LOWER case comparison here to deal with case sensitivity across stats published by source (ORACLE) and target (YB) (in ff workflow)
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/metadataDB.go
```diff
@@ -340,22 +340,22 @@ func (m *MetaDB) GetExportedEventsStatsForTable(schemaName string, tableName str
 	var updates int64
 	var deletes int64
 	// Using SUM + LOWER case comparison here to deal with case sensitivity across stats published by source (ORACLE) and target (YB) (in ff workflow)
-	query := fmt.Sprintf(`select SUM(num_total), SUM(num_inserts), SUM(num_updates), SUM(num_deletes) from %s WHERE LOWER(schema_name)=LOWER('%s') AND LOWER(table_name)=LOWER('%s')`,
+	query := fmt.Sprintf(`select COALESCE(SUM(num_total), 0), COALESCE(SUM(num_inserts),0), COALESCE(SUM(num_updates),0), COALESCE(SUM(num_deletes),0) from %s WHERE LOWER(schema_name)=LOWER('%s') AND LOWER(table_name)=LOWER('%s')`,
```
amit-yb, line 343: Nit: Line too long. Consider splitting the query on multiple lines.

amit-yb (approved):
LGTM.

PR #1057, opened 2023-09-07
print log file path for users to check error information when sqlldr errors out

priyanshi-yb:


yb-voyager/src/tgtdb/oracle.go
```diff
@@ -570,9 +570,9 @@ func (tdb *TargetOracleDB) importBatch(conn *sql.Conn, batch Batch, args *Import
 	var errbuf string
 	outbuf, errbuf, err = sqlldr.RunSqlldr(sqlldrArgs, password)
 
-	if outbuf == "" && errbuf == "" && err != nil {
+	if errbuf != "" || err != nil {
```
amit-yb, line 573: I would guess that only `err != nil` should be required. Are you sure that there are failure cases where `err == nil` and `errbuf != ""`? Does `errbuf != ""` really mean there was any error while executing sqlldr?

amit-yb (approved):
LGTM. I am approving this PR for now. But we should validate the assumption that we're making (as mentioned in the comment).

yb-voyager/src/tgtdb/oracle.go
priyanshi-yb: yes you are right only `err != nil` is required but I guess we should print the `errbuf` in the err msg.

PR #1062, opened 2023-09-07
added a converter func for io.debezium.time.Date type in oracle for streaming from yb to ff

priyanshi-yb:


yb-voyager/src/tgtdb/suites/oracleSuite.go
```diff
@@ -44,6 +44,23 @@ var OraValueConverterSuite = map[string]ConverterFn{
 		}
 		return formattedDate, nil
 	},
+	"io.debezium.time.Date": func(columnValue string, formatIfRequired bool) (string, error) {
+		epochDays, err := strconv.ParseUint(columnValue, 10, 64)
```
makalaaneesh, line 48: I don't think it would be right to assume that it's an unsigned int. 
> `fmt.Printf("date=%s\n", time.Unix(int64(-2), 0).UTC().Format(time.DateOnly))`

date=1969-12-31

pls change it to ParseInt (even for yb suite)

yb-voyager/src/tgtdb/suites/oracleSuite.go
```diff
@@ -44,6 +44,23 @@ var OraValueConverterSuite = map[string]ConverterFn{
...
+		if err != nil {
+			return columnValue, fmt.Errorf("parsing epoch seconds: %v", err)
+		}
+		epochSecs := epochDays * 24 * 60 * 60
+		parsedTime := time.Unix(int64(epochSecs), 0).UTC()
+		oracleDateFormat := "02-01-2006" //format: DD-MON-YYYY
+		formattedDate := parsedTime.Format(oracleDateFormat)
+		if err != nil {
```
makalaaneesh, line 56: Looks like there's no new err to check here?

makalaaneesh (commented):
Added couple of comments. Also unsure of how this is working if the date format in sqlldr file is DD-MM-YY . Pls check that as well.

yb-voyager/src/tgtdb/suites/oracleSuite.go
makalaaneesh: @shubham-yb we should test dates prior to epoch as well 😅

makalaaneesh (approved):
LGTM

PR #1073, opened 2023-09-08
BUG UI: in the snapshot export report for PostgreSQL with multiple schemas in source-db-schema reports the complete list on the UI

priyanshi-yb:
…

### DIFF
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -254,12 +254,10 @@ func displayExportedRowCountSnapshot() {
 			if source.Schema != "" {
 				tableParts := strings.Split(key, ".")
 				table := tableParts[0]
-				schema := source.Schema
+				schema := getDefaultSourceSchemaName()
 				if len(tableParts) > 1 {
 					schema = tableParts[0]
 					table = tableParts[1]
-				} else if source.DBType == POSTGRESQL {
-					schema = getDefaultSourceSchemaName()
 				}
 				uitable.AddRow(schema, table, exportedRowCount[key])
 			} else {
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/common.go
```diff
@@ -252,7 +252,16 @@ func displayExportedRowCountSnapshot() {
 		sort.Strings(keys)
 		for _, key := range keys {
 			if source.Schema != "" {
-				uitable.AddRow(source.Schema, key, exportedRowCount[key])
+				tableParts := strings.Split(key, ".")
+				table := tableParts[0]
+				schema := source.Schema
```
sanyamsinghal, line 257: nit/optional:
```
schema := getDefaultSourceSchemaName()
and remove `else if` since sourceDBType is handled by above function
```

sanyamsinghal (approved):

yb-voyager/cmd/common.go
```diff
@@ -252,7 +252,16 @@ func displayExportedRowCountSnapshot() {
...
+				if len(tableParts) > 1 {
+					schema = tableParts[0]
+					table = tableParts[1]
+				} else if source.DBType == POSTGRESQL {
+					schema = getDefaultSourceSchemaName()
+				}
```
amit-yb, line 263: If `source.Schema` has just one schema, then that should be the "default" instead of "public".

yb-voyager/cmd/common.go
priyanshi-yb: Oh yes, good point. Thanks!

amit-yb (approved):
LGTM.

PR #1099, opened 2023-09-13
Added cutover command to the live migration automation script

shubham-yb:


priyanshi-yb (approved):
LGTM. Can you please see if you can modify the `live-run-tesh.sh` to catch the import data error as well as fail if it throws an error?

PR #1100, opened 2023-09-14
Finalize import stats only when --disable-pb is not set

priyanshi-yb:
Bug when `--disable-pb` is used on the importer side and cutover happens `defer.finalizeStats()` will return a nil pointer exception as the stats table isn't initialized for this case.

### DIFF
--- a/yb-voyager/cmd/live_migration.go
+++ b/yb-voyager/cmd/live_migration.go
@@ -65,8 +65,9 @@ func streamChanges() error {
 	if err != nil {
 		return fmt.Errorf("failed to initialize stats reporter: %w", err)
 	}
-	defer finalizeStats()
+
 	if !disablePb {
+		defer finalizeStats()
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 		go updateExportedEventsStats(ctx)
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/live_migration.go
```diff
@@ -65,8 +65,9 @@ func streamChanges() error {
 	if err != nil {
 		return fmt.Errorf("failed to initialize stats reporter: %w", err)
 	}
-	defer finalizeStats()
+
 	if !disablePb {
+		defer finalizeStats()
```
makalaaneesh, line 70: Probably, doesn't matter right now, but it would be read better to call this _after_ calling `go statsReporter.ReportStats(ctx), i.e. you initialize, then defer finalize (as opposed to the other way around)

makalaaneesh (approved):
LGTM with a small comment!

yb-voyager/cmd/live_migration.go
makalaaneesh: For example, if we add any error handling(where we return err) in future between this statement, and `go statsReporter.ReportStats()` , we can run into the same issue

priyanshi-yb: Hmm got it makes sense

PR #1108, opened 2023-09-15
[Refactoring] Moved import data state related methods from `TargetDB` to `ImportDataState`

amit-yb:


yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -38,18 +40,89 @@ type TargetDB interface {
...
+	Row
+	Next() bool
+	Close()
+}
+
+type Row interface {
+	Scan(dest ...interface{}) error
+}
+
+type sqlRowsToTgtdbRowsAdapter struct {
```
makalaaneesh, line 70: I see the value in Tx interface, since we can define an Exec that returns both rowsAffected, err. 

here, it looks like we're just forwarding each method to the underlying sql.Rows. Is it just for future-proofing in case some other DB's pkg doesn't adhere to sql.rows?

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -38,18 +40,89 @@ type TargetDB interface {
...
+	rows *sql.Rows
```
makalaaneesh, line 71: Could we just embed sql.Rows?

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -38,18 +40,89 @@ type TargetDB interface {
...
 }
 
+func (s *sqlRowsToTgtdbRowsAdapter) Close() {
+	_ = s.rows.Close()
```
makalaaneesh, line 75: Intentional ignore of error here?

makalaaneesh (commented):
Added some comments. 
I remember @ShivanshGahlot mentioning that there was a difference in the exec functions of oracle vs yugabyte (in that one required a semicolon at the end, and one did not) - pls verify that that is handled. 

Also, pls run the oracle tests on jenkins.

yb-voyager/src/tgtdb/target_db_interface.go
amit-yb: The `sql.Rows.Close()` method does not match what is expected by the `tgtdb.Rows` interface (which essentially is a subset of `pgx.Rows`). So the "adapter" class, defines a wrapper on the `sql.Rows.Close()` to drop the error.

yb-voyager/src/tgtdb/target_db_interface.go
amit-yb: Done. Redefined the `Close()` method for the reason mentioned in the previous answer.

yb-voyager/src/tgtdb/oracle.go
```diff
@@ -82,6 +82,45 @@ func (tdb *TargetOracleDB) Init() error {
 	return err
 }
 
+func (tdb *TargetOracleDB) Query(query string) (Rows, error) {
+	rows, err := tdb.conn.QueryContext(context.Background(), query)
+	if err != nil {
+		return nil, fmt.Errorf("run query %q on oracle %s: %s", query, tdb.tconf.Host, err)
+	}
+	return &sqlRowsToTgtdbRowsAdapter{Rows: rows}, nil
```
sanyamsinghal, line 90: why was this conversion required?

makalaaneesh (approved):
LGTM

PR #1119, opened 2023-09-28
Support for fall-back

makalaaneesh:
- new fall-back commands
- cutover flag --with-fall-back that will automatically start fall-back
- resume logic for fall-back setup: first segment that has not been imported by target/fb-db-importer
- Archiver to also include fb-db-importer in importer count
- Fb-db-importer stats to correctly report remaining events (only consider events exported by target-db-importer)

DBZM PR: https://github.com/yugabyte/debezium/pull/93


priyanshi-yb (commented):
LGTM!

yb-voyager/cmd/constants.go
```diff
@@ -46,9 +46,11 @@ const (
 	TARGET_DB                     = "target"
 	FF_DB                         = "ff"
 	FF_DB_IMPORTER_ROLE           = "ff_db_importer"
+	FB_DB_IMPORTER_ROLE           = "fb_db_importer"
 	TARGET_DB_IMPORTER_ROLE       = "target_db_importer"
 	SOURCE_DB_EXPORTER_ROLE       = "source_db_exporter"
-	TARGET_DB_EXPORTER_ROLE       = "target_db_exporter"
+	TARGET_DB_EXPORTER_FF_ROLE    = "target_db_exporter_ff"
+	TARGET_DB_EXPORTER_FB_ROLE    = "target_db_exporter_fb"
```
amit-yb, line 53: Why do we need to distinguish between these two roles?

yb-voyager/cmd/fallBackStatus.go
```diff
@@ -0,0 +1,53 @@
+/*
```
amit-yb, line 1: Should we rename the file as `fallBackSwitchoverStatus.go`?

yb-voyager/cmd/importData.go
```diff
@@ -134,8 +134,8 @@ func startFallforwardSynchronizeIfRequired() {
 	if err != nil {
 		utils.ErrExit("could not fetch MigrationstatusRecord: %w", err)
 	}
-	if !msr.FallForwarDBExists {
-		utils.PrintAndLog("No fall-forward db exists. Exiting.")
+	if !msr.FallForwardEnabled && !msr.FallbackEnabled {
+		utils.PrintAndLog("No fall-forward/back enabled. Exiting.")
```
amit-yb, line 138: Nit: `fall-forward / fall-back`

amit-yb (approved):
LGTM with some minor nits.

yb-voyager/cmd/constants.go
makalaaneesh: I created a new exporter role here, because the trigger which the exporter deals with is different. Depending on whether we are in fall-forward/fall-back, the exporter needs to monitor for and subsequently create different triggers

Lmk if you think this can be dealt with in a better way.. It felt simpler to me to have the trigger mapped to the exporterRole 1:1

PR #1130, opened 2023-10-05
Added a hidden option --profile to start a pprof server for profiling

priyanshi-yb:


makalaaneesh (approved):
LGTM! @amit-yb any concerns with committing this to main? It's a hidden flag, so no visibility to the user.

PR #1132, opened 2023-10-05
CLI HELP: minor formatting change

makalaaneesh:
The options are just better formatted within parantheses. 
before: 
`--output-format string       allowed report formats: html | txt | json | xml (default "txt")`
after:
`--output-format string       format in which report will be generated: (html, txt, json, xml) (default "txt")` 
Examples:
docker pull image


priyanshi-yb (approved):
LGTM

PR #1138, opened 2023-10-10
Bug fix: migration uuid being printed more than once on console

sanyamsinghal:
- return if migrationUUID already present, instead of retrieving


yb-voyager/cmd/common.go
```diff
@@ -429,6 +429,9 @@ func initMetaDB() {
 
 // sets the global variable migrationUUID after retrieving it from MigrationStatusRecord
 func retrieveMigrationUUID() error {
+	if migrationUUID != uuid.Nil {
+		return nil
```
amit-yb, line 433: `return migrationUUID`???

shubham-yb (approved):

PR #1139, opened 2023-10-10
Added a flag object-types-list in export-schema

ShivanshGahlot:


amit-yb (approved):
LGTM

yb-voyager/src/srcdb/source.go
```diff
@@ -18,41 +18,44 @@ package srcdb
...
+	Uri                      string        `json:"uri"`
+	NumConnections           int           `json:"num_connections"`
+	VerboseMode              bool          `json:"verbose_mode"`
+	TableList                string        `json:"table_list"`
+	ExcludeTableList         string        `json:"exclude_table_list"`
+	UseOrafce                utils.BoolStr `json:"use_orafce"`
+	CommentsOnObjects        utils.BoolStr `json:"comments_on_objects"`
+	DBVersion                string        `json:"db_version"`
+	StrExportObjectTypesList string        `json:"str_export_object_types_list"`
+	ExportObjectTypesList    []string      `json:"export_object_types_list"`
```
sanyamsinghal, line 58: instead of storing it as a variable  in source struct we can define a method on source for this.
`source.GetExportObjectTypesList()` which will do the same.

I suggest that because it will avoid storing extra and redundant information in the MSR since it is stored and fetched multiple times in a migration.

ShivanshGahlot: Skipped `ExportObjectTypesList` while creating the MSR json

PR #1144, opened 2023-10-11
Avoid storing metaInfo about IndexesInfo if TABLE object type is skiped during schema export

sanyamsinghal:


yb-voyager/cmd/exportSchema.go
```diff
@@ -147,7 +147,11 @@ func clearSchemaIsExported() {
 }
 
 func updateIndexesInfoInMetaDB() {
-	log.Infof("updating indexes info in meta db")
+	log.Infof("updating indexes info in metaDB")
+	if !utils.ContainsString(source.ExportObjectTypesList, "TABLE") {
```
amit-yb, line 151: What if the user has already migrated TABLEs using some previous invocation?

sanyamsinghal: I didn't get your question.
If TABLEs are already exported in the previous invocation then the INDEXES dependent on them would also have been exported in that invocation itself.

amit-yb (approved):
LGTM

PR #1153, opened 2023-10-17
Removed 2.14 and misc changes

shubham-yb:


priyanshi-yb (approved):
LGTM

.github/workflows/mysql-migtests.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   run-mysql-migration-tests:
     strategy:
       matrix:
-        version: [2.14.13.0-b13,2.19.2.0-b121,2.16.7.0-b59,2.18.3.0-b75]
+        version: [2.19.2.0-b121,2.16.7.0-b59,2.18.3.0-b75]
```
sanyamsinghal, line 13: We should test and support from lowest versions of 2.16.x and 2.18.x, for example: 2.16.0
Because we user might be using any lower version and our docs also says 2.16.x.x

.github/workflows/pg-migtests.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   run-pg-migration-tests:
     strategy:
       matrix:
-        version: [2.14.13.0-b13,2.19.2.0-b121,2.16.7.0-b59,2.18.3.0-b75]
+        version: [2.19.2.0-b121,2.16.7.0-b59,2.18.3.0-b75]
```
sanyamsinghal, line 13: same as above

sanyamsinghal (changes requested):

sanyamsinghal (approved):

PR #1156, opened 2023-10-17
Modifying table list to include all child partitions only and change the event to include the root table name 

priyanshi-yb:
Oracle pipelines- 
1. ora2pg -https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/1259
2. Debezium - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/1260
3. live-migration - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/1261


yb-voyager/src/srcdb/postgres.go
```diff
@@ -469,3 +469,32 @@ func generateSSLQueryStringIfNotExists(s *Source) string {
 func (pg *PostgreSQL) GetServers() []string {
 	return []string{pg.source.Host}
 }
+
+func (pg *PostgreSQL) GetChildPartitions(tableName *sqlname.SourceName) []*sqlname.SourceName {
```
makalaaneesh, line 473: refactoring opportunity: (now or later): since both of the postgres/yugabyte structs share a lot of code, we can consider reusing the common code somehow (maybe separating out the logic into separate functions? Maybe yugabyte embeds postgres?)

yb-voyager/src/srcdb/postgres.go
```diff
@@ -469,3 +469,32 @@ func generateSSLQueryStringIfNotExists(s *Source) string {
...
+	childPartitions := make([]*sqlname.SourceName, 0)
+	query := fmt.Sprintf(`SELECT
+    nmsp_child.nspname  AS child_schema,
+    child.relname       AS child
+FROM pg_inherits
```
makalaaneesh, line 478: Are we sure pg_inherits table ONLY includes partition-type relationships?

makalaaneesh (commented):
Nice work! Added couple of nits and one question on the PG query

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: Yes, It stores the inheritance related information of the tables - 

> The catalog pg_inherits records information about table inheritance hierarchies. There is one entry for each direct child table in the database. (Indirect inheritance can be determined by following chains of entries.)

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: Yes, I think we discussed this earlier as well, we should do this but I think later is better because that will need a lot of code changes and testing for synchronize cmd, etc. I can add it to tech debt and we can pick this up after the release ?

yb-voyager/src/srcdb/postgres.go
makalaaneesh: @priyanshi-yb looks like "inheritance" in PG is a lot more than just partitions.. https://www.postgresql.org/docs/current/tutorial-inheritance.html

I think we should get more clarity here. Either we find a way to perform the above logic ONLY for partition-type inheritance, or we confirm that even in case of other inheritance relationships, this logic still is correct..

yb-voyager/src/srcdb/postgres.go
makalaaneesh: yep yep, pls create a jira 👍

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: Yes sure, let me understand this more and figure out the proper thing for our use case.

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: https://yugabyte.atlassian.net/browse/DB-8375

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: I have verified that this Query is still correct in the case of other Inheritance relationships with the example given in the above link so we are good with this.
On a side note, YugabyteDB doesn't support INHERITS at this moment ([GH issue](https://github.com/yugabyte/yugabyte-db/issues/1129)).

yb-voyager/src/srcdb/srcdb.go
```diff
@@ -41,10 +41,11 @@ type SourceDB interface {
 	FilterEmptyTables(tableList []*sqlname.SourceName) ([]*sqlname.SourceName, []*sqlname.SourceName)
 	GetColumnsWithSupportedTypes(tableList []*sqlname.SourceName, useDebezium bool, isStreamingEnabled bool) (map[*sqlname.SourceName][]string, []string)
 	GetTableColumns(tableName *sqlname.SourceName) ([]string, []string, []string)
-	IsTablePartition(table *sqlname.SourceName) bool
+	ParentTableOfPartition(table *sqlname.SourceName) string
 	GetColumnToSequenceMap(tableList []*sqlname.SourceName) map[string]string
 	GetAllSequences() []string
 	GetServers() []string
+	GetChildPartitions(table *sqlname.SourceName) []*sqlname.SourceName
```
amit-yb, line 48: If there are hundreds of tables that need to be exported, the "get root" and "get leaf" will be called those many number of times (which can be slow).

In future we may need to optimise this by having only one call to `GetTablePartitionTree() map[string][]string`. The implementation of the function makes only one metadata query on the source db to fetch all tables and their relationships and constructs a "tree" in memory.

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -185,19 +194,61 @@ func prepareDebeziumConfig(tableList []*sqlname.SourceName, tablesColumnList map
...
+	requiredForSource := source.DBType == "postgresql" || source.DBType == "yugabytedb"
+	if !requiredForSource  {
+		return tableList, nil
 	}
 
-	filteredTableList := []*sqlname.SourceName{}
+	modifiedTableList := []*sqlname.SourceName{}
+	//TODO: test when we upgrade to PG13+ as partitions are handled with root table 
+	//Refer- https://debezium.zulipchat.com/#narrow/stream/302529-community-general/topic/Connector.20not.20working.20with.20partitions
 	for _, table := range tableList {
```
amit-yb, line 207: Simplify the implementation as suggested on the slack.

yb-voyager/src/dbzm/config.go
```diff
@@ -45,6 +45,7 @@ type Config struct {
 	SchemaNames                 string
 	TableList                   []string
 	ColumnSequenceMap           []string
+	PartitionsToRootTableMap    []string
```
amit-yb, line 48: Let's call this as `TableRenameMapping string`

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -96,6 +100,10 @@ func prepareDebeziumConfig(tableList []*sqlname.SourceName, tablesColumnList map
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to prepare ssl params for debezium: %w", err)
 	}
+	var partitionsToRootTable []string
+	for partition, rootTable := range PartitionsToRootTableMap {
+		partitionsToRootTable = append(partitionsToRootTable, fmt.Sprintf("%s:%s", partition, rootTable))
+	}
```
amit-yb, line 106:
```
	tableRenameMapping := strings.Join(lo.MapToSlice(PartitionsToRootTableMap, func(k, v string) string {
		return fmt.Sprintf("%s:%s", k, v)
	}), ",")
```

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -185,19 +194,61 @@ func prepareDebeziumConfig(tableList []*sqlname.SourceName, tablesColumnList map
...
+
+func GetRootTableOfPartition(table *sqlname.SourceName) (*sqlname.SourceName, error){
+	parentTable := source.DB().ParentTableOfPartition(table)
+	if parentTable == "" {
+		return table, nil
+	}
+	defaultSourceSchema, noDefaultSchema := getDefaultSourceSchemaName()
+	if noDefaultSchema {
+		return nil, fmt.Errorf("default schema not found")
+	}
```
amit-yb, line 236: Use `table.SchemaName.Unquoted` as the "default schema name" to pass to the `NewSourceNameFromMaybeQualifiedName()`.

amit-yb (commented):
Few more comments.

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -414,3 +414,32 @@ func (yb *YugabyteDB) GetServers() []string {
...
+	childPartitions := make([]*sqlname.SourceName, 0)
+	query := fmt.Sprintf(`SELECT
+    nmsp_child.nspname  AS child_schema,
+    child.relname       AS child
+FROM pg_inherits
+    JOIN pg_class parent            ON pg_inherits.inhparent = parent.oid
+    JOIN pg_class child             ON pg_inherits.inhrelid   = child.oid
+    JOIN pg_namespace nmsp_parent   ON nmsp_parent.oid  = parent.relnamespace
+    JOIN pg_namespace nmsp_child    ON nmsp_child.oid   = child.relnamespace
+WHERE parent.relname='%s' AND nmsp_parent.nspname = '%s' `, tableName.ObjectName.MinQuoted, tableName.SchemaName.MinQuoted)
```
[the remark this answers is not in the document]
priyanshi-yb: Oh yes, I Just checked for quoted tables catalog table containing unquoted names. I used the one used earlier in a catalog query to check the table partition. I will make sure to check all these occurrences and fix them. Thanks

yb-voyager/src/srcdb/srcdb.go
priyanshi-yb: hmm yes, makes sense. will add a TODO.

yb-voyager/cmd/exportDataDebezium.go
priyanshi-yb: Actually, no can't use the schema of the table as the default schema because if the partition is non-public the result from the db query will include a qualified name always and only in the case of public schema it will not be qualified. So, we have to use default only and partitions can be cross schemas as well so we can't consider a table's schema to be the same as its root table or vice versa.

amit-yb (approved):
LGTM.

PR #1157, opened 2023-10-18
Bug fix for export data status while fetching event counts from metadb

priyanshi-yb:
2. Removed `LOG_MINING_FLUSH` table from the list when it stays in one migration and comes up when we do start-clean


makalaaneesh (approved):
LGTM

PR #1167, opened 2023-10-26
Bug fix [DB-8364]: Not update source db type from Target Exporter, only SOURCE_DB_EXPORTER_ROLE is allowed

sanyamsinghal:


yb-voyager/cmd/exportData.go
```diff
@@ -507,10 +507,14 @@ func clearDataIsExported() {
...
 
 func updateSourceDBConfInMSR() {
+	if exporterRole != SOURCE_DB_EXPORTER_ROLE {
+		return
+	}
 	metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
 		if record.SourceDBConf == nil {
 			record.SourceDBConf = source.Clone()
 			record.SourceDBConf.Password = ""
+			record.SourceDBConf.Uri = ""
```
priyanshi-yb, line 517: can we also make this change of Uri on the import side for target-db/ff-db as in case of ff-db URI is getting set in msr.

priyanshi-yb (approved):
LGTM with one comment!

amit-yb (approved):
LGTM.

PR #1168, opened 2023-10-26
Implemented `yb-voyager live-migration report` command to  print consolidated report of live migration workflow

priyanshi-yb:


yb-voyager/cmd/exportData.go
```diff
@@ -168,7 +168,8 @@ func exportData() bool {
 			if err != nil {
 				utils.ErrExit("failed to create trigger file after data export: %v", err)
 			}
-			displayExportedRowCountSnapshotAndChanges()
+			utils.PrintAndLog("\nRun the following command to get the current report of the migration:\n" +
+				color.CyanString("yb-voyager live-migration report --export-dir %s\n", exportDir))
```
amit-yb, line 172: Use `%q` format specifier for the path. Fix all such occurrences.

yb-voyager/cmd/liveMigrationReportCommand.go
```diff
@@ -0,0 +1,264 @@
...
+			tableExportStatus = &dbzm.TableExportStatus{
+				TableName:                tableName,
+				SchemaName:               schemaName,
+				ExportedRowCountSnapshot: 0,
+				FileName:                 "",
+			}
+		}
+		row.SnapshotRowCount = tableExportStatus.ExportedRowCountSnapshot
+		source = *msr.SourceDBConf
+		sourceSchemaCount := len(strings.Split(source.Schema, "|"))
```
amit-yb, line 118: This should be done only once before the beginning of the loop.

yb-voyager/cmd/liveMigrationReportCommand.go
```diff
@@ -0,0 +1,264 @@
...
+		row.TableName = table
+		if sourceSchemaCount <= 1 {
+			schemaName = ""
+			row.TableName = tableName
+		}
+		row.DBType = "Source"
+		err := updateRowForOutCounts(&row, tableName, schemaName) //source OUT counts
```
amit-yb, line 125: `updateExportedEventsCountInTheRow()`

yb-voyager/cmd/liveMigrationReportCommand.go
```diff
@@ -0,0 +1,264 @@
...
+		if err != nil {
+			utils.ErrExit("error while getting exported events counts for source DB: %w\n", err)
+		}
+		if fBEnabled {
+			err = updateRowForInCounts(&row, tableName, schemaName, msr.SourceDBAsTargetConf) //fall back IN counts
```
amit-yb, line 130: `updateImportedEventsCountInTheRow()`.

yb-voyager/cmd/liveMigrationReportCommand.go
```diff
@@ -0,0 +1,264 @@
...
+			if err != nil {
+				utils.ErrExit("error while getting imported events for source DB in case of fall-back: %w\n", err)
+			}
+		}
+
+		reportTable.AddRow(row.TableName, row.DBType, row.SnapshotRowCount, row.InsertsOut, row.UpdatesOut, row.DeletesOut, row.InsertsIn, row.UpdatesIn, row.DeletesIn, getFinalRowCount(row))
```
amit-yb, line 136: Factor it out in a separate function `addRowToTheTable(row, reportTable)`.

yb-voyager/cmd/liveMigrationReportCommand.go
```diff
@@ -0,0 +1,264 @@
...
+	DeletesOut       int64
+}
+
+var fBEnabled, fFEnabled bool
+
+func liveMigrationStatusCmdFn(msr *metadb.MigrationStatusRecord) {
+	fBEnabled = msr.FallbackEnabled
+	fFEnabled = msr.FallForwardEnabled
+	tableList := msr.TableListExportedFromSource
+	reportTable := uitable.New()
```
amit-yb, line 90: Should we rename it as `uitbl`?

yb-voyager/cmd/liveMigrationReportCommand.go
```diff
@@ -0,0 +1,264 @@
...
+	switch row.DBType {
+	case "Target":
+		importerRole = TARGET_DB_IMPORTER_ROLE
+	case "Fall Forward":
+		importerRole = FF_DB_IMPORTER_ROLE
+	case "Source":
+		importerRole = FB_DB_IMPORTER_ROLE
+	}
+	//reinitialise targetDB
+	tconf = *targetConf
```
amit-yb, line 180: Will this overwrite the password set in the `tconf`?

amit-yb (approved):
LGTM. 👍

yb-voyager/cmd/liveMigrationReportCommand.go
```diff
@@ -0,0 +1,264 @@
...
+				getFallForwardDBPassword(cmd)
+				migrationStatus.FallForwardDBConf.Password = tconf.Password
+			}
+			if migrationStatus.FallbackEnabled {
+				getSourceDBPassword(cmd)
+				migrationStatus.SourceDBAsTargetConf.Password = source.Password
+			}
+			liveMigrationStatusCmdFn(migrationStatus)
+		} else {
+			utils.ErrExit("export-type 'snapshot-and-changes' is not enabled for this migration")
```
makalaaneesh, line 67: Maybe a more msg like `live-migration report is only applicable when export-type is 'snapshot-and-changes' ?

yb-voyager/cmd/liveMigrationReportCommand.go
```diff
@@ -0,0 +1,264 @@
...
+	UpdatesOut       int64
+	DeletesOut       int64
+}
+
+var fBEnabled, fFEnabled bool
+
+func liveMigrationStatusCmdFn(msr *metadb.MigrationStatusRecord) {
+	fBEnabled = msr.FallbackEnabled
+	fFEnabled = msr.FallForwardEnabled
+	tableList := msr.TableListExportedFromSource
```
makalaaneesh, line 89: Will this include the empty tables?

yb-voyager/cmd/liveMigrationReportCommand.go
```diff
@@ -0,0 +1,264 @@
...
+	reportTable := uitable.New()
+	reportTable.MaxColWidth = 50
+	reportTable.Separator = " | "
+
+	addHeader(reportTable, "TABLE", "DB TYPE", "SNAPSHOT ROW COUNT", "EXPORTED", "EXPORTED", "EXPORTED", "IMPORTED", "IMPORTED", "IMPORTED", "FINAL ROW COUNT")
+	addHeader(reportTable, "", "", "", "INSERTS", "UPDATES", "DELETES", "INSERTS", "UPDATES", "DELETES", "")
+	exportStatusFilePath := filepath.Join(exportDir, "data", "export_status.json")
+	status, err := dbzm.ReadExportStatus(exportStatusFilePath)
```
makalaaneesh, line 97: rename variable to `dbzmStatus` ?

yb-voyager/cmd/liveMigrationReportCommand.go
```diff
@@ -0,0 +1,264 @@
...
+		} else {
+			utils.ErrExit("export-type 'snapshot-and-changes' is not enabled for this migration")
+		}
+	},
+}
+
+type rowData struct {
+	TableName        string
+	DBType           string
+	SnapshotRowCount int64
```
makalaaneesh, line 75: I wonder if we should have two fields "Exported SNapshot" and "Imported Snapshot". 
That would be consistent with the inserts/updates/deletes and also make it more clear that the snapshot is exporteed from the source and imported into target(and ff)

makalaaneesh (approved):
Nicely done! LGTM!

yb-voyager/cmd/liveMigrationReportCommand.go
priyanshi-yb: No, we are first taking the password set in `tconf` to set it back in msr.TargetConf in the starting after getting the password and then directly using that later.
```
if streamChanges {
  getTargetPassword(cmd)
  migrationStatus.TargetDBConf.Password = tconf.Password
  if migrationStatus.FallForwardEnabled {
  getFallForwardDBPassword(cmd)
  migrationStatus.FallForwardDBConf.Password = tconf.Password
  }
  if migrationStatus.FallbackEnabled {
  getSourceDBPassword(cmd)
  migrationStatus.SourceDBAsTargetConf.Password = source.Password
}
```

yb-voyager/cmd/liveMigrationReportCommand.go
priyanshi-yb: Yes

yb-voyager/cmd/liveMigrationReportCommand.go
priyanshi-yb: Hmm...you are suggesting having an 'Exported snapshot', and an 'Imported snapshot' separate fields but then we never export snapshots from DB other than `Source`. So for that case, it will be 0 0.
I think that it should be implicit that the Snapshot row count for Source is exported and Target will be imported only but for inserts/updates/deletes this can be the other way around as well. so having explicit Exported/Imported mention makes sense there.

priyanshi-yb: Created a ticket - https://yugabyte.atlassian.net/browse/DB-8658, will do this change in the separate PR.

PR #1169, opened 2023-10-27
Bug Fix [DB-8045]: Support sequence resuming in live migration for Oracle/YugabyteDB GENERATED BY DEFAULT IDENTITY columns

sanyamsinghal:
https://yugabyte.atlassian.net/browse/DB-8045

1. Oracle: `ALTER TABLE table MODIFY column GENERATED BY DEFAULT AS IDENTITY(START WITH LIMIT VALUE)`
> Here, the `START WITH LIMIT VALUE` clause takes care of setting the value to max value of the `IDENTITY` column.
So the rest of the part of this DDL won't have any affect if column is already `GENERATED BY DEFAULT as IDENTITY`

2. YugabyteDB: `ALTER TABLE table ALTER COLUMN column SET GENERATED BY DEFAULT`
> Similar to oracle, also found as per the testing when this ALTER is run, the column value gets set to the next unique value of the sequence attached to it.


amit-yb (approved):
LGTM.

amit-yb (approved):
LGTM.

PR #1173, opened 2023-10-30
Fall-forward/fall-back commands graceful error

makalaaneesh:
- validateExportDir for all commands except `version`
- validate if metaDB is not created for relevant commands


amit-yb (approved):
LGTM

PR #1174, opened 2023-10-31
Fix Fall-back import: ignore event if importerRole = FB_DB and event exporter role is not TARGET_DB_FB

makalaaneesh:


amit-yb (approved):
LGTM.

PR #1179, opened 2023-11-01
[DB-8651] Bug fix: unlockExportDir() not being called for import/export data command in case of fall-forward/fall-back enabled

sanyamsinghal:
- calling unlockExportDir() manually before switching the process to fall-forward/fall-back sync/setup commands

This bug wa causing exportData and importData lockfile to be leftover even after everything is complete in case of fall-back and fall-forward, which this PR fixes.

Jira Ticket: https://yugabyte.atlassian.net/browse/DB-8651


priyanshi-yb (approved):
LGTM

PR #1180, opened 2023-11-01
Added a small check to not display which objects are being added as a dependency when object-type-list flag is not provided

ShivanshGahlot:


shubham-yb (approved):
LGTM

PR #1188, opened 2023-11-08
Bug fix: Migration State was not getting cleared from target DB

sanyamsinghal:
- introduced tdb.isQueryResultNonEmpty() in place of tdb.isQueryResultEmpty() which was getting checked previously without negation inside isTableExist() and isSchemaExist()


priyanshi-yb (approved):
LGTM

PR #1189, opened 2023-11-08
Fix: case where live-migration report is run when only export is started or snapshot import is going on

priyanshi-yb:


yb-voyager/cmd/liveMigrationReportCommand.go
```diff
@@ -229,7 +234,17 @@ func updateImportedEventsCountsInTheRow(row *rowData, tableName string, schemaNa
 
 	eventCounter, err := state.GetImportedEventsStatsForTable(tableName, migrationUUID)
 	if err != nil {
-		return fmt.Errorf("get imported events stats for table %q for DB type %s: %w", tableName, row.DBType, err)
+		if !strings.Contains(err.Error(), "cannot assign NULL to *int64") {
+			return fmt.Errorf("get imported events stats for table %q for DB type %s: %w", tableName, row.DBType, err)
+		} else {
+			//in case import streaming is not started yet, metadata will not be initialized
+			log.Errorf("streaming ingestion is not started yet for table %q for DB type %s", tableName, row.DBType)
```
amit-yb, line 241: Nit: If you've handled an error so that the code continues further even after the error, then log it as a warning.

amit-yb (approved):
LGTM

PR #1194, opened 2023-11-08
Changed `live-migration report` command to `get data-migration-report`

priyanshi-yb:
Also, changed the DB type names on the UI table from Source, Target and Fall Forward -> source, target and source-replica to match it with the new CLI proposal where we are using import-data-to-source/target/source-replica. 
https://yugabyte.atlassian.net/browse/DB-8835

### DIFF
--- a/yb-voyager/cmd/getCommand.go
+++ b/yb-voyager/cmd/getCommand.go
@@ -21,12 +21,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var liveMigrationCommand = &cobra.Command{
-	Use:   "live-migration",
-	Short: "This command will display the report of any live migration workflow",
-	Long:  ``,
+var getCommand = &cobra.Command{
+	Use:   "get",
+	Short: "Gets the report of the migration.",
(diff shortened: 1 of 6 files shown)


yb-voyager/cmd/getCommand.go
```diff
@@ -21,12 +21,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var liveMigrationCommand = &cobra.Command{
-	Use:   "live-migration",
-	Short: "This command will display the report of any live migration workflow",
-	Long:  ``,
+var getCommand = &cobra.Command{
+	Use:   "get",
+	Short: "Gets the report of the migration.",
```
makalaaneesh, line 26: `Get migration reports` ? For now, we can specify "reports" at this level, but if we add any other subcommands under "get", we should remember to change this.

makalaaneesh (approved):
LGTM

yb-voyager/cmd/getCommand.go
priyanshi-yb: Yes, correct

PR #1187, opened 2023-11-08
Ux changes for FF/FB

makalaaneesh:

PR #1195, opened 2023-11-09
Introduced lockfile package to abstract out all operations related to locking export-dir for multiple identical voyager commands

sanyamsinghal:


yb-voyager/src/lockfile/lockfile.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return -1, fmt.Errorf("failed to parse PID from lockfile %q: %w", l.fpath, err)
+	}
+	return l.cmdPID, nil
+}
+
+func (l *Lockfile) IsPIDActive() bool {
+	pid, err := l.GetCmdPID()
+	if err != nil {
+		return false
```
priyanshi-yb, line 56: How are we considering the error from `GetCmdPID` as `PID` not active?

priyanshi-yb (approved):
LGTM with few comments.

yb-voyager/cmd/endMigrationCommand.go
```diff
@@ -462,7 +459,6 @@ func checkIfEndCommandCanBePerformed(msr *metadb.MigrationStatusRecord) {
 						log.Warnf("stopping ongoing voyager command %q with PID=%d: %v", ongoingCmd, ongoingCmdPID, err)
 					}
 				}
-				time.Sleep(time.Second * 2) // wait for the ongoing commands to completely exit
```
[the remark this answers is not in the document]
sanyamsinghal: it is handled in this function:
```
// this function wait for process to exit after signalling kill to it
func killProcessWithPID(pid int) error {
	process, _ := os.FindProcess(pid) // Always succeeds on Unix systems

	err := process.Signal(syscall.SIGUSR2)
	if err != nil {
		return fmt.Errorf("sending SIGUSR2 signal to process with PID=%d: %w", pid, err)
	}

	// Reference: https://mezhenskyi.dev/posts/go-linux-processes/
	// Poll for 10 sec to make sure process is terminated
	// here process.Signal(syscall.Signal(0)) will return error only if process is not running
	for i := 0; i <= 10; i++ {
		time.Sleep(time.Second * 1)
		err = process.Signal(syscall.Signal(0))
		if err != nil {
			return nil
		}
	}
	return nil
}
```

yb-voyager/src/lockfile/lockfile.go
sanyamsinghal:
```
func (l *Lockfile) GetCmdPID() (int, error) {
	if l.cmdPID != -1 {
		return l.cmdPID, nil
	}

	bytes, err := os.ReadFile(l.fpath)
	if err != nil {
		return -1, fmt.Errorf("failed to read lockfile %q: %w", l.fpath, err)
	}
	l.cmdPID, err = strconv.Atoi(strings.Trim(string(bytes), " \n"))
	if err != nil {
		return -1, fmt.Errorf("failed to parse PID from lockfile %q: %w", l.fpath, err)
	}
	return l.cmdPID, nil
}
```

the case in which GetCmdPID() can fail is when it fails to read the pid from file which will happen when file doesn't exists because of either the command is not running or got exited/completed in the meantime.

priyanshi-yb: Got it! thanks.

yb-voyager/cmd/root.go
```diff
@@ -96,18 +97,19 @@ func startPprofServer() {
...
+	"yb-voyager import data status",
+	"yb-voyager export data status",
+	"yb-voyager cutover status",
+	"yb-voyager fall-forward status",
+	"yb-voyager cutover initiate",
+	"yb-voyager fall-forward switchover",
+	"yb-voyager end",
+	"yb-voyager end migration",
+}
+
```
priyanshi-yb, line 111: this list needs to be modified with this #1187. cc: @makalaaneesh

priyanshi-yb (approved):

yb-voyager/cmd/root.go
makalaaneesh: Correct. 
Get rid of 
- yb-voyager fall-forward status
- yb-voyager cutover initiate
- yb-voyager fall-forward switchover

Add:
- yb-voyager initiate cutover to target/source/source-replica

sanyamsinghal: cool.
Also need to remove `"yb-voyager cutover status"` then.

sanyamsinghal: so this is the new list: 
```

var noLockNeededList = []string{
	"yb-voyager version",
	"yb-voyager import data status",
	"yb-voyager export data status",
	"yb-voyager initiate cutover to source",
	"yb-voyager initiate cutover to target",
	"yb-voyager initiate cutover to source-replica",
	"yb-voyager end",
	"yb-voyager end migration",
}
```

sanyamsinghal: cc @priyanshi-yb @makalaaneesh ^^

PR #1208, opened 2023-11-23
Removed unnecessary flags from data-migration-report and minor cli help changes

priyanshi-yb:


yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -283,7 +283,8 @@ func getFinalRowCount(row rowData) int64 {
 
 func init() {
 	getCommand.AddCommand(getDataMigrationReportCmd)
-	registerCommonGlobalFlags(getDataMigrationReportCmd)
```
makalaaneesh, line 286: I understand we want to get rid of "yes" and "send-diagnostics" for this command, but we seem to be duplicating the export-dir flag definition in multiple places. (cutover related commands as well have their separate definition). 

How about a function for registering the export-dir flag?

priyanshi-yb: Yes makes sens

yb-voyager/cmd/importDataToSource.go
```diff
@@ -24,7 +24,8 @@ import (
 
 var importDataToSourceCmd = &cobra.Command{
 	Use:   "source",
-	Short: "Import data into the source DB to prepare for fall-back.",
+	Short: "Import data into the source DB to prepare for fall-back.\n" +
+	"For more details and examples, visit https://docs.yugabyte.com/preview/yugabyte-voyager/reference/data-migration/import-data/#import-data-to-source",
```
makalaaneesh, line 28: Maybe it would be more useful to link to the workflow page for fall-back?  The reference is almost always same as the help msg of the command, so it might not be that useful. 
Same for fall-forward. 

What do you think?

makalaaneesh (approved):
LGTM

yb-voyager/cmd/importDataToSource.go
priyanshi-yb: yes, the workflow page is linking back to the reference page, so better to link that in the help. Thanks

PR #1209, opened 2023-11-23
Bug fix: Nil pointer exception if export data status command is executed during import data file workflow

sanyamsinghal:
- updated the source db type in meta DB while preparing the export-dir for import data file cmd


amit-yb (approved):
LGTM

PR #1211, opened 2023-11-27
checking the live migration in the report/status commands with only "export-type", removed dbzm mode

priyanshi-yb:


makalaaneesh (approved):
LGTM

PR #1212, opened 2023-11-27
Bug fixes for special handling of stopping archive command from end migration

sanyamsinghal:
[Bug fix 1]: Stopping of 'export data from target' in case of end migration handling archiver termination

[Bug fix 2]: Voyager End Migration not waiting for ongoing commands to stop after 10sec

[Bug fix 3]: Previously we were just exiting if there we no segments to Archive, but that can be wrong in case the importer is not started or importer is lagging/slow in importing the segments so that they are eligible for archiving
- now, new check is introduced to verify if the only unarchived segment is the ongoing segment and then stop the archive changes command

Debezium side PR: https://github.com/yugabyte/debezium/pull/101

### DIFF
--- a/yb-voyager/cmd/archiveChangesCommand.go
+++ b/yb-voyager/cmd/archiveChangesCommand.go
@@ -25,6 +25,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/tebeka/atexit"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/metadb"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
 )
@@ -221,9 +222,10 @@ func (m *EventSegmentCopier) Run() error {
 		}
 
 		// Note/TODO: last incomplete segment will remain unarchived
-		if segmentsToArchive == nil && StopArchiverSignal {
+		// It won't wait for the importer to finish importing and archiving the segments
+		if StopArchiverSignal && len(segmentsToArchive) == 0 {
 			utils.PrintAndLog("\n\nReceived signal to terminate due to end migration command.\nArchiving changes completed. Exiting...")
-			return nil
+			atexit.Exit(0)
 		}
 
 		for _, segment := range segmentsToArchive {
--- a/yb-voyager/cmd/archiveChangesCommand.go
+++ b/yb-voyager/cmd/archiveChangesCommand.go
@@ -47,9 +47,6 @@ Note that: even if some changes are applied to the target databases, they are de
 	Run: archiveChangesCommandFn,
 }
 
-var archivingCopier *EventSegmentCopier
-var archivingDeleter *EventSegmentDeleter
-
 func archiveChangesCommandFn(cmd *cobra.Command, args []string) {
 	if moveDestination != "" && deleteSegments {
 		utils.ErrExit("only one of the --move-to and --delete-changes-without-archiving should be set")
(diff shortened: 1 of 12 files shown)


yb-voyager/cmd/endMigrationCommand.go
```diff
@@ -569,8 +569,10 @@ func checkIfEndCommandCanBePerformed(msr *metadb.MigrationStatusRecord) {
 func stopVoyagerCommands(msr *metadb.MigrationStatusRecord, lockFiles []*lockfile.Lockfile) {
 	if msr.ArchivingEnabled {
 		exportDataLockFile := getLockFileForCommand(lockFiles, "export data")
+		exportDataFromTargetLockFile := getLockFileForCommand(lockFiles, "export data from target")
```
makalaaneesh, line 572: From the looks of it, we only stop "export data" or "export data from target", but there is also "export data from source". (which is the same as "export data" but different from a command perspective.

yb-voyager/src/metadb/metadataDB.go
```diff
@@ -460,6 +460,32 @@ func (m *MetaDB) GetSegmentsToBeDeleted() ([]utils.Segment, error) {
...
+	}
+	if len(segments) == 0 {
+		return -1, nil
+	}
+	return segments[len(segments)-1].Num, nil
+}
+
+func (m *MetaDB) GetOngoingSegmentNum() (int, error) {
+	// sample query: SELECT MAX(segment_no) FROM queue_segment_meta WHERE imported_by_target_db_importer + imported_by_ff_db_importer = 0;
+	predicate := "imported_by_target_db_importer + imported_by_ff_db_importer + imported_by_fb_db_importer = 0"
```
ShivanshGahlot, line 478: Shouldn't the predicate be:
`imported_by_target_db_importer + imported_by_ff_db_importer + imported_by_fb_db_importer != importer_count`

yb-voyager/cmd/archiveChangesCommand.go
```diff
@@ -47,6 +48,9 @@ Note that: even if some changes are applied to the target databases, they are de
 	Run: archiveChangesCommandFn,
 }
 
+var archivingCopier *EventSegmentCopier
+var archivingDeleter *EventSegmentDeleter
```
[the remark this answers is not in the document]
sanyamsinghal: since they are now moved to global var level. So i wanted their name to be specific instead of a generic name so that to avoid potential duplicacy of var names and bugs in future.

amit-yb (commented):
Also modify the `EventSegmentDeleter` to properly return if stop signal is set and there are all segments are deleted.

yb-voyager/cmd/archiveChangesCommand.go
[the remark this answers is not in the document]
amit-yb: For now, I would suggest restoring the names. We can always rename them if such a situation arises in future.

yb-voyager/cmd/endMigrationCommand.go
```diff
@@ -566,11 +566,32 @@ func checkIfEndCommandCanBePerformed(msr *metadb.MigrationStatusRecord) {
 	}
 }
 
+func getLockFileForCommand(lockFiles []*lockfile.Lockfile, cmdName string) *lockfile.Lockfile {
+	for _, lockFile := range lockFiles {
+		if lockFile.GetCmdName() == cmdName {
+			return lockFile
+		}
+	}
+	return nil
```
amit-yb, line 575: Nit: `lo.Find()`?

yb-voyager/cmd/endMigrationCommand.go
```diff
@@ -579,64 +600,73 @@ func stopVoyagerCommands(msr *metadb.MigrationStatusRecord, lockFiles []*lockfil
...
-	return nil
+}
+
+func stopDataExportCommand(lockFile *lockfile.Lockfile) {
+	if lockFile == nil || !lockFile.IsPIDActive() {
+		return
+	}
+
+	metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
+		record.EndMigrationRequested = true
```
amit-yb, line 656: Add a comment here stating that dbzm plugin detects this flag change and then gracefully exits which in turn causes the `export data` command to exit cleanly.

amit-yb (commented):
LGTM with minor nits.

amit-yb (approved):

PR #1216, opened 2023-11-28
Bug fix: Using same ImportCount for Fall-Forward DB even after the exporter_role is switched to target_db_exporter i.e. cutover to target 

sanyamsinghal:
- due to which GetSegmentsToArchive() was not detected the segments for archiving

Bug fix: Added check condition 'archived = 0' in GetSegmetnsToBeArchived() in case of restart and resume

### DIFF
--- a/yb-voyager/src/metadb/metadataDB.go
+++ b/yb-voyager/src/metadb/metadataDB.go
@@ -433,9 +433,10 @@ func (m *MetaDB) GetExportedEventsStatsForTableAndExporterRole(exporterRole stri
 	}, nil
 }
 
+
 func (m *MetaDB) GetSegmentsToBeArchived(importCount int) ([]utils.Segment, error) {
-	// sample query: SELECT segment_no, file_path FROM queue_segment_meta WHERE imported_by_target_db_importer + imported_by_ff_db_importer = 2 AND archived = 0 ORDER BY segment_no;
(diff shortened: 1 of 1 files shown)


yb-voyager/src/metadb/metadataDB.go
```diff
@@ -454,8 +456,8 @@ func (m *MetaDB) GetSegmentsToBeDeleted() ([]utils.Segment, error) {
 }
 
 func (m *MetaDB) GetPendingSegments(importCount int) ([]utils.Segment, error) {
-	// sample query: SELECT segment_no, file_path FROM queue_segment_meta WHERE imported_by_target_db_importer + imported_by_ff_db_importer < importCount ORDER BY segment_no;
-	predicate := fmt.Sprintf(`imported_by_target_db_importer + imported_by_ff_db_importer + imported_by_fb_db_importer < %d`, importCount)
+	predicate := fmt.Sprintf(`(exporter_role == 'source_db_exporter' AND (imported_by_target_db_importer + imported_by_ff_db_importer + imported_by_fb_db_importer < %d)) OR
+		(exporter_role LIKE 'target_db_exporter%%' AND (imported_by_target_db_importer + imported_by_ff_db_importer + imported_by_fb_db_importer < 1))`, importCount)
```
amit-yb, line 460: In future we should use "generated columns" feature of sqlite (https://www.sqlite.org/gencol.html) to add a column `imported_count` to mean `imported_by_target_db_importer + imported_by_ff_db_importer + imported_by_fb_db_importer`

sanyamsinghal: yeah good idea.

PR #1235, opened 2023-12-05
[PG OFFLINE EXPORT DATA] Fix double quote single table in table-list 

makalaaneesh:
In cases where there is a single table provided in the table-list, there is a bug. 
The ini file generated(after executing the template) for pg_dump args has the following: `table='public."TEST_SIMPLE"'`
When we parse this ini file, by default, the single quotes surrounding the value is dropped. 

This results in pg_dump command being generated as `--table=public."TEST_SIMPLE"`, (expected: `table='public."TEST_SIMPLE"'`) . 

This PR fixes the bug by passing the option `PreserveSurroundedQuote` to the ini file loader. 

This case works perfectly fine when there are more than one tables. In such a case, the ini value in file is of the following - `table='public."TEST_SIMPLE"' --table='public.kv'`. This is NOT interpreted as surrounding quotes by ini, and therefore, it doesn't drop the  quotes.

### DIFF
--- a/migtests/scripts/run-test-export-data.sh
+++ b/migtests/scripts/run-test-export-data.sh
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ $# -ne 1 ]
+then
+	echo "Usage: $0 TEST_NAME"
+	exit 1
+fi
+
+set -x
+
+export TEST_NAME=$1
+
+export REPO_ROOT="${PWD}"
+export SCRIPTS="${REPO_ROOT}/migtests/scripts"
+export TESTS_DIR="${REPO_ROOT}/migtests/tests"
+export TEST_DIR="${TESTS_DIR}/${TEST_NAME}"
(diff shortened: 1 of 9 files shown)


priyanshi-yb (approved):
LGTM

migtests/scripts/run-test-export-data.sh
```diff
@@ -0,0 +1,79 @@
...
+	
+	step "End Migration: clearing metainfo about state of migration from everywhere."
+	end_migration --yes
+
+	step "Clean up"
+	./cleanup-db
+	rm -rf "${EXPORT_DIR}/*"
+	run_ysql yugabyte "DROP DATABASE IF EXISTS ${TARGET_DB_NAME};"
+}
+
```
priyanshi-yb, line 78: we can run the whole migration as well, and in that case, we can use `run-test.sh`??

makalaaneesh: I purposely went with only export-data. Reason being - we already have a test that tests entire migration of case sensitive table names. 
In this case, the bug was only in export-data, so it would be redundant to test out the whole migration.

PR #1241, opened 2023-12-08
Test latest ora2pg

priyanshi-yb:


priyanshi-yb:
Will refer to this one in the future when we test the new latest of ora2pg. Closing it for now

PR #1242, opened 2023-12-12
Bug fix: Docker containers using host namespace for their PID so that…

sanyamsinghal:
… the docker container processes can interact with each and signal to stop in case of end migration command

### DIFF
--- a/docker/yb-voyager-docker
+++ b/docker/yb-voyager-docker
@@ -151,7 +151,7 @@ then
 	tty="-it"
 fi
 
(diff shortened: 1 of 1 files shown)


sanyamsinghal:
Jenkins Automation Tests:
1. Live Migration(using dockerized installation of voyager): https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/1530/

PR #1249, opened 2023-12-14
Bug fix: Source DB is not cleaned up if --save-migration-reports flag is provided as false

sanyamsinghal:
- currently, a global var for detecting streaming mode is used which is populated only from the save migration reports section(if enabled)


sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/1550/

amit-yb (approved):
LGTM

PR #1254, opened 2023-12-20
Make import events work for multiple target Schemas in PG

makalaaneesh:
While testing pg live migration for non-public schemas, it does not work. The reason that it does not work is that while importing events, we essentially ignore the schema name in the event, and rely on the targetSchema. This makes sense for oracle because we should insert into the targetSchema and it is guaranteed that there is going to be only ONE schema. This is not the case for pg. 

After this PR:
- if target-db-schema is passed in import-data  for source=PG, we throw an error. It is simply not allowed to change the schema. 
- Both multiple schemas/single schema is supported from PG. While converting the event, we set the appropriate schema name. For oracle migrations, we set the schema to tconf.schema. for PG migrations, we leave it unchanged (because we don't allow changing the schema)


amit-yb (approved):
LGTM

PR #1259, opened 2023-12-22
Added support for PostgreSQL as TargetDB

priyanshi-yb:
Oracle pipeline -https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/1594/


yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -436,17 +448,17 @@ func checkAndHandleSnapshotComplete(status *dbzm.ExportStatus, progressTracker *
...
-	if err != nil {
-		return false, fmt.Errorf("failed to write data file descriptor: %w", err)
-	}
-	log.Infof("snapshot export is complete.")
-	err = renameDbzmExportedDataFiles()
-	if err != nil {
-		return false, fmt.Errorf("failed to rename dbzm exported data files: %v", err)
-	}
 	if !isTargetDBExporter(exporterRole) {
+		setDataIsExported()
```
makalaaneesh, line 452: FYI while merging: I added a condition here to check if config.SnapshotMode != "never" (need this even in case of sourceDBExporter beccause we now use pg_dump. so can't use `isTargetDBExporter`)

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -514,6 +526,11 @@ func renameDbzmExportedDataFiles() error {
 			newFilePath = filepath.Join(exportDir, "data", status.Tables[i].SchemaName+"."+tableName+"_data.sql")
 		}
 
+		if !utils.FileOrFolderExists(oldFilePath) && utils.FileOrFolderExists(newFilePath) { //In case of restarts rename should not be done again else it will error out
```
makalaaneesh, line 529: this case is not required anymore because it's handled in `checkAndHandleSnapshotComplete`, correct?

yb-voyager/cmd/exportDataFromTarget.go
```diff
@@ -74,13 +74,18 @@ func initSourceConfFromTargetConf() error {
...
 	}
+	sourceDBConf := msr.SourceDBConf
 	targetConf := msr.TargetDBConf
 	source.DBType = targetConf.TargetDBType
 	source.Host = targetConf.Host
 	source.Port = targetConf.Port
 	source.User = targetConf.User
 	source.DBName = targetConf.DBName
-	source.Schema = targetConf.Schema
+	if sourceDBConf.DBType == POSTGRESQL {
```
makalaaneesh, line 84: What happens if:
- source db has schema = foo
- user has decided to migrate all tables into schema = bar (passed it via --target-db-schema)

We are making an assumption here that it is unchanged.
(was just discussing this with Amit in some other context)...

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -215,6 +215,10 @@ func updateImportedEventsCountsInTheRow(row *rowData, tableName string, schemaNa
 		return fmt.Errorf("data file descriptor not found at %s for snapshot", dataFileDescriptorPath)
 	}
 
+	if schemaName != "public" && schemaName != "" { //multiple schema specific
```
makalaaneesh, line 218: You should also add the case "if source db type is PG"  I guess? This will break for cases of non-pg where schemaName is not public, right?

yb-voyager/cmd/import.go
```diff
@@ -166,7 +166,7 @@ func registerSourceReplicaDBAsTargetConnFlags(cmd *cobra.Command) {
 		"[For Oracle Only] Name of TNS Alias you wish to use to connect to Oracle instance. Refer to documentation to learn more about configuring tnsnames.ora and aliases")
 
 	cmd.Flags().StringVar(&tconf.Schema, "source-replica-db-schema", "",
-		"schema name in Source-Replica DB") // TODO: add back note after we suppport PG/Mysql - `(Note: works only for source as Oracle and MySQL, in case of PostgreSQL you can ALTER schema name post import)`
+		"schema name in Source-Replica DB (Note: works only for source as Oracle, in case of PostgreSQL schemas remain same as of source)")
```
makalaaneesh, line 169: Was just testing this for import data - it works if you pass a different schema!

yb-voyager/cmd/import.go
```diff
@@ -272,6 +272,8 @@ func validateTargetPortRange() {
 			tconf.Port = ORACLE_DEFAULT_PORT
 		} else if tconf.TargetDBType == YUGABYTEDB {
 			tconf.Port = YUGABYTEDB_YSQL_DEFAULT_PORT
+		} else if tconf.TargetDBType == POSTGRESQL {
```
makalaaneesh, line 275: pls add to cli arg documentation, if not already

makalaaneesh (commented):
Added some initial comments. 
I think the cases of non-public/multiple schemas) needs to be handled carefully. Will try to think it through..

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -436,17 +448,17 @@ func checkAndHandleSnapshotComplete(status *dbzm.ExportStatus, progressTracker *
...
-	err := writeDataFileDescriptor(exportDir, status)
-	if err != nil {
-		return false, fmt.Errorf("failed to write data file descriptor: %w", err)
-	}
-	log.Infof("snapshot export is complete.")
-	err = renameDbzmExportedDataFiles()
-	if err != nil {
-		return false, fmt.Errorf("failed to rename dbzm exported data files: %v", err)
-	}
 	if !isTargetDBExporter(exporterRole) {
```
amit-yb, line 451: How about checking `isSourceDBExporter(exporterRole)`?

yb-voyager/cmd/exportDataDebezium.go
amit-yb, line 529: You will only need to check the later part of the condition.

yb-voyager/cmd/exportDataFromTarget.go
priyanshi-yb: This case is not possible `--target-db-schema` is not supported for PG migration, it is migrated to the same schema as everything is fully-qualified with same schema name while we export using pg_dump. This has to be same as source.

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -0,0 +1,807 @@
...
+func (pg *TargetPostgreSQL) InitConnPool() error {
+	tconfs := []*TargetConf{pg.tconf}
+	var targetUriList []string
+	for _, tconf := range tconfs {
+		targetUriList = append(targetUriList, tconf.Uri)
+	}
+	log.Infof("targetUriList: %s", utils.GetRedactedURLs(targetUriList))
+
+	if pg.tconf.Parallelism == 0 {
+		pg.tconf.Parallelism = fetchDefaultParllelJobs(tconfs) //TODO:
```
amit-yb, line 213: Fix this TODO and the one on 220 -- these are critical parts.

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -0,0 +1,807 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("scan column name: %w", err)
+		}
+		result = append(result, colName)
+	}
+	return result, nil
+}
+
+func (pg *TargetPostgreSQL) IsNonRetryableCopyError(err error) bool {
+	return true //TODO:
```
amit-yb, line 467: At least put `err != nil`

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -0,0 +1,807 @@
+/*
```
amit-yb, line 1: File a ticket to factor out common code between PG and YB.

amit-yb (approved):
LGTM. Good work!

yb-voyager/cmd/exportDataFromTarget.go
priyanshi-yb: `target-db-schema` doesn't work in the case of PG migration, we have this code in `importSchema.go:258`, where we handle this target-db-schema based on source-db-types. 
```
func createTargetSchemas(conn *pgx.Conn) {
	var targetSchemas []string
	tconf.Schema = strings.ToLower(strings.Trim(tconf.Schema, "\"")) //trim case sensitivity quotes if needed, convert to lowercase
	switch sourceDBType {
	case "postgresql": // in case of postgreSQL as source, there can be multiple schemas present in a database
		source = srcdb.Source{DBType: sourceDBType}
		targetSchemas = utils.GetObjectNameListFromReport(analyzeSchemaInternal(), "SCHEMA")
         ....
         
     for _, targetSchema := range targetSchemas {
		//check if target schema exists or not
         ....
	if sourceDBType != POSTGRESQL { // with the new schema list flag, pg_dump takes care of all schema creation DDLs
		schemaExists := checkIfTargetSchemaExists(conn, tconf.Schema)
		createSchemaQuery := fmt.Sprintf("CREATE SCHEMA %s", tconf.Schema)
```

yb-voyager/cmd/getDataMigrationReportCommand.go
priyanshi-yb: Yes, correct. I wanted to get this checked and thought this would be checked in the automation of oracle live but I just realised we are not using non-public schema in those cases. Will fix one test case in a separate PR. Now we need it as we are dealing with schema at a lot of places with PG support.

yb-voyager/cmd/common.go
```diff
@@ -513,3 +515,14 @@ func hideExportFlagsInFallForwardOrBackCmds(cmd *cobra.Command) {
 		}
 	}
 }
+
+func getDefaultPGSchema(schema string) (string, bool) {
+	schemas := strings.Split(schema, "|")
+	if len(schemas) == 1 {
+		return source.Schema, false
```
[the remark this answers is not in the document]
priyanshi-yb: It indicates whether there is a default schema i.e. public in source or not so we should not qualify any unqualified table/object with public instead error out if it is unqualified.

yb-voyager/src/tgtdb/postgres.go
priyanshi-yb: Filed https://yugabyte.atlassian.net/browse/DB-9381

yb-voyager/cmd/exportDataDebezium.go
priyanshi-yb: Not required anymore.

makalaaneesh (approved):
LGTM with the comments

PR #1261, opened 2023-12-27
Bug fix for cutover to target --prepare-for-fallback true  if user ran it accidentally and answer no in the prompt

priyanshi-yb:
1. Fixes the issue where user accidentally ran `initiate cutover to target --prepare-for-fallback t` and answer No in the prompt, in that case `msr.FallbackEnabled` was still getting set and hence get data-migration-report was erroring out.
3. added a hidden flag `--yes` for cutover prompt.

### DIFF
--- a/yb-voyager/cmd/cutoverToTargetCmd.go
+++ b/yb-voyager/cmd/cutoverToTargetCmd.go
@@ -53,12 +53,9 @@ var cutoverToTargetCmd = &cobra.Command{
 				utils.ErrExit("cannot prepare for fall-back. Fall-forward workflow already enabled.")
 			}
 		}
-		err = InitiateCutover("target")
+		err = InitiateCutover("target", bool(prepareForFallBack))
 		if err != nil {
 			utils.ErrExit("failed to initiate cutover: %v", err)
-		}
-		if prepareForFallBack {
-			updateFallBackEnabledInMetaDB()
 		}
 	},
 }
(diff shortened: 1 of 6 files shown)


yb-voyager/cmd/cutoverToTargetCmd.go
```diff
@@ -52,13 +52,14 @@ var cutoverToTargetCmd = &cobra.Command{
...
 				utils.ErrExit("cannot prepare for fall-back. Fall-forward workflow already enabled.")
 			}
-			updateFallBackEnabledInMetaDB()
 		}
 		err = InitiateCutover("target")
 		if err != nil {
 			utils.ErrExit("failed to initiate cutover: %v", err)
 		}
-
+		if prepareForFallBack {
```
makalaaneesh, line 60: I think the updation of record.CutoverToTargetRequested and record.FallBackEnabled should happen atomically. 

Theoretically, there is a possibility that marking record.CutoverToTargetRequested causes the exporter to shut down before record.FallBackEnabled is even set, since they are happening one after the other. (I think that's probably the reason I decided to mark the fallBackEnabled before). 
Practically, it's very unlikely, but I think it's better to still group these two operations together atomically.

priyanshi-yb: Yeah, makes sense. Made the change. thanks

makalaaneesh (approved):
LGTM!

PR #1265, opened 2024-01-03
Support sequences in pg_dump+dbzm live migration 

makalaaneesh:
- When starting dbzm for the cdc phase, extract sequence last values from pg_dump output and init dbzm with those values. This is needed to handle scenarios where there are no events corresponding to a table in the cdc phase (in which case debezium will not be able to track the max value of the corresponding sequences). 
- Support case sensitivity in sequence names

Caveat:
- sequences that are not owned by any column are not supported.

### DIFF
--- a/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/SequenceObjectUpdater.java
+++ b/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/SequenceObjectUpdater.java
@@ -37,6 +37,7 @@ public class SequenceObjectUpdater {
         es.setSequenceMaxMap(this.sequenceMax);
     }
 
+    // columnSequenceMapstring: public.t1.id:public."t1_id_seq",public.cst1.id:public."cSS1"
     public void initColumnSequenceMap(String columnSequenceMapString){
         if (columnSequenceMapString == null){
             return;
@@ -60,6 +61,8 @@ public class SequenceObjectUpdater {
             this.sequenceMax.put(sequenceName, this.sequenceMax.getOrDefault(sequenceName, (long)0));
         }
     }
+
+    // initSequenceMapString: public."cSS1":4,public."t1_id_seq":9
     public void initSequenceMax(String initSequenceMapString, ConcurrentMap<String, Long> sequenceMax){
         this.sequenceMax = new ConcurrentHashMap<>();
         if (initSequenceMapString != null){
--- a/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/SequenceObjectUpdater.java
+++ b/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/SequenceObjectUpdater.java
@@ -18,24 +18,28 @@ import java.util.concurrent.ConcurrentMap;
 public class SequenceObjectUpdater {
     private static final Logger LOGGER = LoggerFactory.getLogger(SequenceObjectUpdater.class);
     public static String propertyName = "column_sequence.map";
+    public static String initSequenceMaxpropertyName = "sequence.max.map";
     String dataDirStr;
     String sourceType;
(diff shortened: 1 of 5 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -264,6 +277,39 @@ func exportPGSnapshotWithPGdump(ctx context.Context, cancel context.CancelFunc,
...
+			if err != nil {
+				return nil, fmt.Errorf("parse %s to int in line - %s: %v", setValArgsArr[1], sequencesTextFileData[i], err)
+			}
+			seqNameRaw := setValArgsArr[0][1 : len(setValArgsArr[0])-1]
+			seqName := sqlname.NewSourceNameFromQualifiedName(seqNameRaw)
+			sequenceValueMap[seqName] = seqVal
+		}
+	}
+	return sequenceValueMap, nil
+}
```
amit-yb, line 311: Sometimes using smaller+idiomatic names can lead to more readable code. The name of the function itself provides some context for the names inside its scope. The context can be utilised to reduce the local name lengths.
```
func getPGDumpSequencesAndValues() (map[*sqlname.SourceName]int64, error) {
	result := map[*sqlname.SourceName]int64{}
	path := filepath.Join(exportDir, "data", "postdata.sql")
	bs, err := os.ReadFile(path)
	if err != nil {
		utils.ErrExit("read file %q: %v", path, err)
	}
	lines := strings.Split(string(bs), "\n")

    // Sample line: SELECT pg_catalog.setval('public.actor_actor_id_seq', 200, true);
	re := regexp.MustCompile(`(?i)SELECT.*setval\((?P<args>.*)\)`)

	for _, line := range lines {
		matches := re.FindStringSubmatch(line)
		if len(matches) == 0 {
			continue
		}
		idx := re.SubexpIndex("args")
		if idx > len(matches) {
			return nil, fmt.Errorf("invalid index %d for matches - %s for line %s", idx, matches, lines[i])
		}
		match := matches[idx]
		args := strings.Split(match, ",")

		seqNameRaw, _ := strconv.Unquote(args[0])
		seqName := sqlname.NewSourceNameFromQualifiedName(seqNameRaw)

		seqVal, err := strconv.ParseInt(strings.TrimSpace(args[1]), 10, 64)
		if err != nil {
			return nil, fmt.Errorf("parse %s to int in line - %s: %v", args[1], lines[i], err)
		}
		result[seqName] = seqVal
	}
	return result, nil
}
```

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/SequenceObjectUpdater.java
```diff
@@ -62,6 +60,27 @@ public void initColumnSequenceMap(String columnSequenceMapString){
             this.sequenceMax.put(sequenceName, this.sequenceMax.getOrDefault(sequenceName, (long)0));
         }
     }
+    public void initSequenceMax(String initSequenceMapString, ConcurrentMap<String, Long> sequenceMax){
+        this.sequenceMax = new ConcurrentHashMap<>();
```
amit-yb, line 64: A comment showing sample contents of the `initSequenceMapString` would really help here.

amit-yb (approved):
LGTM

yb-voyager/cmd/exportData.go
```diff
@@ -264,6 +277,39 @@ func exportPGSnapshotWithPGdump(ctx context.Context, cancel context.CancelFunc,
...
+	if err != nil {
+		utils.ErrExit("read file %q: %v", sequenceFilePath, err)
+	}
+
+	sequencesTextFileData := strings.Split(string(seqeuencesTextFileDataBytes), "\n")
+	numLines := len(sequencesTextFileData)
+	setvalRegex := regexp.MustCompile(`(?i)SELECT.*setval\((?P<args>.*)\)`)
+
+	for i := 0; i < numLines; i++ {
+		matches := setvalRegex.FindStringSubmatch(sequencesTextFileData[i])
```
sanyamsinghal, line 293: a comment explaining the regexp matching etc with an example would help.

PR #1267, opened 2024-01-04
Added value converters for JSON and ENUM types for PG source

priyanshi-yb:


yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -189,6 +189,20 @@ var YBValueConverterSuite = map[string]ConverterFn{
...
+			return columnValue, nil
+		}
+	},
+	"io.debezium.data.Enum": func(columnValue string, formatIfRequired bool) (string, error) {
+		if formatIfRequired {
+			return fmt.Sprintf("'%s'", columnValue), nil
+		} else {
+			return columnValue, nil
+		}
+	},
```
amit-yb, line 205: This function definition is repeating multiple times. How about defining a `func` at the top -- named `quoteForUpdateCommands()`? Then you can simply use the function name instead of repeating the function body.

yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -189,6 +189,20 @@ var YBValueConverterSuite = map[string]ConverterFn{
 			return columnValue, nil
 		}
 	},
+	"io.debezium.data.Json": func(columnValue string, formatIfRequired bool) (string, error) {
```
amit-yb, line 192: How about enhancing test automation to also have json and enum values?

priyanshi-yb: Its already there in PG tests and the issue caught in CDC but we don't have live migration datatypes automation for that yet.

amit-yb (approved):
LGTM

PR #1272, opened 2024-01-09
Fixed Bug: Progress bar hangs in import data when empty table is present in snapshot

priyanshi-yb:
1. fixes https://yugabyte.atlassian.net/browse/DB-9513, In PG live migration, where we use pg_dump for snapshot we don't skip passing empty tables in table-list to pg_dump and pg_dump gives a datafile for that, and hence when importing on target the progress bar for those empty tables hangs.  FIX: skipping those tables while creating `importFileTasks` 
2. fixed printing of table-list twice on UI for PG live migration

### DIFF
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -238,6 +238,8 @@ func exportData() bool {
 		}
 		return true
 	} else {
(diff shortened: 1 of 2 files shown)


amit-yb (approved):
LGTM

PR #1274, opened 2024-01-10
Fix dbzm bug: null in column with default

makalaaneesh:
- Use `getWithoutDefault` instead of `get` from KafkaConnectStruct. The default values will already be handled at the DB insert time, and all the WAL records will already have the default values populated.


makalaaneesh:
https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/1684/pipeline/

amit-yb (approved):
LGTM

PR #1278, opened 2024-01-16
UX changes in data-migration-report for large set of tables

priyanshi-yb:
fixes - 
1. Added pagination in the report printed in `get data-migration-report` with max 10 rows in each table - https://yugabyte.atlassian.net/browse/DB-9629
2. Added a print statement before printing the report so that the user can gets an idea that something is running -https://yugabyte.atlassian.net/browse/DB-9618

### DIFF
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -106,7 +106,7 @@ func getMappingForTableNameVsTableFileName(dataDirPath string, noWait bool) map[
 
 	pgRestorePath, err := srcdb.GetAbsPathOfPGCommand("pg_restore")
 	if err != nil {
-		utils.ErrExit("could not get absolute path of pg_restore command: %w", err)
+		utils.ErrExit("could not get absolute path of pg_restore command: %s", err)
 	}
 	pgRestoreCmd := exec.Command(pgRestorePath, "-l", dataDirPath)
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/common.go
```diff
@@ -106,7 +106,7 @@ func getMappingForTableNameVsTableFileName(dataDirPath string, noWait bool) map[
 
 	pgRestorePath, err := srcdb.GetAbsPathOfPGCommand("pg_restore")
 	if err != nil {
-		utils.ErrExit("could not get absolute path of pg_restore command: %v", pgRestorePath)
+		utils.ErrExit("could not get absolute path of pg_restore command: %w", err)
```
amit-yb, line 109:
```
func ErrExit(formatString string, args ...interface{}) {
	fmt.Fprintf(os.Stderr, formatString+"\n", args...)
	log.Errorf(formatString+"\n", args...)
	atexit.Exit(1)
}
```
The `Fprintf()` does not support `%w` format specifier. `%w` works only in the `fmt.Errorf()`. If you need to print error objects, go for `%s`.

amit-yb (approved):
LGTM with a minor comment.

yb-voyager/cmd/common.go
priyanshi-yb: Oh yes, thanks!

PR #1281, opened 2024-01-18
Updated the YB debezium connector version in installer script to incorporate decimal fix for precise mode

priyanshi-yb:
Updated the connector version and also added config for yb debezium to use precise mode for DECIMAL. 
Added an automation test for offline atleast so that we can remember to keep it in cases to consider while enhancing live migration test cases.


amit-yb (approved):
LGTM. Thanks for taking this to the completion 👍

PR #1291, opened 2024-01-24
Added ssl-mode to dbzm config

ShivanshGahlot:


yb-voyager/src/dbzm/config.go
```diff
@@ -304,7 +305,7 @@ func (c *Config) String() string {
 			triggerDirPath)
 		if c.SSLRootCert != "" {
```
makalaaneesh, line 306: 1. Don't think it'd be right to add sslmode only if the RootCert is passed. Find a way to pass in all the required config. (check the pg config code)
2. Let's throw an unsupported error if the user passes the key/cert in `export-data-from-target`

priyanshi-yb: > Let's throw an unsupported error if the user passes the key/cert in export-data-from-target
User can't pass these TargetConf related flags in the `export data from target`, It is same what is passed to import data case.

makalaaneesh: Ah yes! Good point!
In that case, we're essentially ignoring those key/cert params. We should take explicit confirmation from the user that it's okay to ignore these ssl-params OR at the very least warn the user that we are ignoring. 

Since these are security-related params, we need to be careful.

priyanshi-yb: yeah, makes sense.

ShivanshGahlot: 1. Added a warning for the user if ssl cert and key are passed saying that they will be ignored.
2. The sslmode will be set only when rootcert is passed. Otherwise supplying an empty root cert raises an error.
3. Also made the `--target-ssl-mode` flag available in export data from target since the mode '`prefer`' is not supported by `yb debezium`. So, if the user ends up giving that mode in import data, after cutover an error will be raised to the user saying that '`prefer`' mode is not supported for yb and they can start the command with a new `--target-ssl-mode` flag.

yb-voyager/cmd/exportDataFromTarget.go
```diff
@@ -86,7 +86,9 @@ func initSourceConfFromTargetConf() error {
 	} else {
 		source.Schema = targetConf.Schema
 	}
-	source.SSLMode = targetConf.SSLMode
+	if source.SSLMode == "" {
```
priyanshi-yb, line 89: Why do we need this check?

yb-voyager/src/dbzm/config.go
```diff
@@ -303,9 +305,15 @@ func (c *Config) String() string {
 			c.ExporterRole,
 			triggerDirPath)
 		if c.SSLRootCert != "" {
-			conf += fmt.Sprintf(yugabyteSSLConfigTemplate,
-				c.SSLRootCert)
-		} //TODO test SSL for other methods for yugabytedb
+			if c.SSLMode == "prefer" {
+				utils.ErrExit("Error: SSL mode 'prefer' is not supported for 'export data from target' from yugabytedb. Please restart 'export data from target' with a different --target-ssl-mode flag.")
```
priyanshi-yb, line 309:
```
Error: SSL mode 'prefer' is not supported for 'export data from target'. Please restart 'export data from target' with a different --target-ssl-mode flag.
```

priyanshi-yb (approved):
LGTM with few minor comments

yb-voyager/cmd/exportDataFromTarget.go
priyanshi-yb: Got it, My bad! it is required if a value is passed in a flag.

PR #1294, opened 2024-01-25
Bug Fix[DB-9759]: End Migration cmd if failed after saving schema analysis report, on re-running the cmd it prints no schema report as it is not executed as part of migration workflow

sanyamsinghal:
- moved the check for schema report if already backed up before the check of looking for generated report


amit-yb (approved):
LGTM

PR #1297, opened 2024-01-29
Enabling detection and handling for all 4 possible event conflicts

sanyamsinghal:
- Possible conflicts are: Delete-Insert, Delete-Update, Update-Insert, Update-Update

### DIFF
--- a/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
+++ b/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
@@ -277,9 +277,6 @@ class KafkaConnectRecordParser implements RecordParser {
                 if (valueAndSet == null){
                     continue;
                 }
-            } else if (r.op.equals("d") && valueAndSet == null){
-                // in case of deletes we are using before struct which contains only delta for yb
-                continue;
             }
             
             if (!valueAndSet.getBoolean("set")){
@@ -307,12 +304,12 @@ class KafkaConnectRecordParser implements RecordParser {
             }
             Object afterFieldValue = after.getWithoutDefault(f.name());
             Object beforeFieldValue = null;
+
+            r.addAfterValueField(f.name(), afterFieldValue);
             if (!(r.op.equals("c") || r.op.equals("r"))){ 
                 // before is null for create events
-                // before can also be null if REPLICA IDENTITY is not set as 'FULL' (maybe in case of snapshot-only mode)
                 beforeFieldValue = before.getWithoutDefault(f.name());
             }
-            r.addAfterValueField(f.name(), afterFieldValue);
             r.addBeforeValueField(f.name(), beforeFieldValue);
         }
     }
--- a/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
+++ b/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
@@ -237,7 +237,7 @@ class KafkaConnectRecordParser implements RecordParser {
             } else {
                 fieldValue = key.getWithoutDefault(f.name());
             }
-            r.addAfterKeyField(f.name(), fieldValue);
+            r.addKeyField(f.name(), fieldValue);
 
         }
     }
--- a/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
+++ b/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
@@ -251,19 +251,71 @@ class KafkaConnectRecordParser implements RecordParser {
      */
     protected void parseValueFields(Struct value, Record r) {
         Struct after = value.getStruct("after");
-        // TODO: error handle before is NULL
         Struct before = value.getStruct("before");
+<<<<<<< HEAD
 
         // in case of delete, the before struct contains the values required in cases
         // like DELETE-INSERT conflicts(unique key columns)
         if (r.op.equals("d") && before != null) {
             after = before;
+=======
+        
+        if (sourceType.equals("yb")) {
+            parseValueFieldsForYB(after, r);
+        } else {
+            parseValueFieldsForOthers(after, before, r);
+>>>>>>> 4af14ba (code cleanup and error handling)
         }
+    }
+
+    private void parseValueFieldsForYB(Struct after, Record r) {
         if (after == null) {
             return;
         }
 
         for (Field f : after.schema().fields()) {
+            // TODO: write a proper transformer for this logic
+            // values in the debezium connector are as follows:
+            // "val1" : {
+            //  "value" : "value for val1 column",
+            //  "set" : true
+            //}
+            Struct valueAndSet = after.getStruct(f.name());
+            if (r.op.equals("u")) {
+                // in the default configuration of the stream, for an update, the fields in the after struct
+                // are only the delta fields, therefore, it is possible for a field to not be there.
+                if (valueAndSet == null){
+                    continue;
+                }
+            } else if (r.op.equals("d") && valueAndSet == null){
+                // in case of deletes we are using before struct which contains only delta for yb
+                continue;
+            }
+            
+            if (!valueAndSet.getBoolean("set")){
+                continue;
+            }
+            Object afterFieldValue = valueAndSet.getWithoutDefault("value");
+            r.addAfterValueField(f.name(), afterFieldValue);
+        }
+    }
+
+    private void parseValueFieldsForOthers(Struct after, Struct before, Record r) {
+        if (r.op.equals("d")){ // after is null for delete events
+            for (Field f : before.schema().fields()) {
+                Object beforeFieldValue = before.getWithoutDefault(f.name());
+                r.addBeforeValueField(f.name(), beforeFieldValue);
+            }
+            return;
+        }
+
+        for (Field f : after.schema().fields()) {
+            if (r.op.equals("u")) {
+                if (Objects.equals(after.get(f), before.get(f))) {
+                    // no need to record this as field is unchanged
+                    continue;
+                }
+            }
             Object beforeFieldValue = null;
             Object afterFieldValue;
             if (sourceType.equals("yb")){
--- a/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
+++ b/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
@@ -114,7 +114,6 @@ class KafkaConnectRecordParser implements RecordParser {
                 parseKeyFields(key, r);
             }
             parseValueFields(value, r);
-
             return r;
         } catch (
 
@@ -252,7 +251,6 @@ class KafkaConnectRecordParser implements RecordParser {
     protected void parseValueFields(Struct value, Record r) {
         Struct after = value.getStruct("after");
         Struct before = value.getStruct("before");
-        
         if (sourceType.equals("yb")) {
             parseValueFieldsForYB(after, r);
         } else {
@@ -300,7 +298,6 @@ class KafkaConnectRecordParser implements RecordParser {
             }
             return;
         }
-
         for (Field f : after.schema().fields()) {
             if (r.op.equals("u")) {
                 if (Objects.equals(after.get(f), before.get(f))) {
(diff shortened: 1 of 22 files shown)


sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/1792/

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
```diff
@@ -189,54 +189,72 @@ protected void parseKeyFields(Struct key, Record r) {
...
-                if (r.op.equals("u")) {
-                    if (Objects.equals(after.get(f), before.get(f))) {
-                        // no need to record this as field is unchanged
-                        continue;
-                    }
-                }
-                fieldValue = after.getWithoutDefault(f.name());
+            Object afterFieldValue = after.getWithoutDefault(f.name());
+            Object beforeFieldValue = null;
+            if (!r.op.equals("c") && before != null) {
```
makalaaneesh, line 251: How about just check `if before != null` here, and then validate above before line 242 that for update events, before struct should be non-null?

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
```diff
@@ -189,54 +189,72 @@ protected void parseKeyFields(Struct key, Record r) {
...
+            Struct valueAndSet = after.getStruct(f.name());
+            if (r.op.equals("u")) {
+                // in the default configuration of the stream, for an update, the fields in the after struct
+                // are only the delta fields, therefore, it is possible for a field to not be there.
+                if (valueAndSet == null){
                     continue;
                 }
-                
-                if (!valueAndSet.getBoolean("set")){
+            } else if (r.op.equals("d") && valueAndSet == null){
```
makalaaneesh, line 220: As of now, since we are not enabling before image, maybe we can just remove this block? 
Anyways, at the beginning of this func, if after == null, we return

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
```diff
@@ -189,54 +189,72 @@ protected void parseKeyFields(Struct key, Record r) {
...
         Struct after = value.getStruct("after");
-        // TODO: error handle before is NULL
         Struct before = value.getStruct("before");
         
-        // in case of delete, the before struct contains the values required in cases like DELETE-INSERT conflicts(unique key columns)
-        if (r.op.equals("d") && before != null) {
-            after = before;
+        if (sourceType.equals("yb")) {
+            parseValueFieldsForYB(after, r);
+        } else {
```
makalaaneesh, line 196: Thanks for segregating this. It was getting quite confusing :)

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/Record.java
```diff
@@ -12,20 +12,29 @@ public class Record {
...
-    public ArrayList<Object> keyValues = new ArrayList<>();
-    public ArrayList<String> valueColumns = new ArrayList<>();
-    public ArrayList<Object> valueValues = new ArrayList<>();
+
+     // Value information for 'before' struct
+    public ArrayList<String> beforeValueColumns = new ArrayList<>();
+    public ArrayList<Object> beforeValueValues = new ArrayList<>();
+
+    // Key and value information for 'after' struct
+    public ArrayList<String> afterKeyColumns = new ArrayList<>();
```
makalaaneesh, line 21: There is only a single key, so you can continue calling  this `keyColumns` and `keyValues`

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -107,10 +159,12 @@ func (c *ConflictDetectionCache) eventsConfict(event1, event2 *tgtdb.Event) bool
 	}
 	uniqueKeyColumns := c.tableToUniqueKeyColumns[maybeQualifiedName]
 	for _, column := range uniqueKeyColumns {
-		// if the unique key column value is same, then we have a conflict
-		if *event1.Fields[column] == *event2.Fields[column] {
+		if event1.BeforeFields[column] == nil || event2.Fields[column] == nil {
```
makalaaneesh, line 162: event1 and event2 names are a bit confusing. 
How about `cachedEvent` (which you already use), and `incomingEvent`?

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -107,10 +159,12 @@ func (c *ConflictDetectionCache) eventsConfict(event1, event2 *tgtdb.Event) bool
...
+			return false
+		}
+		if utils.CompareWithoutQuotes(*event1.BeforeFields[column], *event2.Fields[column]) {
```
makalaaneesh, line 165: This is much more clear now. you will be comparing the "before" value of the cachedEvent to the "after" value of the incomingEvent to see if they match 👍

yb-voyager/src/utils/utils.go
```diff
@@ -487,3 +487,9 @@ func ReadTableNameListFromFile(filePath string) ([]string, error) {
 	}
 	return list, nil
 }
+
+func CompareWithoutQuotes(s1, s2 string) bool {
+	s1 = strings.Trim(s1, "\"'")
+	s2 = strings.Trim(s2, "\"'")
+	return strings.EqualFold(s1, s2)
```
makalaaneesh, line 494: Looks like `EqualFold` does some sort of case-insensitive matching. Why do we need this? 
And why do we even need to remove quotes? 
Why won't a simple string match work here?

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
sanyamsinghal: yeah, some leftover code. will remove it.

yb-voyager/cmd/conflictDetectionCache.go
sanyamsinghal: yeah good idea

yb-voyager/src/utils/utils.go
sanyamsinghal: We need to remove quotes because cached event can be quotes due to that `formatIfRequired` check and incoming event is not, and vice-versa.

sanyamsinghal: EqualFold() should not be there, two strings with same value but different case-sensitivity will be different. 
Thanks, good catch! Stupid Github copilot :)

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
sanyamsinghal: I changed that to:
```
if (!(r.op.equals("c") || r.op.equals("r"))){ 
                // before is null for create events
                // before can also be null if REPLICA IDENTITY is not set as 'FULL' (maybe in case of snapshot-only mode)
                beforeFieldValue = before.getWithoutDefault(f.name());
}
```

We can have a simple check of `before != null` but i thought this will be more meaningful and easier to understand in code.

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -96,21 +148,23 @@ func (c *ConflictDetectionCache) eventsConfict(event1, event2 *tgtdb.Event) bool
 
 		For now, we just check if the event is from same table then we consider it as a conflict
 	*/
-	if isTargetDBExporter(event2.ExporterRole) {
-		log.Infof("conflict detected for table %s, between event1(vsn=%d) and event2(vsn=%d)", event1.TableName, event1.Vsn, event2.Vsn)
+	if isTargetDBExporter(incomingEvent.ExporterRole) {
+		log.Infof("conflict detected for table %s, between event1(vsn=%d) and event2(vsn=%d)", cachedEvent.TableName, cachedEvent.Vsn, incomingEvent.Vsn)
 		return true
 	}
```
amit-yb, line 154: For events exported from target stall the pipeline...
- if the cachedEvent is DELETE; or
- if the cachedEvent is UPDATE and both the cached and incoming events are modifying same unique column (without actually checking their values).

amit-yb (commented):
couple of minor comments.

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/Record.java
sanyamsinghal: got it.

yb-voyager/src/tgtdb/event.go
```diff
@@ -269,6 +270,19 @@ func (event *Event) getTableName(targetSchema string) string {
...
+	if event.Op != "u" {
+		return false
+	}
+
+	for _, col := range uniqueKeyCols {
+		if _, ok := event.Fields[col]; ok {
+			return true
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
sanyamsinghal: actually it will be:
```
event.Op == "u" && len(uniqueKeyCols) > 0 && lo.Some(lo.Keys(event.Fields), uniqueKeyCols)
``

File: yb-voyager/src/utils/utils.go (hunk 7 above)
Comment (makalaaneesh) [reply yb-voyager/src/utils/utils.go]: Regarding, quotes, it could potentially lead to some false positive conflicts. Let's say val1 = `"abc"` and val2= `abc` . They are not equal but will be deemed equal. 

Nevertheless, I think this is too rare a case to be bothered with, especially because functionally nothing will break. 
Pls just add a comment indicating why we need to match after removing quotes, and the possibility of false-positives that we are ignoring!

File: yb-voyager/src/tgtdb/event.go (hunk 9 above)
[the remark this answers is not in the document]
Comment (sanyamsinghal) [reply yb-voyager/src/tgtdb/event.go]: done

File: yb-voyager/cmd/conflictDetectionCache.go (hunk 8 above)
Comment (amit-yb) [reply yb-voyager/cmd/conflictDetectionCache.go]: This is not totally correct: assume if there are two uniq keys UK1 and UK2; and if `cachedEvent` is modifying `UK1` and `incomingEvent` is modifying `UK2`, the code incorrectly assumes a conflict. The correct condition should be:
```
ukList := lo.Intersect(cachedEventCols, uniqueKeyColumns)
if lo.Some(incomingEventCols, ukList) {
  conflict = true
}
```

File: yb-voyager/src/utils/utils.go (hunk 7 above)
Comment (sanyamsinghal) [reply yb-voyager/src/utils/utils.go]: @makalaaneesh i actually added a definite check. There won't be any false positives. 
Please take a look at recent commits.

File: yb-voyager/cmd/conflictDetectionCache.go [hunk 10]
```diff
@@ -85,34 +138,69 @@ func (c *ConflictDetectionCache) RemoveEvents(batch *tgtdb.EventBatch) {
...
+			if lo.Some(incomingEventCols, ukList) {
+				conflict = true
+			}
+		}
 
-	maybeQualifiedName := event1.TableName
-	if event1.SchemaName != "public" {
-		maybeQualifiedName = fmt.Sprintf("%s.%s", event1.SchemaName, event1.TableName)
+		if conflict {
+			log.Infof("conflict detected for table %s, between event1(vsn=%d) and event2(vsn=%d)", cachedEvent.TableName, cachedEvent.Vsn, incomingEvent.Vsn)
```
amit-yb, line 173: There is a similar log message in the caller as well.

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -85,34 +138,69 @@ func (c *ConflictDetectionCache) RemoveEvents(batch *tgtdb.EventBatch) {
...
 			log.Infof("conflict detected for table %s, column %s, between value of event1(vsn=%d, colVal=%s) and event2(vsn=%d, colVal=%s)",
-				maybeQualifiedName, column, event1.Vsn, *event1.Fields[column], event2.Vsn, *event2.Fields[column])
+				maybeQualifiedName, column, cachedEvent.Vsn, *cachedEvent.BeforeFields[column], incomingEvent.Vsn, *incomingEvent.Fields[column])
 			return true
 		}
 	}
 	return false
 }
+
+func CompareConflictingEventValues(cachedEvent, incomingEvent *tgtdb.Event, column string) bool {
```
amit-yb, line 191: Would you need all this complexity if you move the call to `WaitUntilNoConflict()` before the call to `ConvertEvent()`?

sanyamsinghal: hmm possible, let me verify that

makalaaneesh (approved):

yb-voyager/cmd/conflictDetectionCache.go
sanyamsinghal: yeah, one in the caller just tells that we are waiting for the conflict to resolve but the ones in this function also tells the column on which the values are conflicting and the conflicting value.
I think we can keep both for now

PR #1298, opened 2024-01-30
DB-9825 Bug Fix: Conflicting events export from Target(YB): Yb-cdc doesn't export  the events in same sequence in which they were executed on YB

sanyamsinghal:
- setting transaction.order=true in the config file which orders the events based on commit time

Introduced a new flag unique-key-columns-exists in command
- flag is true by default which will enable transaction ordering for the events export from target so that conflicts events are correctly order in the eventQueue


amit-yb (approved):
LGTM. Thanks for finding and fixing this issue! 👍

sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/1853/

yb-voyager/cmd/exportDataFromTarget.go
```diff
@@ -55,6 +58,12 @@ func init() {
 	registerTargetDBAsSourceConnFlags(exportDataFromTargetCmd)
 	registerExportDataFlags(exportDataFromTargetCmd)
 	hideExportFlagsInFallForwardOrBackCmds(exportDataFromTargetCmd)
+
+	BoolVar(exportDataFromTargetCmd.Flags(), &uniqueKeyColumnsExists, "unique-key-columns-exists", true,
```
amit-yb, line 62: Let's rename the flag as `--transaction-ordering`. Even in the presence of uniq keys, a user may decide to go full throttle (i.e. without transaction ordering) if their app never modifies UKs or modifies in such a way that it never causes conflict.
Presence of unique keys in not the root problem, it is the way in which they are modified (if at all).

yb-voyager/cmd/exportDataFromTarget.go
```diff
@@ -55,6 +58,12 @@ func init() {
...
+		"Set to true if the target DB has unique key columns."+
+			"Note: This will impact the performance of events streaming from target DB because of transaction ordering enabled(mandatory to resolve conflicting events)\n"+
+			"Set to false if the unique key columns does not exists in the target tables for higher throughput.\n"+
+			"To learn more about conflict resolution by voyager, Refer: https://github.com/yugabyte/yb-voyager/blob/main/yb-voyager/cmd/conflictDetectionCache.go\n")
```
amit-yb, line 66:
```
Setting the flag to `false` disables the transaction ordering. This speeds up change data capture from target yugabytedb. Disable transaction ordering only if the tables under migration do not have unique keys or the app does not modify/reuse the unique keys.
```

amit-yb (approved):
LGTM with couple of important suggestions.

PR #1299, opened 2024-01-31
create publication manually before replication slot  to avoid event loss

makalaaneesh:
As per pg semantics, we are supposed to create publication object before replication slot. Doing it the other way around was causing unexplained loss of events from debezium. 

Ref: 	 https://www.postgresql.org/message-id/flat/e0885261-5723-7bab-f541-e6a260f50328%402ndquadrant.com#a5f257b667575719ad98c59281f3e191


amit-yb (approved):
LGTM

PR #1302, opened 2024-01-31
DB-9827 synchronize flushing of snapshot writers in flusher thread with closing snapshot files

makalaaneesh:
WIthout synchronization, there is a possibility of the flusher thread trying to flush an already closed file stream.


amit-yb (approved):
LGTM

PR #1303, opened 2024-02-02
Report case-sensitive and partition tables as not supported in PG live migration

priyanshi-yb:


makalaaneesh:
@priyanshi-yb pls add one more case:
we don't honor  "target-db-schema" in import-data if migrating from PG, so if arg "target-db-schema" is changed in import-data, and sourceDBType is PG, throw an error.

priyanshi-yb:
> @priyanshi-yb pls add one more case: we don't honor "target-db-schema" in import-data if migrating from PG, so if arg "target-db-schema" is changed in import-data, and sourceDBType is PG, throw an error.

The code is already there but was not getting executing because of a bug, I have fixed It in the multiple schema PR - https://github.com/yugabyte/yb-voyager/pull/1254/files

yb-voyager/cmd/exportData.go
```diff
@@ -348,6 +347,36 @@ func getPGDumpSequencesAndValues() (map[*sqlname.SourceName]int64, error) {
...
+	for _, table := range finalTableList {
+		if table.ObjectName.MinQuoted != table.ObjectName.Unquoted {
+			caseSensitiveTables = append(caseSensitiveTables, table.Qualified.MinQuoted)
+		} 
+		if source.DB().ParentTableOfPartition(table) == "" { //For root tables only
+			if len(source.DB().GetPartitions(table)) > 0 {
+				partitionedTables = append(partitionedTables, table.Qualified.MinQuoted)
+			}
+		} else {	
+			if source.TableList != "" { //If user has passed table list with leaf partition
```
amit-yb, line 363: Do we really need this check? As long as it is a partition it is not supported.

yb-voyager/cmd/exportData.go
```diff
@@ -348,6 +347,36 @@ func getPGDumpSequencesAndValues() (map[*sqlname.SourceName]int64, error) {
...
+	}
+	if len(caseSensitiveTables) > 0 {
+		if len(partitionedTables) > 0 {
+			utils.PrintAndLog("The following partitioned tables are not supported for live migration: %v", partitionedTables)
+		}
+		utils.ErrExit("The following case sensitive tables are not supported for live migration: %v", caseSensitiveTables)
+	} 
+	if len(partitionedTables) > 0 {
+		utils.ErrExit("The following partitioned tables are not supported for live migration: %v", partitionedTables)
+	}
```
amit-yb, line 376:
```
if len(caseSensitiveTables) == 0 && len(partitionedTables) == 0 {
  return
}
utils.PrintAndLog("This voyager release does not support live-migration with case sensitive or partitioned tables. You can exclude the following tables using the --exclude-table-list argument:")
utils.PrintAndLog("Case sensitive table names: %s", caseSensitiveTables)
utils.PrintAndLog("Partition/Partitioned tables names: %s", partitionedTables)
```

amit-yb (commented):
Couple of minor comments.

yb-voyager/cmd/exportData.go
priyanshi-yb: This was to make the partitions list efficient for users where if table-list is not provided then when we fetch the table list from the source, it also contains child partitions as well along with root partitions, in that case, we need not report about the child partitions only root table name is sufficient to be excluded in the list.
But now I realize that this will be not sufficient if a user is using only the exclude-table list with the root table, we will still have the table list from the source (which will eventually have the child partitions). 
Thanks, will fix it!

amit-yb (approved):
LGTM 👍

PR #1306, opened 2024-02-02
DB-9858 Wait for 2 min for yb cdc initialization to complete. 

makalaaneesh:
1. Add a sleep of 2 min to wait for 2 min for yb cdc initialization to complete. 
2. Only then, set msr fields to indicate that export-data-from-target has started. As a result cutover status remains "INITIATED" as long as we are waiting for yb cdc to initialize.


amit-yb (approved):
LGTM

PR #1311, opened 2024-02-06
Fix: qualify public table names for target PG by default schema in multiple schemas case

priyanshi-yb:


yb-voyager/src/tgtdb/postgres.go
```diff
@@ -285,9 +285,21 @@ outer:
...
+		return schema, false
+	} else if slices.Contains(schemas, "public") {
+		return "public", false
+	} else {
+		return "", true
+	}
+}
+
 func (pg *TargetPostgreSQL) qualifyTableName(tableName string) string {
+	defaultSchema, _ := getDefaultPGSchema(pg.tconf.Schema)
```
amit-yb, line 300: What about cases where there isn't a clear "default" schema?

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -285,9 +285,21 @@ outer:
 	return nil
 }
 
+func getDefaultPGSchema(schema string) (string, bool) {
```
amit-yb, line 288: Add a comment stating the meaning of the `bool` return value?

amit-yb (approved):
LGTM

yb-voyager/src/tgtdb/postgres.go
priyanshi-yb: hmm, handled it now.

PR #1313, opened 2024-02-06
DB-9772 Remove "export data failed" msg when process shutdown requested by end migration

makalaaneesh:
export-data does not report failures if `ProcessShutdownRequested==true` which is the case also when end-migration command is run.


amit-yb (approved):
LGTM

PR #1319, opened 2024-02-07
Add recommendation to run end-migration to prevent replication slot from filling up disk space

makalaaneesh:
Note: cutover also removes the slot, but it's simpler to just ask the user to run end-migration (which also removes the slot) as it also covers the case where the user aborts the migration mid-way.

### DIFF
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -30,11 +30,11 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/fatih/color"
-	"golang.org/x/exp/slices"
 	"github.com/samber/lo"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/tebeka/atexit"
+	"golang.org/x/exp/slices"
 
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/callhome"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/cp"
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -293,6 +293,9 @@ func exportPGSnapshotWithPGdump(ctx context.Context, cancel context.CancelFunc,
 	if err != nil {
 		return fmt.Errorf("export snapshot: failed to create replication slot: %v", err)
 	}
+	utils.PrintAndLog(color.YellowString(fmt.Sprintf("Created replication slot '%s' on source PG database. "+
```
amit-yb, line 296: Can we make it bold and blink?
```
// Base attributes
const (
	Reset Attribute = iota
	Bold       <----------------
	Faint
	Italic
	Underline
	BlinkSlow   <-----------
	BlinkRapid
	ReverseVideo
	Concealed
	CrossedOut
)
```

amit-yb (approved):
LGTM with a suggestion.

yb-voyager/cmd/exportData.go
```diff
@@ -293,6 +293,9 @@ func exportPGSnapshotWithPGdump(ctx context.Context, cancel context.CancelFunc,
...
+		"Be sure to run 'end migration' command after completing/aborting this migration to drop the replication slot. "+
```
priyanshi-yb, line 297: What about the `cutover to target` in the msg?

priyanshi-yb: Oh just saw the note in description. 👍

priyanshi-yb: @makalaaneesh,  this might give the user the impression that the slot is open till the end migration is done if we don't tell them that this is also removed in the cutover phase.

makalaaneesh: @priyanshi-yb True! I just thought it would be simpler this way, instead of getting into more details. But I suppose we can just say "Be sure to run cutover to target/end migration command after completing/aborting this migration ...."

yb-voyager/cmd/exportData.go
makalaaneesh: Added bold. 
For some reason, blinkSlow/blinkRapid wasn't working.

PR #1321, opened 2024-02-07
[#20974] Query to push UpdateRowCountEvent to yugabyted db fails

gargsans-yb:
Summary:
* Added a condition to check for empty list of UpdateRowCountEvents before sending them.


amit-yb (approved):
LGTM

PR #1322, opened 2024-02-07
Made fixes related to sslmode in export data from target

ShivanshGahlot:


yb-voyager/cmd/importData.go
```diff
@@ -160,11 +160,17 @@ func startExportDataFromTargetIfRequired() {
...
+	if tconf.SSLMode == "prefer" || tconf.SSLMode == "allow" {
+		utils.PrintAndLog(color.RedString("Warning: SSL mode '%s' is not supported for 'export data from target' yet. Downgrading it to 'disable'.\nIf you dont want these settings you can restart the 'export data from target' with the --target-ssl-mode flag.", source.SSLMode))
+		tconf.SSLMode = "disable"
+	}
 	cmd := []string{"yb-voyager", "export", "data", "from", "target",
 		"--export-dir", exportDir,
 		"--table-list", strings.Join(importTableList, ","),
 		fmt.Sprintf("--transaction-ordering=%t", transactionOrdering),
 		fmt.Sprintf("--send-diagnostics=%t", callhome.SendDiagnostics),
+		"--target-ssl-mode", tconf.SSLMode,
```
makalaaneesh, line 173: Should we also add root-cert arg to export-data-from-target? 
What happens if 
- someone uses "prefer" in import-data-to-target, with no value for target-root-cert flag
- Now, after cutover, export-data-from-target starts with sslmode=disable. 
- User is not happy with this behaviour, and wants to change sslmode to "require". But now, that will not work because there is no way to specify root-cert and root-cert is mandatory for yb cdc connector

yb-voyager/cmd/importData.go
```diff
@@ -160,11 +160,17 @@ func startExportDataFromTargetIfRequired() {
 	importTableList := getImportTableList(tableListExportedFromSource)
 
 	lockFile.Unlock() // unlock export dir from import data cmd before switching current process to ff/fb sync cmd
+
+	if tconf.SSLMode == "prefer" || tconf.SSLMode == "allow" {
+		utils.PrintAndLog(color.RedString("Warning: SSL mode '%s' is not supported for 'export data from target' yet. Downgrading it to 'disable'.\nIf you dont want these settings you can restart the 'export data from target' with the --target-ssl-mode flag.", source.SSLMode))
```
makalaaneesh, line 165: If you dont want these settings you can restart the 'export data from target' with a different value for --target-ssl-mode flag.

yb-voyager/src/dbzm/config.go
```diff
@@ -310,16 +310,9 @@ func (c *Config) String() string {
 			c.RunId,
 			c.ExporterRole)
 		if c.SSLRootCert != "" {
```
makalaaneesh, line 312: Let's not use this check. 
In all conditions, we should use the user-provided ssl mode and set it in dbzm config. In the current logic, we would only set it, if root-cert was not nil. This is a bit dangerous although it might not cause any issues at the moment. 

Instead, we can add a validation (in the exportDataFromTarget init) that if ssl mode = [require, verify, verify-ca, verify-full], ssl-root-cert should also be provided. 

Here, while preparing the dbzm config, 
- blindly set the ssl mode
- if rootcert is not nil, set rootcert

ShivanshGahlot: Done

yb-voyager/cmd/exportDataFromTarget.go
```diff
@@ -80,12 +95,14 @@ func initSourceConfFromTargetConf() error {
 	} else {
 		source.Schema = targetConf.Schema
 	}
-	if source.SSLMode == "" {
-		source.SSLMode = targetConf.SSLMode
+	if targetConf.SSLCertPath != "" || targetConf.SSLKey != "" {
```
makalaaneesh, line 98: if sslmode is "disable", we shouldn't even invoke this prompt

yb-voyager/cmd/exportDataFromTarget.go
```diff
@@ -63,6 +67,17 @@ func init() {
 		"Setting the flag to `false` disables the transaction ordering. This speeds up change data capture from target YugabyteDB. Disable transaction ordering only if the tables under migration do not have unique keys or the app does not modify/reuse the unique keys.")
 }
 
+func verifySSLFlags() error {
+	if source.SSLMode != "disable" && source.SSLMode != "require" && source.SSLMode != "verify-ca" && source.SSLMode != "verify-full" {
+		return fmt.Errorf("invalid SSL mode '%s' for 'export data from target'. Please restart 'export data from target' with the --target-ssl-mode flag with one of these modes: 'disable', 'require', 'verify-ca', 'verify-full'", source.SSLMode)
+
+	}
+	if (source.SSLMode == "require" || source.SSLMode == "verify-ca" || source.SSLMode == "verify-full") && source.SSLRootCert == "" {
```
makalaaneesh, line 75: Again, you can use `lo.Contains`to make the code more readable

yb-voyager/cmd/importData.go
```diff
@@ -160,11 +160,18 @@ func startExportDataFromTargetIfRequired() {
...
+		utils.PrintAndLog(color.RedString("Warning: SSL mode '%s' is not supported for 'export data from target' yet. Downgrading it to 'disable'.\nIf you dont want these settings you can restart the 'export data from target' with a different value for --target-ssl-mode and --target-ssl-root-cert flag.", source.SSLMode))
+		tconf.SSLMode = "disable"
+	}
 	cmd := []string{"yb-voyager", "export", "data", "from", "target",
 		"--export-dir", exportDir,
 		"--table-list", strings.Join(importTableList, ","),
 		fmt.Sprintf("--transaction-ordering=%t", transactionOrdering),
 		fmt.Sprintf("--send-diagnostics=%t", callhome.SendDiagnostics),
+		"--target-ssl-mode", tconf.SSLMode,
+		"--target-ssl-root-cert", tconf.SSLRootCert,
```
makalaaneesh, line 174: Only add this param if it's non-empty

yb-voyager/cmd/exportDataFromTarget.go
```diff
@@ -63,6 +67,17 @@ func init() {
 		"Setting the flag to `false` disables the transaction ordering. This speeds up change data capture from target YugabyteDB. Disable transaction ordering only if the tables under migration do not have unique keys or the app does not modify/reuse the unique keys.")
 }
 
+func verifySSLFlags() error {
+	if source.SSLMode != "disable" && source.SSLMode != "require" && source.SSLMode != "verify-ca" && source.SSLMode != "verify-full" {
```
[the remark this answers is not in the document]
ShivanshGahlot: Done

yb-voyager/cmd/exportDataFromTarget.go
ShivanshGahlot: Done

yb-voyager/cmd/exportDataFromTarget.go
ShivanshGahlot: Done

yb-voyager/cmd/importData.go
ShivanshGahlot: Done

makalaaneesh (approved):
LGTM!

PR #1323, opened 2024-02-07
Bug fix: Bit(n) or Bit varying (n) datatype Converter Function of target YB to handle bit values of length less than n

priyanshi-yb:
Issue - 
```
create table x(id serial primary key, b BIT(10));
insert into x(b) values (B'0101010101');
```
This value `0101010101` on the source will error out in import with error `ERROR: bit string length 9 does not match type bit(10) (SQLSTATE 22026)` as the value converter function was converting is - `101010101` instead of `0101010101` so need to add the zeros on left for the extra length.  

Fix - This Patch fixes it by using the length of that column present in schema dumped by debezium. So now passing debezium schema to the converter function.
1. Changed Converter Function to accept debezium schema in parameter 
3. Moved SchemaRegistry from dbzm package to utils package to resolve circular dependency

Jenkins - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/1903/


yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -112,31 +116,42 @@ var YBValueConverterSuite = map[string]ConverterFn{
...
+			bitVal = fmt.Sprintf("%0*s", bitLen, bitVal) // Left padding with zeros
+		}
+
 		if formatIfRequired {
-			return fmt.Sprintf("'%b'", data), nil
+			return fmt.Sprintf("'%s'", bitVal), nil
 		} else {
-			return fmt.Sprintf("%b", data), nil
+			return bitVal, nil
 		}
```
amit-yb, line 134: Simplify:
```
quote := lo.Ternary(formatIfRequired, "'", "") 
strLength, ok := dbzmSchema.Parameters["length"]
length := lo.Ternary(ok, lo.Must(strings.Atoi(strLength)), BIT_VARYING_MAX_LEN)
return fmt.Sprintf("%s%0*b%s", quote, length, data, quote)
```

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -202,20 +208,20 @@ func (conv *DebeziumValueConverter) convertMap(eventSchema string, tableName str
...
+			colType, colDbzmSchema, err = conv.schemaRegistrySource.GetColumnType(strings.ToUpper(tableName), strings.ToUpper(column), conv.shouldFormatAsPerSourceDatatypes())
 			//assuming table name/column name is case insensitive TODO: handle this case sensitivity properly
 			if err != nil {
 				return fmt.Errorf("fetch column schema: %w", err)
 			}
 		}
 		converterFn := conv.valueConverterSuite[colType]
 		if converterFn != nil {
-			columnValue, err = converterFn(columnValue, formatIfRequired)
+			columnValue, err = converterFn(columnValue, formatIfRequired, colDbzmSchema) // Fix: Pass the value of colDbzmSchema instead of the pointer
```
amit-yb, line 224: Why the `Fix:` comment? We should be passing pointer.

amit-yb (approved):
LGTM with a few minor comments. Good work! 👍

yb-voyager/src/dbzm/valueConverter.go
priyanshi-yb: Yes, nothing needs to be done here, not sure how this comment is here probably github copilot. Will remove it.

yb-voyager/src/tgtdb/suites/oracleSuite.go
```diff
@@ -23,10 +23,12 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
 )
 
 var OraValueConverterSuite = map[string]ConverterFn{
-	"DATE": func(columnValue string, formatIfRequired bool) (string, error) {
+	"DATE": func(columnValue string, formatIfRequired bool, dbzmSchema *utils.ColumnSchema) (string, error) {
```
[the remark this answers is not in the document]
priyanshi-yb: Yeah, at other places I have kept it like this only but in some functions we are using another one inside `quoteIfRequired()` so need to pass this to that as well so kept for those ones.

PR #1325, opened 2024-02-08
User facing  message change for erroring out when different voyager versions are used on an export-dir

priyanshi-yb:


amit-yb (approved):
LGTM

PR #1344, opened 2024-02-29
Removed the check for fall back from PG live migration

shubham-yb:
Removed the check from cutover to target which denies switching to the fall back workflow in PG live migration.


makalaaneesh (approved):
lgtm

PR #1346, opened 2024-03-02
Implement NameRegistry

amit-yb:


yb-voyager/src/namereg/namereg_test.go
```diff
@@ -0,0 +1,636 @@
...
+		tableNames []string
+		expected   *NameTuple
+	}{
+		{[]string{
+			// YB side variants:
+			`table1`, `"table1"`, `public.table1`, `public."table1"`, `public."TABLE1"`, `public.TABLE1`,
+			// Oracle side variants:
+			`TABLE1`, `"TABLE1"`, `SAKILA.TABLE1`, `SAKILA."TABLE1"`, `SAKILA."table1"`, `SAKILA.table1`,
+		}, table1},
+		{[]string{"table2", "TABLE2"}, table2},
```
makalaaneesh, line 479: If you add `TaBLe2` to this list, the test still passes. Do we want that? 
Basically, ORACLE DB has the following tables:
- TABLE2 (case insensitive)
- Table2 (case sensitive)

Now, if someone passes `TaBLe2` , is it safe to lookup and return the case insensitive match ? Maybe the user wanted `Table2` and made a mistake in specifying the case. Maybe we should fail here?

yb-voyager/src/namereg/namereg.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+	match2, err := filepath.Match(pattern, nv.Unqualified.Quoted)
+	if err != nil {
+		return false, fmt.Errorf("invalid pattern: %s", pattern)
+	}
+	return match2, nil
+}
+
+// <SourceTableName, TargetTableName>
+type NameTuple struct {
```
makalaaneesh, line 410: NameTuple is a little ambiguou/confusing imo 😅 
How about SourceOrTargetName ?

yb-voyager/src/namereg/namereg_test.go
amit-yb:
```


casedb=# create table foo (k int primary key, v int);
CREATE TABLE
casedb=# create table "Foo" (k int primary key, v int);
CREATE TABLE
casedb=# \dt
             List of relations
 Schema |     Name      | Type  |  Owner
--------+---------------+-------+----------
 public | ALL_UPPERCASE | table | postgres
 public | Foo           | table | postgres
 public | MixedCase     | table | postgres
 public | all_lowercase | table | postgres
 public | foo           | table | postgres
(5 rows)

casedb=# select * from foo;
 k | v
---+---
(0 rows)

casedb=# select * from "Foo";
 k | v
---+---
(0 rows)

casedb=# select * from FOO;
 k | v
---+---
(0 rows)

```
Take a look at the above scenario: the db didn't say that `FOO` is ambiguous -- it picks the case insensitive variant. Whenever user needs precision they have to spell the exact case: with quotes on psql shell and without quotes on voyager cli.

The choice is also important in case of fall-back case:
Oracle: `SAKILA.TABLE1`, `SAKILA.Table1`
YB: `ybsakila.table1`, `ybsakila.Table1`
An event exported from YB for table `ybsakila.table1`. We __have__ to map the name to `SAKILA.TABLE1`.

amit-yb (commented):
Addressed review comments.

yb-voyager/src/namereg/namereg_test.go
makalaaneesh: As discussed offline, 

if input to Lookup is quoted, we should force exact match. This does not happen in the current implementation. It falls back to the case insensitive match. 
- Lookup(fOo) should pick the case insensitive match. (foo)
- Lookup("Foo") should pick "Foo" 
- Lookup("fOo") should fail.

yb-voyager/src/namereg/namereg.go
```diff
@@ -0,0 +1,527 @@
...
```
makalaaneesh, line 431: - We should try and move this to sqlname package. That way srcdb/tgtdb can freely use this. (Will try this out when Priaynshi/I are integrating this into code)
- The "Mode" field seems a bit wrong at this layer. It should be at the lookup layer, and finally what we get is a NameTuple that knows what CurrentName is mapped to.

makalaaneesh (approved):
LGTM with few comments.

yb-voyager/src/namereg/namereg.go
```diff
@@ -0,0 +1,506 @@
...
+	case len(parts) == 1:
+		schemaName = ""
+		tableName = tableNameArg
+	case len(parts) == 2:
+		schemaName = parts[0]
+		tableName = parts[1]
+	default:
+		return nil, fmt.Errorf("invalid table name: %s", tableNameArg)
+	}
+	if schemaName == reg.DefaultSourceSideSchemaName() || schemaName == reg.DefaultYBSchemaName {
```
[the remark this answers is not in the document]
makalaaneesh: @amit-yb  planning to change this block to `if schemaName == reg.DefaultSourceDBSchemaName || schemaName == reg.DefaultSourceReplicaDBSchemaName || schemaName == reg.DefaultYBSchemaName` 

Consider the case of fall-forward import-data to source-replica (with different schema for source-replica `SAKILA_REPLICA`

During the  snapshot and event data in the beginning before cutover, lookup will be for `SAKILA.TABLE1`, but we want to get the SourceName to be `SAKILA_REPLICA.TABLE1` 

Do you see any issues?

PR #1347, opened 2024-03-04
Bug fixes: Partitions for PG adding leaf partitions in case table-list passed by user has root table and load data via root table into target

priyanshi-yb:
This patch includes two bug fixes in the case of pg_dump for offline and live migration's snapshot phase.
example - 
```
customers (root table) -> cust_other, cust_part11, cust_part12, cust_part22, cust_part21 (leaf partitions)
```

1. In case the user has passed the root table in table-list `--table-list 'customers'`, no data will be exported as pg_dump dumps the data for leaf partitions only, so in such case, we need to get the leaf partitions of this table and then pass those to pg_dump.
2. While importing data of partitions in target we use the name with which the export data has dumped the data which is leaf partitions but in case the user has changed the partition strategy, list, or names this will not work so in any situation the root table is the only source of truth and we should load data in target via root table.

Jenkins pipelines for Oracle and live migration tests -
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2033/

### DIFF
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -176,20 +176,20 @@ func exportData() bool {
 		dfd.Save()
 		os.Exit(0)
 	}
-
-	finalTableList, err = addLeafPartitionsInTableList(finalTableList)
+	var partitionsToRootTableMap map[string]string
+	partitionsToRootTableMap, finalTableList, err = addLeafPartitionsInTableList(finalTableList)
 	if err != nil {
 		utils.ErrExit("failed to add the leaf partitions in table list: %w", err)
 	}
 
-	metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
-		record.RenameTablesMap = strings.Join(lo.MapToSlice(partitionsToRootTableMap, func(k, v string) string {
-			return fmt.Sprintf("%s:%s", k, v)
-		}), ",")
-	})
+	if source.DBType == POSTGRESQL {
+		metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
+			record.RenameTablesMap = partitionsToRootTableMap
+		})
+	}
 
 	if changeStreamingIsEnabled(exportType) || useDebezium {
-		config, tableNametoApproxRowCountMap, err := prepareDebeziumConfig(finalTableList, tablesColumnList)
+		config, tableNametoApproxRowCountMap, err := prepareDebeziumConfig(partitionsToRootTableMap, finalTableList, tablesColumnList)
 		if err != nil {
 			log.Errorf("Failed to prepare dbzm config: %v", err)
 			return false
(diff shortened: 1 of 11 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -177,6 +177,17 @@ func exportData() bool {
 		os.Exit(0)
 	}
 
+	finalTableList, err = addLeafPartitionsInTableList(finalTableList)
+	if err != nil {
+		utils.ErrExit("failed to add the leaf partitions in table list: %w", err)
+	}
+
+	metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
+		record.RenameTablesMap = strings.Join(lo.MapToSlice(partitionsToRootTableMap, func(k, v string) string {
```
makalaaneesh, line 186: why stringify it? We should be okay to store it as a map itself, right? You wouldn't need the logic in `getRenameTableMap` then.

yb-voyager/cmd/exportData.go
```diff
@@ -272,6 +283,78 @@ func exportData() bool {
...
+	//Refer- https://debezium.zulipchat.com/#narrow/stream/302529-community-general/topic/Connector.20not.20working.20with.20partitions
+	for _, table := range tableList {
+		rootTable, err := GetRootTableOfPartition(table)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get root table of partition %s: %v", table.Qualified.MinQuoted, err)
+		}
+		allLeafPartitions := GetAllLeafPartitions(table)
+		switch true {
+		case len(allLeafPartitions) == 0 && rootTable != table: //leaf partition
+			partitionsToRootTableMap[table.Qualified.Unquoted] = rootTable.Qualified.MinQuoted // Unquoted->MinQuoted map as debezium uses Unquoted table name
```
makalaaneesh, line 316: Shall we get rid of this global variable `partitionsToRootTableMap` ? We can construct that map and return it from this function. 

The caller can then do whatever they want with it. 
in this case the caller will:
1. store in msr. 
2. pass it along to `prepareDebeziumConfig()`.

yb-voyager/src/srcdb/postgres.go
```diff
@@ -501,7 +501,13 @@ WHERE parent.relname='%s' AND nmsp_parent.nspname = '%s' `, tableName.ObjectName
 		if err != nil {
 			utils.ErrExit("Error in scanning for child partitions of table=%s: %v", tableName, err)
 		}
-		partitions = append(partitions, sqlname.NewSourceName(childSchema, childTable))
+		if tableName.ObjectName.MinQuoted != tableName.ObjectName.Unquoted {
```
makalaaneesh, line 504: Is this a reliable way to figure out if a name is case-sensitive? I'm not entirely clear on what MinQuoted is.

makalaaneesh (approved):
Added a few minor comments. LGTM! Thanks for the description.

makalaaneesh:
@priyanshi-yb pls add automation test-cases for these fixes.
For 1, we can re-use an existing test-case folder, just add a new test-case run and pass an explicit table list via env 
For 2, you'll have to manually change the schema, I guess? not so sure how to do it in the current test setup.

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: Yes, the `MinQuoted` means that the table name will be quoted if required which means for PG/YB in case of case-sensitive table name, in `MinQuoted` it will be quoted.

makalaaneesh (approved):

PR #1348, opened 2024-03-05
Clean up pgx usage in ExecuteBatch

makalaaneesh:
1. Use `tx.SendBatch` instead of `conn.SendBatch`. It's harmless to use `conn.SendBatch` inside a transaction, but it's less confusing this way. Reference - https://github.com/jackc/pgx/issues/879#issuecomment-1974900237
2. ~~Defer BatchResults.Close() so that we close the batch even if there is an error. As per [docs](https://github.com/jackc/pgx/blob/master/batch.go#L97), best practice to close before re-using the connection. Again, this is not really an issue because we don't re-use the same connection if there's an error, but it's cleaner this way.~~ Defer does not work. Need to call close in both scenarios - normal as well as error because we use the same connection to Exec metadata queries in the same function,.

### DIFF
--- a/yb-voyager/src/tgtdb/postgres.go
+++ b/yb-voyager/src/tgtdb/postgres.go
@@ -570,8 +570,8 @@ func (pg *TargetPostgreSQL) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
 
 		br := tx.SendBatch(ctx, &ybBatch)
 		defer func() {
-			if err = br.Close(); err != nil {
-				log.Errorf("error closing batch: %v", err)
+			if closeErr := br.Close(); closeErr != nil {
+				log.Errorf("error closing batch: %v", closeErr)
 			}
 		}()
 		for i := 0; i < len(batch.Events); i++ {
--- a/yb-voyager/src/tgtdb/postgres.go
+++ b/yb-voyager/src/tgtdb/postgres.go
@@ -568,17 +568,18 @@ func (pg *TargetPostgreSQL) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
 			}
 		}
 
-		br := conn.SendBatch(ctx, &ybBatch)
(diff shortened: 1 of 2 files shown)


sanyamsinghal (approved):
lgtm

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -568,17 +568,25 @@ func (pg *TargetPostgreSQL) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
...
+				log.Errorf("error closing batch: %v", closeErr)
+				return closeErr
+			}
+			return nil
+		}
 		for i := 0; i < len(batch.Events); i++ {
 			_, err := br.Exec()
 			if err != nil {
 				log.Errorf("error executing stmt for event with vsn(%d): %v", batch.Events[i].Vsn, err)
+				closeBatch()
```
sanyamsinghal, line 583: it will be safer to just log the `closeBatch()` error here if its not nil.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -555,17 +555,25 @@ func (yb *TargetYugabyteDB) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
...
+				log.Errorf("error closing batch: %v", closeErr)
+				return closeErr
+			}
+			return nil
+		}
 		for i := 0; i < len(batch.Events); i++ {
 			_, err := br.Exec()
 			if err != nil {
 				log.Errorf("error executing stmt for event with vsn(%d): %v", batch.Events[i].Vsn, err)
+				closeBatch()
```
sanyamsinghal, line 570: same as above

sanyamsinghal (approved):

yb-voyager/src/tgtdb/postgres.go
makalaaneesh: closeBatch already logs the error 👍

PR #1349, opened 2024-03-05
Added check for replica identity of tables in export data from PG

ShivanshGahlot:


yb-voyager/cmd/exportData.go
```diff
@@ -201,6 +201,11 @@ func exportData() bool {
 			// 3. start debezium with configration to read changes from the created replication slot, publication.
 
 			if !dataIsExported() { // if snapshot is not already done...
+				err = checkIfTablesHaveReplicationIdentityFull(finalTableList)
```
makalaaneesh, line 204: This check can be done one block above, (line 202). Irrespective of whether snapshot is done, if it is PG and live -migration, we should perform this check

yb-voyager/src/srcdb/postgres.go
```diff
@@ -669,3 +669,31 @@ func (pg *PostgreSQL) GetNonPKTables() ([]string, error) {
 	}
 	return nonPKTables, nil
 }
+
+func (pg *PostgreSQL) GetTablesWithReplicaIdentityNotFull(tableList []*sqlname.SourceName) ([]string, error) {
+	var tablesWithReplicaIdentityNotFull []string
+	var qualifiedTableNames []string
+	for _, table := range tableList {
+		qualifiedTableNames = append(qualifiedTableNames, fmt.Sprintf("'%s'", table.Qualified.MinQuoted))
```
makalaaneesh, line 677: pls check that this works for case-sensitive tables as well. "FoO" for example. Check what the value for this table is in c.relname. Is it with quotes, without quotes? I'm assuming it should be without quotes, in which case MinQuoted would be wrong?

yb-voyager/src/srcdb/srcdb.go
```diff
@@ -51,6 +51,7 @@ type SourceDB interface {
 	GetTableToUniqueKeyColumnsMap(tableList []*sqlname.SourceName) (map[string][]string, error)
 	ClearMigrationState(migrationUUID uuid.UUID, exportDir string) error
 	GetNonPKTables() ([]string, error)
+	GetTablesWithReplicaIdentityNotFull(tableList []*sqlname.SourceName) ([]string, error)
```
makalaaneesh, line 54: We shouldn't have a func with this name in the SourceDB interface, because "ReplicaIdentity" is PG-specific. I see two options:
1. Remove it from the interface; Have the func `checkIfTablesHaveReplicationIdentityFull` use the concrete PG struct by using [type assertion](https://go.dev/tour/methods/15)
2.  Declare a generic method like `ValidateTablesReadyForLiveMigration` in the interface, and return a generic set of validation results for a set of tables. (map[string]string[])

yb-voyager/cmd/exportData.go
ShivanshGahlot: Done

yb-voyager/src/srcdb/postgres.go
ShivanshGahlot: Leaving it out now since the current live migration doesnt support case sensitive tables

yb-voyager/src/srcdb/srcdb.go
ShivanshGahlot: Chose the 2nd option since it was easier to implement.

ShivanshGahlot: We can add further checks in the future if necessary in the same function

makalaaneesh (approved):
LGTM

PR #1350, opened 2024-03-07
Init NameRegistry

makalaaneesh:
- Initialize NameRegistry in import-data, export-data, import-data-to-file
- Move NameTuple, ObjectName to sqlname to enable use of these objects in srcdb/tgtdb (to avoid cyclic dependency).

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -111,6 +111,11 @@ func importDataCommandFn(cmd *cobra.Command, args []string) {
 	sourceDBType = GetSourceDBTypeFromMSR()
 	sqlname.SourceDBType = sourceDBType
 
+	if tconf.TargetDBType == YUGABYTEDB {
+		tconf.Schema = strings.ToLower(tconf.Schema)
+	} else if tconf.TargetDBType == ORACLE && !utils.IsQuotedString(tconf.Schema) {
+		tconf.Schema = strings.ToUpper(tconf.Schema)
+	}
 	tdb = tgtdb.NewTargetDB(&tconf)
 	err := tdb.Init()
 	if err != nil {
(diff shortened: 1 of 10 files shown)


yb-voyager/src/namereg/namereg.go
```diff
@@ -287,16 +299,18 @@ func (reg *NameRegistry) LookupTableName(tableNameArg string) (*NameTuple, error
 			}
 		}
 	}
-	// Set the current table name based on the mode.
-	ntup.SetMode(reg.role)
-	if ntup.SourceName == nil && ntup.TargetName == nil {
+	// // Set the current table name based on the mode.
```
priyanshi-yb, line 302: Remove the commented lines if not required.

yb-voyager/cmd/importData.go
```diff
@@ -109,6 +110,24 @@ func importDataCommandFn(cmd *cobra.Command, args []string) {
...
+		tconf.Schema = strings.ToLower(tconf.Schema)
+	} else if tconf.TargetDBType == ORACLE && !utils.IsQuotedString(tconf.Schema) {
+		tconf.Schema = strings.ToUpper(tconf.Schema)
+	}
+	tdb = tgtdb.NewTargetDB(&tconf)
+	err := tdb.Init()
+	if err != nil {
+		utils.ErrExit("Failed to initialize the target DB: %s", err)
+	}
+	defer tdb.Finalize()
```
priyanshi-yb, line 124: here calling `defer tdb.Finalize()`, will run this `Finalize()` even after the `export data from target` is completed ? as we have this at the end of this function. I think we should call this in `importData()` itself?
```
if changeStreamingIsEnabled(importType) {
		startExportDataFromTargetIfRequired()
	}
```

priyanshi-yb (approved):
LGTM.

yb-voyager/cmd/importData.go
makalaaneesh: did not add to `importdata()` because it is called by import data file command as well. 
added it in `importDataCommandFn` after calling `importData`

PR #1353, opened 2024-03-11
Partitions support for PG live migration, several UX changes wrt partitions (offline and live)

priyanshi-yb:
This patch changes the UX of how to report the partition's progress in the report/status commands which now represent partitions everywhere with a root table as per the following.
1. Import should be unaware of partitions at all
2. Rename the Export display of table list or everything for leaf->root
3. Store any metadata after export with root table only (such as DataFileDescriptor, ExportSnapshotStatus, MSR.TableListExportedFromSource)
4. Renamed the RenameTablesMap -> SourceRenameTablesMap and storing one more map in MSR for TargetRenameTablesMap for target's leaf parititon -> root mapping to be able to rename that

Testing: Added automation for parititons live migration / with fall-forward / with fall-back


yb-voyager/cmd/common.go
```diff
@@ -225,14 +234,21 @@ func displayExportedRowCountSnapshot(snapshotViaDebezium bool) {
 					schema = tableParts[0]
 					table = tableParts[1]
 				}
-				uitable.AddRow(schema, table, exportedRowCount[key])
+				displayTableName := table
+				if source.DBType == POSTGRESQL && leafPartitions[key] != nil && msr.IsExportTableListSet {
```
makalaaneesh, line 238: What is the purpose of `msr.IsExportTableListSet`

yb-voyager/cmd/common.go
```diff
@@ -195,20 +195,29 @@ func GetTableRowCount(filePath string) map[string]int64 {
 	return tableRowCountMap
 }
 
-func getExportedRowCountSnapshot(exportDir string) map[string]int64 {
+func getExportedRowCountSnapshot(exportDir string) (map[string]int64, map[string][]string) {
 	tableRowCount := map[string]int64{}
+	leafPartitions := map[string][]string{}
 	for _, fileEntry := range datafile.OpenDescriptor(exportDir).DataFileList {
-		tableRowCount[fileEntry.TableName] += fileEntry.RowCount
+		renamedTable := renameTableIfRequired(fileEntry.TableName)
```
makalaaneesh, line 202: nit: it'll be cleaner if `renameTableIfRequired` can return two values (renamedTable, didRenameHappen-bool). Then you can use the second value directly instead of having to compare original vs renamed table in every use of this func

yb-voyager/cmd/common.go
```diff
@@ -225,14 +234,21 @@ func displayExportedRowCountSnapshot(snapshotViaDebezium bool) {
 					schema = tableParts[0]
 					table = tableParts[1]
 				}
-				uitable.AddRow(schema, table, exportedRowCount[key])
+				displayTableName := table
```
makalaaneesh, line 237: I see this logic being used in multiple places. Can we define a func getDisplayTableName(table, sourceDBType, leafPartitions, ... ) ?

yb-voyager/cmd/common.go
```diff
@@ -669,3 +720,61 @@ func renameTableIfRequired(table string) string {
...
 	return table
 }
+
+
+func getExportedSnapshotRowsMap(tableList []string, exportSnapshotStatus *ExportSnapshotStatus) (map[string]int64, map[string]string, error) {
+	snapshotRowsMap := make(map[string]int64)
+	snapshotStatusMap := make(map[string]string)
+	for _, table := range tableList {
+		renamedTable := renameTableIfRequired(table)
+		snapshotStatusMap[renamedTable] = exportSnapshotStatus.Tables[table].Status
```
makalaaneesh, line 730: What do we display here as part of "status"? It'd be technically wrong I guess because we're simply taking the last leaf table's status to be the root table's status?

yb-voyager/cmd/live_migration.go
```diff
@@ -54,7 +55,12 @@ func streamChanges(state *ImportDataState, tableNames []string) error {
 	log.Infof("NUM_EVENT_CHANNELS: %d, EVENT_CHANNEL_SIZE: %d, MAX_EVENTS_PER_BATCH: %d, MAX_INTERVAL_BETWEEN_BATCHES: %d",
 		NUM_EVENT_CHANNELS, EVENT_CHANNEL_SIZE, MAX_EVENTS_PER_BATCH, MAX_INTERVAL_BETWEEN_BATCHES)
 	tdb.PrepareForStreaming()
-	err := state.InitLiveMigrationState(migrationUUID, NUM_EVENT_CHANNELS, bool(startClean), tableNames)
+
+	renamedTables := make([]string, 0)
+	for _, tableName := range tableNames {
+		renamedTables = append(renamedTables, renameTableIfRequired(tableName))
+	}
+	err := state.InitLiveMigrationState(migrationUUID, NUM_EVENT_CHANNELS, bool(startClean), renamedTables)
```
makalaaneesh, line 63: Why are we renaming here? Because essentially we're passing renameconfig to dbzm, so we will always get root tables? Pls add a comment here to explain..
Also, should we do a unique before passing? I assume renamedTables might have duplicates
?

yb-voyager/cmd/exportData.go
```diff
@@ -176,6 +176,30 @@ func exportData() bool {
 		utils.ErrExit("failed to add the leaf partitions in table list: %w", err)
 	}
 
+	leafPartitions := make(map[string][]string)
```
makalaaneesh, line 179: What display problem are we solving here exactly?

yb-voyager/cmd/importDataStatusCommand.go
```diff
@@ -154,11 +155,55 @@ func prepareImportDataStatusTable() ([]*tableMigStatusOutputRow, error) {
...
 	}
 
+	renamedRowsForPartitions := make(map[string]*tableMigStatusOutputRow, 0)
+
 	for _, dataFile := range dataFileDescriptor.DataFileList {
 		row, err := prepareRowWithDatafile(dataFile, state)
 		if err != nil {
 			return nil, fmt.Errorf("prepare row with datafile: %w", err)
 		}
+		renamedTable := renameTableIfRequired(row.tableName)
```
makalaaneesh, line 165: Not exactly sure what is happening here.. I thought we already modified the import data status related functions?

yb-voyager/cmd/common.go
priyanshi-yb: This is to handle the second case if a subset of the table-list is passed, So I am just doing a simple thing to print the leaf partitions in case the table list is passed to export nothing extra for the subset and all. I think if a user has passed a table-list (event the root table is passed) then we should tell what all will be exported or imported.

yb-voyager/cmd/common.go
priyanshi-yb: Yes will do that, this is the code duplication mainly I was talking about.

yb-voyager/cmd/common.go
priyanshi-yb: Yeah, I had fixed it properly in import data status. Missed here will do.

yb-voyager/cmd/exportData.go
priyanshi-yb: The same with the table list we display on UI for users to see what is being exported. As the actual table list is with leaf partitions but to display we should show the root as per what we had decided.

yb-voyager/cmd/importDataStatusCommand.go
priyanshi-yb: No this is the only place where I am modifying the import data status related functions.

priyanshi-yb: I am doing the same thing I did at other places clubbing all the leaf partition rows to root table one. Getting all the leaf partitions, clubbing the count of imported leaf partitions for that root, etc... 
Let me know if you to connect to discuss this.

yb-voyager/cmd/exportData.go
```diff
@@ -181,14 +181,54 @@ func exportData() bool {
...
+			partitions := strings.Join(leafPartitions[table], ", ")
+			return fmt.Sprintf("%s (%s)", table, partitions)
+		}
+		return table
+	})
+	fmt.Printf("num tables to export: %d\n", len(tableListToDisplay))
+	utils.PrintAndLog("table list for data export: %v", tableListToDisplay)
+
+	//add root tables in the list for catalog queries to work fine with both leaf and root tables
+	defaultSchema, _ := GetDefaultPGSchema(source.Schema, "|")
```
makalaaneesh, line 222: see if you can add this to the function `addLeafPartitionsInTableList` ?

yb-voyager/cmd/exportData.go
```diff
@@ -605,6 +638,15 @@ func exportDataOffline(ctx context.Context, cancel context.CancelFunc, finalTabl
 	}
 
 	source.DB().ExportDataPostProcessing(exportDir, tablesProgressMetadata)
+	if source.DBType == POSTGRESQL {
```
makalaaneesh, line 641: Ideally,  we should do this in ExportDataPostProcessing of PostgresDB itself. Is there anything blocking that? 

THe datafiledescriptor is written by pg_dump layer itself i believe. 
Not so sure about the exportSnapshotStatusFile

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -59,23 +59,24 @@ func prepareDebeziumConfig(partitionsToRootTableMap map[string]string, tableList
...
 	tableNameToApproxRowCountMap := getTableNameToApproxRowCountMap(tableList)
 
 	var dbzmTableList, dbzmColumnList []string
 	for _, table := range tableList {
+		t := table.Qualified.MinQuoted
+		if source.DBType == POSTGRESQL && table.SchemaName.MinQuoted == "public" {
+			t = table.ObjectName.MinQuoted
+		}
+		if leafPartitions[t] != nil {
+			//In case of debezium offline migration of PG, tablelist should not have root and leaf both so not adding root table in table list
```
makalaaneesh, line 71: just checking, this flow will kick in even for live migration, correct?

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -59,23 +59,24 @@ func prepareDebeziumConfig(partitionsToRootTableMap map[string]string, tableList
...
+			continue
+		}
 		dbzmTableList = append(dbzmTableList, table.Qualified.Unquoted)
```
makalaaneesh, line 74: Wait you don't seem to be using t. Is that intended? t  is just for skipping root tables?

yb-voyager/cmd/importDataStatusCommand.go
```diff
@@ -93,10 +94,10 @@ func runImportDataStatusCmd() error {
 			uiTable.AddRow(row.tableName, row.fileName, row.status, totalCount, importedCount, perc)
 		} else {
 			if i == 0 {
-				addHeader(uiTable, "TABLE", "FILE", "STATUS", "TOTAL ROWS", "IMPORTED ROWS", "PERCENTAGE")
+				addHeader(uiTable, "TABLE", "STATUS", "TOTAL ROWS", "IMPORTED ROWS", "PERCENTAGE")
```
makalaaneesh, line 97: For import data file, we still show file, correct?

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -226,13 +250,14 @@ func updateExportedSnapshotRowsInTheRow(msr *metadb.MigrationStatusRecord, row *
...
-		row.ExportedSnapshotRows = status.ExportedRowCountSnapshot
+		if schemaName != "public" {
+			tableName = schemaName + "." + tableName
+		}
+		row.ExportedSnapshotRows = exportedSnapshotPGRowsMap[tableName]
 	}
 }
 
-func updateImportedEventsCountsInTheRow(sourceDBType string, row *rowData, tableName string, schemaName string, targetConf *tgtdb.TargetConf, snapshotDataFileDescriptor *datafile.Descriptor) error {
+func updateImportedEventsCountsInTheRow(sourceDBType string, row *rowData, tableName string, schemaName string, targetConf *tgtdb.TargetConf, snapshotDataFileDescriptor *datafile.Descriptor, snapshotImportedRowsMap map[string]int64) error {
```
makalaaneesh, line 260: maybe we can get rid of `snapshotDataFileDescriptor` ?

makalaaneesh (approved):
LGTM with a few nits!

yb-voyager/cmd/common.go
```diff
@@ -637,35 +719,141 @@ func initBaseTargetEvent(bev *cp.BaseEvent, eventType string) {
...
+	renameTablesMap := msr.SourceRenameTablesMap
+	if source.DBType == YUGABYTEDB {
+		renameTablesMap = msr.TargetRenameTablesMap
+	}
+	defaultSchema, noDefaultSchema := GetDefaultPGSchema(schema, "|")
 	if noDefaultSchema && len(strings.Split(table, ".")) <= 1 {
 		utils.ErrExit("no default schema found to qualify table %s", table)
 	}
 	tableName := sqlname.NewSourceNameFromMaybeQualifiedName(table, defaultSchema)
 	fromTable := tableName.Qualified.Unquoted
```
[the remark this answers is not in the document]
makalaaneesh: Yeah. You were using `.Qualified.Unquoted` so I was wondering...

yb-voyager/cmd/exportData.go
priyanshi-yb: Yes, you're right dataFileDescriptor for all non-dbzm cases is written in exportDataPostProcessing. But when I was trying to make changes there, It was a bit more code duplication that way as I already have the rename function in this code path, So I realised this way is easier and less complicated to just update these things once written.

.github/workflows/pg-migtests.yml
```diff
@@ -113,9 +125,12 @@ jobs:
 
       - name: "TEST: pg-partitions"
         run: migtests/scripts/run-test.sh pg/partitions
+        if: matrix.version != '2.16.7.0-b59'
```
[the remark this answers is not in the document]
priyanshi-yb: Because I have added PKs now in the partitions schema which will create some INDEX related DDLs which doesn't work in this by version, we have similar test pg-partitions-with-indexes which also doesn't run for this version

yb-voyager/cmd/exportDataDebezium.go
priyanshi-yb: Yes right, but atleast even if we pass table-list with root to dbzm there won't be any issue as events only be exported on the basis of leaf partitions but in snapshot phase we can't as you were also mentioning that it duplicates the data for parititons and root.

yb-voyager/cmd/exportDataDebezium.go
priyanshi-yb: Yes

yb-voyager/cmd/importDataStatusCommand.go
priyanshi-yb: Right

yb-voyager/cmd/common.go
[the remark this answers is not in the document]
priyanshi-yb: Hmm, Actually this is because the map I create for leaf->root mapping is also used by debezium where debezium only deals with `Unquoted` names even for case sensitivity so had to create it like that.

PR #1375, opened 2024-03-20
Support case-sensitive column names.

amit-yb:


yb-voyager/src/tgtdb/attr_name_registry.go
```diff
@@ -0,0 +1,107 @@
...
+func NewAttributeNameRegistry(tdb TargetDB, tconf *TargetConf) *AttributeNameRegistry {
+	return &AttributeNameRegistry{
+		dbType:    tconf.TargetDBType,
+		tdb:       tdb,
+		tconf:     tconf,
+		attrNames: make(map[string][]string),
+	}
+}
+
+func (reg *AttributeNameRegistry) QuoteIdentifier(schemaName, tableName, columnName string) string {
```
amit-yb, line 30: Let's rename this to `QuoteAttributeName()`

yb-voyager/src/tgtdb/attr_name_registry.go
```diff
@@ -0,0 +1,107 @@
...
+		reg.attrNames[qualifiedTableName] = targetColumns
+	}
+	c, err := reg.findBestMatchingColumnName(columnName, targetColumns)
+	if err != nil {
+		utils.ErrExit("find best matching column name for %q in table %s.%s: %w", columnName, schemaName, tableName, err)
+	}
+	return fmt.Sprintf("%q", c)
+}
+
+func (reg *AttributeNameRegistry) IfRequiredQuoteColumnNames(tableName string, columns []string) ([]string, error) {
```
amit-yb, line 56: Let's rename this to `QuoteAttributeNames()` and reimplement this using calls to `QuoteAttributeName()` in a loop over `columns`.

yb-voyager/src/dbzm/valueConverter.go
```diff
@@ -255,13 +257,17 @@ func (conv *DebeziumValueConverter) GetTableNameToSchema() *utils.StructMap[sqln
 		colSchemaMap := make(map[string]map[string]string)
 
 		for _, col := range tblSchema.Columns {
-			colSchemaMap[col.Name] = col.Schema.Parameters
+			colNameQuoted, err := conv.tdb.QuoteAttributeName(tbl, col.Name)
+			if err != nil {
+				return false, fmt.Errorf("error quoting attribute name %s for table %s: %w", col.Name, tbl, err)
+			}
+			colSchemaMap[colNameQuoted] = col.Schema.Parameters
```
amit-yb, line 264: Are you sure we need quoted column names inside value converter? I guess not. COPY statement (during snapshot load) and IUD statements during streaming are the only two places where we need to quote the column names.

amit-yb (commented):
Posted one comment.

yb-voyager/src/dbzm/valueConverter.go
makalaaneesh: Yeah, a test case failed because of this. 
basically, we are storing a map[column]schema. and we lookup in this map while importing. Now, the lookup key was quoted, but this map had unquoted column names as keys. That's why I had to make this also quoted.

priyanshi-yb (approved):
LGTM

migtests/tests/import-file/run-import-file-test
```diff
@@ -223,9 +223,9 @@ main() {
 			--file-table-map "csv_with_header_test.csv:gcs_csv_with_header" --has-header true
 
 	# Test larger import from GCS buckets
-	step "Import large data file from GCS: volume.csv-> gcs_volume"
```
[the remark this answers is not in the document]
makalaaneesh: volume.csv had header which as comma-delimited, and data which was tab-delimited. 
With the change in this PR, the columns were looked up in the attribute name registry, and was therefore failing. 

I fixed the file, and instead of overwriting, I created a volume1.csv just to be on the safe side!

PR #1378, opened 2024-03-22
Added checks for asking the user for target db password and running import data status in end migration

ShivanshGahlot:


yb-voyager/cmd/importData.go
```diff
@@ -390,6 +390,9 @@ func updateTargetConfInMigrationStatus() {
 }
 
 func importData(importFileTasks []*ImportFileTask) {
+	if importerRole == TARGET_DB_IMPORTER_ROLE {
```
makalaaneesh, line 393: I would say, put this in `importDataCommandFn` ? We set other things in MSR such as `updateFallForwardEnabledInMetaDB` .

yb-voyager/cmd/endMigrationCommand.go
```diff
@@ -400,6 +408,11 @@ func deletePGPublication(msr *metadb.MigrationStatusRecord, source *srcdb.Source
 }
 
 func cleanupTargetDB(msr *metadb.MigrationStatusRecord) {
+	if !msr.ImportDataIntoTargetStarted {
```
makalaaneesh, line 411: see line 417. Looks like we alreay have a check for TargetDBConf. 
I think everwhere, you can use `msr.TargetDBConf != nil` instead of a new flag for `ImportDataIntoTargetStarted` ?

makalaaneesh (approved):
LGTM

PR #1384, opened 2024-03-26
Name registry integration - Import Data + Stats

makalaaneesh:
Integration of https://github.com/yugabyte/yb-voyager/pull/1346 into import and stats code paths.


yb-voyager/src/tgtdb/event.go
```diff
@@ -24,18 +25,56 @@ import (
...
+	e.Vsn = rawEvent.Vsn
+	e.Op = rawEvent.Op
+	e.Key = rawEvent.Key
+	e.Fields = rawEvent.Fields
+	e.BeforeFields = rawEvent.BeforeFields
+	e.ExporterRole = rawEvent.ExporterRole
+	if !e.IsCutoverEvent() {
+		e.TableName, err = namereg.NameReg.LookupTableName(fmt.Sprintf("%s.%s", rawEvent.SchemaName, rawEvent.TableName))
+		if err != nil {
+			return fmt.Errorf("lookup table %s.%s in name registry: %v", rawEvent.SchemaName, rawEvent.TableName, err)
```
amit-yb, line 73: `%w` format specifier for `err`.

yb-voyager/src/tgtdb/event.go
```diff
@@ -65,8 +104,7 @@ func (e *Event) Copy() *Event {
 	return &Event{
 		Vsn:          e.Vsn,
 		Op:           e.Op,
-		SchemaName:   e.SchemaName,
-		TableName:    e.TableName,
+		TableName:    e.TableName, // TODO: TABLENAME do we need to deepcopy?
```
amit-yb, line 107: We never modify a `NameTuple` so shallow-copy should be just fine.

yb-voyager/src/tgtdb/event.go
```diff
@@ -324,15 +355,15 @@ type EventBatch struct {
 	Events             []*Event
 	ChanNo             int
 	EventCounts        *EventCounter
-	EventCountsByTable map[string]*EventCounter
+	EventCountsByTable *utils.StructMap[sqlname.NameTuple, *EventCounter] //map[string]*EventCounter
```
amit-yb, line 358: Remove the trailing comment. Fix all such occurrences if any.

yb-voyager/src/namereg/namereg.go
```diff
@@ -34,6 +32,28 @@ const (
 
 var NameReg NameRegistry
 
+type SourceDbRegistry interface {
+	GetAllTableNamesRaw(schemaName string) ([]string, error)
+}
+
+type YBDBRegistry interface {
```
amit-yb, line 39: We should not use word "Registry" in these interface names. Let's call them just `SourceDBInterface` and `YBDBInterface`.

yb-voyager/src/namereg/namereg.go
```diff
@@ -34,6 +32,28 @@ const (
...
+	FilePath string
+	Role     string
+
+	SourceDBType   string
+	SourceDBSchema string
+	SourceDBName   string
+	SDBReg         SourceDbRegistry
+
+	TargetDBSchema string
+	YBDBReg        YBDBRegistry
```
amit-yb, line 54: Use the same name that is used in `cmd` package i.e. `sdb` and `tdb`.

yb-voyager/src/namereg/namereg.go
```diff
@@ -34,6 +32,28 @@ const (
 
 var NameReg NameRegistry
```
amit-yb, line 33: I would have preferred if the instance was defined in the `cmd` package.

yb-voyager/src/namereg/namereg.go
```diff
@@ -50,42 +70,43 @@ type NameRegistry struct {
...
-	tconf    *tgtdb.TargetConf
-	tdb      tgtdb.TargetDB
+	params NameRegistryParams
+}
+
+func InitNameRegistry(
+	exportDir string, role string,
+	sourceDbType string, sourceDbSchema string, sourceDbName string, targetDbSchema string,
+	sdb SourceDbRegistry,
+	ybdb YBDBRegistry) error {
```
amit-yb, line 80: How about taking `params` as argument? Let the call construct the params object? The whole point of introducing the `params` object was to tame the list of arguments to the `NameRegistry` constructor.

yb-voyager/src/namereg/namereg.go
```diff
@@ -109,28 +130,28 @@ func (reg *NameRegistry) Init() error {
 // TODO: only registry tables from tablelist, not all from schema.
 func (reg *NameRegistry) registerNames() (bool, error) {
 	switch true {
-	case reg.role == SOURCE_DB_EXPORTER_ROLE && reg.SourceDBTableNames == nil:
+	case reg.params.Role == SOURCE_DB_EXPORTER_ROLE && reg.SourceDBTableNames == nil:
```
amit-yb, line 133: How about embedding the `*NameRegistryParams` in the `NameRegistry`? Then we won't need to qualify `reg.params` for every param access.

yb-voyager/src/namereg/namereg_test.go
```diff
@@ -210,47 +212,47 @@ func TestNameRegistryFailedLookup(t *testing.T) {
...
+	// assert.Nil(ntup)
 	assert.ErrorAs(err, &errNameNotFound)
 	assert.Equal(&ErrNameNotFound{ObjectType: "table", Name: "table3"}, errNameNotFound)
 
 	// Missing schema name.
-	ntup, err = reg.LookupTableName("schema1.table1")
+	_, err = reg.LookupTableName("schema1.table1")
 	require.NotNil(err)
-	assert.Nil(ntup)
+	// assert.Nil(ntup)
```
amit-yb, line 224: Remove commented code. Fix all such occurrences.

yb-voyager/src/metadb/metadataDB.go
```diff
@@ -435,6 +437,68 @@ func (m *MetaDB) GetExportedEventsStatsForTableAndExporterRole(exporterRole stri
...
+
+		var ec *tgtdb.EventCounter
+		var found bool
+		ec, found = res.Get(nt)
+		if !found {
+			ec = &tgtdb.EventCounter{
+				TotalEvents: 0,
+				NumInserts:  0,
+				NumUpdates:  0,
+				NumDeletes:  0,
```
amit-yb, line 489: Members are by default initialised to their zero value.

amit-yb (commented):
Reviewed `src/...` packages. Will continue reviewing the `cmd` package tomorrow.

yb-voyager/src/namereg/namereg.go
makalaaneesh: This is a bit tricky. If i embed the fields, then fields like role get persisted automatically, which we don't want (because that will be different across processes). 
 Will look into this.

yb-voyager/src/namereg/namereg.go
makalaaneesh: RIght now, there are places in opther packages that lookup from name registry (event - which is in tgtdb, and schemaregistry), so can't put it in cmd, as it would lead to circular dependency.

This way, name registry is a leaf pkg, and any other pkg needs to only depend on it to use it.

yb-voyager/src/utils/sqlname/nametuple.go
```diff
@@ -143,28 +135,37 @@ func (t *NameTuple) MatchesPattern(pattern string) (bool, error) {
...
+
 //================================================
 
 func quote2(dbType, name string) string {
 	switch dbType {
 	case POSTGRESQL, YUGABYTEDB, ORACLE:
 		return `"` + name + `"`
 	case MYSQL:
+		// TODO:TABLENAME
+		if IsReservedKeywordPG(name) {
```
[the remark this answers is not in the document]
makalaaneesh: test cases passed after removing this 👍

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -100,6 +104,15 @@ func getDataMigrationReportCmdFn(msr *metadb.MigrationStatusRecord) {
 	fBEnabled = msr.FallbackEnabled
 	fFEnabled = msr.FallForwardEnabled
 	tableList := msr.TableListExportedFromSource
+	tableNts, err := getImportTableList(tableList)
```
amit-yb, line 107: `tableNameTups`

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -127,51 +151,66 @@ func getDataMigrationReportCmdFn(msr *metadb.MigrationStatusRecord) {
...
+		namereg.NameReg = *sourceReplicaNameRegistry
+		replicaImportedSnapshotRowsMap, err = getImportedSnapshotRowsMap("source-replica", tableNts)
 		if err != nil {
 			utils.ErrExit("error while getting imported snapshot rows for source-replica DB: %w\n", err)
 		}
+		namereg.NameReg = oldNameReg
 	}
 
-	for i, table := range tableList {
+	for i, nt := range tableNts {
```
amit-yb, line 206: Instead of `nt`, use `nameTup` everywhere.

amit-yb (approved):
LGTM

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -39,6 +38,7 @@ import (
 var targetDbPassword string
 var sourceReplicaDbPassword string
 var sourceDbPassword string
+var sourceReplicaNameRegistry *namereg.NameRegistry
```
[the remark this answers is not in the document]
makalaaneesh: True, but in this case, I am creating a separate name registry for the "IMPORT_TO_SOURCE_REPLICA" role. Can't think of a better name. 
Maybe `nameRegistryForSourceReplicaRole` ?

PR #1389, opened 2024-03-28
select 1 as first stmt in event batch on YB

makalaaneesh:
An additional safety net to workaround an issue in YB where in batched execution, transactions can be retried partially, breaking atomicity.
 SELECT 1 causes the ysql layer to record that data was sent back to the user, thereby, preventing retries.

Additionally, setting the default isolation level to repeatable_read because under read_committed, the SELECT 1 workaround does not work. 
https://yugabyte.slack.com/archives/CAR5BCH29/p1708320808330589


priyanshi-yb (approved):
LGTM!

amit-yb (approved):
LGTM

PR #1394, opened 2024-04-03
Added case sensitivity and reserved words test cases

shubham-yb:
1. Added case sensitivity and reserved words test cases for Oracle, PG and MySQL.
2. The tests are added for offline, live, fall forward and fall back (except MySQL).
3. Minor automation script changes
4. Minor UI change


migtests/scripts/live-migration-run-test.sh
```diff
@@ -126,7 +127,7 @@ main() {
 	step "Archive Changes."
 	archive_changes &
 
-	sleep 30 
+	sleep 60
```
priyanshi-yb, line 130: Is this for any reason in normal and fallback and not in fall-forward?

migtests/tests/pg/case-sensitivity-reserved-words/add-pk-from-alter-to-create
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
```
priyanshi-yb, line 1: Instead of duplicating the file let's put this in the `scripts` dir and call it from there in scripts for all tests where required.

priyanshi-yb (approved):
LGTM

migtests/tests/pg/case-sensitivity-reserved-words/add-pk-from-alter-to-create
priyanshi-yb: Nit right now if it is easy to do then do it, else it's fine for now.

migtests/scripts/live-migration-run-test.sh
shubham-yb: Fall Forward also has this but after trapping the archive changes process. It's set to 120 currently. Let me change to 60.

migtests/tests/pg/case-sensitivity-reserved-words/add-pk-from-alter-to-create
shubham-yb: Yes didn't want to get into this currently. Will do this separately.

PR #1397, opened 2024-04-03
check if event batch is already imported before retry

makalaaneesh:
In certain situations, we get an error on `targetDB.ExecuteBatch`, but eventually the transaction is committed. For example, in Yugabyte, we can get an `rpc timeout` on commit, and the commit eventually succeeds on the DB server. 
Retrying an already executed batch has consequences:
- It can fail with some duplicate / unique key constraint errors
- Stats will double count the events.

This PR ensures that we check if the batch has already been imported (by comparing the last_applied_vsn with the last event's vsn) before retrying.


priyanshi-yb (approved):
LGTM

yb-voyager/cmd/live_migration.go
```diff
@@ -45,6 +46,7 @@ var END_OF_QUEUE_SEGMENT_EVENT = &tgtdb.Event{Op: "end_of_source_queue_segment"}
 var FLUSH_BATCH_EVENT = &tgtdb.Event{Op: "flush_batch"}
 var eventQueue *EventQueue
 var statsReporter *reporter.StreamImportStatsReporter
+var tgtDbQueryMutex sync.Mutex
```
amit-yb, line 49: This is an incorrect place for adding this mutex. If really required, it should be inside the `tdb`.

amit-yb (approved):
LGTM with one comment.

yb-voyager/cmd/live_migration.go
makalaaneesh: I added it because tdb.Query uses the single connection in each targetDB, not the conn pool. 

I agree with your suggestion. it makes more sense in tdb. Should I make all the implementations of tdb.Query/Exec/etc use a mutex?

makalaaneesh:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2240/

priyanshi-yb (approved):

PR #1400, opened 2024-04-04
Removing root tables from the table list in CREATE PUBLICATION query for PG

priyanshi-yb:
PG11 [doesn't support](https://www.postgresql.org/docs/11/sql-createpublication.html#:~:text=Only%20persistent%20base%20tables%20can%20be%20part%20of%20a%20publication.%20Temporary%20tables%2C%20unlogged%20tables%2C%20foreign%20tables%2C%20materialized%20views%2C%20regular%20views%2C%20and%20partitioned%20tables%20cannot%20be%20part%20of%20a%20publication.%20To%20replicate%20a%20partitioned%20table%2C%20add%20the%20individual%20partitions%20to%20the%20publication.) creating publication on a partitioned table. So to be consistent for all the PG versions removing the root table from the list.

### DIFF
--- a/yb-voyager/src/srcdb/postgres.go
+++ b/yb-voyager/src/srcdb/postgres.go
@@ -705,7 +705,7 @@ func (pg *PostgreSQL) DropLogicalReplicationSlot(conn *pgconn.PgConn, replicatio
 	return nil
 }
 
-func (pg *PostgreSQL) CreatePublication(conn *pgconn.PgConn, publicationName string, tableList []sqlname.NameTuple, dropIfAlreadyExists bool) error {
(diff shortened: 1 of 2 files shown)


yb-voyager/src/srcdb/postgres.go
```diff
@@ -714,6 +714,10 @@ func (pg *PostgreSQL) CreatePublication(conn *pgconn.PgConn, publicationName str
 	}
 	tablelistQualifiedQuoted := []string{}
 	for _, tableName := range tableList {
+		if leafPartitions[tableName.ForOutput()] != nil {
```
makalaaneesh, line 717: We should use utils.StructMap for leafPartitions, not `map[string]...`

priyanshi-yb: Yes was thinking of doing it but first thought to fix it. Will change.

makalaaneesh (approved):
LGTM

PR #1401, opened 2024-04-04
Removed extra handling for debezium export data for dumped files and dataFileDescriptor 

priyanshi-yb:
- changed the `partitionsToRootTableMap` passed to debezium to have root table as Unquoted instead of Quoted 
- Removed renaming of debezium dumped exported files (data and schema).
- Removed extra handling of case sensitive and reserved words in dataFileDescriptor for debezium

### DIFF
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -346,14 +346,14 @@ func addLeafPartitionsInTableList(tableList []sqlname.NameTuple) (map[string]str
 		allLeafPartitions := GetAllLeafPartitions(table)
 		switch true {
 		case len(allLeafPartitions) == 0 && rootTable != table: //leaf partition
-			partitionsToRootTableMap[qualifiedCatalogName] = rootTable.ForKey() // Unquoted->MinQuoted map as debezium uses Unquoted table name
+			partitionsToRootTableMap[qualifiedCatalogName] = rootTable.AsQualifiedCatalogName() // Unquoted->Unquoted map as debezium uses Unquoted table name
 			modifiedTableList = append(modifiedTableList, table)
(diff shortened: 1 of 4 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -346,14 +346,14 @@ func addLeafPartitionsInTableList(tableList []sqlname.NameTuple) (map[string]str
 		allLeafPartitions := GetAllLeafPartitions(table)
 		switch true {
 		case len(allLeafPartitions) == 0 && rootTable != table: //leaf partition
-			partitionsToRootTableMap[qualifiedCatalogName] = rootTable.ForKey() // Unquoted->MinQuoted map as debezium uses Unquoted table name
+			partitionsToRootTableMap[qualifiedCatalogName] = rootTable.AsQualifiedCatalogName() // Unquoted->Unquoted map as debezium uses Unquoted table name
```
makalaaneesh, line 349: Let's try to use utils.StructMap wherever we want to use table as a key. 
Using map[string].. is fine but quite error prone. 
If there are any concerns in this case, given that we manually create the name tuples for partitions, let's discuss.

makalaaneesh: in this case, it could be `utils.StructMap[NameTuple, NameTuple]`

makalaaneesh (approved):
LGTM with one comment

yb-voyager/cmd/exportData.go
priyanshi-yb: This map is just used to put it in MSR or dbzm's config and no where else. So anyway then I had to iterate to get string->string map at two places. So I think it should be fine to get it like this itself from here?

PR #1402, opened 2024-04-05
Added support for json output format in report and status commands

priyanshi-yb:
Added automation step for offline and normal live migration to verify the output.


amit-yb (approved):
LGTM

PR #1403, opened 2024-04-08
Store the TableListExportedFromSource in MSR for debezium offline case as well

priyanshi-yb:
1. To avoid any errors or issues, we should store this table-list in msr in all the cases as we rely on this table-list at a few places now.
fixes - https://yugabyte.atlassian.net/browse/DB-10761


amit-yb (approved):
LGTM

PR #1405, opened 2024-04-11
Atomic write of schema file

makalaaneesh:
Atomically write the schema file for each table. 
Non-atomic writes were leading to situations where the importer notices the file(while trying to re-initialize the schema registry), tries to read the file, and fails with an EOF error.


amit-yb (approved):
LGTM

PR #1408, opened 2024-04-12
Dropping Log mining flush table in Oracle with start-clean 

priyanshi-yb:
Pipelines - 
1. https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2302
2. https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2305

### DIFF
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -805,7 +805,7 @@ func checkDataDirs() {
 		if err != nil {
 			utils.ErrExit("Failed to truncate tables in metadb: %s", err)
 		}
-		//For dropping VOYAGER_LOG_MINING_FLUSH_{migrationUUID} table in oracle due a bug in debezium
+		//For dropping VOYAGER_LOG_MINING_FLUSH_{migrationUUID} table in oracle on start-clean
 		err = source.DB().ClearMigrationState(migrationUUID, exportDir)
 		if err != nil {
 			utils.ErrExit("failed to clear migration state: %s", err)
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -805,10 +805,12 @@ func checkDataDirs() {
 		if err != nil {
 			utils.ErrExit("Failed to truncate tables in metadb: %s", err)
 		}
-		//For dropping VOYAGER_LOG_MINING_FLUSH_{migrationUUID} table in oracle on start-clean
-		err = source.DB().ClearMigrationState(migrationUUID, exportDir)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -805,6 +805,13 @@ func checkDataDirs() {
 		if err != nil {
 			utils.ErrExit("Failed to truncate tables in metadb: %s", err)
 		}
+		if changeStreamingIsEnabled(exportType) {
```
sanyamsinghal, line 808: shall we do the same: `ClearMigrationState()` for targetDB also in case of start-clean?

priyanshi-yb: we already do this in importData in `cleanImportState()` on startClean right? are you talking about anything else?

yb-voyager/cmd/exportData.go
```diff
@@ -152,6 +152,13 @@ func exportData() bool {
 		utils.ErrExit("Failed to connect to the source db: %s", err)
 	}
 	defer source.DB().Disconnect()
+	if changeStreamingIsEnabled(exportType) && bool(startClean) {
```
makalaaneesh, line 155: See if you can repurpose `checkDataDirs` to also handle this. Would be better to have all startClean logic in one place. 
- Change the name of the func to `ClearMigrationStateIfRequired` 
- Call it in exportData after init of sourceDB instead of in `exportDataCommandFn`. (double-check that it is okay to do this, and won't cause other issues)

makalaaneesh: Ironically, I'm running into a case with import-data where it doesn't seem possible to put all the startClean Logic in the same place. 
Lmk if you're running into anything similar here!

yb-voyager/cmd/exportData.go
makalaaneesh, line 808: Do we really need this check? 
`ClearMigrationState` should be able to handle both offline/live migration types

makalaaneesh (approved):
LGTM with a minor comment. Pls run all oracle tests on jenkins

yb-voyager/cmd/exportData.go
priyanshi-yb: Yes right now we need this check because we are only dropping the Log mining flush table in this function and that is also just for Oracle, so log mining flush can only be dropped in case of live migration. 
Ideally, the query should be appropriate to drop the table only if exists but the direct if exists clause doesn't exist in Oracle so need to modify the query. So I have changed the query to handle if an exists case after which this condition is not required

PR #1409, opened 2024-04-12
Fix preserving case in dataFileDescriptor for the partitions renaming 

priyanshi-yb:
fixes - https://yugabyte.atlassian.net/issues/?filter=12966

### DIFF
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -218,8 +218,8 @@ func getLeafPartitionsFromRootTable() map[string][]string {
 	}
 	tables := msr.TableListExportedFromSource
 	for leaf, root := range msr.SourceRenameTablesMap {
-		leafTable := sqlname.NewSourceNameFromQualifiedName(leaf)
-		rootTable := sqlname.NewSourceNameFromQualifiedName(root)
+		leafTable := sqlname.NewSourceNameFromQualifiedName((leaf))
+		rootTable := sqlname.NewSourceNameFromQualifiedName(getQuotedFromUnquoted(root))
 		leaf = leafTable.Qualified.MinQuoted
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/common.go
```diff
@@ -802,7 +809,7 @@ func renameTableIfRequired(table string) (string, bool) {
 	fromTable := tableName.Qualified.Unquoted
 
 	if renameTablesMap[fromTable] != "" {
-		table := sqlname.NewSourceNameFromQualifiedName(renameTablesMap[fromTable])
+		table := sqlname.NewSourceNameFromQualifiedName(getQuotedFromUnquoted(renameTablesMap[fromTable]))
```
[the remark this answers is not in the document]
priyanshi-yb: Yes, but I had to use sqlname.SourceName at other places for leaf partitions to avoid manual NameTuple creation so thought to use sqlname.SourceName in these few places to avoid extra lookup which is not required because we anyway lookup when we read dataFileDescriptor.

yb-voyager/cmd/common.go
```diff
@@ -218,8 +218,8 @@ func getLeafPartitionsFromRootTable() map[string][]string {
 	}
 	tables := msr.TableListExportedFromSource
 	for leaf, root := range msr.SourceRenameTablesMap {
-		leafTable := sqlname.NewSourceNameFromQualifiedName(leaf)
```
makalaaneesh, line 221: I forgot exactly what we discussed. Just to re-confirm, this function is only for display purposes, right?

makalaaneesh (approved):
LGTM

yb-voyager/cmd/common.go
priyanshi-yb: Yes right

PR #1411, opened 2024-04-12
 use go-json instead of encoding/json

makalaaneesh:
We used to use go-json as it is more performant compared to encoding/json. 
This change got reverted when we introduced manual unmarshalling.


amit-yb (approved):
LGTM

PR #1412, opened 2024-04-12
Create metadb entry before queue segment file to avoid race condition. 

makalaaneesh:
We need to create entry in metadb before creating file to avoid edge case where importer finds the file but cannot find the corresponding entry in metadb

https://yugabyte.atlassian.net/browse/DB-10838


amit-yb (approved):
LGTM

PR #1415, opened 2024-04-15
Bug fixes for UI/UX issues introduced due to name registry implementation

priyanshi-yb:
Fixes https://yugabyte.atlassian.net/browse/DB-10832 -
1. Display issues while printing table list.
2. Don't throw errors while running `import data status` before `import data`. Just returning empty output.


amit-yb (approved):
LGTM

PR #1421, opened 2024-04-16
Modify CREATE TABLE DDLs to apply schema recommendations during export schema

sanyamsinghal:


yb-voyager/src/utils/utils.go
```diff
@@ -216,6 +216,25 @@ func ParseSchemaAnalysisReport(jsonString string) SchemaReport {
 	return report
 }
 
+func ParseJsonReportFile[T any](path string) (*T, error) {
```
amit-yb, line 219: Hmm... we already have `utils.JsonFile[T].Load(obj *T) error` for this purpose.

yb-voyager/cmd/exportSchema.go
```diff
@@ -167,22 +191,114 @@ func clearSchemaIsExported() {
...
+func applyColocatedVsShardedTableRecommendation(shardingReport *migassessment.ShardingReport) error {
+	filePath := utils.GetObjectFilePath(schemaDir, "TABLE")
+	if !utils.FileOrFolderExists(filePath) {
+		log.Warnf("required schema file %s does not exists, returning without applying the recommendations", filePath)
+		return nil
+	}
+
+	log.Infof("applying colocated vs sharded table recommendation")
+
+	modifiedFilePath := utils.AddSuffixToFilePath(filePath, "modified")
```
amit-yb, line 248: It will be better to change the original file name to something like `<existing-file-name>.orig` and then create a new file with the `<existing-file-name>` containing the modified schema. That way, the code after the `export schema` (e.g. analyze-schema doesn't have to worry about two kinds of files. They will continue working as it is.).

amit-yb (approved):
LGTM

yb-voyager/cmd/exportSchema.go
sanyamsinghal: good idea, this way i don't need to do any changes on the import side.

yb-voyager/src/utils/utils.go
sanyamsinghal: oh ok, i was not aware of that.

PR #1425, opened 2024-04-17
Synchronized access to metadatadbconn 

makalaaneesh:
connection object to metadataDB is being accessed concurrently by two threads. This is unsafe. 
Adding synchronized blocks to avoid concurrent use.


amit-yb (approved):
LGTM

PR #1426, opened 2024-04-17
Added cache to store the best matching column names in attribute name registry

priyanshi-yb:
fixes - https://yugabyte.atlassian.net/browse/DB-10933

1. https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2335/
2. https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2336/

### DIFF
--- a/yb-voyager/src/tgtdb/attr_name_registry.go
+++ b/yb-voyager/src/tgtdb/attr_name_registry.go
@@ -12,20 +12,20 @@ import (
 )
 
 type AttributeNameRegistry struct {
-	dbType    string
-	tdb       TargetDB
-	tconf     *TargetConf
-	attrNames *utils.StructMap[sqlname.NameTuple, []string]
+	dbType                  string
+	tdb                     TargetDB
+	tconf                   *TargetConf
+	attrNames               *utils.StructMap[sqlname.NameTuple, []string]
 	bestMatchingColumnCache *utils.StructMap[sqlname.NameTuple, map[string]string]
-	mu        sync.Mutex
+	mu                      sync.Mutex
 }
 
 func NewAttributeNameRegistry(tdb TargetDB, tconf *TargetConf) *AttributeNameRegistry {
 	return &AttributeNameRegistry{
-		dbType:    tconf.TargetDBType,
-		tdb:       tdb,
-		tconf:     tconf,
-		attrNames: utils.NewStructMap[sqlname.NameTuple, []string](),
+		dbType:                  tconf.TargetDBType,
+		tdb:                     tdb,
+		tconf:                   tconf,
+		attrNames:               utils.NewStructMap[sqlname.NameTuple, []string](),
 		bestMatchingColumnCache: utils.NewStructMap[sqlname.NameTuple, map[string]string](),
 	}
 }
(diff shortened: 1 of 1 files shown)


yb-voyager/src/tgtdb/attr_name_registry.go
```diff
@@ -12,45 +12,60 @@ import (
...
+		tdb:                     tdb,
+		tconf:                   tconf,
+		attrNames:               utils.NewStructMap[sqlname.NameTuple, []string](),
+		bestMatchingColumnCache: utils.NewStructMap[sqlname.NameTuple, map[string]string](),
 	}
 }
 
 func (reg *AttributeNameRegistry) QuoteAttributeName(tableNameTup sqlname.NameTuple, columnName string) (string, error) {
 	var err error
+	reg.mu.Lock()
```
makalaaneesh, line 35: This function would be called by every goroutine for every event, so ideally, we should not lock here. Locking inside the `!ok` block is okay because that will run only once per table. (first event)

yb-voyager/src/tgtdb/attr_name_registry.go
```diff
@@ -12,45 +12,60 @@ import (
...
+			return "", fmt.Errorf("get list of table attributes: %w", err)
 		}
-
-		reg.mu.Unlock()
+		reg.attrNames.Put(tableNameTup, targetColumns)
 	}
-	c, err := reg.findBestMatchingColumnName(columnName, targetColumns)
-	if err != nil {
-		return "", fmt.Errorf("find best matching column name for %q in table %s: %w", columnName, tableNameTup, err)
+	bestMatchingColumnMapForTuple, ok := reg.bestMatchingColumnCache.Get(tableNameTup)
```
makalaaneesh, line 46: It would be cleaner to have this logic abstracted away inside the `findBestMatchingColumnName` function. 
That way the API is clean, the caller just calls `findBestMatchingColumnName` . Whether a cache is used internally or not is not the concern of the caller, it's just an implementation detail. 



```
func findBestMatchingColumnName(){
// if present in cache, return. 
// if not, do the actual matching, put it in cache, and return 
}
```

For example, in python, decorators are used to do this quite nicely. https://wiingy.com/learn/python/memoization-using-decorators-in-python/

makalaaneesh: btw, it might make sense to memoize the result of `QuoteAttributeName` func instead of the `findBestMatchingColumnName` . (you've sort of done that, but you could basically check in cache as the very first line of the function; before you even look for all attr names, etc )

priyanshi-yb: Yeah, makes sense. In Go, there is nothing similar to decorators I believe but can be implemented with separate wrapper functions. Made the change. Please review it and let me know.

makalaaneesh (approved):
LGTM. Pls ask @amit-yb to also take a look once.

yb-voyager/src/tgtdb/attr_name_registry.go
```diff
@@ -6,51 +6,67 @@ import (
...
-			}
-			reg.attrNames.Put(tableNameTup, targetColumns)
+		targetColumns, err = reg.tdb.GetListOfTableAttributes(tableNameTup)
+		log.Infof("columns of table %s in target db: %v", tableNameTup, targetColumns)
+		if err != nil {
+			return "", fmt.Errorf("get list of table attributes: %w", err)
 		}
+		reg.attrNames.Store(tableNameTup.ForKey(), targetColumns)
+	}
+	return reg.withCacheFindBestMatchingColumnName(columnName, targetColumns, tableNameTup)
```
makalaaneesh, line 46: You can lookup from cache at the start of this function (without even having to deal with getting list of table attributes, since you don't really need that for cache lookup). Similar to what amit did in https://github.com/yugabyte/yb-voyager/pull/1433/files

priyanshi-yb:
Closing this as the problem is solved here https://github.com/yugabyte/yb-voyager/pull/1433 in a better way

PR #1427, opened 2024-04-17
Updated the installer script to download tagged release gather-assessment-data tarball during voyager installation

sanyamsinghal:


amit-yb (approved):
LGTM

PR #1431, opened 2024-04-18
Minor UI/UX issues when the status/report commands run before the actual migration commands

priyanshi-yb:


yb-voyager/cmd/importDataStatusCommand.go
```diff
@@ -281,7 +271,7 @@ func prepareRowWithDatafile(dataFile *datafile.FileEntry, state *ImportDataState
 		status = "MIGRATING"
 	}
 	row := &tableMigStatusOutputRow{
-		TableName:          dataFileNt.ForMinOutput(),
+		TableName:          dataFileNt.ForKey(),
```
makalaaneesh, line 274: Curious, why?

makalaaneesh (approved):
LGTM

yb-voyager/cmd/importDataStatusCommand.go
priyanshi-yb: I have removed the importerRole from InitNameRegistry, in this command to avoid all the failures to register YB names, so ForMinOutput won't work as CurrentName can't be set hence I thought it is fine to use this as we are also using it in get data-migration-report.

PR #1432, opened 2024-04-18
Added check to disable default postgresql module if centos/rhel version is higher than 8 in the installer script

ShivanshGahlot:


amit-yb (approved):
LGTM

PR #1439, opened 2024-04-19
Introduced a check in yb-voyager-pg-gather-assessment-data.sh for track_counts to check if read/write stats are being tracked or not, required for calculating reads_per_second and writes_per_second stats

sanyamsinghal:


amit-yb (approved):
LGTM

PR #1442, opened 2024-04-23
Bug fix: Gather Metadata not collecting info for multiple schema if provided

sanyamsinghal:
- In the scripts it was expecting comma-separated but go code was sending pipe-separated, using pipe-separate everywhere to avoid confusion


amit-yb (approved):
LGTM

PR #1444, opened 2024-04-24
Added asssess migration to automation

shubham-yb:


migtests/tests/pg/datatypes/validate
```diff
@@ -46,9 +46,22 @@ EXPECTED_SUM_OF_COLUMN = {
...
+	'datatypes2': 't',
+	'datetime_type2': 't', 
+	'null_and_default' : 't',
+	'decimal_types': 't',
+}
+
 def migration_completed_checks_yb():
 	print("Running tests on YB")
 	yb.run_checks(migration_completed_checks)
+	# Commenting the validation for now due to the large sized source database requirements.
```
makalaaneesh, line 62: Didn't understand why we need to comment this?
Yes, all tables will be determined to be colocated, but that's what you're asserting as well, right?

makalaaneesh (approved):
LGTM with one small question.

migtests/tests/pg/datatypes/validate
shubham-yb: Yes I was confused as to what to do here. 
Removing the comment.

.github/workflows/pg-migtests.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   run-pg-migration-tests:
     strategy:
       matrix:
-        version: [2.21.0.0-b545, 2.16.9.0-b67, 2.18.7.0-b30, 2.20.2.2-b1]
+        version: [2.21.0.0-b545, 2.18.7.0-b30, 2.20.2.2-b1]
```
priyanshi-yb, line 13: Shall we do this change in a separate PR for both the workflows MySQL and PG?
Or 
Do it for both the workflows and mention this in the commit-msg to the main branch and also the reason for removing this version (end of maintenance).

priyanshi-yb (approved):
LGTM

.github/workflows/pg-migtests.yml
shubham-yb: Creating a separate PR.

migtests/tests/pg/datatypes/validate
```diff
@@ -46,9 +46,22 @@ EXPECTED_SUM_OF_COLUMN = {
 EXPECTED_ENUM_VALUES = ['mon', 'tue', 'wed', 'thu', 'fri', 'sat', 'sun']
 EXPECTED_BYTE_TYPE_VALUES = ['3031303130', '2d61626364', '34343538']
 
+#Due to the large sized source database requirements all tables will come up as colocated for now
```
sanyamsinghal, line 49: yeah, we need to find a way so that those size limits in ExperimentDB can be modified for the test case.
Maybe add a hidden flag and provide a updated ExpDB

migtests/tests/pg/datatypes/validate
```diff
@@ -46,9 +46,22 @@ EXPECTED_SUM_OF_COLUMN = {
...
+EXPECTED_COLOCATION_OF_TABLES = {
```
sanyamsinghal, line 50: This might not be useful as such in future. Since ideally we would want to fetch this list of colocated/sharded tables from the generated assessmentReport.json and use that list in our validations.

sanyamsinghal (approved):

migtests/tests/pg/datatypes/validate
shubham-yb: Yes that would be the best case scenario.

migtests/tests/pg/datatypes/validate
shubham-yb: We just need to know which table has which recommendation right? This won't change for every run?
I was thinking of adding a separate validation just for assessmentReport.json in the future.

sanyamsinghal: Nope this can change based on the any change in sizing algorithm in future. That change might not be very frequent. 
But the reports are the source of truth, and we should extract that from it.

PR #1447, opened 2024-04-25
Use .tmp suffix instead of /tmp for schema files generated by debezium

makalaaneesh:
Placing the tmp file in /tmp causes AtomicMoveException in docker (because of having to move across different volumes. Therefore, we place the tmp file in the same directory as that of the final file we want to write.


amit-yb (approved):
LGTM

makalaaneesh:
https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/2395/pipeline/

PR #1448, opened 2024-04-25
Minor fixes in assessment script

makalaaneesh:
- Saving absolute path in SCRIPT_DIR because we cd later and relative paths won't work while listing files of pattern `*.psql`  https://stackoverflow.com/a/246128
- in the example, adding PGPASSWORD env var


amit-yb (approved):
LGTM

PR #1453, opened 2024-04-29
Updated setup script to check for the existence of YB_VOYAGER_SEND_DIAGNOSTICS env var in developers' machine env

sanyamsinghal:


sanyamsinghal:
callhome service now ignores the internal ips

PR #1454, opened 2024-04-30
Export Data: Retrieve unique key info before starting pg dump

makalaaneesh:
In PG live migration flow, we currently 
1. open a connection
1. gather data from server
1. run pg dump
1. gather some more information from server
1. start debezium. 

Because pg_dump can take a long time, we are seeing some connection failures (`broken pipe` for example) in step 4. 
Modifying the logic so that step 4 is done in step2, so that we don't have to use the connection again after pg_dump. 

This is a temporary fix. Meanwhile, we should find out how to properly handle these type of errors either via config/retries/etc.

### DIFF
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -239,6 +239,7 @@ func exportData() bool {
 			log.Errorf("Failed to prepare dbzm config: %v", err)
 			return false
 		}
+		saveTableToUniqueKeyColumnsMapInMetaDB(finalTableList)
 		if source.DBType == POSTGRESQL && changeStreamingIsEnabled(exportType) {
 			// pg live migration. Steps are as follows:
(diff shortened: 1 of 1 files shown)


sanyamsinghal (approved):
LGTM. Please add a one-liner reason why this change in the commit msg. 
In all our commits to main, we should write: `what` +`why`

makalaaneesh:
> LGTM. Please add a one-liner reason why this change in the commit msg. In all our commits to main, we should write: `what` +`why`

Added PR description 👍

makalaaneesh:
Jenkins: https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/2417/pipeline/

PR #1456, opened 2024-04-30
Introduced a check in 'import schema' to ensure if target db is colocated or not

sanyamsinghal:
- gives a YES/NO prompt to users if they want to continue or not


yb-voyager/cmd/importSchema.go
```diff
@@ -118,6 +118,11 @@ func importSchema() {
 		installOrafceIfRequired(conn)
 	}
 
+	if !isYBDatabaseIsColocated(conn) && !utils.AskPrompt(fmt.Sprintf("\nWarning: Target DB '%s' is a non-colocated database, colocated tables can't be created in a non-colocated database.\n", tconf.DBName),
+		"Use a colocated databse if your schema contains colocated tables. Do you still want to continue") {
```
priyanshi-yb, line 122: colocated database (typo).

yb-voyager/cmd/importSchema.go
```diff
@@ -118,6 +118,11 @@ func importSchema() {
 		installOrafceIfRequired(conn)
 	}
 
+	if !isYBDatabaseIsColocated(conn) && !utils.AskPrompt(fmt.Sprintf("\nWarning: Target DB '%s' is a non-colocated database, colocated tables can't be created in a non-colocated database.\n", tconf.DBName),
```
makalaaneesh, line 121: This warning will not be correct if the colocated table list (as per assessment) is non-empty. Maybe we should add that as an additional condition?

makalaaneesh: + one nit: what Neeraj said in the meeting: if the --skip-recommendations was used, this would be redundant. 

Maybe we can handle this later (since it's just a warning, so it's harmless, user can always just continue)

priyanshi-yb (approved):
LGTM

yb-voyager/cmd/exportSchema.go
```diff
@@ -226,27 +225,32 @@ func applyMigrationAssessmentRecommendations() error {
...
+	report, err := ParseJSONToAssessmentReport(assessmentReportPath)
 	if err != nil {
 		return fmt.Errorf("failed to parse json report file %q: %w", assessmentReportPath, err)
 	}
 
-	err = applyColocatedVsShardedTableRecommendation(report.Sizing)
+	shardedTables, err := report.GetShardedTablesRecommendation()
 	if err != nil {
-		return fmt.Errorf("failed to apply colocated vs sharded table recommendation: %w", err)
+		log.Warnf("GetShardedTablesRecommendation: %v", err)
```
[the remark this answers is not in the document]
sanyamsinghal: if you look the `report.GetShardedTablesRecommendation()`
It returns error in case if shardedTables list is null, no other error case.

Is your point is why are we returning that as an error instead of just handing the null value here?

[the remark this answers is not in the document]
makalaaneesh: Oh okay. In that case, yes,  I'd suggest checking for nil value, instead of handling it as an error. (it's not really an error).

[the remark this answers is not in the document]
sanyamsinghal: make sense. will do the change.

yb-voyager/cmd/importSchema.go
sanyamsinghal: yeah, this warning is telling users to use colocated db only if tables are colocated...

yb-voyager/cmd/exportSchema.go
```diff
@@ -226,27 +225,37 @@ func applyMigrationAssessmentRecommendations() error {
...
+func applyShardedTablesRecommendation(shardedTables []string) error {
+	if shardedTables == nil {
+		log.Info("list of sharded tables is null hence all the tables are recommended as colocated")
+		return nil
+	}
+
 	filePath := utils.GetObjectFilePath(schemaDir, "TABLE")
 	if !utils.FileOrFolderExists(filePath) {
-		log.Warnf("required schema file %s does not exists, returning without applying the recommendations", filePath)
+		utils.PrintAndLog("Required schema file %s does not exists, returning without applying Colocated/Sharded Tables recommendation", filePath)
```
makalaaneesh, line 254: Under what circumstances would the table file not exist?

makalaaneesh (approved):
LGTM with one question/comment

yb-voyager/cmd/exportSchema.go
sanyamsinghal: in case there is no table on the source, just some non-table objects like UDF or something, which is rare and might not be the real world.

PR #1458, opened 2024-04-30
Bug fix[DB-11119]: Unsupported Datatypes info in assessment report is coming out to be NULL

sanyamsinghal:
- there was snakecase vs camelcase difference, the headers of csv(fetched from DB) had snakecase but we use camelcase for finally converting it to JSON

### DIFF
--- a/yb-voyager/src/migassessment/common.go
+++ b/yb-voyager/src/migassessment/common.go
@@ -19,9 +19,13 @@ import (
 	"encoding/csv"
 	"encoding/json"
 	"fmt"
-	log "github.com/sirupsen/logrus"
 	"net/http"
 	"os"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 )
 
 var AssessmentMetadataDir string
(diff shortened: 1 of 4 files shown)


yb-voyager/src/migassessment/common.go
```diff
@@ -55,7 +59,12 @@ func LoadCSVDataFile[T any](filePath string) ([]*T, error) {
 
 	for _, record := range records {
 		var tmplRec T
-		bs, err := json.Marshal(record)
+		camelCaseRecord := make(Record)
```
makalaaneesh, line 62: Can't we use json tags here? seems wrong to write this logic in a generic `LoadCSVDataFile` function.

sanyamsinghal: Json tags needs to be camel case format. We are following that for all our json reports.
But here that is the program since in the database(tablename and column names) are snake case formatted. cc @makalaaneesh

sanyamsinghal: Did the changes as discussed. cc @makalaaneesh

makalaaneesh (approved):
LGTM. Thanks for taking this up now itself!

PR #1459, opened 2024-05-02
Use conn pool instead of raw connections in PG/YB source DBs

makalaaneesh:
Recently, a customer ran into an error `write: broken pipe` when trying out live migration (export-data). This happened because we were making some metadata queries after snapshot step, by which time, the connection that we had opened earlier was automatically closed (because of server side settings - 
```
idle_in_transaction_session_timeout = 20000
statement_timeout = 120000
idle_session_timeout = 600000
```

To avoid this, we already moved the logic of metadata query to be called before the snapshot step. (in an older PR)
To add in additional layer of safety, we are making the following changes as part of this PR:
- use the sql.DB interface instead of managing connections ourselves. This is essentially a connection pool, which is concurrency-safe.
- Additionally, we set the SetMaxOpenConns to 1, just to cap the number of connections we open to the source database. We typically never need more than one connection. (we only query some metadata). 
- We also SetConnMaxIdleTime = 5 min. This is to ensure that the pool automatically discards connections if they are idle for longer than 5 min. It is still possible for the server to have lower thresholds for idle connections; this is just some arbitrary (and hopefully conservative) value that we are setting.


makalaaneesh:
https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/2437/pipeline/390/

makalaaneesh:
https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/2546/pipeline/

makalaaneesh:
https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/2556/pipeline/561/

https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/2557/pipeline/168/

yb-voyager/src/srcdb/mysql.go
```diff
@@ -46,6 +47,8 @@ func newMySQL(s *Source) *MySQL {
 
 func (ms *MySQL) Connect() error {
 	db, err := sql.Open("mysql", ms.getConnectionUri())
+	db.SetMaxOpenConns(1)
+	db.SetConnMaxIdleTime(5 * time.Minute)
```
sanyamsinghal, line 51: So what will be the behaviour? In case conn is idle for more than 5 minutes then it will close it and let's the code needs a connection after 6 or 7 minutes, will the pool make a new connection?

yb-voyager/src/srcdb/mysql.go
```diff
@@ -366,6 +369,7 @@ func (ms *MySQL) GetColumnToSequenceMap(tableList []sqlname.NameTuple) map[strin
 		if err != nil {
 			utils.ErrExit("Failed to query %q for auto increment column of %q: %s", query, table.String(), err)
 		}
+		defer rows.Close()
```
sanyamsinghal, line 372: good to log the error here:
```
defer func() {
     closeErr = rows.Close()
     if closeErr != nil {
         ........
    }
}()
```

yb-voyager/src/srcdb/mysql.go
```diff
@@ -376,6 +380,10 @@ func (ms *MySQL) GetColumnToSequenceMap(tableList []sqlname.NameTuple) map[strin
 			sequenceName := fmt.Sprintf("%s_%s_seq", tname, columnName)
 			columnToSequenceMap[qualifiedColumeName] = sequenceName
 		}
+		err = rows.Close()
+		if err != nil {
+			utils.ErrExit("close rows for table %s query %q: %s", table.String(), query, err)
+		}
```
sanyamsinghal, line 386: this is because we need this connection to be free for other parts of code to use?

sanyamsinghal: but wouldn't `sql.DB` package should manage in that case and create/maintain a new connection. And if thats the case you can add a warning to log instead of ErrExit??

sanyamsinghal (commented):
same comments for all the source db types.

yb-voyager/src/srcdb/mysql.go
makalaaneesh: Yes, the pool will automatically create a new connection if required

yb-voyager/src/srcdb/mysql.go
makalaaneesh: The connection is only released back to the pool after we call `rows.Close()` (https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/database/sql/sql.go;l=3424)  . In this case, in every loop iteration, we make a call to the db, so it becomes necessary to close the rows object before the next iteration

sanyamsinghal: yeah that i understand, but is `ErrExit` required to close the connection or we can just add a warning in the logs and move forward?

makalaaneesh: I think it would make sense to `ErrExit`, instead of just logging. 
The way I discovered this issue was that the export-data process was stuck (because of this function)
- In first iteration of the for loop, we access db (get a connection from the pool). We do not call rows.Close, and therefore do not return connection back to pool.
- In second iteration of the for loop, it gets stuck while trying to fetch a connection from pool because max open connections = 1.  

Therefore, I think it would make sense to error out.

sanyamsinghal: ok, cool. 
Also i think is initial step(before any data export), i was more concerned for cases where whole data and everything is exported but the command fails due to some unimportant error while closing the connection, which can be ignored/retried instead of error out.

sanyamsinghal (approved):

yb-voyager/src/srcdb/oracle.go
```diff
@@ -418,6 +437,10 @@ func (ora *Oracle) GetColumnToSequenceMap(tableList []sqlname.NameTuple) map[str
 			qualifiedColumnName := fmt.Sprintf("%s.%s", table.AsQualifiedCatalogName(), columnName)
 			columnToSequenceMap[qualifiedColumnName] = fmt.Sprintf("%s_%s_seq", tname, columnName)
 		}
+		err = rows.Close()
```
priyanshi-yb, line 440: If we have a defer function already for this, why do we need it explicitly? (please check all other places)

yb-voyager/src/srcdb/postgres.go
```diff
@@ -562,6 +600,10 @@ func (pg *PostgreSQL) GetColumnToSequenceMap(tableList []sqlname.NameTuple) map[
 			// quoting sequence name as it can be case sensitive - required during import data restore sequences
 			columnToSequenceMap[qualifiedColumnName] = fmt.Sprintf(`%s."%s"`, schemaName, sequenceName)
 		}
+		err = rows.Close()
```
priyanshi-yb, line 603: same as above comment.

yb-voyager/src/srcdb/oracle.go
makalaaneesh: We need it because we need to close rows object before the next iteration. Check this - https://github.com/yugabyte/yb-voyager/pull/1459#discussion_r1606332604

Defer will only run at the end of the function, so that is just a safety net in this case, really.

priyanshi-yb: Got it, thanks!

PR #1466, opened 2024-05-04
fix in sizing: get experiment DB path in func instead of declared at pkg level

makalaaneesh:
- Moving the variable `experimentDBPath` from global scope to a function. (declaring it globally causes it to be defined at pkg import time, but it depends on a variable `assessmentDir` which is only initialized at run time.


sanyamsinghal (approved):
LGTM

PR #1468, opened 2024-05-06
Removed 2.16 from automation

shubham-yb:
1. Removed 2.16 from the automation since the `colocated` clause is not supported and its reaching EOL.
2. Updated 2.20 version


priyanshi-yb (approved):
LGTM! 
Thanks for the quick PR.

PR #1469, opened 2024-05-06
Modified the CLI for assess-migration command to only mention about PostgreSQL db type

priyanshi-yb:
…SQL mentions

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -30,9 +30,9 @@ import (
 	"text/template"
 
 	"github.com/samber/lo"
-
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/cp"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/metadb"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/migassessment"
@@ -54,8 +54,8 @@ type UnsupportedFeature struct {
 
 var assessMigrationCmd = &cobra.Command{
 	Use:   "assess-migration",
-	Short: "Assess the migration from source database to YugabyteDB.",
-	Long:  `Assess the migration from source database to YugabyteDB.`,
+	Short: "Assess the migration from PostgreSQL source database to YugabyteDB.",
+	Long:  `Assess the migration from PostgreSQL source database to YugabyteDB.`,
 
 	PreRun: func(cmd *cobra.Command, args []string) {
 		validateSourceDBTypeForAssessMigration()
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -74,10 +74,51 @@ var assessMigrationCmd = &cobra.Command{
...
+	// TODO: All sensitive parameters can be taken from the environment variable
+	cmd.Flags().StringVar(&source.Password, "source-db-password", "",
+		"source password to connect as the specified user. Alternatively, you can also specify the password by setting the environment variable SOURCE_DB_PASSWORD. If you don't provide a password via the CLI, yb-voyager will prompt you at runtime for a password. If the password contains special characters that are interpreted by the shell (for example, # and $), enclose the password in single quotes.")
+
+	cmd.Flags().StringVar(&source.DBName, "source-db-name", "",
+		"source database name to be migrated to YugabyteDB")
+
+	cmd.Flags().StringVar(&source.Schema, "source-db-schema", "",
+		"source schema name(s) to export\n"+
+			`Note: in case of multiple schemas, use a comma separated list of schemas: "schema1,schema2,schema3"`)
```
sanyamsinghal, line 99: you can make use of `registerCommonSourceDBConnFlags()` for the above args

sanyamsinghal (approved):
LGTM with a minor comment

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: +1. If it's only about the mentions of "oracle/mysql", see if it's possible to re-declare those flags or override their help msgs _after_ having called `registerSourceDBConnFlags` ?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -54,8 +54,8 @@ type UnsupportedFeature struct {
 
 var assessMigrationCmd = &cobra.Command{
 	Use:   "assess-migration",
-	Short: "Assess the migration from source database to YugabyteDB.",
-	Long:  `Assess the migration from source database to YugabyteDB.`,
+	Short: "Assess the migration from PostgreSQL source database to YugabyteDB.",
```
makalaaneesh, line 57: nit: let's still keep it `source` , but we can frame it as `source (PostgreSQL) database ...` ? 
That way, it would be indicative, that we plan to support other database types in the future.

sanyamsinghal: good point.

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: I think its better to keep it here as this is just for this command otherwise, I will have to add conditions in that function and duplicate the messaging for most of the flags.

priyanshi-yb: will try to do this refactoring thing in parallel to RC in a separate PR and merge this PR for the time being.

PR #1478, opened 2024-05-07
Avoid count(*) in assessment metadata db

makalaaneesh:
- Currently we do not use row counts of tables in sizing logic, and count(*) can be time-consuming for large tables, so commenting it out for now.


sanyamsinghal (approved):

shaharuk-yb (approved):
LGTM. We are not using the row count at the moment.

PR #1480, opened 2024-05-07
Miscellaneous changes in assess-migration command

priyanshi-yb:
1. Changed the SchemaName in all the reports to SchemaName(s)/SchemaNames and in multiple schemas case it is a list of schemas
2. Marked `--source-db-user`, `--source-db-schema`, and `--source-db-name` as mandatory flags.
(fixes - https://yugabyte.atlassian.net/browse/DB-11212)
3.  Added a flag --iops-capture-interval for configuring the sleep between IOPS metadata collection ( fixes - https://yugabyte.atlassian.net/browse/DB-11211 )
4. Preventing the `source-*` flags to be passed in case `--assessment-metadata-dir` is provided.
(fixes - https://yugabyte.atlassian.net/browse/DB-11188)


yb-voyager/src/srcdb/data/gather-assessment-metadata/postgresql/yb-voyager-pg-gather-assessment-metadata.sh
```diff
@@ -48,15 +52,27 @@ if [ "$1" == "--help" ]; then
 fi
 
 # Check if all required arguments are provided
-if [ "$#" -ne 3 ]; then
-    echo "Usage: $SCRIPT_NAME <pg_connection_string> <schema_list> <assessment_metadata_dir>"
-    exit 1
+if [ "$4" != "" ]; then
+    if [ "$#" -ne 4 ]; then
+        echo "Usage: $SCRIPT_NAME <pg_connection_string> <schema_list> <assessment_metadata_dir>"
```
makalaaneesh, line 57: In this usage, (and everwhere else even in examples..), we can add the sleep interval, right?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -64,6 +65,10 @@ var assessMigrationCmd = &cobra.Command{
 		validatePortRange()
 		validateSSLMode()
 		validateAssessmentMetadataDirFlag()
+		cmd.MarkFlagRequired("source-db-user")
```
makalaaneesh, line 68: These Flags can't be marked as mandatory outright, because user can also pass `assessment-metadata-dir` , in which case user should not be allowed to pass these flags. https://yugabyte.atlassian.net/browse/DB-11188

We might have to validate the args manually...

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -127,6 +132,10 @@ func init() {
 	assessMigrationCmd.Flags().StringVar(&assessmentMetadataDirFlag, "assessment-metadata-dir", "",
 		"Directory path where assessment metadata like source DB metadata and statistics are stored. Optional flag, if not provided, "+
 			"it will be assumed to be present at default path inside the export directory.")
+
+	assessMigrationCmd.Flags().Int64Var(&sleepIntervalForIOPS, "sleep-interval", 120,
```
makalaaneesh, line 136: I think `sleep-interval` is unclear for an arg name. It's not obvious why assess-migration would need to sleep.  We might need to think of a more specific argument name. Let's discuss this in standup!

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: Oh yes, thanks for pointing it out. will check for that flag first.

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: Yes, I also couldn't think of better alternative.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -927,7 +927,7 @@ func generateHTMLReport(Report utils.SchemaReport) string {
 	//Broad details
 	htmlstring := "<html><body bgcolor='#EFEFEF'><h1>Database Migration Report</h1>"
 	htmlstring += "<table><tr><th>Database Name</th><td>" + Report.SchemaSummary.DBName + "</td></tr>"
-	htmlstring += "<tr><th>Schema Name</th><td>" + Report.SchemaSummary.SchemaName + "</td></tr>"
+	htmlstring += "<tr><th>Schema Name(s)</th><td>" + Report.SchemaSummary.SchemaNames + "</td></tr>"
```
sanyamsinghal, line 930: did you verify the HTML? Since the variable `SchemaNames` here is a list now, it might need to handled with the `range` clause. You can refer to the `assessment.template` file for that

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -986,7 +986,7 @@ func generateTxtReport(Report utils.SchemaReport) string {
 	txtstring += "| Database Migration Report |\n"
 	txtstring += "+---------------------------+\n"
 	txtstring += "Database Name\t" + Report.SchemaSummary.DBName + "\n"
-	txtstring += "Schema Name\t" + Report.SchemaSummary.SchemaName + "\n"
+	txtstring += "Schema Name(s)\t" + Report.SchemaSummary.SchemaNames + "\n"
```
sanyamsinghal, line 989: will this look better:
```
txtstring += "Schema Name(s)\t" + strings.Join(Report.SchemaSummary.SchemaNames, ", ") + "\n"
```

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: Lets rename it to `--iops-measurement-interval` until we have something better.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -127,6 +132,10 @@ func init() {
...
+		"Sleep interval to be used to calculate IOPS on source database in seconds (default: 120)")
```
sanyamsinghal, line 137: correspondingly update the help message

yb-voyager/src/utils/commonVariables.go
```diff
@@ -78,11 +78,11 @@ type SchemaReport struct {
 }
 
 type SchemaSummary struct {
-	DBName     string     `json:"DbName,omitempty"`
-	SchemaName string     `json:"SchemaName,omitempty"`
-	DBVersion  string     `json:"DbVersion,omitempty"`
-	Notes      []string   `json:"Notes,omitempty"`
-	DBObjects  []DBObject `json:"DatabaseObjects"`
+	DBName      string     `json:"DbName,omitempty"`
+	SchemaNames string     `json:"SchemaName(s),omitempty"`
```
sanyamsinghal, line 82: i was expecting you to use `[]string` here. Since that will be a better representation, and also will be better for outside json parser(unaware of voyager logic) such that they will know it can contain multiple schemas not just one.

yb-voyager/src/srcdb/data/gather-assessment-metadata/postgresql/yb-voyager-pg-gather-assessment-metadata.sh
sanyamsinghal, line 1: @priyanshi-yb can you please remove the password from the URI at line 28-29.
```
'postgresql://username:password@hostname:port/dbname'
```
We don't recommend embedding the password in connection URI

yb-voyager/src/srcdb/data/gather-assessment-metadata/postgresql/yb-voyager-pg-gather-assessment-metadata.sh
```diff
@@ -48,15 +52,27 @@ if [ "$1" == "--help" ]; then
...
+        exit 1
+    fi
+    sleep_interval=$4
+else 
+    sleep_interval=120
+    if [ "$#" -ne 3 ]; then
+        echo "Usage: $SCRIPT_NAME <pg_connection_string> <schema_list> <assessment_metadata_dir>"
+        exit 1
+    fi
 fi
```
sanyamsinghal, line 67:
```
if [ "$#" -lt 3 ]; then
    echo "Usage: $0 <pg_connection_string> <schema_list> <assessment_metadata_dir> [sleep_interval]"
    exit 1
elif [ "$#" -gt 4 ]; then
    echo "Usage: $0 <pg_connection_string> <schema_list> <assessment_metadata_dir> [sleep_interval]"
    exit 1
fi

# Set default sleep interval
sleep_interval=120

# Override default sleep interval if a fourth argument is provided
if [ "$#" -eq 4 ]; then
    sleep_interval=$4
    echo "Using sleep interval: $sleep_interval seconds"
fi
```

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -63,7 +65,21 @@ var assessMigrationCmd = &cobra.Command{
...
+			validateAssessmentMetadataDirFlag()
+			cmd.Flags().VisitAll(func(f *pflag.Flag) {
+				if f.Changed {
+					if strings.Contains(strings.ToLower(f.Name), "source") && f.Name != "source-db-type" {
+						utils.ErrExit("No need of `--source-*` connection related flags as `--assessment-metadata-dir` is provided.\nPlease re-run the command without these flags")
+					}
+				}
+			})
+		} else {
+			cmd.MarkFlagRequired("source-db-user")
```
makalaaneesh, line 78: We are essentially marking the flag as required at runtime, not compile time, here, essentially, right? 
How does that work? Does this throw a validation error here then?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -63,7 +65,21 @@ var assessMigrationCmd = &cobra.Command{
 		validateSourceSchema()
 		validatePortRange()
 		validateSSLMode()
-		validateAssessmentMetadataDirFlag()
+		if cmd.Flags().Changed("assessment-metadata-dir") {
+			validateAssessmentMetadataDirFlag()
+			cmd.Flags().VisitAll(func(f *pflag.Flag) {
+				if f.Changed {
```
makalaaneesh, line 71: While this might work, I'd recommend better list out all the flags that we want to check explicitly :) It'd be less confusing and more clear.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -63,7 +65,21 @@ var assessMigrationCmd = &cobra.Command{
...
+					if strings.Contains(strings.ToLower(f.Name), "source") && f.Name != "source-db-type" {
+						utils.ErrExit("No need of `--source-*` connection related flags as `--assessment-metadata-dir` is provided.\nPlease re-run the command without these flags")
```
makalaaneesh, line 73: Let's phrase it as "cannot pass source-* flags when `-assessment-metadata-dir` is passed" ?

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: Yes, it throws this error 
```
Error: required flag(s) "source-db-schema" not set
```

priyanshi-yb: > We are essentially marking the flag as required at runtime, not compile time, here, essentially, right?

Yes

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: I thought this might be better than listing all the `source-*` flags as those are in the majority and we need to just exclude `--source-db-type` only.

makalaaneesh: (btw, this is a nit, don't block PR on this :D just discussing because of interest)
As a general thumb rule, [explicit tends to be better than implicit.](https://peps.python.org/pep-0020/) The less complicated the logic, the better.  The code becomes more readable as a result. I think one of golang's goals also is simplicity. 

In this example, if you list all the flags here, you immediately find out what all flags are being disallowed. The code becomes very clear as opposed to understanding the condition (containing "source") and then going to the place in the code where the flags are declared. 

Ofc you can argue from the other side - If we add more flags in the future, this logic will automatically incorporate it (but also maybe it might incorrectly block a flag from being passed, and we'll have to come change it.. )

There's no right answer..

priyanshi-yb: hmm okay, makes sense in terms of readability. Will make the change. Thanks!

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -63,7 +65,21 @@ var assessMigrationCmd = &cobra.Command{
...
+					}
+				}
+			})
+		} else {
+			cmd.MarkFlagRequired("source-db-user")
+			cmd.MarkFlagRequired("source-db-name")
+			//Update this later as per db-types TODO
+			cmd.MarkFlagRequired("source-db-schema")
```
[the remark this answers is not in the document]
priyanshi-yb: Yeah, but that was doing some based on db-type so thought we should add that when we start supporting other types.

makalaaneesh (approved):
LGTM

sanyamsinghal (approved):
LGTM, thanks for resolving all the issues in one go.

PR #1484, opened 2024-05-08
Changed installation method of OIC in ubuntu in the installer script

ShivanshGahlot:
Docker build uses an ubuntu image and installs yb-voyager using the installer script on it. The docker build was failing because alien command was failing to convert the sqlplus OIC from rpm to deb. So, using our already converted OIC clients from our own apt repository.


makalaaneesh (approved):
LGTM. Pls run the docker pipeline

makalaaneesh:
@ShivanshGahlot pls also add PR description to explain why we're making this change.

PR #1485, opened 2024-05-08
Modified Migration Assessment HTML Report

sanyamsinghal:
To make the tables scrollable for the sections where a large list of object names is possible

- also added checks in Unsupported Features and Unsupported Datatypes sections to non-empty subsections and display relevant text if nothing is unsupported

-- also moved the Recommendation section above from last

For example this is how the scrollable sections will look:
image
image


shaharuk-yb (approved):
LGTM

PR #1498, opened 2024-05-09
Added check while installing OIC's to skip installation if a higher or equal version is already installed.

ShivanshGahlot:
The installer script installation was failing on our jenkins pipeline as the OICs are already installed over there. It was failing with a downgraded error. To fix that a check has been introduced to skip the installation of individual OICs if their higher or equal version is already installed.


installer_scripts/install-yb-voyager
```diff
@@ -843,16 +843,36 @@ ubuntu_install_oracle_instant_clients() {
...
 
+ubuntu_install_oic() {
+	local package_name="$1"
+    
+    # Check if package is installed
+    if dpkg -l | grep -q "$package_name"; then
+        # Extract installed version
+        installed_version=$(dpkg -l | grep "$package_name" | awk '{print $3}')
+        # Check if installed version is greater than or equal to 21.5.0.0.0-1
+        if dpkg --compare-versions "$installed_version" ">=" "21.5.0.0.0-1"; then
```
sanyamsinghal, line 864: nit: define a variable for `"21.5.0.0.0-1"`

PR #1499, opened 2024-05-09
Introduced pg_query module for parsing and updating the schema as per assessment report recommendations

sanyamsinghal:
- it can handle/update existing 'WITH' clauses to add/append 'COLOCATION = true'
- also handles clauses like ON COMMIT and TABLESPACE if present after the WITH clause

Some key points about minimizing the risk/failures:
1. We are using this SQL parser only for DDLs where we need to apply sharding recommendations(i.e. sharded tables), rest of the tables' DDLs are untouched.
2. In the worst case, if any error occurs due to this new module, we are falling back to the original DDL, and in that case, we will inform the user also about the failure to apply the recommendation for this table, and they can do the change manually.

Module used: https://github.com/pganalyze/pg_query_go


yb-voyager/cmd/exportSchema.go
```diff
@@ -231,11 +231,12 @@ func applyMigrationAssessmentRecommendations() error {
...
 	}
 
-	shardedTables, err := report.GetShardedTablesRecommendation()
+	colocatedTables, err := report.GetColocatedTablesRecommendation()
+	fmt.Printf("colocatedTables: %v\n", colocatedTables)
 	if err != nil {
 		return fmt.Errorf("failed to fetch sharded tables recommendation: %w", err)
 	} else {
-		err := applyShardedTablesRecommendation(shardedTables)
+		err := applyShardedTablesRecommendation(colocatedTables)
```
makalaaneesh, line 239: I assume this is just for testing purposes?

yb-voyager/cmd/exportSchema.go
```diff
@@ -260,23 +261,29 @@ func applyShardedTablesRecommendation(shardedTables []string) error {
...
+			We can pass the whole .sql file as a string also to pg_query.Parse() all the statements at once.
+			But avoiding that also specially for cases where the SQL syntax can be invalid
+		*/
+		// parsing only the Recommended Sharded Tables
+		if sqlInfo.objName != "" && slices.Contains(shardedTables, sqlInfo.objName) {
+			log.Infof("applying sharding recommendation for table=%q", sqlInfo.objName)
+			modifiedSqlStmt, err = applyShardingRecommendation(&sqlInfo)
+			if err != nil {
+				log.Errorf("failed to apply sharding recommendation for table=%q: %v", sqlInfo.objName, err)
+				fmt.Printf("unable to apply sharding recommendation for table=%q, ignoring the recommendation and continue...\n", sqlInfo.objName)
```
makalaaneesh, line 279: Shall we also mention "Pls manually add the clause "WITH COLOCATION=false to table <xyz>" ?

yb-voyager/cmd/exportSchema.go
```diff
@@ -245,7 +246,7 @@ func applyMigrationAssessmentRecommendations() error {
 	return nil
 }
 
-func applyShardedTablesRecommendation(shardedTables []string) error {
+func applyShardedTablesRecommendation(shardedTables []string) (err error) {
```
makalaaneesh, line 249: why give the return type a name? I generally feel like this approach is generally error prone. (especially when it comes to defer funcs where we might want to ignore the error)

yb-voyager/cmd/exportSchema.go
```diff
@@ -309,32 +316,75 @@ func applyShardedTablesRecommendation(shardedTables []string) error {
...
+	if err != nil {
+		return stmt, fmt.Errorf("error parsing the stmt-%s: %v", stmt, err)
+	}
 
-func prependSpacing(sql string, condition bool) string {
-	if condition {
-		return "\n\n" + sql
+	if len(parseTree.Stmts) == 0 {
+		log.Infof("parse tree is empty, returning")
+		return stmt, nil
```
makalaaneesh, line 338: shouldn't we return a non-nil error here?

yb-voyager/cmd/exportSchema.go
```diff
@@ -309,32 +316,75 @@ func applyShardedTablesRecommendation(shardedTables []string) error {
...
 	}
-	return sql
+
+	// Access the first statement directly
+	createStmtNode, ok := parseTree.Stmts[0].Stmt.Node.(*pg_query.Node_CreateStmt)
+	if !ok { // return the original sql if it's not a CreateStmt
+		log.Warnf("stmt=%s is not createTable as per the parse tree, expected tablename=%s", stmt, sqlInfo.objName)
+		return stmt, nil
```
makalaaneesh, line 345: Again here also we should return non-nil error, I feel. Because the assumption of this function is that it will be a CREATE TABLE stmt

makalaaneesh (commented):
Nice!

yb-voyager/cmd/exportSchema.go
sanyamsinghal: correct

yb-voyager/cmd/exportSchema.go
sanyamsinghal: makes sense. I will add

yb-voyager/cmd/exportSchema.go
sanyamsinghal: hmm i give it for specific use cases. Like in this case, i have a common variable defined `modifiedSqlStmt` used under multiple if-else conditions,like this:
```
if ....
         modifiedSqlStmt, err = func1()
else ...
         modifiedSqlStmt, err = func2()
```

so i would have to define err variable before hand. I could do when required - `var err error`

But for the defer case you mentioned, i agree its needs to be taken care of but that anyways we should use a separate variable in that case than the normal errors one.

yb-voyager/cmd/exportSchema.go
sanyamsinghal: `ALTER TABLE tablename ....`
These statements are non-create table, but might have the same name which matches the sharded tables list.

So just to be safe i am not returning any error from this function, just logging it and returning the original stmt

yb-voyager/cmd/exportSchema.go
sanyamsinghal: we don't know yet, when will this be empty. maybe empty strings, some valid syntax which is not covered by the library!

sanyamsinghal:
Reverted last 2 commits which were just to test the automation tests
> - all the tests in automation passed except 2-3 due to some check, verified them manually

yb-voyager/cmd/exportSchema.go
makalaaneesh: Got it. so, all stmts containing the table name are passed to this func

yb-voyager/cmd/exportSchema.go
```diff
@@ -260,23 +260,29 @@ func applyShardedTablesRecommendation(shardedTables []string) error {
...
+			but due to time constraint this module can't be tested thoroughly so relying on the existing as much as possible
+
+			We can pass the whole .sql file as a string also to pg_query.Parse() all the statements at once.
+			But avoiding that also specially for cases where the SQL syntax can be invalid
+		*/
+		// parsing only the Recommended Sharded Tables
+		if sqlInfo.objName != "" && slices.Contains(shardedTables, sqlInfo.objName) {
+			log.Infof("applying sharding recommendation for table=%q", sqlInfo.objName)
+			modifiedSqlStmt, err = applyShardingRecommendation(&sqlInfo)
+			if err != nil {
```
makalaaneesh, line 276: Let's log the original and modified statement. (Only for cases where it is different)?

yb-voyager/cmd/exportSchema.go
makalaaneesh: In that case, warn log is not necessary, it can be an info log

yb-voyager/cmd/exportSchema.go
```diff
@@ -309,32 +315,75 @@ func applyShardedTablesRecommendation(shardedTables []string) error {
...
+		createTableStmt.Options = []*pg_query.Node{
+			{
+				Node: &pg_query.Node_DefElem{
+					DefElem: colocationOption,
+				},
+			},
+		}
+	} else {
+		createTableStmt.Options = append(createTableStmt.Options, &pg_query.Node{
+			Node: &pg_query.Node_DefElem{
```
makalaaneesh, line 364: nit: define the node once

yb-voyager/cmd/exportSchema.go
makalaaneesh: Here you can log (WARN) along with the sqlinfo.stmt?

yb-voyager/cmd/exportSchema.go
```diff
@@ -309,32 +315,75 @@ func applyShardedTablesRecommendation(shardedTables []string) error {
...
-	if condition {
-		return "\n\n" + sql
+	if len(parseTree.Stmts) == 0 {
+		log.Infof("parse tree is empty, returning")
+		return stmt, nil
 	}
-	return sql
+
+	// Access the first statement directly
+	createStmtNode, ok := parseTree.Stmts[0].Stmt.Node.(*pg_query.Node_CreateStmt)
```
makalaaneesh, line 341: Is type assertion the only way of figuring out if it is a CreateStmt ? 
Is there no API on the Node?

makalaaneesh (approved):
LGTM with some nits. Pls run all jenkins pipelines

yb-voyager/cmd/exportSchema.go
sanyamsinghal: ok done

sanyamsinghal:
Jenkins Pipeine: https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2501/

yb-voyager/cmd/exportSchema.go
sanyamsinghal: let me check

sanyamsinghal: i didn't find any, we can look at it later.

yb-voyager/cmd/exportSchema.go
sanyamsinghal: done

PR #1500, opened 2024-05-10
Misc Automation/UI changes

shubham-yb:
The default value of the command is already appended to the help message. We don't need to explicitly mention it as well.
Added a check to see if FailureReasoning is empty or not in assessmentReport.json.


priyanshi-yb (approved):
LGTM, thanks!

PR #1503, opened 2024-05-10
Check for colocated database during import schema but not import schema post-snapshot-import

sanyamsinghal:
- Added check to prompt for colocated database only if the user has executed asses-migration command in their migration workflow

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -232,6 +232,18 @@ func IsMigrationAssessmentDone() (bool, error) {
 	return record.MigrationAssessmentDone, nil
 }
 
+func ClearMigrationAssessmentDone() error {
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -232,6 +232,18 @@ func IsMigrationAssessmentDone() (bool, error) {
 	return record.MigrationAssessmentDone, nil
 }
 
+func ClearMigrationAssessmentDone() error {
+	err := metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
+		if record.MigrationAssessmentDone {
```
priyanshi-yb, line 237: nit: no need for if condition, we can just unset it.

priyanshi-yb (approved):
LGTM

PR #1507, opened 2024-05-10
Bug: Schema modification not getting applied to case sensitive or case sensitive reserved words table

sanyamsinghal:
Discovered some cases where sharding recommendation was not getting applied to the case-sensitive or reserved keywords tables.
For example: for the same table report stored `public.case` and our object name(extract from DDL) had `public."case"` due to `case` being a reserved word.

- Fix: Using the pg_query parser to fetch the table name(not quotes but proper casing) and compare it with the names of recommended sharded tables in the assessmentReport

**Ignore the changes in second commit, its a temporary commit just for testing purpose**

### DIFF
--- a/yb-voyager/cmd/exportSchema.go
+++ b/yb-voyager/cmd/exportSchema.go
@@ -231,7 +231,7 @@ func applyMigrationAssessmentRecommendations() error {
 		return fmt.Errorf("failed to parse json report file %q: %w", assessmentReportPath, err)
 	}
 
-	shardedTables, err := report.GetShardedTablesRecommendation()
+	shardedTables, err := report.GetColocatedTablesRecommendation()
 	if err != nil {
 		return fmt.Errorf("failed to fetch sharded tables recommendation: %w", err)
 	} else {
--- a/yb-voyager/cmd/exportSchema.go
+++ b/yb-voyager/cmd/exportSchema.go
@@ -231,7 +231,7 @@ func applyMigrationAssessmentRecommendations() error {
 		return fmt.Errorf("failed to parse json report file %q: %w", assessmentReportPath, err)
 	}
 
-	shardedTables, err := report.GetColocatedTablesRecommendation()
+	shardedTables, err := report.GetShardedTablesRecommendation()
 	if err != nil {
 		return fmt.Errorf("failed to fetch sharded tables recommendation: %w", err)
 	} else {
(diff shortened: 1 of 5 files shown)


sanyamsinghal:
Jenkins pipeline: https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2517/

yb-voyager/cmd/exportSchema.go
```diff
@@ -261,29 +261,23 @@ func applyShardedTablesRecommendation(shardedTables []string) (err error) {
...
-				fmt.Printf("Please manually add the clause \"WITH (colocation = false)\" to the CREATE TABLE DDL of the '%s' table.\n", sqlInfo.objName)
-				modifiedSqlStmt = sqlInfo.formattedStmt
-			} else {
-				log.Infof("original ddl - %s", sqlInfo.stmt)
-				log.Infof("modified ddl - %s", modifiedSqlStmt)
-			}
-		} else {
-			modifiedSqlStmt = sqlInfo.formattedStmt
+		modifiedSqlStmt, match, err := applyShardingRecommendationIfMatching(&sqlInfo, shardedTables)
+		if match && err != nil {
```
makalaaneesh, line 272: Just a suggestion on re-organizing the if-else flow: (basically divide the error and non-error cases at the highest level.
```
if err != nil{
    if match {
    } else {
    }
} else {
   if match {}
}
```

yb-voyager/cmd/exportSchema.go
```diff
@@ -261,29 +261,23 @@ func applyShardedTablesRecommendation(shardedTables []string) (err error) {
...
+			log.Errorf("failed to apply sharding recommendation for table=%q: %v", sqlInfo.objName, err)
+			fmt.Printf("Unable to apply sharding recommendation for table=%q, continuing without applying...\n", sqlInfo.objName)
+			fmt.Printf("Please manually add the clause \"WITH (colocation = false)\" to the CREATE TABLE DDL of the '%s' table.\n", sqlInfo.objName)
+		} else if err != nil {
+			log.Errorf("failed to apply sharding recommendation for table=%q: %v", sqlInfo.objName, err)
```
makalaaneesh, line 277: In these cases, if err != nil, shouldn't we collate all of these and return an error as part of this function?

yb-voyager/cmd/exportSchema.go
```diff
@@ -320,35 +314,48 @@ func applyShardedTablesRecommendation(shardedTables []string) (err error) {
...
 		log.Infof("stmt=%s is not createTable as per the parse tree, expected tablename=%s", stmt, sqlInfo.objName)
-		return stmt, nil
+		return formattedStmt, false, nil
 	}
 	createTableStmt := createStmtNode.CreateStmt
 
+	// Extract schema and table name
+	relation := createTableStmt.Relation
+	parsedTableName := relation.Schemaname + "." + relation.Relname
+	if !slices.Contains(shardedTables, parsedTableName) {
```
makalaaneesh, line 355: so, table names in `shardedTables` is always expected to be fully qualified, non-quoted?

makalaaneesh:
https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/2520/pipeline/429/

yb-voyager/cmd/exportSchema.go
sanyamsinghal: yeah, for postgresql that is fair expectation i think? What do you say

sanyamsinghal: quotes are needed while creating the objects. For the comparisons we can compare unquoted?

yb-voyager/cmd/exportSchema.go
sanyamsinghal: In the parent function, calling `applyShardedTablesRecommendation` the errors returned are handled differently, it errors out.

`applyShardedTablesRecommendation` can just logs and ignores the error which does not ErrExit.

yb-voyager/cmd/exportSchema.go
makalaaneesh: Yeah comparing unquoted seems okay! I was just confirming 👍

yb-voyager/cmd/exportSchema.go
makalaaneesh: Got it, so we want our default behaviour to be: even if there are any errors in the func `applyShardingRecommendationIfMatching` , we don't want to halt the process. We just log (that we failed to apply the recommendation, and that user can manually apply) and move on. 

Is that correct?

yb-voyager/cmd/exportSchema.go
sanyamsinghal: yeah cool

yb-voyager/cmd/exportSchema.go
sanyamsinghal: yeah, i thought we should keep it till the time this SQL parsing related logic matures. Later on we can halt.

Currently in case of any unexpected behaviors, at least the export schema will be complete and user can unblock themselves.

makalaaneesh: Makes sense

yb-voyager/cmd/exportSchema.go
priyanshi-yb: Just to confirm, as I am not very familiar with the library yet, Is the expectation from `createTableStmt.Relation` also to have an unquoted name always?

sanyamsinghal: Yeah that what i observed with this.

PR #1511, opened 2024-05-14
Minor bug fixes

sanyamsinghal:
-  Ignoring metaDB rollback to log failures/errors in case of Transaction is done/aborted already

- Adding 'ALTER TABLE .. ADD CONSTRAINT ...' in table.sql instead of uncategoried.sql
for example below DDL was present in `uncategorized.sql`

```
--
-- Name: rna constraint_rna_seq_long_n; Type: CHECK CONSTRAINT; Schema: rnacen; Owner: -
--

ALTER TABLE rnacen.rna
    ADD CONSTRAINT constraint_rna_seq_long_n CHECK (((seq_long IS NULL) OR (((len - length(replace(seq_long, 'N'::text, ''::text))))::numeric <= round((0.1 * (len)::numeric))))) NOT VALID;
```

### DIFF
--- a/yb-voyager/src/srcdb/pg_dump_extract_schema.go
+++ b/yb-voyager/src/srcdb/pg_dump_extract_schema.go
@@ -92,7 +92,7 @@ func parseSchemaFile(exportDir string, schemaDir string, exportObjectTypesList [
 	if utils.FileOrFolderExists(filepath.Join(schemaDir, "schema.sql")) { // assess-migration workflow
 		schemaFilePath = filepath.Join(schemaDir, "schema.sql")
 	}
-	
+
 	log.Infof("begun parsing the schema file %q", schemaFilePath)
 	lines := readSchemaFile(schemaFilePath)
 	var delimiterIndexes []int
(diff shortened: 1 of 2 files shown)


yb-voyager/src/srcdb/pg_dump_extract_schema.go
```diff
@@ -135,7 +135,7 @@ func parseSchemaFile(exportDir string, schemaDir string, exportObjectTypesList [
 				objSqlStmts["SEQUENCE"].WriteString(stmts)
 			case "INDEX", "INDEX ATTACH":
 				objSqlStmts["INDEX"].WriteString(stmts)
-			case "TABLE", "DEFAULT", "CONSTRAINT", "FK CONSTRAINT":
+			case "TABLE", "DEFAULT", "CONSTRAINT", "FK CONSTRAINT", "CHECK CONSTRAINT":
```
priyanshi-yb, line 138: Is this observed for some specific case now?  or in-general for Check constraints in the schema.

priyanshi-yb (approved):
LGTM!

yb-voyager/src/srcdb/pg_dump_extract_schema.go
sanyamsinghal: Yeah observed in case of `rna` database;.
Previously i think it came because of CONSTRAINT type, but CHECK CONSTRAINT is also there.

PR #1514, opened 2024-05-14
Enhancement: Optimized the deferred statements retry logic(for dependent ddls) to voluntarily check further iterations of retry is required

sanyamsinghal:
- compare the before and after counts of deferredSqlStmts at each iteration

### DIFF
--- a/yb-voyager/cmd/importSchemaYugabyteDB.go
+++ b/yb-voyager/cmd/importSchemaYugabyteDB.go
@@ -25,7 +25,7 @@ import (
 	"golang.org/x/exp/slices"
 )
 
-var defferedSqlStmts []sqlInfo
+var deferredSqlStmts []sqlInfo
 var failedSqlStmts []string
 
 func importSchemaInternal(exportDir string, importObjectList []string,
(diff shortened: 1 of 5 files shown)


yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -42,44 +42,55 @@ func importSchemaInternal(exportDir string, importObjectList []string,
...
+				log.Infof("failed retry of deferred stmt: %s\n%v", utils.GetSqlStmtToPrint(deferredSqlStmts[j].stmt), err)
+				errString := "/*\n" + err.Error() + "\n*/\n"
+				failedSqlStmtInIthIteration = append(failedSqlStmtInIthIteration, errString+deferredSqlStmts[j].formattedStmt)
+				err = conn.Close(context.Background())
+				if err != nil {
+					log.Warnf("error while closing the connection due to failed deferred stmt: %v", err)
 				}
-				conn.Close(context.Background())
 				conn = newTargetConn()
-				j++
```
makalaaneesh, line 80: I'm not sure if it's safe to move the `j++` from here to the for loop definition. Looks like it is here in the else block only because in the event that we successfully execute the stmt, we change the `deferredSqlStmts` array by removing that element, therefore, in the next iteration, the value of `j` should remain unchanged.

sanyamsinghal: After this operation of deleting the element from slice: 
```
deferredSqlStmts = append(deferredSqlStmts[:j], deferredSqlStmts[j+1:]...)
```

We also break the inner for-loop, so there is no next iteration.

Let me know if i am missing anything.

priyanshi-yb (approved):
LGTM

yb-voyager/cmd/importSchemaYugabyteDB.go
makalaaneesh: Oh right, I see the break! 
But now I'm confused; why do we break when we successfully execute a deferred statement? Shouldn't we continue?

sanyamsinghal: once the jth statement executed successfully, there are chances that some of the ddls before 'j' were dependent on 'j' and can be executed successfully, so as soon as 'j' passes we start over.

PR #1517, opened 2024-05-15
Added code to save ssl flags in a local file in docker script and added code in end migration to remove docker metadata file

ShivanshGahlot:
End migration was failing with docker voyager whenever ssl came into the picture. This was because docker was unable to mount the directories containing ssl certs since no such flags are provided at the time of end migration. To fix this:
* Added provision in the docker script to save the ssl flags in a file /tmp directory whenever it encounters them. This file is moved to export dir and is read at the time of end migration(and other commands) and the the docker script mounts the respective directories.


priyanshi-yb (approved):
LGTM, Please ensure to run all docker tests properly.

docker/yb-voyager-docker
```diff
@@ -54,6 +54,12 @@ for var_name in "${variables[@]}"; do
   fi
 done
 
+flags_to_store=("--source-ssl-cert"
```
makalaaneesh, line 57: We should do the same for target and source-replica ssl flags

yb-voyager/cmd/endMigrationCommand.go
```diff
@@ -782,6 +783,17 @@ func getFreeDiskSpace(path string) (uint64, error) {
 	return freeSpace, nil
 }
 
+func cleanupDockerScriptMetadata() {
```
makalaaneesh, line 786: the `/tmp` dir where the file is stored would be on the host's filesystem, but this code will run inside the container (i.e. a different file system), so I don't think it would correctly delete that file?

Either way, I think since this is a docker-specific logic, it would be good to have this logic inside the `yb-voyager-docker` script if possible.

makalaaneesh: Since it's just a small tmp file, maybe we don't really have to delete as well?

docker/yb-voyager-docker
```diff
@@ -54,6 +54,12 @@ for var_name in "${variables[@]}"; do
...
+"--source-ssl-key"
+"--source-ssl-root-cert")
+
+file_name="/tmp/voyager-docker-metadata.txt"
```
makalaaneesh, line 61: This name should be export-dir/migrationUUID specific. Otherwise, it will interfere with other migrations running on the same host.

makalaaneesh: Consider moving the temp file into the export-dir once you have collected all the arguments. That way it will reside inside the export-dir.

yb-voyager/cmd/endMigrationCommand.go
priyanshi-yb: I agree.

docker/yb-voyager-docker
```diff
@@ -54,8 +54,31 @@ for var_name in "${variables[@]}"; do
...
 exported_vars="${exported_vars} -e ORACLE_HOME=/usr/lib/oracle/21/client64 -e LD_LIBRARY_PATH=/usr/lib/oracle/21/client64/lib -e PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/usr/local/go/bin:/usr/lib/oracle/21/client64/bin"
 
+# if command is like yb-voyager export data. The command name is export data.
+command_name=${argv[0]}-${argv[1]}
+# if argv[2] is "from" or "to" then command name is like import data from source or import data to target
+if [[ ${argv[2]} == "from" || ${argv[2]} == "to" ]]
+then
+	command_name=${argv[0]}-${argv[1]}-${argv[2]}-${argv[3]}
+fi
+file_name="/tmp/docker-metadata-${command_name}.txt"
```
makalaaneesh, line 79: let's use a variable name that is more specific? say `command_metadata_file_name` ?

docker/yb-voyager-docker
```diff
@@ -152,11 +188,32 @@ do
     i=$(( $i + 1))
 done
 
+# If the metadata file exists then move it to the export_dir and overwrite any existing file.
+if [ -f "$file_name" -a -n "$export_dir" ]
+then
+	mv $file_name $export_dir
```
makalaaneesh, line 194: Shall we put it inside export-dir/meta? 
note that you might have to handle the case where the export-dir is empty at this time..

makalaaneesh (approved):
Nice! LGTM with a couple of nits

docker/yb-voyager-docker
```diff
@@ -54,8 +54,31 @@ for var_name in "${variables[@]}"; do
...
 exported_vars="${exported_vars} -e ORACLE_HOME=/usr/lib/oracle/21/client64 -e LD_LIBRARY_PATH=/usr/lib/oracle/21/client64/lib -e PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/usr/local/go/bin:/usr/lib/oracle/21/client64/bin"
 
+# if command is like yb-voyager export data. The command name is export data.
+command_name=${argv[0]}-${argv[1]}
+# if argv[2] is "from" or "to" then command name is like import data from source or import data to target
+if [[ ${argv[2]} == "from" || ${argv[2]} == "to" ]]
+then
+	command_name=${argv[0]}-${argv[1]}-${argv[2]}-${argv[3]}
+fi
+command_metadata_file_name="/tmp/docker-metadata-${command_name}.txt"
```
makalaaneesh, line 79: Just realised, we should add a random txt to this filename, otherwise it will conflict with other migrations run on the same host.

PR #1521, opened 2024-05-20
Integrated assess-migration command in the end to end automation test workflow

sanyamsinghal:


migtests/scripts/functions.sh
```diff
@@ -708,4 +708,64 @@ validate_failure_reasoning() {
...
+
+fetch_colocated_tables() {
+    local json_file=$1
+    jq '.Sizing.SizingRecommendation.ColocatedTables // []' "$json_file"
+}
+
+# Function to move a specified percentage of tables from colocated to sharded
+move_tables() {
+    local json_file=$1
+    local percentage=$2
```
makalaaneesh, line 739: (Just a discussion, not a strong recommendation) 
I understand the intention behind randomizing the set of tables in colocated/sharded by percentags, but usually, I think it's better to have tests be deterministic, i.e. you should be able to reproduce a a test case failure by just running it. 
By depending on percentages, it makes it more complicated to do so. (although you could just modify it to 100% or 0% when we need to repro.

Alternatively, what if we just mark all tables are sharded? That way we get to test our DDL modification for all table types.

sanyamsinghal: @makalaaneesh  Regarding 100% Sharded tables, so that it modify all the tables is good from perspective of testing.

With that just my concern was that we should cover this general case(combination of colocated and sharded tables) with higher priority and not miss it

shubham-yb: Just a suggestion. We can make the percentage a random value as well (maybe from a specific set). That will ensure 0, 100 and anything in between.
Maybe choose from a set of 0, 25, 50, 75, 100?

makalaaneesh: ^ that would make it even more non-deterministic :D

migtests/tests/pg/partitions/fix-schema
```diff
@@ -10,3 +10,7 @@ sed -i  's/p2\.sydney/p2.sydney_region/g' $TEST_DIR/export-dir/schema/tables/tab
 sed -i  's/p2\.boston/p2.boston_region/g' $TEST_DIR/export-dir/schema/tables/INDEXES_table.sql
 sed -i  's/p2\.london/p2.london_region/g' $TEST_DIR/export-dir/schema/tables/INDEXES_table.sql
 sed -i  's/p2\.sydney/p2.sydney_region/g' $TEST_DIR/export-dir/schema/tables/INDEXES_table.sql
+
+sed -i  's/p2\.boston/p2.boston_region/g' $TEST_DIR/export-dir/assessment/reports/assessmentReport.json
```
makalaaneesh, line 14: @sanyamsinghal pls add a comment in this file explaining the reason for the change in names (i.e. to test the scenario of manual change in partition names)

makalaaneesh (approved):
LGTM. Added a comment on how we could improve coverage. Can discuss and take it up later. 
@shubham-yb pls also take a look

migtests/tests/pg/datatypes/validateAfterChanges
```diff
@@ -96,6 +96,10 @@ def change_expected_values():
 def migration_completed_checks_yb():
 	print("Running tests on YB")
 	yb.run_checks(migration_completed_checks)
+	yb.run_checks(YB_specific_checks)
```
shubham-yb, line 99: Is this check redundant in `validateAfterChanges` ? There won't be any schema based changes to the tables.

migtests/tests/pg/basic-non-public-live-test/validateAfterChanges
```diff
@@ -73,6 +73,10 @@ def change_expected_values():
 def migration_completed_checks_yb():
 	print("Running tests on YB")
 	yb.run_checks(migration_completed_checks)
+	yb.run_checks(YB_specific_checks)
```
shubham-yb, line 76: We should add it for the dependent-ddls and codependent schemas tests. Those will cover nice scenarios. We can add more later.

migtests/scripts/functions.sh
sanyamsinghal: yeah makes sense to have a separate test script for schema migration

migtests/tests/pg/datatypes/validateAfterChanges
sanyamsinghal: done

migtests/tests/pg/basic-non-public-live-test/validateAfterChanges
sanyamsinghal: i didn't get this comment @shubham-yb.
I have removed the checks from `validateAfterChanges` for now

migtests/tests/pg/partitions/fix-schema
sanyamsinghal: done

migtests/tests/pg/basic-non-public-live-test/validateAfterChanges
shubham-yb: NVM, I can add it in a separate PR.

PR #1522, opened 2024-05-21
Optimised get data-migration-report command to initialise target(s) and fetch events stats at once

priyanshi-yb:
This patch changes the way of fetching events imported stats in target from per table to fetch in a single query for the complete table list. It reduces the time taken in generating report for around 25 tables from 1min52s to 5s.

Jenkins pipeline - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2579


yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -316,42 +332,54 @@ func updateImportedEventsCountsInTheRow(sourceDBType string, row *rowData, table
 	}
 	if importerRole == SOURCE_REPLICA_DB_IMPORTER_ROLE {
 		var err error
-		tblName := tableNameTup.ForKey()
-		tableNameTup, err = nameRegistryForSourceReplicaRole.LookupTableName(tblName)
-		if err != nil {
-			return fmt.Errorf("lookup %s in source replica name registry: %v", tblName, err)
+		sourceReplicaTups := make([]sqlname.NameTuple, len(tableNameTups))
```
makalaaneesh, line 335: Why are we calling lookup again here? I'm assuming it's because we now want to lookup in the context of SOURCE_REPLICA_DB_IMPORTER_ROLE? so that the state lookup happens properly? But role is tied to name registry object, right? 

Pls add a comment explaining

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -316,42 +332,54 @@ func updateImportedEventsCountsInTheRow(sourceDBType string, row *rowData, table
...
+	switch row.DBType {
+	case "target":
+		importerRole = TARGET_DB_IMPORTER_ROLE
+	case "source-replica":
+		importerRole = SOURCE_REPLICA_DB_IMPORTER_ROLE
+	case "source":
+		importerRole = SOURCE_DB_IMPORTER_ROLE
+	}
+	if importerRole == SOURCE_REPLICA_DB_IMPORTER_ROLE {
+		var err error
```
makalaaneesh, line 370: Pls add a comment here as well. I forgot why exactly we need a separate name registry in this case...

Since we are looking up nametuple from the map, wouldn't it work even without changing the tableNameTup to be from the nameRegistryForSourceReplicaRole? i.e. because the map.Get would use ForKey, it should work even if the nametuples are different no?

makalaaneesh (approved):
LGTM! 
I didn't see any changes to run-test*.sh, you only added the expected report json files. Will they automatically be validated?

priyanshi-yb:
> I didn't see any changes to run-test*.sh, you only added the expected report JSON files. Will they automatically be validated?

@makalaaneesh  Yes, validation for all the workflows and all status/report commands is already there, just that if the expected file is not found we are not erroring out right now. Once we add all the expected files for tests we can then error out if the expected file is not found.

PR #1523, opened 2024-05-21
Added Oracle live migration SSL tests

shubham-yb:


migtests/scripts/functions.sh
```diff
@@ -432,6 +442,12 @@ import_data_to_source_replica() {
 	then
 		args="${args} --source-replica-db-schema ${SOURCE_REPLICA_DB_SCHEMA}"
 	fi
+	if [ "${SOURCE_DB_ORACLE_CDB_TNS_ALIAS}" != "" ]
```
makalaaneesh, line 445: this should jsut be `SOURCE_DB_ORACLE_TNS_ALIAS` not with `CDB`?

migtests/scripts/functions.sh
```diff
@@ -432,6 +442,12 @@ import_data_to_source_replica() {
...
+	then
+		args="${args} --oracle-tns-alias ORCLPDB1SSL"
```
makalaaneesh, line 447: You should be taking the actual env var here instead of hard-coding

migtests/tests/oracle/ssl/cleanup-db
```diff
@@ -6,4 +6,10 @@ set -x
...
 
 echo "Deleting existing data in source database"
-run_sqlplus_as_schema_owner ${SOURCE_DB_NAME} ${TESTS_DIR}/oracle/utils/delete_full_schema
\ No newline at end of file
+run_sqlplus_as_schema_owner ${SOURCE_DB_NAME} ${TESTS_DIR}/oracle/utils/delete_full_schema
+
+if [ "${SOURCE_REPLICA_DB_SCHEMA}" != "" ]
+then
+	echo "Deleting existing data in FF database"
+	run_sqlplus ${SOURCE_DB_NAME} ${SOURCE_REPLICA_DB_SCHEMA} ${SOURCE_REPLICA_DB_PASSWORD} ${TESTS_DIR}/oracle/utils/delete_full_schema
```
makalaaneesh, line 14: We don't need to `run_sqlplus_as_schema_owner` ?

migtests/tests/oracle/ssl/validate
```diff
@@ -1,39 +1,41 @@
...
+	oracle.run_checks(migration_completed_checks, db_type="source_replica")
 
 def migration_completed_checks(tgt):
 	# Validation to check for matching row counts
-	got_row_count = tgt.row_count_of_all_tables("public")
+	got_row_count = tgt.row_count_of_all_tables(db_schema)
 	for table_name, row_count in EXPECTED_ROW_COUNT.items():
 		print(f"table_name: {table_name}, row_count: {got_row_count[table_name]}")
 		assert row_count == got_row_count[table_name]
-#=============================================================================
```
makalaaneesh, line 27: Why remove this validation?

migtests/tests/oracle/ssl/validateAfterChanges
```diff
@@ -0,0 +1,63 @@
...
+		migration_completed_checks_fb()
+	else:
+		migration_completed_checks_yb()
+
+EXPECTED_ROW_COUNT = {
+	'test_simple': 2,
+}
+
+
+EXPECTED_ROW_COUNT_FF = {
```
makalaaneesh, line 29: EXPECTED_ROW_COUNT_FF_FB ? since you are using same values in both cases...

migtests/scripts/functions.sh
shubham-yb: Same as above. Since `SOURCE_DB_ORACLE_TNS_ALIAS` is only getting set in case of offline migration, had to hard-code it here. Can try change this behaviour by adding this variable as well to live_env.sh but then the export-data command will need to have more conditions.

migtests/tests/oracle/ssl/cleanup-db
shubham-yb: Just copied it from the other tests. Also checked it in the automation. Seems to be working fine without superuser.

migtests/tests/oracle/ssl/validate
shubham-yb: We don't have datatype checking in the Oracle library yet and didn't want to complicate this validation. So removed it. Can add it if it seems necessary.

migtests/scripts/functions.sh
shubham-yb: For live migration `SOURCE_DB_ORACLE_CDB_TNS_ALIAS` is required so setting that and not `SOURCE_DB_ORACLE_TNS_ALIAS`. It was getting a bit tricky to run all types of combinations within the same test. Thus the introduction of live_env.sh.

makalaaneesh: env var `SOURCE_REPLICA_DB_ORACLE_TNS_ALIAS=ORCLPDB1SSL`

migtests/tests/oracle/ssl/validate
makalaaneesh: Seems alright. This is only for ssl tests. data types mapping will get validated in datatypes test anyway..

migtests/scripts/functions.sh
makalaaneesh: https://github.com/yugabyte/yb-voyager/pull/1523/files#r1628896181

makalaaneesh (approved):
LGTM

PR #1524, opened 2024-05-23
Bug fix: In Oracle live migration log mining flush table should be passed in upper case to debezium

priyanshi-yb:
1. The issue is when we resume export data and use the same log mining flush table, debezium is erroring out while creating the table again with the error `name already exist` as it checks first if the table exists or not and that check is failing to detect the existence of this table because of lowercase chars present in the table name we pass to debezium. Fixes https://yugabyte.atlassian.net/browse/DB-11433
2. Another issue for resumption is that if the user and schema of Oracle are the same for the migration then this log mining flush table is created in the Schema for which the migration is happening but the name registry wasn't aware of it in starting as it is created during the migration. So on resumption, the table list we fetch from Oracle will have that table name but the lookup will error out saying table not found in name registry. Fixes - https://yugabyte.atlassian.net/browse/DB-11434

Jenkins runs - 
1. Live migration tests with. oracle offline tests - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2593/
4. Oracle offline test with debezuim - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2592/

### DIFF
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -232,6 +232,7 @@ func exportData() bool {
 	utils.PrintAndLog("table list for data export: %v", tableListToDisplay)
 
 	//finalTableList is with leaf partitions and root tables after this in the whole export flow to make all the catalog queries work fine
+
 	if changeStreamingIsEnabled(exportType) || useDebezium {
 		config, tableNametoApproxRowCountMap, err := prepareDebeziumConfig(partitionsToRootTableMap, finalTableList, tablesColumnList, leafPartitions)
 		if err != nil {
@@ -547,7 +548,13 @@ func getFinalTableColumnList() (map[string]string, []sqlname.NameTuple, *utils.S
 	tableListFromDB := source.DB().GetAllTableNames()
 	var err error
 	var fullTableList []sqlname.NameTuple
+	logMiningFlushTable := utils.GetLogMiningFlushTableName(migrationUUID)
 	for _, t := range tableListFromDB {
+		if source.DBType == ORACLE {
+			if t.ObjectName.Unquoted == logMiningFlushTable {
+				continue
+			}
+		}
 		schema, table := t.SchemaName.Unquoted, t.ObjectName.Unquoted
 		defaultSchemaName, _ := getDefaultSourceSchemaName()
 		//For partitions case there is no defined mapping and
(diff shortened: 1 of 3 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -547,7 +548,14 @@ func getFinalTableColumnList() (map[string]string, []sqlname.NameTuple, *utils.S
 	tableListFromDB := source.DB().GetAllTableNames()
 	var err error
 	var fullTableList []sqlname.NameTuple
+	logMiningFlushTable := utils.GetLogMiningFlushTableName(migrationUUID)
 	for _, t := range tableListFromDB {
+		if source.DBType == ORACLE {
```
makalaaneesh, line 553: How about adding this logic to `GetAllTableNames` and `GetAllTableNamesRaw` of the Oracle implementation of source db? That way it will get filtered out of everywhere..

We probably don't need exact match, we can check for prefix..? (since migrationUUID won't be available there.)

priyanshi-yb: yeah, if an exact match is not required then its better in `GetAllTableNamesRaw`.

makalaaneesh (approved):
LGTM

yb-voyager/src/srcdb/oracle.go
```diff
@@ -142,6 +142,10 @@ func (ora *Oracle) GetAllTableNamesRaw(schemaName string) ([]string, error) {
 		if err != nil {
 			return nil, fmt.Errorf("error in scanning query rows for table names: %v", err)
 		}
+		if strings.HasPrefix(tableName, "VOYAGER_LOG_MINING_FLUSH_") {
```
sanyamsinghal, line 145: I was just thinking should we use the exact table name here or not.

Let's say two separate migration(different uuid) are running for the same schema, in that case we don't want any migration's the flush table to comeup, right?

So then it is fine.

sanyamsinghal (approved):
LGTM

yb-voyager/src/srcdb/oracle.go
priyanshi-yb: yeah right, now with this we will ignore all such tables prefixed with `VOYAGER_LOG_MINING_FLUSH`

PR #1536, opened 2024-06-05
Fix comparison logic in live migration batch rows affected

makalaaneesh:
Currently, we seem to be comparing the wrong values of rows affected with the batch inserts/updates/deletes. 
While this won't functionally cause any problems, this will lead to :
1. False positive warning logs
2. Very quick growth in logs because of unnecessary logging. (~ 400MB per hour in one experimental run)

Testing:
- manually tested:
    - before: for all batches, we were getting a log
    - now: not getting a log by default; should only come up in an actual discrepancy


priyanshi-yb (approved):
LGTM!
Thanks for fixing this!

PR #1537, opened 2024-06-05
Changing index creation workflow from "post-snapshot-import" phase of import-schema to default phase.

makalaaneesh:
Changing index creation workflow from "post-snapshot-import" phase of import-schema to default phase. 

**Reason:** 
Based on perf experiments, it was shown that pre-creating indexes and then loading data was faster as compared to loading data and backfilling indexes. 
The theory here is:
- Index creation is serial(parallel indexes cannot be created on YB as of now). Furthermore, as a result, there will be as many full table scans as there are no. of indexes 
- On the other hand, writing to indexes at the time of INSERT/UPDATE is done in parallel. 

Testing: 
- Added schema validation to test cases to validate that indexes are created at import-schema default phase
- Manually tested that if an index is invalid, and import-schema is run again, it will drop it and recreate it (Changed some code related to this logic, so manually tested again)


yb-voyager/cmd/importData.go
```diff
@@ -1143,6 +1143,11 @@ func beforeIndexCreation(sqlInfo sqlInfo, conn **pgx.Conn, objType string) error
 		return fmt.Errorf("extract qualified index name from DDL [%v]: %w", sqlInfo.stmt, err)
 	}
 	if invalidTargetIndexesCache == nil {
+		tdb = tgtdb.NewTargetDB(&tconf)
```
makalaaneesh, line 1146: @sanyamsinghal because I moved indexes to the import-schema by default, I have to create targetDB here. I can't create it at the beginning of import-schema, because Init checks if the schema is present, and at the time, the schema objects haven't yet been created. 

I'm not too happy about creating targetDB here, because it seems out of place. Any suggestions? 
One approach I was thinking of was: let's not have `InvalidIndexes` be part of target_db_interface. Let us use the connection object that is part of this function to determine the invalid indexes. Right now, we don't really care about importing schema/indexes to source-replicas; it's only YB we care about. 
cc: @priyanshi-yb

priyanshi-yb: Agreed, for the approach having this `InvalidIndexes` not in target_db_interface. As for only `import-schema` we need this and it is not a hard requirement to be a part of target_db_iterface.

makalaaneesh:
jenkins: https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/2683/pipeline/

priyanshi-yb (approved):
LGTM

makalaaneesh:
oracle tests - https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/2712/pipeline/459/

PR #1546, opened 2024-06-06
Fix the deferred logic to not override the failed statements caught apart from deferred statements

priyanshi-yb:
**Issue:**
In case a few statements can't be deferred have issues, there is a bug in the current deferred logic which is overriding the non-deferred statements with just the deferred ones before putting them in the `failed.sql` while using `--continue-on-error true`.
https://yugabyte.atlassian.net/browse/DB-11653

**Fix:** 
Fixed the logic not to override `failedSqlStmts` where we put the failed statements non-deferred ones and only appending the final deferred statements that can't be executed in the `failedSqlStmts`.

**Tests:**
Manual testing is done, can't add automation for `continue-on-error` right now. In the future, we can have a test for `continue-on-error` and validate the `failed.sql` for expected failures. cc: @shubham-yb

### DIFF
--- a/yb-voyager/cmd/importSchemaYugabyteDB.go
+++ b/yb-voyager/cmd/importSchemaYugabyteDB.go
@@ -20,7 +20,10 @@ import (
 	"fmt"
 	"path/filepath"
 	"strings"
+	"time"
 
+	"github.com/fatih/color"
+	"github.com/jackc/pgx/v4"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/exp/slices"
 
@@ -28,7 +31,7 @@ import (
 )
 
 var deferredSqlStmts []sqlInfo
-var failedSqlStmts []string
+var finalFailedSqlStmts []string
 
 func importSchemaInternal(exportDir string, importObjectList []string,
 	skipFn func(string, string) bool) {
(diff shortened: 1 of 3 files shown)


yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -89,10 +91,11 @@ func importDeferredStatements() {
 		} else if beforeDeferredSqlCount == afterDeferredSqlCount {
 			// no need for further iterations since the deferred list will remain same
 			log.Infof("none of the deferred statements executed successfully in the %d iteration", i)
-			failedSqlStmts = failedSqlStmtInIthIteration
+			finalFailedDeferredStmts = failedSqlStmtInIthIteration
 			break
 		}
 	}
+	failedSqlStmts = append(failedSqlStmts, finalFailedDeferredStmts...)
```
sanyamsinghal, line 98: how about renaming global variable: `failedSqlStmts` -> `finalFailedSqlStmts`

yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -22,8 +22,9 @@ import (
 	"strings"
 
 	log "github.com/sirupsen/logrus"
-	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
 	"golang.org/x/exp/slices"
+
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
```
sanyamsinghal, line 27: remove extra newline.

sanyamsinghal (approved):
LGTM, good catch and thanks for fixing.

yb-voyager/cmd/importSchemaYugabyteDB.go
priyanshi-yb: its automatically added by my editor's go formatter

PR #1547, opened 2024-06-06
Fix analyze-schema reporting the false positive case for "uuid-ossp" extension not supported

priyanshi-yb:
fixes - https://yugabyte.atlassian.net/browse/DB-11650
Tests: Added case in automation tests for this extension.


sanyamsinghal (approved):

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -56,7 +56,7 @@ var (
 	anything                     = `.*`
 	ws                           = `[\s\n\t]+`
 	optionalWS                   = `[\s\n\t]*` //optional white spaces
-	ident                        = `[a-zA-Z0-9_."]+`
+	ident                        = `[a-zA-Z0-9_."-]+`
```
makalaaneesh, line 59: Worth noting that `-` is not a valid character in an identifier in postgres. 
It just becomes valid when inside quotes. 
I don't know if there's an easier way to solve this.
```
SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($). 
...
Quoted identifiers can contain any character, except the character with code zero. (To include a double quote, write two double quotes.) This allows constructing table or column names that would otherwise not be possible, such as ones containing spaces or ampersands. The length limitation still applies.
```
https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

makalaaneesh: chatgpt gave a regex :)  https://chatgpt.com/share/b5ab5cf7-e96d-48bb-be87-d522fee54827

`^(?![0-9])[A-Za-z_][A-Za-z0-9_$]{0,62}$|^"([^"\x00]|"")*"$`

See if this works @priyanshi-yb ? Will avoid potential future issues (although we should move away from regex to parser anyway ... )

priyanshi-yb: Yeah, you're right but I think modifying this regex to handle all such cases like having these chars only with quotes and all is a bit complicated, the regex you shared isn't working and I have also tried a few. 
Also, I think we anyway don't support object names with special chars completely cc: @shubham-yb https://docs.yugabyte.com/preview/yugabyte-voyager/known-issues/general-issues/#exporting-data-with-names-for-tables-functions-procedures-using-special-characters-whitespaces-fails , so I think this regex should be fine for now to fix this extension case and we can take these special chars in object names with quotes as an end-to-end case later? What do you think?

priyanshi-yb: Filed a low priority ticket - https://yugabyte.atlassian.net/browse/DB-11707

PR #1548, opened 2024-06-06
Modifying call-home structure to send the information of each run of a command 

priyanshi-yb:
**Description:** 
This patch modifies the call-home structure to send the information of each run of a command to call-home database this will give a log book or time-series kind of view of migration including the multiple run of a same command in multiple records and not updating a single record.
https://yugabyte.atlassian.net/browse/DB-11554
 The structure looks like this - 
```
CREATE TABLE diagnostics (
	migration_uuid UUID,
	phase_start_time TIMESTAMP WITH TIME ZONE,
	source_db_details JSONB,
	target_db_details JSONB,
	yb_voyager_version TEXT,
	migration_phase TEXT,
	phase_payload JSONB,
	migration_type TEXT,
	time_taken_sec BIGINT,
	status TEXT,
	PRIMARY KEY (migration_uuid, phase_start_time, migration_phase)
);
```
In this structure, we are sending the source/target necessary details in a JSON format to the database. And another column `phase_payload` which stores the necessary or relevant information of a phase-specific input or output parameters refer to the document attached below for more detailed information.

1. This patch sends the payload for all the offline migration commands and its workflow with/without debezium. https://yugabyte.atlassian.net/browse/DB-11563
2. This patch also adds support for phase-specific information such as - source_db_size, migration type (offline, live..), and assessment_report information like unsupported features/datatypes, index/table size stats.. https://yugabyte.atlassian.net/browse/DB-11566
3. Added support for dev/testing purpose local call home service end points ENV variables to provide the testing locally the call-home changes.
```
LOCAL_CALL_HOME_SERVICE_HOST, LOCAL_CALL_HOME_SERVICE_PORT
```
 
Doc for reference - https://docs.google.com/document/d/1lZd6ih83wrV5W-Jt09WzdANoN6KZALkq5WJs78U1apI/edit#heading=h.ig5c7dnaclcn

#=================================================================

**Tests:** 

Automation - TODO
Manual - 
I have tested offline migration completely for PG, MySQL and Oracle with/without debezium with send-diagnostics on.
More cases mentioned in the docs here - https://docs.google.com/document/d/1lZd6ih83wrV5W-Jt09WzdANoN6KZALkq5WJs78U1apI/edit#heading=h.vwhyx9psdpre


yb-voyager/src/callhome/diagnostics.go
```diff
@@ -33,35 +32,122 @@ import (
...
+	phase_start_time TIMESTAMP WITH TIME ZONE,
+	source_db_details JSONB,
+	target_db_details JSONB,
+	yb_voyager_version TEXT,
+	migration_phase TEXT,
+	phase_payload JSONB,
+	migration_type TEXT,
+	time_taken_sec NUMERIC(20,2),
+	status TEXT,
+	PRIMARY KEY (migration_uuid, phase_start_time, migration_phase)
```
makalaaneesh, line 57: For live migration /long running processes, where we aim to send multiple rows for a command, this won't suffice, correct? 

I think it might be worth thinking about this right now, even if don't implement live migration. That way, the schema will be fixed, and we just have to enhance voyager later on..

yb-voyager/cmd/exportData.go
```diff
@@ -127,29 +129,119 @@ func exportDataCommandFn(cmd *cobra.Command, args []string) {
...
 		log.Error("Export of data failed.")
+		packAndSendExportDataPayload(ERROR)
 		atexit.Exit(1)
 	}
 }
 
+func packAndSendExportDataPayload(status string) {
+	//TODO: send this INPROGRESS status in some fixed interval for long running export data
+
+	if !callhome.SendDiagnostics {
```
makalaaneesh, line 152: Can you add this logic to callhome.SendPayload function as an extra layer of protection?

yb-voyager/cmd/importData.go
```diff
@@ -577,7 +577,54 @@ func importData(importFileTasks []*ImportFileTask) {
...
+			importDataPayload.TotalRows += value
+			if value > importDataPayload.LargestTableRows {
+				importDataPayload.LargestTableRows = value
+			}
+			return true, nil
+		})
+	}
+	importDataPayloadBytes, err := json.Marshal(importDataPayload)
+	if err != nil {
+		log.Errorf("error in parsing the export data payload: %v", err)
```
makalaaneesh, line 621: Makes sense that we **ONLY** log the errors and do not return/fail. However, I think it would make sense to prefix the errors with "callhome:" so that we don't confuse them with actual errors. I don't suppose there is another way to determine that the errors are from call home, right now, is there? 

if you're making the change, pls make it for all the relevant functions and logs

yb-voyager/cmd/importData.go
```diff
@@ -577,7 +577,54 @@ func importData(importFileTasks []*ImportFileTask) {
...
+	if err != nil {
+		log.Errorf("error in parsing sourcedb details: %v", err)
+	}
+	payload.TargetDBDetails = string(bytes)
+	payload.MigrationPhase = IMPORT_DATA_PHASE
+	importDataPayload := callhome.ImportDataPhasePayload{
+		ParallelJobs: int64(tconf.Parallelism),
+		StartClean:   bool(startClean),
 	}
+	importRowsMap, err := getImportedSnapshotRowsMap("target")
```
makalaaneesh, line 607: We should try and avoid hitting DB from these functions. (because of what Hemant had said about exit handlers potentially getting stuck or taking too long). That could happen in this case if it takes too long to get data from the DB. 

Would it be easy to capture the relevant information and store it in our StatsReporter, and this function can just fetch it in memory? 

(I could only find this case where we are hitting a DB during the packAndSend* functions. If there are other cases, let's discuss those as well)

yb-voyager/src/srcdb/postgres.go
```diff
@@ -463,6 +463,16 @@ func (pg *PostgreSQL) GetCharset() (string, error) {
 	return encoding, nil
 }
 
+func (pg *PostgreSQL) GetDatabaseSize() (int64, error) {
+	var dbSize int64
+	query := fmt.Sprintf("select pg_database_size('%s'); ", pg.source.DBName)
```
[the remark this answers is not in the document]
priyanshi-yb: I have read about all three of them, and most are generally fast as they all read the database's metadata to give size but the Oracle one where I am accessing `DBA_SEGMENTS` is summing up all the segments in the schema so if those are large in number it might become a bit slow. But I don't think there can be any other option for this in Oracle.

yb-voyager/cmd/common.go
```diff
@@ -989,3 +990,38 @@ func (ar *AssessmentReport) GetClusterSizingRecommendation() string {
...
 }
+
+// ==========================================================================
+
+func createCallhomePayload() callhome.Payload {
+	var payload callhome.Payload
+	payload.MigrationUUID = migrationUUID
+	payload.PhaseStartTime = startTime.UTC().Format("2006-01-02T15:04:05.999999")
+	payload.YBVoyagerVersion = utils.YB_VOYAGER_VERSION
+	payload.TimeTakenSec = time.Since(startTime).Seconds()
```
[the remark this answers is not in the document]
priyanshi-yb: Yes right, I am going modify this function and its once I will work on sending the payload for long-running commands

yb-voyager/src/callhome/diagnostics.go
priyanshi-yb: > For live migration /long running processes, where we aim to send multiple rows for a command, this won't suffice, correct?

so for this what I was thinking is that when I will send the payload for a log running command I will modify the start time to the time just after the previous payload was sent and accordingly update the timeTaken. So that way this PK would suffice. What do you say?

yb-voyager/cmd/importData.go
priyanshi-yb: Yes sure that's a good idea to distinguish the logs. Thanks! will make the changes

yb-voyager/cmd/importSchema.go
```diff
@@ -56,7 +57,11 @@ var importSchemaCmd = &cobra.Command{
 
 	Run: func(cmd *cobra.Command, args []string) {
 		tconf.ImportMode = true
-		importSchema()
+		err := importSchema()
+		if err != nil {
+			packAndSendImportSchemaPayload(ERROR, err.Error())
```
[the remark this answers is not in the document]
priyanshi-yb: Yes was thinking of fixing it but missed. Will do.

yb-voyager/cmd/exportData.go
```diff
@@ -127,29 +129,119 @@ func exportDataCommandFn(cmd *cobra.Command, args []string) {
...
+
+	exportDataPayloadBytes, err := json.Marshal(exportDataPayload)
+	if err != nil {
+		log.Errorf("error in parsing the export data payload: %v", err)
+	}
+	payload.PhasePayload = string(exportDataPayloadBytes)
+	payload.Status = status
+
+	callhome.SendPayload(&payload)
+	callHomePayloadSent = true
```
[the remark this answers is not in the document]
priyanshi-yb: Here I am preventing of upserting the COMPLETE status from being updated to EXIT by the exit handler as exit handlers will run every time even in success cases right? So I am using the bool to not send the payload if its already sent. I know I might have to revisit this in case I am sending the payload multiple times.

yb-voyager/cmd/importData.go
priyanshi-yb: `state.getImportedSnapshotRowsMap` is not hitting DB it's only using local importDataState `.D/.P...` files only.

> (I could only find this case where we are hitting a DB during the packAndSend* functions. If there are other cases, let's discuss those as well)

No there are no other cases where I am calling in these functions, all the DB-related calls are outside of this function in normal flow and then using that info here only if that is populated. Even in this case, this is just local files if there won't be any files or any negative case it will just return 0.

yb-voyager/src/srcdb/postgres.go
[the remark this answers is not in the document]
makalaaneesh: Okay. Either way this would be a one-time operation at the beginning of export-schema right?

yb-voyager/src/callhome/diagnostics.go
makalaaneesh: I think that might be a bit confusing. (it might seem like the user stop/start command multiple times). I think we should have a clearer way of representing this.
Maybe we can add another timestamp field - collected_at, which would be different for paylaods sent at different times. but then time_taken does not make sense for these type of payloads...

We can discuss this once in the stand-up 👍

priyanshi-yb: Made this change in #1555

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -74,70 +160,33 @@ func ReadEnvSendDiagnostics() {
...
-	}
-	jsonBuf, err := os.ReadFile(jsonFilePath)
-	if err != nil {
-		log.Errorf("Error while reading diagnostics.json file: %v", err)
-		return
+func readCallHomeServiceEnv() {
+	host := os.Getenv("LOCAL_CALL_HOME_SERVICE_HOST")
+	port := os.Getenv("LOCAL_CALL_HOME_SERVICE_PORT")
+	if host != "" {
+		CALL_HOME_SERVICE_HOST = host
```
sanyamsinghal, line 167: nit: 
```
CALL_HOME_SERVICE_HOST = lo.Ternary(host != "", host, CALL_HOME_SERVICE_HOST)
```

yb-voyager/src/srcdb/postgres.go
```diff
@@ -463,6 +463,17 @@ func (pg *PostgreSQL) GetCharset() (string, error) {
...
+	err := pg.db.QueryRow(query).Scan(&dbSize)
+	if err != nil {
+		return 0, fmt.Errorf("error in querying database encoding: %w", err)
+	}
+	fmt.Printf("size : %v", dbSize)
```
sanyamsinghal, line 473: i believe this will be removed before merge.

yb-voyager/src/srcdb/mysql.go
```diff
@@ -247,6 +247,16 @@ func (ms *MySQL) GetCharset() (string, error) {
...
 }
 
+func (ms *MySQL) GetDatabaseSize() (int64, error) {
+	var dbSize int64
+	query := fmt.Sprintf("SELECT SUM(data_length + index_length) FROM information_schema.tables WHERE table_schema = '%s'", ms.source.DBName)
+	err := ms.db.QueryRow(query).Scan(&dbSize)
+	if err != nil {
+		return 0, fmt.Errorf("error in querying database encoding: %w", err)
+	}
+	return dbSize, nil
```
sanyamsinghal, line 257: good to log the `dbSize` for each db type

yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -78,12 +82,16 @@ func executeSqlFile(file string, objType string, skipFn func(string, string) boo
 			}
 		}
 
-		err := executeSqlStmtWithRetries(&conn, sqlInfo, objType)
+		err = executeSqlStmtWithRetries(&conn, sqlInfo, objType)
 		if err != nil {
 			conn.Close(context.Background())
 			conn = nil
+			if !bool(tconf.ContinueOnError) && !slices.Contains(deferredSqlStmts, sqlInfo) {
```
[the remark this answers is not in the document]
priyanshi-yb: @makalaaneesh, As discussed made this change in the `executeSqlStmtWithRetries` fn itself.

yb-voyager/src/srcdb/postgres.go
```diff
@@ -463,6 +463,17 @@ func (pg *PostgreSQL) GetCharset() (string, error) {
 	return encoding, nil
 }
 
+func (pg *PostgreSQL) GetDatabaseSize() (int64, error) {
+	var dbSize int64
+	query := fmt.Sprintf("select pg_database_size('%s'); ", pg.source.DBName)
```
sanyamsinghal, line 468: @priyanshi-yb in case of postgres, we also take input of schema-list from user.
So user is migrating a subset of schema then taking full db size is not correct.

Sharing some sample queries given by chatgpt for doing that. Please verify them before using.
1.
```
SELECT
    nspname AS schema_name,
    pg_size_pretty(SUM(pg_total_relation_size(pg_class.oid))) AS size
FROM
    pg_namespace
JOIN
    pg_class ON pg_class.relnamespace = pg_namespace.oid
WHERE
    nspname IN ('schema1', 'schema2', 'schema3') -- Add your schema names here
GROUP BY
    nspname
ORDER BY
    nspname;
```

2. 
```
WITH schema_tables AS (
    SELECT table_schema AS schema_name,
           table_name
    FROM information_schema.tables
    WHERE table_schema IN ('schema1', 'schema2', 'schema3') -- Add your schema names here
),
schema_sizes AS (
    SELECT schema_name,
           ROUND(SUM(pg_total_relation_size(schema_name || '.' || table_name)) / 1024 / 1024, 2) AS size_mb
    FROM schema_tables
    GROUP BY schema_name
)
SELECT schema_name, size_mb
FROM schema_sizes;

```

yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -168,6 +166,9 @@ func executeSqlStmtWithRetries(conn **pgx.Conn, sqlInfo sqlInfo, objType string)
 				return err
 			}
 		}
+		(*conn).Close(context.Background())
```
makalaaneesh, line 169: Ah I see. when there is an error, we want to close the connection anyway, whether we report it or not, correct? 

nit: How about we write the following on line 156 (as soon as we enter the err != nil block). That way, you won't have to have the close connection logic in two places (one here, and one on line 87)
```
defer func() {
    (*conn).Close(context.Background())
    *conn = nil
}
```

makalaaneesh: or if you can close it there itself, no need of a defer.

makalaaneesh (approved):
LGTM with a nit

yb-voyager/cmd/importSchema.go
```diff
@@ -192,7 +209,46 @@ func importSchema() {
 	importSchemaCompleteEvent := createImportSchemaCompletedEvent()
 	controlPlane.ImportSchemaCompleted(&importSchemaCompleteEvent)
 
-	callhome.PackAndSendPayload(exportDir)
+	return nil
+}
+
+func packAndSendImportSchemaPayload(status string, errMsg string) {
```
sanyamsinghal, line 215: minor nit: some newlines in the function at meaningful points will increase the readability of the code.

yb-voyager/cmd/importSchema.go
```diff
@@ -192,7 +209,46 @@ func importSchema() {
...
+	payload := createCallhomePayload()
+	payload.MigrationPhase = IMPORT_SCHEMA_PHASE
+	payload.Status = status
+	targetDBDetailsBytes, err := json.Marshal(targetDBDetails)
+	if err != nil {
+		log.Errorf("callhome: error in parsing sourcedb details: %v", err)
+	}
+	payload.TargetDBDetails = string(targetDBDetailsBytes)
+	var errorsList []string
+	for _, stmt := range finalFailedSqlStmts {
```
sanyamsinghal, line 228: consider adding a comment explaining this for-loop(split and trim parts) with an example.

yb-voyager/cmd/importSchema.go
```diff
@@ -192,7 +209,46 @@ func importSchema() {
...
+		parts := strings.Split(stmt, "*/\n")
+		errorsList = append(errorsList, strings.Trim(parts[0], "/*\n"))
```
sanyamsinghal, line 230: have you tested this part, what about sql statements? I dont see the code for ignoring that.
Are they being ignored or considered as a part of error msg?

yb-voyager/cmd/exportData.go
```diff
@@ -127,29 +128,78 @@ func exportDataCommandFn(cmd *cobra.Command, args []string) {
...
+	if !callhome.SendDiagnostics {
+		return
+	}
+	payload := createCallhomePayload()
+
+	switch exportType {
+	case SNAPSHOT_ONLY:
+		payload.MigrationType = OFFLINE
+	case SNAPSHOT_AND_CHANGES:
+		payload.MigrationType = LIVE_MIGRATION
```
sanyamsinghal, line 160: How/where are we telling in case FF/FB is enabled?

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -33,35 +32,122 @@ import (
...
+}
+
+type TargetDBDetails struct {
+	Host      string `json:"host"`
+	DBVersion string `json:"db_version"`
+	NodeCount int    `json:"node_count"`
+	Cores     int    `json:"cores"`
+}
+
+type AssessMigrationPhasePayload struct {
```
sanyamsinghal, line 89: We should send `SchemaSummary` also here. Because we wont be having the analyze schema output at this point.

yb-voyager/cmd/common.go
```diff
@@ -989,3 +992,83 @@ func (ar *AssessmentReport) GetClusterSizingRecommendation() string {
...
+		packAndSendExportDataPayload(EXIT)
+	case importDataCmd.CommandPath():
+		packAndSendImportDataPayload(EXIT)
+	case endMigrationCmd.CommandPath():
+		packAndSendEndMigrationPayload(EXIT)
+		//..more cases
+	}
+}
+
+func updateExportSnapshotDataStatsInPayload(exportDataPayload *callhome.ExportDataPhasePayload) {
```
sanyamsinghal, line 1031: please add a comment explaining this function

yb-voyager/cmd/common.go
```diff
@@ -989,3 +992,83 @@ func (ar *AssessmentReport) GetClusterSizingRecommendation() string {
...
+				}
+				return true, nil
+			})
+		}
+		switch source.DBType {
+		case POSTGRESQL:
+			exportDataPayload.ExportDataMechanism = "pg_dump"
+		case ORACLE, MYSQL:
+			exportDataPayload.ExportDataMechanism = "ora2pg"
+		}
```
sanyamsinghal, line 1072: wont this override the action of line no:1046
```
exportDataPayload.ExportDataMechanism = "debezium"
```

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: I was thinking this `total_db_size` in source_db_details is the total database size of the pg for which migration is happening not only the schemas that are getting migrated because if we go with schemas specific then it is already present in assess-migration's table/index sizing stats in some way. So having this total_db_size as the total pg db size is fine. We can this one in standup with others as well and if required I can modify this later.

yb-voyager/cmd/importSchema.go
priyanshi-yb: Yes, Its only taking the error msg from each entry and not the SQL stmts.

yb-voyager/cmd/exportData.go
priyanshi-yb: It will come up with live migration-specific upcoming PR as anyways I can't get the information directly here about whether it will be with FF/FB it has to be extra which is TODO

yb-voyager/cmd/common.go
priyanshi-yb: No, its in else block which is for non-debezium case

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -33,35 +32,122 @@ import (
...
+type Payload struct {
+	MigrationUUID    uuid.UUID `json:"migration_uuid"`
+	PhaseStartTime   string    `json:"phase_start_time"`
+	SourceDBDetails  string    `json:"source_db_details"`
+	TargetDBDetails  string    `json:"target_db_details"`
+	YBVoyagerVersion string    `json:"yb_voyager_version"`
+	MigrationPhase   string    `json:"migration_phase"`
+	PhasePayload     string    `json:"phase_payload"`
+	MigrationType    string    `json:"migration_type"`
+	TimeTakenSec     float64   `json:"time_taken_sec"`
```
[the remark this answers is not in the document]
sanyamsinghal: I understand that but for those quickly completed cmds we don't even need the time. Its more meaningful for cmds like export data....
You can set default to 1sec if it is coming up as 0, or just roundoff mill seconds to 1 seconds while actual putting it in the payload

yb-voyager/src/srcdb/postgres.go
sanyamsinghal: Sure we can discuss this in standup.
But how is it of any use to consider for the schemas we are not migrating them.

Callhome user might draw wrong conclusions if some large schemas are present which are not migrated

yb-voyager/cmd/common.go
sanyamsinghal: oh ok, i misread it

sanyamsinghal (approved):
LGTM

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -53,7 +53,7 @@ CREATE TABLE diagnostics (
 	migration_phase TEXT,
 	phase_payload JSONB,
 	migration_type TEXT,
-	time_taken_sec NUMERIC(20,2),
+	time_taken_sec bigint,
```
sanyamsinghal, line 56: even `int`or `unsigned int` is enough here.
Because `2147483647 seconds` --> `~68 years`

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -68,7 +68,7 @@ type Payload struct {
 	MigrationPhase   string    `json:"migration_phase"`
 	PhasePayload     string    `json:"phase_payload"`
 	MigrationType    string    `json:"migration_type"`
-	TimeTakenSec     float64   `json:"time_taken_sec"`
+	TimeTakenSec     int64     `json:"time_taken_sec"`
```
sanyamsinghal, line 71: similarly here

sanyamsinghal (commented):
@priyanshi-yb a small comment regarding `bigint`

yb-voyager/src/callhome/diagnostics.go
priyanshi-yb: yeah, makes sense. Made the change

PR #1555, opened 2024-06-11
Added support to send callhome payload for import-data-file and periodically for long running data commands

priyanshi-yb:
**Description :**

This patch is follow-up to the #1548 for adding few more enhancements -

1. Added support for `import data file` to send call-home payload and also added the specific information relevant for bulk data load from file workflow like file storage type (AWS, GCS, Azure, local), and datafile input parameters.  
https://yugabyte.atlassian.net/browse/DB-11687

3. Added support for sending the payload periodically multiple times for long-running data commands like export/import data and import data file.
https://yugabyte.atlassian.net/browse/DB-11565
    -  for this to accommodate updated the table schema to add a column uniquely identify a payload for a phase having same start_time -
    
          ```
          CREATE TABLE diagnostics (
	          migration_uuid UUID,
	          phase_start_time TIMESTAMP WITH TIME ZONE,
	          collected_at TIMESTAMP WITH TIME ZONE,
	          ....
	          PRIMARY KEY (migration_uuid, phase_start_time, migration_phase, collected_at)
          );
          ```
    
    - Kept interval as 20 mins for sending payload for long-running command. 
Example of both them is here https://docs.google.com/document/d/1lZd6ih83wrV5W-Jt09WzdANoN6KZALkq5WJs78U1apI/edit#heading=h.m3i7aohk2mwo 

3. Modified PG query for `GetDatabaseSize()` to return the size of schemas that are passed.

#=========================================================================================

**Tests:**

Automation - TODO
Manual - Tested the import data file command to send payload properly with all parameters and long-running data commands (export/import data and import data file) to send payload at the interval specified with IN-PROGRESS status.

Jenkins runs - 
1. https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2702
2. https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2703


yb-voyager/cmd/common.go
```diff
@@ -922,6 +923,34 @@ func getImportedSnapshotRowsMap(dbType string) (*utils.StructMap[sqlname.NameTup
 	return snapshotRowsMap, nil
 }
 
+func getImportedSizeMap() (*utils.StructMap[sqlname.NameTuple, int64], error) { //used for import data file case right now
```
makalaaneesh, line 926: Isn't there an existing function that we can use for this (that import-data status would be using) ?

priyanshi-yb: Not exactly, import-data-status is handling this differently by going to each datafile and preparing table to display. we can't use it.

yb-voyager/cmd/common.go
```diff
@@ -1076,3 +1108,26 @@ func updateExportSnapshotDataStatsInPayload(exportDataPayload *callhome.ExportDa
 		}
 	}
 }
+
+func sendCallhomePayloadAtIntervals(ctx context.Context) {
```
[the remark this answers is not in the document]
priyanshi-yb: yeah for 2nd point, I also don't have strong points towards tickers. I can use for loop with a sleep as well not a problem.

yb-voyager/cmd/common.go
```diff
@@ -1076,3 +1108,26 @@ func updateExportSnapshotDataStatsInPayload(exportDataPayload *callhome.ExportDa
...
+	sendTicker := time.NewTicker(20 * time.Minute) //TODO: confirm if this is fine
```
sanyamsinghal: i feel 20mins is lot, 10-15mins is fine i think

yb-voyager/cmd/common.go
sanyamsinghal: how/when  is this function closed? 
Is it possible that command has also completed but this func will still be sending `IN-PROGRESS` ?

yb-voyager/cmd/constants.go
```diff
@@ -65,13 +65,22 @@ const (
...
-	OFFLINE              = "offline"
-	LIVE_MIGRATION       = "live migration"
+	OFFLINE        = "offline"
+	LIVE_MIGRATION = "live migration"
+	BULK_DATA_LOAD = "bulk data load from flat files"
+
+	AWS_S3      = "AWS-S3"
+	GCS_BUCKETS = "GCS-Buckets"
+	AZURE_BLOBS = "Azure-blob-storage"
+	LOCAL_DISK  = "Local-disk"
```
sanyamsinghal, line 77: Might be minor nit, but using mix and non-defined casing(upper/lower) might be confusing and hard to remember at user's end who is consuming this info, built an app based on it.

My approach is to use smaller cases wherever possible, unless its really required to use upper case.
Here everything i think can be lowercase, like `aws-s3`, gcs-buckets`, `local-disk`

yb-voyager/cmd/importDataFileCommand.go
```diff
@@ -322,6 +326,68 @@ func checkAndParseEscapeAndQuoteChar() {
 
 }
 
+func packAndSendImportDataFilePayload(status string) {
+	if !callhome.SendDiagnostics {
+		return
+	}
```
sanyamsinghal, line 332: just thinking, instead of having `SendDiagnostics` flag check everywhere, can we let this functions run.
And at the end, where we send the payload, just check this flag and proceed based on that.

One problem with that approach i think is unnecessary processing if we don't need to send, but depends how costly that is....

What do you think?

yb-voyager/cmd/importDataFileCommand.go
```diff
@@ -322,6 +326,68 @@ func checkAndParseEscapeAndQuoteChar() {
...
+		HasHeader:  bool(hasHeader),
+		Delimiter:  delimiter,
+		EscapeChar: escapeChar,
+		QuoteChar:  quoteChar,
+		NullString: nullString,
+	}
+	parametersBytes, err := json.Marshal(dataFileParameters)
+	if err != nil {
+		log.Errorf("callhome: error in parsing dataFileParameters: %v", err)
+	}
```
sanyamsinghal, line 352: @priyanshi-yb since `DataFileParameters` is a struct, how about defining a method in it.
`func (dfp *DataFileParameters) MarshalledJsonString() string`
or some better method name....

which can be directly used:
```
importDataFilePayload := callhome.ImportDataFilePhasePayload{
		ParallelJobs:       int64(tconf.Parallelism),
		StartClean:         bool(startClean),
		DataFileParameters: dataFileParameters.MarshalledJsonString(),
	}
```

You can similarly do this for other structs also, this will abstract out the code at the cmd.go level.

yb-voyager/cmd/importDataFileCommand.go
```diff
@@ -322,6 +326,68 @@ func checkAndParseEscapeAndQuoteChar() {
...
+			if value > importDataFilePayload.LargestTableSize {
+				importDataFilePayload.LargestTableSize = value
+			}
+			return true, nil
+		})
+	}
+	importDataFilePayloadBytes, err := json.Marshal(importDataFilePayload)
+	if err != nil {
+		log.Errorf("callhome: error in parsing the export data payload: %v", err)
+	}
```
sanyamsinghal, line 383: similar suggestion as above for `importDataFilePayload` struct

yb-voyager/cmd/common.go
```diff
@@ -922,6 +923,34 @@ func getImportedSnapshotRowsMap(dbType string) (*utils.StructMap[sqlname.NameTup
...
+	for dataFilePath, nt := range dataFilePathNtMap {
+		byteCount, err := state.GetImportedByteCount(dataFilePath, nt)
+		if err != nil {
+			return nil, fmt.Errorf("could not fetch snapshot row count for table %q: %w", nt, err)
+		}
+		exisitingByteCount, _ := snapshotRowsMap.Get(nt)
+		snapshotRowsMap.Put(nt, exisitingByteCount+byteCount)
+	}
+	return snapshotRowsMap, nil
+}
```
sanyamsinghal: can we merge these two for-loops?

yb-voyager/cmd/common.go
```diff
@@ -903,7 +903,37 @@ func getImportedSnapshotRowsMap(dbType string) (*utils.StructMap[sqlname.NameTup
...
+func getImportedSizeMap() (*utils.StructMap[sqlname.NameTuple, int64], error) { //used for import data file case right now
+	importerRole = IMPORT_FILE_ROLE
+	state := NewImportDataState(exportDir)
+	dataFileDescriptor, err := prepareDummyDescriptor(state)
+	if err != nil {
+		return nil, fmt.Errorf("prepare dummy descriptor: %w", err)
+	}
+	snapshotRowsMap := utils.NewStructMap[sqlname.NameTuple, int64]()
+	dataFilePathNtMap := map[string]sqlname.NameTuple{}
+	for _, fileEntry := range dataFileDescriptor.DataFileList {
```
sanyamsinghal, line 936: can we merge these two for-loops?

```
for _, fileEntry := range dataFileDescriptor.DataFileList

for dataFilePath, nt := range dataFilePathNtMap
```

yb-voyager/cmd/common.go
priyanshi-yb: okay , can make it to 15 mins

yb-voyager/cmd/common.go
priyanshi-yb: Yes, handled already that now that there is one var keeps track of the complete payload sent `callhomeCompletePayloadSent` and uses it to terminate this function.

yb-voyager/cmd/importDataFileCommand.go
priyanshi-yb: I think it's better to have this extra check as to prevent even packing the payload to avoid any unnecessary or unknown failures.

yb-voyager/cmd/constants.go
priyanshi-yb: okay, I think only these four are the ones I kept mixed cases, will make it lower.

yb-voyager/cmd/importDataFileCommand.go
priyanshi-yb: Yeah, thats nice. Thanks. I have implemented a single generic function for all types to marshall.

yb-voyager/cmd/common.go
priyanshi-yb: Done.

sanyamsinghal (approved):
LGTM

PR #1556, opened 2024-06-11
Added support for detecting Unsupported Features(Virtual Columns, Unsupported Index types, Inherited Types) of Oracle

sanyamsinghal:
- Added support for reporting Unsupported Datatypes in Oracle


sanyamsinghal:
Jenkins Pipelines:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2709/

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -561,45 +567,92 @@ func getAssessmentReportContentFromAnalyzeSchema() (err error) {
...
 
-func fetchUnsupportedFeaturesOfSourceInYb(schemaAnalysisReport utils.SchemaReport) ([]UnsupportedFeature, error) {
-	switch source.DBType {
-	case POSTGRESQL:
-		return fetchUnsupportedFeaturesForPG(schemaAnalysisReport)
-	case ORACLE:
-		return nil, nil
-	default:
-		panic("unsupported source db type")
+var filterIssues = func(featureName, issueReason string, schemaAnalysisReport utils.SchemaReport, unsupportedFeatures *[]UnsupportedFeature) {
```
makalaaneesh, line 588: I would suggest defining this as a proper function. Don't see why it has to be a variable? 
we can call it `findAndAddUnsupportedFeature`

makalaaneesh, line 588: You shouldn't have to pass a pointer to an array. You are only appending.
For slices, the actual array will be effectively passed by reference.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -561,45 +567,92 @@ func getAssessmentReportContentFromAnalyzeSchema() (err error) {
...
+	log.Infof("fetching unsupported features for Oracle...")
+	unsupportedFeatures := make([]UnsupportedFeature, 0)
+	filterIssues("Compound Triggers", COMPOUND_TRIGGER_ISSUE_REASON, schemaAnalysisReport, &unsupportedFeatures)
+	return unsupportedFeatures, nil
+}
+
+var OracleUnsupportedIndexTypes = []string{"CLUSTER INDEX", "DOMAIN INDEX", "BITMAP INDEX", "FUNCTION-BASED DOMAIN INDEX", "IOT - TOP INDEX", "NORMAL/REV INDEX", "FUNCTION-BASED NORMAL/REV INDEX"}
+
+func fetchUnsupportedObjectTypes() ([]UnsupportedFeature, error) {
+	if source.DBType != ORACLE {
```
makalaaneesh, line 619: In pg also we have multiple types of indexes(INET) and inherited tables, etc that are not supported right? Where are those caught? I would have expected them also to be here itself.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -631,10 +684,19 @@ func fetchColumnsWithUnsupportedDataTypes() ([]utils.TableColumnsDataTypes, erro
 		allColumnsDataTypes = append(allColumnsDataTypes, columnDataTypes)
 	}
 
-	// filter columns with unsupported data types using srcdb.PostgresUnsupportedDataTypesForDbzm
-	pgUnsupportedDataTypes := srcdb.PostgresUnsupportedDataTypesForDbzm
+	var sourceUnsupportedDataTypes []string
+	switch source.DBType {
+	case POSTGRESQL:
+		sourceUnsupportedDataTypes = srcdb.PostgresUnsupportedDataTypesForDbzm
```
makalaaneesh, line 690: This might not be entirely accurate, right? (we seem to be picking the unsupported types for dbzm...) 

I'm not 100% sure how to report this, because it's a little complicated, let's dicuss this

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -631,10 +684,19 @@ func fetchColumnsWithUnsupportedDataTypes() ([]utils.TableColumnsDataTypes, erro
...
+	case ORACLE:
+		sourceUnsupportedDataTypes = srcdb.OracleUnsupportedDataTypes
+	default:
+		panic(fmt.Sprintf("invalid source db type %q", source.DBType))
+	}
+
+	// filter columns with unsupported data types using sourceUnsupportedDataTypes
 	for i := 0; i < len(allColumnsDataTypes); i++ {
-		if utils.ContainsAnySubstringFromSlice(pgUnsupportedDataTypes, allColumnsDataTypes[i].DataType) {
+		if utils.ContainsAnySubstringFromSlice(sourceUnsupportedDataTypes, allColumnsDataTypes[i].DataType) {
```
makalaaneesh, line 699: why are we doing substring match here? Why not exact match?

yb-voyager/src/srcdb/data/gather-assessment-metadata/oracle/index-to-table-mapping.sqlplus
```diff
@@ -30,7 +30,8 @@ ON COMMIT PRESERVE ROWS AS
             WHERE owner = '&1'
             AND constraint_type IN ('P', 'U') -- ignoring primary and unique keys
         )
-        AND i.index_name NOT LIKE 'SYS_%';
+        AND i.index_name NOT LIKE 'SYS_%'
+        AND i.index_name NOT LIKE 'DR$%';
```
makalaaneesh, line 34: What is the context here? Pls add a comment

yb-voyager/src/srcdb/data/gather-assessment-metadata/oracle/object-type-mapping.sqlplus
```diff
@@ -34,7 +34,29 @@ ON COMMIT PRESERVE ROWS AS
...
+    FROM
+        dba_types
+    WHERE
+        owner = '&1'
+        AND supertype_name IS NOT NULL
+    UNION ALL
+    SELECT
+        owner AS schema_name,
+        table_name || '.' || column_name AS object_name,
+        'VIRTUAL COLUMN' AS object_type
```
makalaaneesh, line 52: Does it make sense to add Virtual columns to "objects" ? It seems to be more suitable for table-columns-data-types?

yb-voyager/src/srcdb/oracle.go
```diff
@@ -512,7 +512,7 @@ func (ora *Oracle) GetColumnsWithSupportedTypes(tableList []sqlname.NameTuple, u
 		var unsupportedColumnNames []string
 		for i := 0; i < len(columns); i++ {
 			isUdtWithDebezium := (dataTypesOwner[i] == sname) && useDebezium // datatype owner check is for UDT type detection as VARRAY are created using UDT
-			if isUdtWithDebezium || utils.ContainsAnySubstringFromSlice(oracleUnsupportedDataTypes, dataTypes[i]) {
+			if isUdtWithDebezium || utils.ContainsAnySubstringFromSlice(OracleUnsupportedDataTypes, dataTypes[i]) {
```
makalaaneesh, line 515: Even this is something we need to think about. This will only come up when some one runs export-data, not when assess-migration is run.  

In general, the list of unsupported data types depends on a bunch of factors: 
- pg_dump/debezium (for PG)
- offline or live (for both PG and oracle)
- forward migration or fall-forward or fall-back. 

Maybe for now it's okay to not mention the specifics, but let us discuss so that we are all on the same page about it.

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: I actually talked to @priyanshi-yb  about this, since we also use Substring match in oracle export data codepath.
The reason which she mentioned and makes sense to me also::
> Ex like as we nvarchar2 which can also contain its limit like NVARCHAR2 (100) something like this and I am not sure but I think in db it's type name will be complete with limits as well so we should catch this as well using substring

PR #1558, opened 2024-06-13
[DB-11611] Added logging in assessment gather metadata bash script for both Oracle and PostgreSQL

sanyamsinghal:
- will help in case of debugging any failure during the assess-migration command


sanyamsinghal:
Jenkins:
1. https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2718/

migtests/scripts/live-migration-fallb-run-test.sh
```diff
@@ -67,7 +67,10 @@ main() {
 
 	step "Assess Migration"
 	if [ "${SOURCE_DB_TYPE}" = "postgresql" ]; then
-		assess_migration
+		assess_migration || {
+			cat_log_file "yb-voyager-assess-migration.log"
+			cat_file ${EXPORT_DIR}/assessment/metadata/yb-voyager-assessment.log
```
makalaaneesh, line 72: why is the log file under `assessment/metadata` ?

makalaaneesh: I see, I get it now. You only have access to the assessment metadir from the assessment script. That makes sense 👍  . 
Nit: maybe put it inside a /log dir?

makalaaneesh (approved):
LGTM with a small nit. 
I'm assuming that the core logic of the assessment script did not change( you only moved it to a main func), so not reviewing that part.

migtests/scripts/live-migration-fallb-run-test.sh
sanyamsinghal: yeah, but one more `/log` dir will increase the nesting, i thought to keep it outside so that its visible to users upfront. Plus we wont have more than one log file, same is appended everytime...

makalaaneesh: sounds good

PR #1561, opened 2024-06-18
Add support in YugabyteD assess-migration payload to also send some extra details with json report

priyanshi-yb:
Description:
This patch modifies the payload sent to YugabyteD by adding a few extra details in addition to the assessment report such as Source sizing details (total db size, total index size, total table size, total table row count), target recommendations (total colocated size, total sharded size ), analyze-schema issues, and Migration Complexity (TODO to figure out the algorithm).
https://yugabyte.atlassian.net/browse/DB-11775

Tests:
Tested PG/Oracle source for assess-migration.
Automation- N/A


yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -352,15 +352,49 @@ func createMigrationAssessmentStartedEvent() *cp.MigrationAssessmentStartedEvent
...
-		utils.PrintAndLog("Failed to serialise the assessment report to json (ERR IGNORED): %s", err)
+		utils.PrintAndLog("Failed to get the colocated tables recommendation from assessmentReport: %v", err)
+	}
+	if assessmentReport.TableIndexStats != nil {
+		for _, stat := range *assessmentReport.TableIndexStats {
+			if stat.IsIndex {
+				finalReport.SourceSizeDetails.TotalIndexSize += *stat.SizeInBytes
+			} else {
+				finalReport.SourceSizeDetails.TotalTableSize += *stat.SizeInBytes
+				finalReport.SourceSizeDetails.TotalTableRowCount += *stat.RowCount
```
makalaaneesh, line 375: hope Nikhil is aware, that row count is now just 0? (we don't actually fetch row count)

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -352,15 +352,49 @@ func createMigrationAssessmentStartedEvent() *cp.MigrationAssessmentStartedEvent
...
+				if slices.Contains(colocatedTables, fmt.Sprintf("%s.%s", stat.SchemaName, stat.ObjectName)) {
```
makalaaneesh, line 376: hope we won't run into case sensitivity issues here? Maybe we should have represented it as NameTuples?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -352,15 +352,49 @@ func createMigrationAssessmentStartedEvent() *cp.MigrationAssessmentStartedEvent
...
+		ConversionIssues:      schemaAnalysisReport.Issues,
+	}
+	colocatedTables, err := assessmentReport.GetColocatedTablesRecommendation()
 	if err != nil {
-		utils.PrintAndLog("Failed to serialise the assessment report to json (ERR IGNORED): %s", err)
+		utils.PrintAndLog("Failed to get the colocated tables recommendation from assessmentReport: %v", err)
+	}
+	if assessmentReport.TableIndexStats != nil {
+		for _, stat := range *assessmentReport.TableIndexStats {
+			if stat.IsIndex {
```
makalaaneesh, line 371: It might make sense to define these methods `getTotalIndexSize`, `getTotalTableSize` , etc on the AssessmentReport struct itself (it would be the right abstraction, and would be useful if we need these calculations somewhere else..)

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: Yes

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: I thought of having a single loop of iterating this `TableIndexStats` for all the calculation, otherwise we have to iterate it 4 times if we have to separate it out for these, I think that might not be a good idea in case there are a lot of objects (tables and indexes). Maybe I can separate this whole calculation in separate function and return all the values?

priyanshi-yb: Made the change, to separate out this in a separate function.

yb-voyager/cmd/common.go
```diff
@@ -924,6 +924,12 @@ func getImportedSnapshotRowsMap(dbType string) (*utils.StructMap[sqlname.NameTup
 	return snapshotRowsMap, nil
 }
 
+func safeDereferenceInt64(ptr *int64) int64 {
+	if ptr != nil {
+		return *ptr
+	}
+	return 0
+}
```
sanyamsinghal, line 932: For now it is fine, but how about moving it to `utils.go` so that it can be used anywhere in future, no cyclic dependency issue will be there.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -327,15 +327,74 @@ func createMigrationAssessmentStartedEvent() *cp.MigrationAssessmentStartedEvent
...
+				details.TotalTableSize += safeDereferenceInt64(stat.SizeInBytes)
+				details.TotalTableRowCount += safeDereferenceInt64(stat.RowCount)
+				if slices.Contains(colocatedTables, fmt.Sprintf("%s.%s", stat.SchemaName, stat.ObjectName)) {
+					details.TotalColocatedSize += safeDereferenceInt64(stat.SizeInBytes)
+				} else {
+					details.TotalShardedSize += safeDereferenceInt64(stat.SizeInBytes)
+				}
+			}
+		}
+	}
```
sanyamsinghal, line 394: how about defining methods on `AssessmentReport` struct for `GetTotalRowCount()`, `GetTotalColocationSize()` etc...

sanyamsinghal: I see Aneesh also had same comment, i get your point of avoid loops, it just that methods make it more structures and abstract.
Also the time complexity wise, theoretically i don't think a lot of difference - O(N) vs O(4N)

priyanshi-yb: I already made a change to separate it out in a way that will calculate size details and return all of them, so anything can be as per requirement. Also, we can more parameters to it later if required without having a separate func of each of them

PR #1563, opened 2024-06-19
Bug fix: Incorrect messaging of assess-migration cmd in case of MySQL db type saying it as supported and PG/Oracle as unsupported

sanyamsinghal:
- fixed the error messaging


shubham-yb (approved):
LGTM

PR #1565, opened 2024-06-19
Updated SQLPlus scripts to not use GTT which required CREATE TABLE privilege to the migration user, rather use inline views for fetching metadata

sanyamsinghal:
Automation tests were failing because `ybvoyager` (migration user) does not have CREATE TABLE privileges, so during gather script execution the assess migration command was failing repeatedly.


priyanshi-yb (approved):
LGTM

PR #1568, opened 2024-06-20
Sizing Calc: Updated import time calculation for sharded tables based on indexes on the table

shaharuk-yb:
- Updated import time calculation for sharded tables based on indexes on the table. For sharded table:
- Import time is calculated for every table and according to the number of indexes, the multiplication factor is applied on import time. Size of the index is not taken into consideration. (In colocated case, we are still using a old logic which works on cumulative size of the colocated tables + respective indexes)
- The multiplication factor calculation is like this(roughly derived from the experiment data):
1 index ==> importTime * 2
2 indexes ==> importTime * 3
indexes >= 3 ==> importTime * (numberOfIndexes * 0.8) 
- added unit tests to test the new function.


yb-voyager/src/migassessment/sizing.go
```diff
@@ -863,6 +875,151 @@ func calculateTimeTakenAndParallelJobsForImport(tableName string, dbObjects []So
...
+
+	table: Metadata for the database table for which the multiplication factor is to be calculated.
+	sourceIndexMetadata: A slice containing metadata for the indexes in the database.
+
+Returns:
+
+	float64: The multiplication factor for import time based on the number of indexes on the table.
+*/
+func getMultiplicationFactorForImportTimeBasedOnIndexes(table SourceDBMetadata, sourceIndexMetadata []SourceDBMetadata) float64 {
+	var numberOfIndexesOnTable float64 = 0
```
hbhanawat, line 1003: We should not hardcode this. We should have the index numbers in database and use them for calculation. Things can change release over release

hbhanawat: We cam merge this PR for Dev build. But the change has to happen before the release

hbhanawat (approved):
Approved but please create a task for my comment

yb-voyager/src/migassessment/sizing.go
shaharuk-yb: sure. Created another task for this.

PR #1573, opened 2024-06-21
sizing calc: handle null values for reads_per_sec and writes_per_sec

shaharuk-yb:
- handle null values in columns **reads**, **writes**, **reads_per_sec** and **writes_per_sec** in table **table_index_stats** from **assessment.db**(source database)
- using ternary operator to treat null as 0 in the code.
- adjusted unit tests accordingly to use the updated data type
- Associated JIRA task: https://yugabyte.atlassian.net/browse/DB-11877

### DIFF
--- a/yb-voyager/src/migassessment/sizing_test.go
+++ b/yb-voyager/src/migassessment/sizing_test.go
@@ -358,12 +358,12 @@ func TestShardingBasedOnTableSizeAndCount_NoColocatedTables(t *testing.T) {
 func TestShardingBasedOnOperations_CanSupportOpsRequirement(t *testing.T) {
 	// Define test data
 	sourceIndexMetadata := []SourceDBMetadata{
-		{ObjectName: "table1", Size: 10.0, ReadsPerSec: 100, WritesPerSec: 50},
+		{ObjectName: "table1", Size: 10.0, ReadsPerSec: sql.NullInt64{Valid: true, Int64: 100}, WritesPerSec: sql.NullInt64{Valid: true, Int64: 50}},
 	}
 	recommendation := map[int]IntermediateRecommendation{
 		4: {
 			ColocatedTables: []SourceDBMetadata{
-				{ObjectName: "table1", Size: 10.0, ReadsPerSec: 100, WritesPerSec: 50},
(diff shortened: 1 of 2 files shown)


yb-voyager/src/migassessment/sizing_test.go
```diff
@@ -459,8 +459,8 @@ func TestCheckShardedTableLimit_WithinLimit(t *testing.T) {
 	recommendation := map[int]IntermediateRecommendation{
 		16: {
 			ShardedTables: []SourceDBMetadata{
-				{SchemaName: "public", ObjectName: "table1", Size: 10.0, ReadsPerSec: 100, WritesPerSec: 50},
-				{SchemaName: "public", ObjectName: "table2", Size: 10.0, ReadsPerSec: 100, WritesPerSec: 50},
+				{SchemaName: "public", ObjectName: "table1", Size: 10.0, ReadsPerSec: sql.NullInt64{Valid: true, Int64: 100}, WritesPerSec: sql.NullInt64{Valid: true, Int64: 50}},
+				{SchemaName: "public", ObjectName: "table2", Size: 10.0, ReadsPerSec: sql.NullInt64{Valid: true, Int64: 100}, WritesPerSec: sql.NullInt64{Valid: true, Int64: 50}},
```
sanyamsinghal, line 463: I see this repeated multiple times: 
1. `sql.NullInt64{Valid: true, Int64: 100}`
2. `sql.NullInt64{Valid: true, Int64: 50}}`

Maybe a global const variable or something and reusing it at all places will be helpful?

sanyamsinghal (approved):
LGTM with a minor comment.
Just check if using a common/global variable will be better or not.

yb-voyager/src/migassessment/sizing_test.go
shaharuk-yb: I agree that some values are reused in test case and we can create a constant. But, what I've learned over the years is that the test data should be isolated and self explanatory. Also, there are tests which are using different values. Hence, decided to add multiple occurrences. I'm also thinking of creating a separate fixtures/data file for this test data, this is something I plan to do in next cycle.

PR #1574, opened 2024-06-21
Improved logging in case of Oracle SQLPlus scripts to print the error on console and in LOG_FILE

sanyamsinghal:
previously the real error message was getting suppressed because for SPOOLing we disable the terminal outputs in sqlplus, just SPOOL to the csv file

Before this change, it was hard to debug if there turns out any issue in the SQL query for metadata gathering since the actual sql error message was not visible.


priyanshi-yb (approved):
LGTM

PR #1580, opened 2024-06-24
Change enable-upsert default to false

makalaaneesh:
- Change default value of enable-upsert to false as it is unsafe when secondary indexes are present. Added warning to help msg as well 
- In snapshot phase of import-data, we always check if the batch has already been imported. If so, we skip it.
- In CDC phase of import-data, if there are duplicate events (because debezium ensures at-least-once delivery), then we were relying on upsert_mode to ensure that we do not fail. Since now, upsert_mode will be set to false by default, I have added the ON CONFLICT DO NOTHING clause to INSERT statements.

### DIFF
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -207,7 +207,7 @@ func registerImportDataCommonFlags(cmd *cobra.Command) {
 		"path of the file containing the list of the source db table names to import data")
 
 	BoolVar(cmd.Flags(), &tconf.EnableUpsert, "enable-upsert", false,
-		"Enable UPSERT mode on target tables")
+		"Enable UPSERT mode on target tables. WARNING: Ensure that tables on target YugabyteDB do not have secondary indexes. If a table has secondary indexes, setting this flag to true may lead to corruption of the indexes. (default false)")
 	BoolVar(cmd.Flags(), &tconf.UsePublicIP, "use-public-ip", false,
 		"Use the public IPs of the nodes to distribute --parallel-jobs uniformly for data import (default false)\n"+
 			"Note: you might need to configure database to have public_ip available by setting server-broadcast-addresses.\n"+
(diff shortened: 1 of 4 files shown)


shubham-yb (commented):
LGTM. Please wait for @priyanshi-yb to take a look as well.

yb-voyager/cmd/import.go
```diff
@@ -206,8 +206,8 @@ func registerImportDataCommonFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&tableListFilePath, "table-list-file-path", "",
 		"path of the file containing the list of the source db table names to import data")
 
-	BoolVar(cmd.Flags(), &tconf.EnableUpsert, "enable-upsert", true,
-		"Enable UPSERT mode on target tables")
+	BoolVar(cmd.Flags(), &tconf.EnableUpsert, "enable-upsert", false,
```
priyanshi-yb, line 209: We use this Flag for streaming in Oracle targetDB for upserting any insert and marking this false here itself will not be a good idea.

yb-voyager/src/tgtdb/event.go
```diff
@@ -282,7 +282,7 @@ func (event *Event) getPreparedInsertStmt(tdb TargetDB, targetDBType string) (st
 	columns := strings.Join(columnList, ", ")
 	values := strings.Join(valueList, ", ")
 	stmt := fmt.Sprintf(insertTemplate, event.TableNameTup.ForUserQuery(), columns, values)
-	if targetDBType == POSTGRESQL {
+	if targetDBType == POSTGRESQL || targetDBType == YUGABYTEDB {
```
priyanshi-yb, line 285: Either we should use add the flag check here as well or remove it from everywhere and use upsert behavior directly without the check for events.

priyanshi-yb (approved):
LGTM

PR #1582, opened 2024-06-25
Support case sensitive/reserved table names in assess-migration

makalaaneesh:


makalaaneesh:
Dropping this for now. Needs more implementation to support for partitions, will take this up later.

PR #1587, opened 2024-06-26
Added test for a non-public target-db-schema in import data file tests

priyanshi-yb:


migtests/scripts/functions.sh
```diff
@@ -456,7 +456,6 @@ import_data_file() {
 		--target-db-host ${TARGET_DB_HOST} \
 		--target-db-port ${TARGET_DB_PORT} \
 		--target-db-user ${TARGET_DB_USER} \
-		--target-db-schema "public" \
```
makalaaneesh, line 459: is the default "public" if not specified? 
Should we maybe use an env var SIMILAR to the other args?

makalaaneesh (approved):
LGTM with a comment

migtests/scripts/functions.sh
priyanshi-yb: > is the default "public" if not specified?
yes its default public

PR #1588, opened 2024-06-26
sizing calc: db-11948: handle null values for size_in_bytes column in source metadata table_index_stats

shaharuk-yb:
- associated JIRA task: https://yugabyte.atlassian.net/browse/DB-11948
- handles null values for size_in_bytes column
- adjusted unit tests accordingly
- tested by setting null in that column.


sanyamsinghal (approved):
lgtm, thanks for the quick fix!

PR #1589, opened 2024-06-26
[Callhome] Added CommandLineArgs field in the phase payload to send the cli arguments changed in the command 

priyanshi-yb:
1. Added CommandLineArgs field in the phase payload to send the cli arguments changes in the command 
2. Misc Callhome fixes-
       - Nil pointer handling for statsReporter in import data payloads
       - In case of debezium export data parallel-jobs should be stored as 1.
       - Populated MigrationType field for end-migration
       - Populated `node_count` and `total_cores` in import schema

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -605,8 +605,9 @@ func packAndSendImportDataPayload(status string) {
 	payload.TargetDBDetails = callhome.MarshalledJsonString(targetDBDetails)
 	payload.MigrationPhase = IMPORT_DATA_PHASE
 	importDataPayload := callhome.ImportDataPhasePayload{
-		ParallelJobs: int64(tconf.Parallelism),
-		StartClean:   bool(startClean),
+		ParallelJobs:    int64(tconf.Parallelism),
+		StartClean:      bool(startClean),
+		CommandLineArgs: cliArgsString,
 	}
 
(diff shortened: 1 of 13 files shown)


yb-voyager/src/callhome/diagnostics.go
```diff
@@ -129,13 +131,15 @@ type ExportDataPhasePayload struct {
 	TotalExportedEvents int64  `json:"total_exported_events,omitempty"`
 	EventsExportRate    int64  `json:"events_export_rate_3m,omitempty"`
 	LiveWorkflowType    string `json:"live_workflow_type,omitempty"`
+	CommandLineArgs     string `json:"command_line_args,omitempty"`
```
ShivanshGahlot, line 134: Do we need omitempty here? We haven't used it in other structs.

yb-voyager/cmd/importData.go
```diff
@@ -107,6 +107,8 @@ var importDataToTargetCmd = &cobra.Command{
 }
 
 func importDataCommandFn(cmd *cobra.Command, args []string) {
+
+	importPhase = dbzm.MODE_SNAPSHOT
```
ShivanshGahlot, line 111: Why have we exactly moved it?

ShivanshGahlot (approved):
LGTM

PR #1590, opened 2024-06-26
sizing calc: DB-11954: added constant to prefer_remote_file, set to false by default

shaharuk-yb:
- associated JIRA task: https://yugabyte.atlassian.net/browse/DB-11954
- added a constant to prefer using remote file or not. By default it is set to false, meaning that it will use the local file.
- In future, it needed, we will just have to change the constant value.


makalaaneesh (approved):
LGTM

PR #1591, opened 2024-06-26
incorporate change in assessment report to have unqualified names

makalaaneesh:


yb-voyager/cmd/exportSchema.go
```diff
@@ -396,8 +396,7 @@ func applyShardingRecommendationIfMatching(sqlInfo *sqlInfo, shardedTables []str
 	case ORACLE:
 		// TODO: handle case-sensitivity properly
 		for _, shardedTable := range shardedTables {
-			parts := strings.Split(shardedTable, ".")
-			if strings.ToLower(parts[1]) == parsedTableName {
+			if strings.ToLower(shardedTable) == parsedTableName {
```
sanyamsinghal, line 399: can we add a comment/note here, that assuming that shardedTable will not be fully qualified in case of oracle.

Just thinking from perspective of supported multiple oracle schemas in future.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -405,7 +405,7 @@ func (ar *AssessmentReport) CalculateSizeDetails() (SizeDetails, error) {
 			} else {
 				details.TotalTableSize += utils.SafeDereferenceInt64(stat.SizeInBytes)
 				details.TotalTableRowCount += utils.SafeDereferenceInt64(stat.RowCount)
-				if slices.Contains(colocatedTables, fmt.Sprintf("%s.%s", stat.SchemaName, stat.ObjectName)) {
+				if slices.Contains(colocatedTables, stat.ObjectName) {
```
priyanshi-yb, line 408: But in case of PG we need qualified so will have to add handling as per source here

sanyamsinghal, line 408: @makalaaneesh why are we changing this? At this time there will be fully qualified names present.

I had added a function just before generating report for removing the schema names `postProcessingOfAssessmentReport()`

sanyamsinghal: ignore my comment, it will be required

makalaaneesh: At this time when the function is called, postprocessing is already done. (because this is done at the last when we want to send the event to callhome

priyanshi-yb: Right

makalaaneesh: Yeah..good point. Made the change. 
I think we went off in the wrong direction here. I'm not satisfied with this PR, but going ahead in the interest of time. The change for removing schemas should ideally happen at the html layer. and the struct/json should have remained intact. 
We should pick this up post release.

makalaaneesh:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2832/

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: Yes, right

priyanshi-yb (approved):
LGTM

PR #1592, opened 2024-06-26
Added SSL to Oracle Assess Migration

shubham-yb:


migtests/scripts/functions.sh
```diff
@@ -684,19 +684,24 @@ EOF
...
 	"
 	if [ "${SOURCE_DB_SCHEMA}" != "" ]
 	then
 		args="${args} --source-db-schema ${SOURCE_DB_SCHEMA}"
 	fi
+	if [ "${SOURCE_DB_ORACLE_TNS_ALIAS}" != "" ]
+	then
+		args="${args} --oracle-tns-alias ${SOURCE_DB_ORACLE_TNS_ALIAS}"
+	else
+		args="${args} --source-db-host ${SOURCE_DB_HOST} --source-db-port ${SOURCE_DB_PORT}"
```
sanyamsinghal, line 703: `--source-db-port` is already being considered above

migtests/scripts/functions.sh
```diff
@@ -684,19 +684,24 @@ EOF
...
 		--send-diagnostics=false --yes
 		--start-clean t
-		--iops-capture-interval 0
+		--iops-capture-interval 1
 	"
 	if [ "${SOURCE_DB_SCHEMA}" != "" ]
 	then
 		args="${args} --source-db-schema ${SOURCE_DB_SCHEMA}"
 	fi
+	if [ "${SOURCE_DB_ORACLE_TNS_ALIAS}" != "" ]
```
sanyamsinghal, line 699: Question: If TNS_ALIAS is provided, does SCHEMA will still be required?

sanyamsinghal (changes requested):

migtests/scripts/functions.sh
shubham-yb: Ideally schema shouldn't be required. But there's other dependencies in our test structure for it. So it's kept that way for now. 
We can restructure it later if need be.

shubham-yb:
Added in another PR
https://github.com/yugabyte/yb-voyager/pull/1601

PR #1604, opened 2024-07-02
Altered password for test_schema

shubham-yb:
TEST_SCHEMA is used in Oracle to run the tests on in Jenkins. This is a pre-created schema in the AMI.
The password gets expired after some time. Adding an alter statement to the create user step to avoid the situation.
We can look for a better fix later if required.

### DIFF
--- a/migtests/scripts/oracle/create_oracle_user
+++ b/migtests/scripts/oracle/create_oracle_user
@@ -15,6 +15,7 @@ begin
   end if;
 end;
 /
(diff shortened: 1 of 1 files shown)


shubham-yb:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/2897/

sanyamsinghal (approved):

PR #1610, opened 2024-07-03
Fix import data streaming logic

makalaaneesh:
Fixed import data logic. 
After import data snapshot phase, we were assuming that `no debezium =>  offline migration with pg_dump/ora2pg`. This assumption is wrong because for live migration from postgres source, we use pg_dump for the snapshot phase. 

While this might not affect real-world scenarios, because by the time import-data snapshot is complete, export-data streaming phase would have started (i.e. debezium would have started). However, this logic is wrong. Moreover, this was leading to intermittent failures in test runs (where import-data snapshot import finished before export-data started debezium, leading to import-data assuming that it is offline migration)  

Testing:
- Manually tested export data status for regression (pg, oracle with and w/o debezium)
- Current test automation pipelines should be enough to validate live migration correctness. In fact, intermittent failures should not be observed after this PR is merged.

### DIFF
--- a/yb-voyager/cmd/exportDataStatusCommand.go
+++ b/yb-voyager/cmd/exportDataStatusCommand.go
@@ -57,7 +57,11 @@ var exportDataStatusCmd = &cobra.Command{
 		if err != nil {
 			utils.ErrExit("initializing name registry: %v", err)
 		}
-		useDebezium = dbzm.IsDebeziumForDataExport(exportDir)
+		msr, err := metaDB.GetMigrationStatusRecord()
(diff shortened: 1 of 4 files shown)


makalaaneesh:
https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/2914/pipeline/

priyanshi-yb (approved):
LGTM

PR #1613, opened 2024-07-26
Bug Fix: In case startTime is not initialised do not send the call home diagnostics

priyanshi-yb:
fixes - https://yugabyte.atlassian.net/browse/DB-12040

### DIFF
--- a/yb-voyager/cmd/analyzeSchema.go
+++ b/yb-voyager/cmd/analyzeSchema.go
@@ -1103,7 +1103,10 @@ func packAndSendAnalyzeSchemaPayload(status string) {
 	if !callhome.SendDiagnostics {
 		return
 	}
-	payload := createCallhomePayload()
+	payload, shouldSend := createCallhomePayload()
+	if !shouldSend {
+		return
+	}
 	payload.MigrationPhase = ANALYZE_PHASE
 	var callhomeIssues []utils.Issue
 	for _, issue := range schemaAnalysisReport.Issues {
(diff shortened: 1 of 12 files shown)


yb-voyager/cmd/analyzeSchema.go
```diff
@@ -1103,7 +1103,10 @@ func packAndSendAnalyzeSchemaPayload(status string) {
 	if !callhome.SendDiagnostics {
 		return
 	}
-	payload := createCallhomePayload()
+	payload, shouldSend := createCallhomePayload()
```
makalaaneesh, line 1106: It seems wrong for `createCallhomePayload` to return a value that tells you whether you should send the payload or not. `createCallhomePayload` should only deal with creating the payload. 

How about:
- define another function `shouldSendCallhome()` which returns false if either callhome.sendDiagnostics is false OR startTime is uninitialized. 
- call that function on line 1103.

makalaaneesh: Also add a comment explaining that startTime being uninitialized is a proxy for detecting that export-dir wasn't set up properly.

priyanshi-yb: Yeah makes sense.

yb-voyager/cmd/common.go
```diff
@@ -1049,6 +1049,13 @@ func (ar *AssessmentReport) GetClusterSizingRecommendation() string {
 
 // ==========================================================================
 
+func shouldSendCallhome() bool {
+	//here checking the startTime is initialised or not as can be seen in root.go we
+	//initialise startTime once we have logging setup after all initial checks
```
makalaaneesh, line 1054: nit: let's also add comments based on your findings for why uninitialized startTime in payload causes issues in callhome server?

makalaaneesh (approved):
LGTM!

PR #1617, opened 2024-08-07
Implement assess-migration-bulk command

sanyamsinghal:
- Added assess-migration-bulk command for doing bulk schema assessments with a single command.
- Parses fleet config file for multiple schema assessments.
- Runs assess-migration for each schema

Sample `bulkAssessmentReport.html`, here in this if you click on a particular assessmentReport then it will open up in a new tab in the browser
1. All successful assessments
image

2. Success and Failed assessments
image


yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -0,0 +1,311 @@
...
+		return nil, err
+	}
+
+	return dbConfigs, nil
+}
+
+/*
+Format: covers both SSL and non-SSL cases
+
+	<dbtype>,<hostname>,<port>,<service_name>,<sid>,<tns_alias>,<username>,<password>,<schema>
```
priyanshi-yb, line 204: Ques: With this structure, the user will have to specify all but in the case of non-SSL, SSL params will be empty?

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -0,0 +1,311 @@
...
+	cmdArgs := buildCommandArguments(dbConfig, exportDirPath)
+	if err := os.MkdirAll(exportDirPath, 0755); err != nil {
+		return fmt.Errorf("creating export-directory %q for schema %q: %w", exportDirPath, dbConfig.Schema, err)
+	}
+
+	execCmd := exec.Command(os.Args[0], cmdArgs...)
+	execCmd.Stdout = os.Stdout
+	execCmd.Stderr = os.Stderr
+	execCmd.Stdin = os.Stdin
+	log.Infof("executing the cmd: %s", execCmd.String())
```
priyanshi-yb, line 136: Please take care of the password param in logging as I see we are passing it with flag to assess-migration.

yb-voyager/cmd/root.go
```diff
@@ -66,29 +66,46 @@ Refer to docs (https://docs.yugabyte.com/preview/migrate/) for more details like
...
-			go sendCallhomePayloadAtIntervals()
-		}
 
-		log.Infof("Start time: %s\n", startTime)
-		if metaDBIsCreated(exportDir) {
-			initMetaDB()
-		}
-		if perfProfile {
-			go startPprofServer()
+		if currentCommand == "yb-voyager assess-migration-bulk" {
```
priyanshi-yb, line 70: I think here we can have separate function for bulk commands identification (right now it will have only one but maybe future we plan to add more we can just it to that)
`isBulkCommand(cmd)`.

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	err = generateBulkAssessmentReport(dbConfigs)
+	if err != nil {
+		utils.ErrExit("failed to generate bulk assessment report: %s", err)
+	}
+}
+
+func executeAssessment(dbConfig AssessMigrationDBConfig) error {
+	log.Infof("executing assessment for schema %q", dbConfig.Schema)
```
makalaaneesh, line 125: I would have imagined we could do this without having to create a new process. 
Can you explore if we can do that? 

One positive of that is that we don't have to deal with process handling (ensuring child is killed, different environments, etc. 

One negative is: we probably can't run all of them in parallel(if we ever need to) because there are a lot of global variables.

yb-voyager/cmd/templates/bulkAssessmentReport.template
```diff
@@ -0,0 +1,85 @@
...
+        }
+        a:hover {
+            text-decoration: underline;
+        }
+    </style>
+</head>
+<body>
+    <div class="container">
+        <h1>Oracle Bulk Assessment Report</h1>
+        <table>
```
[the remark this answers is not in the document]
sanyamsinghal: added screenshot

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: Yes @makalaaneesh, initially i tried without a new process, using assessMigrationCmd.Run() and PreRun() functions internally.

Couple of challenges i faced because of which i chose to go with this approach(simpler):
1. We need a separate/independent migrationUUID and metaDB for each migration
2. Conflicts due to different directory structure for this command, got complicated to handle that in code.
     - When i was calling assessMigratioCmd.Run() and PreRun() from assessMigrationBulkCmd.go then it was not calling Root.PreRun() by default, which is mandatory for all the checks to happen and i believe that is because Root.PreRun() is called once which was happening for bulkCmd but not subsequent assessMig commands
     - assessMigrationCmd was not properly picking up the flags, which i was passing to its Run(cmd, args) as arguments.

In this some might be just solved by proper implementation but i think 1 and 2.1 points are still a bigger problem.

So using a new process approach seemed simpler and better to me. Let me know what do you think

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: yes right. 
What i am think to add here is a csv header so that:
- file is more readable
- flexiblity for user to define the parameters here

yb-voyager/cmd/root.go
```diff
@@ -66,29 +66,46 @@ Refer to docs (https://docs.yugabyte.com/preview/migrate/) for more details like
...
+			setControlPlane()
+		} else {
+			validateExportDirFlag()
+			schemaDir = filepath.Join(exportDir, "schema")
+			if shouldLock(cmd) {
+				lockFPath := filepath.Join(exportDir, fmt.Sprintf(".%sLockfile.lck", GetCommandID(cmd)))
+				lockFile = lockfile.NewLockfile(lockFPath)
+				lockFile.Lock()
+			}
+			InitLogging(exportDir, cmd.Use == "status", GetCommandID(cmd))
```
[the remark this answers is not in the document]
sanyamsinghal: yeah we can, but i don't much gain from doing that.
Also once we have call-home also implemented for this command, most of the code will be similar in both cases, then it will be easy to merge this code.

yb-voyager/cmd/assessMigrationCommandBulk.go
makalaaneesh: Yeah, i think our code in cmd package might not be very friendly to re-use easily. Makes sense I guess. 
Let's just discuss this implementation  with the whole team once in stand-up to see if there are other opinions.

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -0,0 +1,316 @@
...
+
+	execCmd := exec.Command(os.Args[0], cmdArgs...)
+	// password either has to be provided via fleet_config_file or can be provided at run-time by the user.
+	// user setting the env var route is not supported for assess-migration-bulk command
+	execCmd.Env = append(os.Environ(), "SOURCE_DB_PASSWORD="+dbConfig.Password)
+	execCmd.Stdout = os.Stdout
+	execCmd.Stderr = os.Stderr
+	execCmd.Stdin = os.Stdin
+	log.Infof("executing the cmd: %s", execCmd.String())
+	err := execCmd.Run()
```
makalaaneesh, line 139: This should already be implemented (via the atexit handlers), but pls test and ensure that if we ctrl-c interrupt/ SIGTERM the main voyager process, the child process is also shutdown.

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -0,0 +1,316 @@
...
+	dbConfigs, err := parseFleetConfigFile(fleetConfigPath)
+	if err != nil {
+		utils.ErrExit("failed to parse fleet config file: %v", err)
+	}
+
+	for _, dbConfig := range dbConfigs {
+		utils.PrintAndLog("\nAssessing '%s' schema\n", dbConfig.Schema)
+		err = executeAssessment(dbConfig)
+		if err != nil {
+			log.Errorf("failed to assess migration for schema %s: %v", dbConfig.Schema, err)
```
makalaaneesh, line 109: where do these logs go? I'm assuming there is a separate log file for the bulk command store in bulk-dir/logs/

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+/*
+Format: covers both SSL and non-SSL cases
+
+	<dbtype>,<hostname>,<port>,<service_name>,<sid>,<tns_alias>,<username>,<password>,<schema>
+*/
+func parseFleetConfigLine(line string) AssessMigrationDBConfig {
+	config := strings.Split(line, ",")
+	return AssessMigrationDBConfig{
```
makalaaneesh, line 205: we should validate no. of arguments/length of array before processing it

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -0,0 +1,316 @@
...
+		SID:         config[4],
+		TnsAlias:    config[5],
+		User:        config[6],
+		Password:    config[7],
+		Schema:      config[8],
+	}
+}
+
+func getAssessmentExportDirPath(dbConfig AssessMigrationDBConfig) string {
+	switch {
```
makalaaneesh, line 219: You can reuse `getDatabaseIdentifier` instead of writing another switch here

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -0,0 +1,316 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get relative path for %s schema assessment report: %w", dbConfig.Schema, err)
+		}
+		var assessmentDetail = AssessmentDetail{
+			Schema:             dbConfig.Schema,
+			DatabaseIdentifier: getDatabaseIdentifier(dbConfig),
+			ReportPath:         assessmentReportRelPath,
+			Status:             COMPLETE,
+		}
+		if !utils.FileOrFolderExists(assessmentReportPath) {
```
makalaaneesh, line 262: What if the command fails after generating the report? It should still be considered an ERROR, I think. We should probably get this information (or error or not) from the command.Run output, rather than interpreting it based on the presence of assessment report.

yb-voyager/main.go
```diff
@@ -35,6 +35,7 @@ func main() {
 	registerSignalHandlers()
 	atexit.Register(cmd.PackAndSendCallhomePayloadOnExit)
 	atexit.Register(cmd.CleanupChildProcesses)
+	atexit.Register(restoreTerminalState) // ensure terminal is always restored
```
makalaaneesh, line 38: Implementation of `restoreTerminalState` has a call to `utils.ErrExit`, which will call `atexit.Exit()` potentially causing an infinite loop! 
Exit handlers should be very light. See if you can replace the call to utils.ErrExit with just a log perhaps?

yb-voyager/cmd/assessMigrationCommandBulk.go
makalaaneesh: Ah I see you have a TODO 👍

yb-voyager/cmd/assessMigrationCommandBulk.go
makalaaneesh: Header makes sense I think! So, in that case, it is possible for a user to only specify a susbet of all the possible arguments as the names of the arguments will be in header. Correct?

sanyamsinghal: yeah specifying a subset, and ignoring the not required ones will be possible with the header.

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: yeah will do that along with the change of header for config file.

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: sure, actually i have moved some of these functionality under methods of dbConfig struct, in the next PR - https://github.com/yugabyte/yb-voyager/pull/1624/files#diff-2e05f57efc8625d316cbff0d6180f3b8d6b651b6e8a321f81d6beb9606411074R1009-R1020

Is it okay if i do the change there itself.

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: hmm we can do that, will have to store all the results/err for each assessment.

Currently, i am doing two checks - assessmentReportExistsOrNot + checking MSR.MigrationAssessmentDone flag, which i thought is also full proof since the flag is set at the end of assess-migration command.
What do you think?

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: https://yugabyte.atlassian.net/browse/DB-12385

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: yeah let me verify

yb-voyager/main.go
sanyamsinghal: yeah log.Errorf() or log.Fatalf() should be fine. Will test and do the change.

sanyamsinghal: Great catch!! I manually tested, utils.ErrExit() was getting into a never-ending loop.

With log.Fatalf() it will just log the error and exit. 
But nothing will be printed on the console.

 I am thinking to go with log.Errorf(), since this err won't fail the migration, just the terminal stdin/stdout will be in wrong state.

yb-voyager/cmd/assessMigrationCommandBulk.go
makalaaneesh: using metadb flag should work! But here, you seem to only be checking if assessmentReportExistsOrNot?

yb-voyager/main.go
makalaaneesh: Makes sense! Just logging the error and moving on should be fine

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: yeah got confused this with the same check at other place, there i am doing both.

Will add that here also.

yb-voyager/cmd/assessMigrationCommandBulk.go
makalaaneesh: Overall, let's test the following cases:
- ctrl-c on terminal, both child process, and parent process should get interupted. 
- kill SIGINT/SIGTERM directly to parent voyager process; Again, both parent process and child processes should exit. 
- kill SIGINT/SIGTERM to child process. Parent process should not die, it should continue to the next assessment (depending on continue-on-error)

sanyamsinghal: Yeah it works as expected now. 
For the third one, it is anyways proceeding to the next assessment, since there is no error as such.

makalaaneesh: >For the third one, it is anyways proceeding to the next assessment, since there is no error as such. 

Oh, but ideally we should detect, and report that error. Would be wrong to silently continue (especially if someone passed --continue-on-error false)

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -112,6 +113,11 @@ func assessMigrationBulk() {
 				break
 			}
 		}
+		if ProcessShutdownRequested {
+			log.Info("Shutting down as SIGINT/SIGTERM received...")
+			log.Infof("sleep for 10 seconds for exit handlers to execute")
+			time.Sleep(time.Second * 10)
```
makalaaneesh, line 119: Here, after 10s if exit handlers did not successfully complete, we will continue with the next assessment; This should not happen. Let's just return from the function here after sleeping

sanyamsinghal: ok

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: The challenge here for to exit as error, the parent process is not getting any error from the child function call. so it assumes that everything worked fine.

yb-voyager/cmd/assessMigrationCommandBulk.go
makalaaneesh: When cleaning child processes, we have a timeout of 60s (`err := ShutdownProcess(pid, 60)`); Should we wait for at least 60 seconds here?

sanyamsinghal: 60 seconds will be too much here.
In `ShutdownProcess` it is like max 60sec, voyager can exit before

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: added to TO, will do it in the pending PR.

sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/3113/

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: as discussed, better approach will be just return and the main thread waiting for the cleanup goroutine to exit.

Will be completely fixed under this ticket - https://yugabyte.atlassian.net/issues/DB-12388

yb-voyager/cmd/assessMigrationCommandBulk.go
priyanshi-yb: I believe there was some discussion in the standup regarding this one to send exit code 1  from child when interrupted to parent, are we going to take that also in separate PR?

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: done in the other PR - https://github.com/yugabyte/yb-voyager/pull/1624

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: handled for now with the help of sleep. 
Separate ticket for proper handling the issue - https://yugabyte.atlassian.net/issues/DB-12388

makalaaneesh (approved):
LGTM

priyanshi-yb (approved):
LGTM.

PR #1622, opened 2024-08-08
[sizing calc][DB-12358] improve import time calculation - integrate impact of row count on import time.

shaharuk-yb:
## Changes
- added column ```row_count``` to experiment data tables ```colocated_load_time``` and ```sharded_load_time```
- added impact of row count on import time
- added unit tests 

### Description of the change:
- modified the import time calculation by preferring maximum import time when calculated wrt table size and row count.
#### Example1:
- INPUT: Experiment data having same table size but different(increasing) row count (and higher import time respectively). 
- OUTPUT: import time will be calculated wrt both size and row count, but in this case having higher row count will be referred and used as max import time.

#### Example2:
- INPUT: Experiment data having different table size but same row count
- OUTPUT: import time will be calculated wrt both size and row count, but in this case having higher size will be referred and used as max import time.


sonalsagarwal (approved):
LGTM

PR #1623, opened 2024-08-08
Bug fix: Check for TargetDB colocation only when assessment recommendations are applied, otherwise ignore.

sanyamsinghal:
- assume a case where user didn't apply recommendations and doesn't want to have target db colocated also, then this prompt is of no use.


makalaaneesh (approved):
LGTM!

PR #1630, opened 2024-08-13
Fix: Let upsert mode warning show up in import-data-file workflow 

makalaaneesh:
Move error msg to importData function so that it is applicable to both import data as well as import-data-file

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -394,7 +394,10 @@ func importData(importFileTasks []*ImportFileTask) {
 	}
 
 	if (importerRole == TARGET_DB_IMPORTER_ROLE || importerRole == IMPORT_FILE_ROLE) && (tconf.EnableUpsert) {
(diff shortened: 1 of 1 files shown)


priyanshi-yb (approved):
LGTM

yb-voyager/cmd/importData.go
```diff
@@ -397,6 +393,10 @@ func importData(importFileTasks []*ImportFileTask) {
 		utils.ErrExit("failed to get migration UUID: %w", err)
 	}
 
+	if (importerRole == TARGET_DB_IMPORTER_ROLE || importerRole == IMPORT_FILE_ROLE) && (tconf.EnableUpsert) {
+		utils.PrintAndLog(color.RedString("WARNING: Ensure that tables on target YugabyteDB do not have secondary indexes. If a table has secondary indexes, setting --enable-upsert to true may lead to corruption of the indexes."))
```
sanyamsinghal, line 397: If user has already enabled upsert, but based on this warning they want to disable it.

Shall we have it as a prompt so that user can exit and restart if required?

makalaaneesh: Good point, let me do this.

PR #1633, opened 2024-08-14
Fix: Go static checks are failing with go version format mismatch with latest staticcheck package

priyanshi-yb:
Go `Run staticcheck` is failing with latest `go/tools/cmd/staticcheck` package with an error - 
```
go: honnef.co/go/tools/cmd/staticcheck@latest (in honnef.co/go/tools@v0.5.0): go.mod:3: invalid go version '1.22.1': must match format 1.23
```
which seems to be not working with current go version we use, so restricting the static check package to use the last release before the latest i.e. [2023.1.7](https://github.com/dominikh/go-tools/releases/tag/2023.1.7)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -36,5 +36,5 @@ jobs:
     - name: Run staticcheck
       run: |
         cd yb-voyager
-        go install honnef.co/go/tools/cmd/staticcheck@latest
+        go install honnef.co/go/tools/cmd/staticcheck@2023.1.7
         staticcheck ./...


.github/workflows/go.yml
```diff
@@ -36,5 +36,5 @@ jobs:
     - name: Run staticcheck
       run: |
         cd yb-voyager
-        go install honnef.co/go/tools/cmd/staticcheck@latest
+        go install honnef.co/go/tools/cmd/staticcheck@2023.1.7
```
sanyamsinghal, line 39: just to confirm: this is second latest release?

sanyamsinghal (approved):
LGTM.
Nit: If there is a bug tracking the issue on go repository, also add that to the commit message.

.github/workflows/go.yml
priyanshi-yb: Right, excluding RC release of latest

PR #1639, opened 2024-08-19
Updated the target YugabyteDB versions

shubham-yb:


sanyamsinghal (approved):
LGTM.
Add in the final commit message, what versions are removed/added

PR #1640, opened 2024-08-19
[DB-12385] Introduce CSV header for flexibility and enhance fleet config file validation

sanyamsinghal:
- Added flexibility to support case-insensitive and customizable CSV headers.
- Implemented comprehensive validation to ensure the fleet config file exists, is not empty, and is correctly formatted as CSV.
- Introduced checks for mandatory fields and validated that all fields in the header are allowed.
- Updated parsing logic to validate the structure of each line in the config file.
- Renamed service_name field to db_name in fleet config file

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommandBulk.go
+++ b/yb-voyager/cmd/assessMigrationCommandBulk.go
@@ -17,10 +17,11 @@ limitations under the License.
 package cmd
 
 import (
-	"bufio"
 	_ "embed"
+	"encoding/csv"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -42,6 +43,13 @@ var assessMigrationBulkCmd = &cobra.Command{
 	Use:   "assess-migration-bulk",
 	Short: "Bulk Assessment of multiple schemas across one or more Oracle database instances",
 	Long:  "Bulk Assessment of multiple schemas across one or more Oracle database instances",
+
+	PreRun: func(cmd *cobra.Command, args []string) {
+		err := validateFleetConfigFile(fleetConfigPath)
+		if err != nil {
+			utils.ErrExit("%s", err.Error())
+		}
+	},
 
 	Run: func(cmd *cobra.Command, args []string) {
 		assessMigrationBulk()
@@ -185,21 +193,61 @@ func parseFleetConfigFile(filePath string) ([]AssessMigrationDBConfig, error) {
 	}
 	defer file.Close()
 
-	scanner := bufio.NewScanner(file)
-	for scanner.Scan() {
-		line := scanner.Text()
-		if len(line) == 0 {
-			continue
+	reader := csv.NewReader(file)
+	reader.FieldsPerRecord = -1 // Allow variable number of fields per record
+
+	header, err := reader.Read()
+	if err != nil {
+		return nil, fmt.Errorf("failed to read fleet config file header: %w", err)
+	}
+	header = normalizeFleetConfFileHeader(header)
+
+	for {
+		record, err := reader.Read()
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return nil, fmt.Errorf("failed to read: %w", err)
 		}
-		dbConfig := parseFleetConfigLine(line)
+
+		dbConfig := createDBConfigFromRecord(record, header)
 		dbConfigs = append(dbConfigs, dbConfig)
 	}
 
-	if err := scanner.Err(); err != nil {
-		return nil, err
-	}
+	// scanner := bufio.NewScanner(file)
+	// for scanner.Scan() {
+	// 	line := scanner.Text()
+	// 	if len(line) == 0 {
+	// 		continue
+	// 	}
+	// 	dbConfig := parseFleetConfigLine(line)
+	// 	dbConfigs = append(dbConfigs, dbConfig)
+	// }
+
+	// if err := scanner.Err(); err != nil {
+	// 	return nil, err
+	// }
 
 	return dbConfigs, nil
+}
+
+func createDBConfigFromRecord(record []string, header []string) AssessMigrationDBConfig {
+	configMap := make(map[string]string)
+	for i, field := range header {
+		configMap[field] = record[i]
+	}
+
+	return AssessMigrationDBConfig{
+		DbType:      configMap["dbtype"],
+		Host:        configMap["hostname"],
+		Port:        configMap["port"],
+		ServiceName: configMap["service_name"],
+		SID:         configMap["sid"],
+		TnsAlias:    configMap["tns"],
+		User:        configMap["username"],
+		Password:    configMap["password"],
+		Schema:      configMap["schema"],
+	}
 }
 
 /*
@@ -207,20 +255,20 @@ Format: covers both SSL and non-SSL cases
 
 	<dbtype>,<hostname>,<port>,<service_name>,<sid>,<tns_alias>,<username>,<password>,<schema>
 */
-func parseFleetConfigLine(line string) AssessMigrationDBConfig {
-	config := strings.Split(line, ",")
-	return AssessMigrationDBConfig{
-		DbType:      config[0],
-		Host:        config[1],
-		Port:        config[2],
-		ServiceName: config[3],
-		SID:         config[4],
-		TnsAlias:    config[5],
-		User:        config[6],
-		Password:    config[7],
-		Schema:      config[8],
-	}
-}
+// func parseFleetConfigLine(line string) AssessMigrationDBConfig {
+// 	config := strings.Split(line, ",")
+// 	return AssessMigrationDBConfig{
+// 		DbType:      config[0],
+// 		Host:        config[1],
+// 		Port:        config[2],
+// 		ServiceName: config[3],
+// 		SID:         config[4],
+// 		TnsAlias:    config[5],
+// 		User:        config[6],
+// 		Password:    config[7],
+// 		Schema:      config[8],
+// 	}
+// }
 
 const REPORT_PATH_NOTE = "To automatically apply the recommendations, continue the migration steps(export-schema, import-schema, ..) using the auto-generated export-dirs.</br> " +
 	"If using a different export-dir, specify report path in export-schema cmd with `--assessment-report-path` flag  to apply the recommendations."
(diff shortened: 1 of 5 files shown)


sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/3152/

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -51,10 +59,18 @@ var assessMigrationBulkCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(assessMigrationBulkCmd)
 
+	const fleetConfigFileHelp = `
+Path to the CSV file with connection parameters for schema(s) to be assessed.
+Fields (case-insensitive): dbtype, hostname, port, service_name, sid, tns_alias, username, password, schema.
```
makalaaneesh, line 64: let's keep the style consistent: `db_type`, `host_name` ?

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -185,41 +204,73 @@ func parseFleetConfigFile(filePath string) ([]AssessMigrationDBConfig, error) {
...
+	validConfig := false
+	for _, mandatoryCombo := range mandatoryHeaderFields {
+		missingOrEmptyFields := false
+		for _, field := range mandatoryCombo {
+			if val, ok := configMap[field]; !ok || val == "" {
+				missingOrEmptyFields = true
+				break
+			}
+		}
+		if !missingOrEmptyFields {
```
makalaaneesh, line 254: this seems wrong. You want to break if `missingOrEmptyFields=true` not false.

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -185,41 +204,73 @@ func parseFleetConfigFile(filePath string) ([]AssessMigrationDBConfig, error) {
...
+			validConfig = true
+			break
+		}
 	}
+	if !validConfig {
+		return nil, fmt.Errorf("record does not contain a valid combination of mandatory fields")
```
makalaaneesh, line 260: let's give more information here. Which mandatory fields are missing.. and in which row.

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -176,6 +192,9 @@ func buildCommandArguments(dbConfig AssessMigrationDBConfig, exportDirPath strin
 	return args
 }
 
+/*
+Sample header: <dbtype>,<hostname>,<port>,<service_name>,<sid>,<tns_alias>,<username>,<password>,<schema>
```
makalaaneesh, line 196: where is dbname? I might have missed this in the previous PR. Did you chose to use "service_name" instead of this? That is an oracle-specific detail, so I think we shouldn't change our interface altogether. 
We should have db_name for now, not service_name. (consistent with all the other comands). When we revisit these args, we can change them across the board. Till then we should be consistent.

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -336,13 +387,97 @@ func validateBulkAssessmentDirFlag() {
...
-		check if value valid or not,
-		expected/mandatory params are passed
-		strip any trailing spaces
-	func validateFleetConfigFilePath() {}
-*/
+var fleetConfigFileHeaderFields = []string{"dbtype", "hostname", "port", "service_name", "sid",
+	"tns_alias", "username", "password", "schema"}
+
+// TODO: verify/update this list
+var mandatoryHeaderFields = [][]string{
```
makalaaneesh, line 394: Seems like a confusing way to implement this IMO. 
We should have just one list of mandatory field. Additionally, we can ensure that only one of (db_name,sid,tns_alias) is passed.

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -185,41 +204,73 @@ func parseFleetConfigFile(filePath string) ([]AssessMigrationDBConfig, error) {
...
+		record, err := reader.Read()
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return nil, fmt.Errorf("failed to read line %d: %w", lineNum, err)
+		}
 
-	if err := scanner.Err(); err != nil {
-		return nil, err
+		if len(record) != len(header) {
```
makalaaneesh, line 223: DO we need this here if we have already validated it?

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -185,41 +204,73 @@ func parseFleetConfigFile(filePath string) ([]AssessMigrationDBConfig, error) {
...
-		DbType:      config[0],
-		Host:        config[1],
-		Port:        config[2],
-		ServiceName: config[3],
-		SID:         config[4],
-		TnsAlias:    config[5],
-		User:        config[6],
-		Password:    config[7],
-		Schema:      config[8],
+	// Check that at least one of the mandatory field combinations is non-empty and valid
```
makalaaneesh, line 244: Again, this has already been validated, so we might not need this here?

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: yeah sure.
I am just not sure about `hostname` and `username`, whether there should be any `-` in them or not.

Also just thinking whether we should following the same voyager flags here:
`hostname` --> `db_host`
`username` --> `db_user`

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: yeah no need, will remove it.

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: this is verified for the header only, in `validateFleetConfigFile` function.
Here doing that for the records.

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: No, missingOrEmptyFields=false means that all the fields from a `mandatoryCombo` are available in that record.
So we don't need to check anymore, and can break the loop.

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: done

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: simplified the logic

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: done

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: done

yb-voyager/cmd/assessMigrationCommandBulk_test.go
```diff
@@ -0,0 +1,220 @@
...
+		name              string
+		fileContent       string
+		expectedErrSuffix string
+	}{
+		// Single line Record Tests
+		{
+			name: "Missing Mandatory Field in Record (username, schema)",
+			fileContent: `dbtype,username,schema,dbname
+oracle,,,dms`,
+			expectedErrSuffix: "mandatory fields missing in the record: 'username', 'schema'",
```
makalaaneesh, line 177: let's also specify the line no. of the record?

yb-voyager/cmd/assessMigrationCommandBulk_test.go
```diff
@@ -0,0 +1,220 @@
+package cmd
```
makalaaneesh, line 1: thanks for this 🙏  Much better than functional tests

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -185,41 +210,71 @@ func parseFleetConfigFile(filePath string) ([]AssessMigrationDBConfig, error) {
...
-		Password:    config[7],
-		Schema:      config[8],
+	// Check if mandatory fields are present and non-empty
+	missingFields := []string{}
+	for _, field := range fleetConfRequiredFields {
+		if val, ok := configMap[field]; !ok || val == "" {
+			missingFields = append(missingFields, field)
+		}
 	}
+	hasDBIdentifier := false
```
makalaaneesh, line 253: We should additionally validate that only ONE db identifier is passed (db_name, sid, tns_alias)

makalaaneesh: would be good to validate here as well; but okay to take it up as an enhancement later.

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -51,10 +59,18 @@ var assessMigrationBulkCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(assessMigrationBulkCmd)
 
+	const fleetConfigFileHelp = `
+Path to the CSV file with connection parameters for schema(s) to be assessed.
```
makalaaneesh, line 63: pls change the help message as well to reflect the new header fields names

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -176,6 +198,9 @@ func buildCommandArguments(dbConfig AssessMigrationDBConfig, exportDirPath strin
 	return args
 }
 
+/*
+Sample header: <dbtype>,<hostname>,<port>,<service_name>,<sid>,<tns_alias>,<username>,<password>,<schema>
```
makalaaneesh, line 202: change the comment with new header names

yb-voyager/src/utils/utils.go
```diff
@@ -332,6 +332,21 @@ func ContainsAnySubstringFromSlice(slice []string, s string) bool {
 	return false
 }
 
+func IsSubset(subset []string, set []string) bool {
```
makalaaneesh, line 335: not used anywhere anymore. let's remove?

makalaaneesh (approved):
LGTM with a few comments

sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/3169/

PR #1641, opened 2024-08-20
Added an option to use the new yb logical replication connector

ShivanshGahlot:
- Added a flag named `--use-logical-replication-yb-connector` in `inititate cutover to target`. The `default value is false` and if set true then the user can start up export data from target using the new YB connector instead of the old gRPC connector.
- Added code relevant to using the logical replication connector such as config file changes, creation and deletion of replication and publication slots. This new code path works only if the flag is on. Otherwise, the default code path for `gRPC` connector is picked up.
- Also made changes in the installer script such that both the `postgres` and `new YB` connector are kept in their individual folders instead of the `debezium-server/lib` directory. 
- Made changes in the `run.sh` and the voyager code such that the `run.sh` receives the path of the connector to be used and adds it to the `classpath`.
- Made changes to package `debezium` `2.5.2` instead of the current `2.3.3`


yb-voyager/src/dbzm/dbzm.go
```diff
@@ -83,8 +84,16 @@ func (d *Debezium) Start() error {
 	if err != nil {
 		return err
 	}
+
+	YB_OR_PG_CONNECTOR_PATH := filepath.Join(DEBEZIUM_DIST_DIR, "pg-connector")
```
makalaaneesh, line 88: This should work, no?

```
if targetDBExporterRole and d.Config.UseLogicalReplicationYBConnector {
		YB_OR_PG_CONNECTOR_PATH = filepath.Join(DEBEZIUM_DIST_DIR, "yb-connector")
else 
                 YB_OR_PG_CONNECTOR_PATH := filepath.Join(DEBEZIUM_DIST_DIR, "pg-connector")
```

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -831,3 +834,88 @@ func (yb *YugabyteDB) GetNonPKTables() ([]string, error) {
 	}
 	return nonPKTables, nil
 }
+
+func (yb *YugabyteDB) GetReplicationConnection() (*pgconn.PgConn, error) {
```
makalaaneesh, line 838: Let me know if there are any major differences between this and the existing PG implementation..

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -831,3 +834,88 @@ func (yb *YugabyteDB) GetNonPKTables() ([]string, error) {
...
+		err := yb.DropPublication(publicationName)
+		if err != nil {
+			return fmt.Errorf("drop publication: %v", err)
+		}
+	}
+	tablelistQualifiedQuoted := []string{}
+	for _, tableName := range tableList {
+		_, ok := leafPartitions.Get(tableName)
+		if ok {
+			//In case of partiitons, tablelist in CREATE PUBLICATION query should not have root
```
makalaaneesh, line 898: Pls manually verify partitions behavior for YB in CREATE PUBLICATION. It may or may not be the same as PG.

yb-voyager/src/dbzm/dbzm.go
ShivanshGahlot: I think this might make it easier to read:
```
        var YB_OR_PG_CONNECTOR_PATH string
	if isTargetDBExporter(d.ExporterRole) {
		if d.Config.UseLogicalReplicationYBConnector {
			// In case of logical replication connector we need the path /opt/yb-voyager/debezium-server/yb-connector
			YB_OR_PG_CONNECTOR_PATH = filepath.Join(DEBEZIUM_DIST_DIR, "yb-connector")
		} else {
			// In case of gRPC connector the DEBEZIUM_DIST_DIR is set to debezium-server-1.9.5, hence the connector path is empty to avoid any errors in run.sh
			YB_OR_PG_CONNECTOR_PATH = ""
		}
	} else {
		// In case of source db exporter we need the path /opt/yb-voyager/debezium-server/pg-connector
		YB_OR_PG_CONNECTOR_PATH = filepath.Join(DEBEZIUM_DIST_DIR, "pg-connector")
	}
```

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Its only the sleep that I have introduced in the dropReplicationSlot function that is different from PG

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Okay

ShivanshGahlot (commented):
Made the suggested changes

yb-voyager/cmd/exportData.go
```diff
@@ -517,6 +532,46 @@ func GetAllLeafPartitions(table sqlname.NameTuple) []sqlname.NameTuple {
 	return allLeafPartitions
 }
 
+func createYBReplicationSlotAndPublication(ctx context.Context, config *dbzm.Config, finalTableList []sqlname.NameTuple, tablesColumnList *utils.StructMap[sqlname.NameTuple, []string], leafPartitions *utils.StructMap[sqlname.NameTuple, []string]) error {
```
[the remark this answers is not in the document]
makalaaneesh: Yeah, but you should NOT drop the slot if already present. That is wrong. Then you will lose events. You should re-use the slot.

debezium-server-voyager/debezium-server-voyager-dist/src/main/resources/distro/run.sh
```diff
@@ -13,12 +13,26 @@ if [ -z "$1" ]; then
...
     exit 1;
 fi
+
 PROPERTIES_FILE_PATH="$1"
 if [ ! -f "$PROPERTIES_FILE_PATH" ]; then
     echo "$PROPERTIES_FILE_PATH does not exist."
     exit 1;
 fi
 
+if [ -z "$2" ]; then
```
makalaaneesh, line 23: Pls add comments here to explain why we're doing all of this.

debezium-server-voyager/debezium-server-voyager-dist/src/main/resources/distro/run.sh
```diff
@@ -13,12 +13,26 @@ if [ -z "$1" ]; then
...
+    echo "No YB or PG connector path provided.";
```
makalaaneesh, line 24: Let's not design this as a positional argument but rather as an optional argument.  
A good default is to pick up the pg_connector. one can override it by setting an env variable.

That way, even the code gets simplified while invoking it in voyager. 
```
if targetDBexporterRole and !useLogicalReplicationConnector: 
    debezium_dist_dir = 1.9.5
else: 
    debezium_dist_dir = default path 
    
if targetDBexporterRole and useLogicalReplicationConnector : 
    set env var
```

yb-voyager/cmd/exportData.go
```diff
@@ -362,14 +363,17 @@ func exportData() bool {
 
 		if changeStreamingIsEnabled(exportType) {
 			log.Infof("live migration complete, proceeding to cutover")
-			if isTargetDBExporter(exporterRole) {
+			msr, err := metaDB.GetMigrationStatusRecord()
+			if err != nil {
+				utils.ErrExit("get migration status record: %v", err)
+			}
+			if isTargetDBExporter(exporterRole) && !msr.UseLogicalReplicationYBConnector {
```
makalaaneesh, line 370: shall we drop replication slot here as well? Do you see any concerns now that we sleep?

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -158,44 +164,60 @@ func prepareDebeziumConfig(partitionsToRootTableMap map[string]string, tableList
...
 			}
-			if startClean {
-				err = ybCDCClient.DeleteStreamID()
-				if err != nil {
-					return nil, nil, fmt.Errorf("failed to delete stream id: %w", err)
+
+			config.ReplicationSlotName = msr.YBReplicationSlotName
+			config.PublicationName = msr.YBPublicationName
+		} else {
+			if exportType == CHANGES_ONLY {
```
makalaaneesh, line 182: can we have a separate function for this else block

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -445,9 +467,16 @@ func checkAndHandleSnapshotComplete(config *dbzm.Config, status *dbzm.ExportStat
...
 
-				err := utils.WaitForLineInLogFile(logFilePath, "Beginning to poll the changes from the server", 3*time.Minute)
-				if err != nil {
-					return false, fmt.Errorf("failed to poll for message in log file: %w", err)
+				// In case of the logical replication connector, we don't need to wait for the log message.
+				// The momemnt a replication slot is created, we are guaranteed that we will receive events.
+				// In the case of the old connector, there is no such explicit operation to 'start' CDC.
+				// It used to happen implicitly, which is why we have to wait for a log message.
+				if !msr.UseLogicalReplicationYBConnector {
+					pollingMessage := "Beginning to poll the changes from the server"
```
makalaaneesh, line 475: The printAndLog and logFilePath (lines 467 and 468) should also be inside this if block

yb-voyager/src/dbzm/config.go
```diff
@@ -251,6 +278,9 @@ debezium.source.database.sslrootcert=%s
 func (c *Config) String() string {
 	dataDir := filepath.Join(c.ExportDir, "data")
 	offsetFile := filepath.Join(dataDir, "offsets.dat")
+	if isTargetDBExporter(c.ExporterRole) {
```
makalaaneesh, line 281: Instead of checking, just set the offsetFile path to "offsets.<exportRole>.data". That will automatically take care of separating it out.

debezium-server-voyager/debezium-server-voyager-dist/src/main/resources/distro/run.sh
makalaaneesh: I now see the reason why it can't be an optional arg (because we use the same script for 1.9.5 but there we won't have the connector paths logic.)

makalaaneesh: Let's call the argument `YB_OR_PG_CONNECTOR_PATH` here as well.

ShivanshGahlot (commented):
Made the suggested changes

debezium-server-voyager/debezium-server-voyager-dist/src/main/resources/distro/run.sh
```diff
@@ -58,4 +75,5 @@ else
     DEBUGGER="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005"
 fi
 
-exec "$JAVA_BINARY" $DEBEZIUM_OPTS $JAVA_OPTS -Xmx3g $DEBUGGER -cp "$RUNNER"$PATH_SEP"conf"$PATH_SEP$LIB_PATH -Dquarkus.config.locations=$PROPERTIES_FILE_PATH -Doracle.net.tns_admin=$TNS_ADMIN io.debezium.server.Main
\ No newline at end of file
+echo "LIB_PATH=$LIB_PATH"
```
makalaaneesh, line 78: remove echo

yb-voyager/cmd/endMigrationCommand.go
```diff
@@ -493,13 +493,65 @@ func cleanupTargetDB(msr *metadb.MigrationStatusRecord) {
...
+	}
+
 	if msr.YBCDCStreamID == "" {
 		log.Info("yugabytedb cdc stream id is not set. skipping deleting stream id")
 		return
 	}
 	deleteCDCStreamIDForEndMigration(tconf)
 }
 
+func deleteYBReplicationSlotAndPublication(msr *metadb.MigrationStatusRecord, source srcdb.Source) (err error) {
```
makalaaneesh, line 530: let's make the signature of this function be (replicationSlotName, publicationName, source)

yb-voyager/cmd/endMigrationCommand.go
```diff
@@ -493,13 +493,65 @@ func cleanupTargetDB(msr *metadb.MigrationStatusRecord) {
...
+	log.Info("deleting yb replication slot: ", msr.YBReplicationSlotName)
+	ybDB, ok := source.DB().(*srcdb.YugabyteDB)
+	if !ok {
+		return fmt.Errorf("unable to cast source db to yugabytedb")
+	}
+
+	if msr.YBReplicationSlotName != "" && source.DBType == YUGABYTEDB {
```
makalaaneesh, line 537: `if msr.YBReplicationSlotName != ""` 
This check can be done much earlier on line 496 before even invoking this function.

yb-voyager/cmd/exportData.go
```diff
@@ -362,14 +363,28 @@ func exportData() bool {
...
-				err = ybCDCClient.DeleteStreamID()
-				if err != nil {
-					utils.ErrExit("failed to delete stream id after data export: %v", err)
+				if !msr.UseLogicalReplicationYBConnector {
+					err = ybCDCClient.DeleteStreamID()
+					if err != nil {
+						utils.ErrExit("failed to delete stream id after data export: %v", err)
+					}
+				} else {
+					if msr.TargetDBConf == nil {
```
makalaaneesh, line 377: I don't really understand why we need this check?

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -62,7 +62,8 @@ func findDebeziumDistribution(sourceDBType string) error {
 			return err
 		}
 	}
-	if sourceDBType == "yugabytedb" {
+
+	if sourceDBType == "yugabytedb" && !useLogicalReplicationConnector {
```
makalaaneesh, line 66: we should check for is targetDBexporter or not rather than source=yb

makalaaneesh (approved):
LGTM!

PR #1644, opened 2024-08-21
Bug Fix[DB-12388]: Bulk Assessment command continues even after child process (assess-migration) is interrupted

sanyamsinghal:
-  added sleep of 120sec in main() in case of ProcessShutdownRequested to handle the race condition
- changed exit code from 0 -> 1 in at.Exit(1), so that if there is a script or parent process calling a voyager cmd, then in case of SIGINT/SIGTERM it should error out.

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommandBulk.go
+++ b/yb-voyager/cmd/assessMigrationCommandBulk.go
@@ -26,7 +26,6 @@ import (
 	"path/filepath"
 	"strings"
 	"text/template"
-	"time"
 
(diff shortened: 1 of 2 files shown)


sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/3164/

makalaaneesh (approved):
LGTM. However, let's just run it by the team once before merging

PR #808, opened 2023-04-04
Refactored schema parser code for better sql file parsing

sanyamsinghal:
Separated out the cases into SQL:
1. with code block
2. without code block

Added an automation test verifying the issue: [#16576](https://github.com/yugabyte/yugabyte-db/issues/16576)

PR #1647, opened 2024-08-22
Fix for detecting end of sql statement

makalaaneesh:
https://yugabyte.atlassian.net/issues/DB-12261?jql=ORDER%20BY%20created%20DESC


makalaaneesh:
oracle tests - https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/3172/pipeline/

makalaaneesh:
btw @sanyamsinghal i found this ! 😁 https://github.com/yugabyte/yb-voyager/pull/808#discussion_r1159653468

migtests/tests/pg/views-and-rules/pg_views_and_rules_automation.sql
```diff
@@ -56,6 +56,7 @@ select * from test_views.v2;
 
 create or replace view test_views.v3 as select a.first_name,b.last_name from test_views.view_table1 a inner join test_views.view_table2 b using(id);
 
+create or replace view test_views.v4 as select (a.first_name || ' ' || a.last_name || ';'::text) as full_name from test_views.view_table1 a;
```
sanyamsinghal, line 59: @makalaaneesh this is a single line DDL which i think would have worked already.
Add a test cases like this where one of the lines has `;` in it.

```
CREATE OR REPLACE VIEW view test_views.v4 as 
SELECT 
 (a.first_name || ' ' || a.last_name || ';'::text) as full_name 
 from test_views.view_table1 a;
```

```
 SELECT l._assay_key,
    l.sequencenum,
    g._gellanestructure_key,
    g._gellane_key,
    g._emapa_term_key,
    g._stage_key,
    g.creation_date,
    g.modification_date,
    s.term,
    t.stage,
    ((('TS'::text || ((t.stage)::character varying(5))::text) || ';'::text) || s.term) AS displayit
```

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -1101,13 +1101,17 @@ func isEndOfSqlStmt(line string) bool {
...
+		return true
+	}
 
-	cmtStartIdx := strings.Index(line, "--")
+	cmtStartIdx := strings.LastIndex(line, "--")
 	if cmtStartIdx != -1 {
 		line = line[0:cmtStartIdx] // ignore comment
 		line = strings.TrimRight(line, " ")
 	}
-	return strings.Contains(line, ";")
```
sanyamsinghal, line 1110: Thinking of some more cases:

1. DDLs ending with multi-line SQL comments
```
create table temp (
id int,
name varchar
); /* multi-line sql comment*/
```

2. DDLs with multi-line SQL comments in between
```
create table temp (
/* multi-line sql comment*/ id int,
name varchar
);
```

I don't think these are handled before this change also.

migtests/tests/pg/views-and-rules/pg_views_and_rules_automation.sql
makalaaneesh: @sanyamsinghal when pg_dump exports schema, it comes as multi-line. So, for example,view.sql for this statement is: 
```
CREATE VIEW test_views.v4 AS
 SELECT ((((a.first_name)::text || ' '::text) || (a.last_name)::text) || ';'::text) AS full_name
   FROM test_views.view_table1 a;
   ```
   
  This covers the multi-line case

Comment (makalaaneesh) [reply migtests/tests/pg/views-and-rules/pg_views_and_rules_automation.sql:59]: `);` is the base case (every table is output like this in pg_dump), so it is already covered.

File: yb-voyager/cmd/analyzeSchema.go (hunk 2 above)
Comment (makalaaneesh) [reply yb-voyager/cmd/analyzeSchema.go:1110]: Good point about multi-line comments. 
```
create table temp (
id int,
name varchar
); /* multi-line sql comment*/
```
Surprisingly, this works right now. Not because it succesfully determines it as the end of an sql statement; it does not. It just moves on, until it finds the end of the next statement. (essentially two statements get merged into one, and that's okay because it's valid sql).

Comment (makalaaneesh) [reply yb-voyager/cmd/analyzeSchema.go:1110]: @sanyamsinghal the fix for multi-line comments is not straight-forward. We can easily detect multi-line comments as well. However, consider the following case 
```
1: CREATE TABLE
2: ...
3 ); /* multi line 
4: comment */
```

Here we can detect that line 3 is the end of the CREATE TABLE stmt, however pg-parser and analyze-schema then report it the statement (lines 1-3) as a PG unsupported syntax, because the multi-line comment is not closed properly. 

I would suggest that we pick this task up later. lmk.

sanyamsinghal: yeah, multi-line comments would be required to be handled separlety.
Please create a ticket for these cases.

sanyamsinghal (approved):
lgtm

yb-voyager/cmd/analyzeSchema.go
makalaaneesh: https://yugabyte.atlassian.net/browse/DB-12598

PR #1650, opened 2024-08-22
[DB-12491] Add validation and enhancements to assess-migration-bulk command

sanyamsinghal:
- Implemented checks for both JSON and HTML reports to decide whether the assessment for that schema is done or not.
- Added a sample fleet configuration file in the help message for better user guidance.
- Introduced global flags `--yes` and `--send-diagnostics` to the bulk assessment command for consistent flag usage across commands.
- Added validation for `source-db-type` in the fleet configuration file, restricting it to 'oracle' only.
- Included unit tests to verify proper validation of `source-db-type` in both single and multi-line config file.

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommandBulk.go
+++ b/yb-voyager/cmd/assessMigrationCommandBulk.go
@@ -185,7 +185,7 @@ func buildCommandArguments(dbConfig AssessMigrationDBConfig, exportDirPath strin
 		args = append(args, "--source-db-name", dbConfig.DbName)
 	} else {
 		// special handling due to issue https://yugabyte.atlassian.net/browse/DB-12481
-		args = append(args, "--source-db-name", `""`)
+		args = append(args, "--source-db-name", "")
 	}
 	if dbConfig.TnsAlias != "" {
 		args = append(args, "--oracle-tns-alias", dbConfig.TnsAlias)
--- a/yb-voyager/cmd/assessMigrationCommandBulk.go
+++ b/yb-voyager/cmd/assessMigrationCommandBulk.go
@@ -28,7 +28,6 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/samber/lo"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
@@ -186,7 +185,7 @@ func buildCommandArguments(dbConfig AssessMigrationDBConfig, exportDirPath strin
 		args = append(args, "--source-db-name", dbConfig.DbName)
 	} else {
 		// special handling due to issue https://yugabyte.atlassian.net/browse/DB-12481
-		args = append(args, `--source-db-name ""`)
+		args = append(args, "--source-db-name", `""`)
 	}
 	if dbConfig.TnsAlias != "" {
 		args = append(args, "--oracle-tns-alias", dbConfig.TnsAlias)
@@ -204,7 +203,11 @@ func buildCommandArguments(dbConfig AssessMigrationDBConfig, exportDirPath strin
 	// Always safe to use --start-clean to cleanup if there is some state from previous runs in export-dir
 	// since bulk command has separate check to decide beforehand whether the report exists or assessment needs to be performed.
 	args = append(args, "--start-clean", "true")
-	args = append(args, "--yes", lo.Ternary(utils.DoNotPrompt, "true", "false"))
+
+	if utils.DoNotPrompt {
+		args = append(args, "--yes")
+	}
+
 	return args
 }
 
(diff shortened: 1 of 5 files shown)


yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -193,6 +199,11 @@ func buildCommandArguments(dbConfig AssessMigrationDBConfig, exportDirPath strin
 	if dbConfig.Port != "" {
 		args = append(args, "--source-db-port", dbConfig.Port)
 	}
+
+	// Always safe to use --start-clean and --yes to cleanup if there is some state from previous runs in export-dir
+	// since bulk command has separate check to decide beforehand whether the existing report or need to be executed
```
sanyamsinghal, line 204: // since bulk command has separate check to decide beforehand whether the report exists or assessment need to be performed.

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -193,6 +199,11 @@ func buildCommandArguments(dbConfig AssessMigrationDBConfig, exportDirPath strin
 	if dbConfig.Port != "" {
 		args = append(args, "--source-db-port", dbConfig.Port)
 	}
+
+	// Always safe to use --start-clean and --yes to cleanup if there is some state from previous runs in export-dir
+	// since bulk command has separate check to decide beforehand whether the report exists or assessment needs to be performed.
+	args = append(args, "--start-clean", "true",
+		"--yes", "true")
```
makalaaneesh, line 206: Why always  defaulting yes to true. Let's propagate the yes arg value from the bulk command to the child assess-migration command instead?

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -262,6 +273,11 @@ func createDBConfigFromRecord(record []string, header []string) (*AssessMigratio
 		return nil, fmt.Errorf("mandatory fields missing in the record: '%s'", strings.Join(missingFields, "', '"))
 	}
 
+	// Check if the source-db-type is supported (only 'oracle' allowed)
+	if dbType, ok := configMap[SOURCE_DB_TYPE]; ok && strings.ToLower(dbType) != ORACLE {
```
makalaaneesh, line 277: dbtype is a mandatory argument, and it is checked above. You need not check with `_, ok` syntax.

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -281,13 +297,13 @@ const REPORT_PATH_NOTE = "To automatically apply the recommendations, continue t
 func generateBulkAssessmentReport(dbConfigs []AssessMigrationDBConfig) error {
 	log.Infof("generating bulk assessment report")
 	for _, dbConfig := range dbConfigs {
-		assessmentReportPath := dbConfig.GetAssessmentReportPath()
+		assessmentReportPath := dbConfig.GetAssessmentReportBasePath()
```
makalaaneesh, line 300: Let's change the variable names here and on line 309 to indicate that it's only the base path. I was confused first..
Let's also add a comment to indicate that the full path will be set by the json/html report functions

sanyamsinghal: makes sense

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: done

yb-voyager/cmd/assessMigrationCommandBulk.go
```diff
@@ -200,10 +201,10 @@ func buildCommandArguments(dbConfig AssessMigrationDBConfig, exportDirPath strin
 		args = append(args, "--source-db-port", dbConfig.Port)
 	}
 
-	// Always safe to use --start-clean and --yes to cleanup if there is some state from previous runs in export-dir
+	// Always safe to use --start-clean to cleanup if there is some state from previous runs in export-dir
 	// since bulk command has separate check to decide beforehand whether the report exists or assessment needs to be performed.
-	args = append(args, "--start-clean", "true",
-		"--yes", "true")
+	args = append(args, "--start-clean", "true")
+	args = append(args, "--yes", lo.Ternary(utils.DoNotPrompt, "true", "false"))
```
makalaaneesh, line 207: Does `--yes false/true` work? Pls verify this. 
nit: an easier, more readable way is just 
```
if utils.DoNotPrompt: 
    args = append(args, "--yes")
```

makalaaneesh (approved):
LGTM with one comment

yb-voyager/cmd/assessMigrationCommandBulk.go
sanyamsinghal: oh you are right, that does work. thanks
I supposed this flag to be BoolString as we have for `--start-clean true`...

PR #1653, opened 2024-08-23
Added a script to do dependency checks before proceeding with installation of provided packages in an air-gapped environment

ShivanshGahlot:
- The script checks all the required dependencies for `yum`, `apt` and `cpan` depending on the OS.
- If any dependency is missing or not installed, it is listed out by the script and then the script exits.
- If all the dependencies are met, the script proceeds with the installation of `ora2pg`, `debezium` and yb-voyager.
- To skip the checking of dependencies, `--force-install` flag can be passed.
- To do only dependency check, `--check-only-dependency `flag can be passed.


shubham-yb:
LGTM from a functional perspective.

shubham-yb (approved):

PR #1654, opened 2024-08-23
Bug fix: Assessment report path in bulk assessment report if assessment status=ERROR

sanyamsinghal:
- previous .html and .json was coming as report path which actually should be empty


makalaaneesh (approved):
LGTM

PR #1662, opened 2024-08-27
Report unsupported data types and unsupported data types for live migration separately in assessment report.

makalaaneesh:
https://yugabyte.atlassian.net/browse/DB-12245

Essentially, we segregate the unsupported types into two lists
1. data types that are not supported by YB (will be reported in unsupported data types section)
2. data types that are not supported by live migration. (will be reported in migration caveats)

in future, we can add more lists for fall-forward/back unsupported data types as well. 

Also fixed a bug in one of the catalog queries to get column names, data types for a table. 

image


yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -901,18 +903,18 @@ func fetchColumnsWithUnsupportedDataTypes() ([]utils.TableColumnsDataTypes, erro
...
 
 	var sourceUnsupportedDataTypes []string
+	var liveMigrationUnsupportedDataTypes []string
 	switch source.DBType {
 	case POSTGRESQL:
-		sourceUnsupportedDataTypes = srcdb.PostgresUnsupportedDataTypesForDbzm
-		//adding XID here as data export import should not fail as YugabyteDB supports inserts on it but its functions are not supported
-		sourceUnsupportedDataTypes = append(sourceUnsupportedDataTypes, "xid")
+		sourceUnsupportedDataTypes = srcdb.PostgresUnsupportedDataTypes
+		liveMigrationUnsupportedDataTypes = srcdb.PostgresUnsupportedDataTypesForDbzm
```
priyanshi-yb, line 917: Can we return only the list of types unsupported in live migration (excluding the ones that are completely unsupported), we can just do that by getting the difference of these two lists for `liveMigrationUnsupportedDatatypes`? 
 this will avoid the duplication of the details in the report as we are already listing some types as unsupported which are completely not supported and then we should list only the ones not supported in live migration in the caveats section.

priyanshi-yb (approved):
LGTM

PR #1665, opened 2024-08-27
Fix: Display of all object names in schema summary of analyze report 

priyanshi-yb:
Description:

1. Changed the type of Object set in the SummaryMap from a Map to an Array so also list the multiple object names which is possible for INDEX, TRIGGER, and POLICY objects. 
For the counts to be proper, fixed some of the invalidCount for these object types to set the object as invalid with proper name.
 https://yugabyte.atlassian.net/browse/DB-11409
3. Fixed a bug where `InvalidCount` for Indexes for getting set for all the indexes as in `reportCreateIndexStorageParamenter()` func the `invalidCount` was set for all indexes instead it should be done inside check condition of storage parameter. 
2. Updated the Assessment report to not put MigrationCaveats section in case there are no caveats.
3. Modified the `IssueType` for `Foreign table` and `Alter partition table to add PK` issues to `migration_caveats`.


yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -993,9 +993,12 @@ func addMigrationCaveatsToAssessmentReport(unsupportedDataTypesForLiveMigration
 			}
 			migrationCaveats = append(migrationCaveats, UnsupportedFeature{"Unsupported Data Types for Live Migration", columns, false, "", UNSUPPORTED_DATATYPES_FOR_LIVE_MIGRATION_ISSUE})
 		}
-
-		if len(migrationCaveats) > 0 {
-			assessmentReport.MigrationCaveats = migrationCaveats
+		for _, caveat := range migrationCaveats {
```
makalaaneesh, line 996: pls explain what is being done here and add TODO for proper solution.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -43,7 +43,7 @@ import (
 type summaryInfo struct {
 	totalCount   int
 	invalidCount map[string]bool
-	objSet       map[string]bool
+	objSet       []string
```
makalaaneesh, line 46: Add a todo here as well. Ultimately we want to still use a set, not an array

makalaaneesh (approved):
LGTM with a couple of comments

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: https://yugabyte.atlassian.net/browse/DB-12602

PR #1667, opened 2024-08-27
Miscellaneous changes in the air gapped installation script

ShivanshGahlot:
- Added --help flag to list out all the dependencies required on the system and also list the links from which oracle instant clients can be downloaded
- Bug fixed: If binutils was not installed then the script displayed that the minimum required version for binutils was not found. Fixed it to display binutils not installed.
- If a package is not found installed then the version of the required package is also printed
- Added a prompt to ask for the user's consent before starting to install the packages.

### DIFF
--- a/installer_scripts/install-voyager-airgapped.sh
+++ b/installer_scripts/install-voyager-airgapped.sh
@@ -4,37 +4,37 @@ ARGS_LINUX=$@
 LOG_FILE=/tmp/install-yb-voyager.log
 CHECK_ONLY_DEPENDENCIES="false"
 FORCE_INSTALL="false"
+PRINT_DEPENDENCIES="false"
 
 centos_yum_package_requirements=(
-  "gcc:min:0"
-  "make:min:0"
-  "sqlite:min:0"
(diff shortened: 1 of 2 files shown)


installer_scripts/install-voyager-airgapped.sh
```diff
@@ -186,6 +195,12 @@ check_binutils_version() {
 	# Example output of "ld -v" on Ubuntu:
 	# GNU ld (GNU Binutils for Ubuntu) 2.38
 	version=$(ld -v | awk '{print $NF}' | awk -F '-' '{print $1}')
+    if [ -z "$version" ]; then
+        echo ""
+        echo -e "\e[31mERROR: binutils not found. Please install binutils version > ${min_required_version}.\e[0m"
+        binutils_wrong_version=1
+        return
+    fi
```
shubham-yb, line 203: Can we add this block to our installer script as well.

ShivanshGahlot: Added

shubham-yb (approved):

PR #1669, opened 2024-08-28
Fix for import-data-to-source-replica clearing out yb table names in name registry

makalaaneesh:
https://yugabyte.atlassian.net/browse/DB-12600

We introduce the ability to re-register YB table names in name registry to support start-clean. 
But this should only be performed for TARGET_DB_IMPORTER, not SOURCE_REPLICA_DB_IMPORTER or SOURCE_DB_IMPORTER.


priyanshi-yb (approved):
LGTM

PR #1670, opened 2024-08-28
Implement callhome for assess-migration-bulk command

sanyamsinghal:


yb-voyager/cmd/root.go
```diff
@@ -77,7 +76,14 @@ Refer to docs (https://docs.yugabyte.com/preview/migrate/) for more details like
 			}
 			InitLogging(bulkAssessmentDir, cmd.Use == "status", GetCommandID(cmd))
 			startTime = time.Now()
+			log.Infof("Start time: %s\n", startTime)
+
+			if callhome.SendDiagnostics {
+				createCLIArgsString(cmd)
+				go sendCallhomePayloadAtIntervals()
```
makalaaneesh, line 83: There is nothing really that changes frequently in bulk-assessment. Don't think we need this.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -159,7 +159,7 @@ func packAndSendAssessMigrationPayload(status string, errMsg string) {
 
 	assessPayload := callhome.AssessMigrationPhasePayload{
 		UnsupportedFeatures:  callhome.MarshalledJsonString(assessmentReport.UnsupportedFeatures),
-		UnsupportedDataTypes: callhome.MarshalledJsonString(assessmentReport.UnsupportedDataTypes),
+		UnsupportedDatatypes: callhome.MarshalledJsonString(assessmentReport.UnsupportedDataTypes),
```
makalaaneesh, line 162: Why the field name change?

yb-voyager/src/metadb/metadataDB.go
```diff
@@ -70,6 +70,11 @@ func CreateAndInitMetaDBIfRequired(exportDir string) error {
 }
 
 func createMetaDBFile(path string) error {
+	err := utils.EnsureDir(path)
```
makalaaneesh, line 73: Curious, what prompted you to add this? Any scenario which caused an issue?

sanyamsinghal: yeah, in case of other cmds, we have 
```
def CreateMigrationProjectIfNotExists(): 
       which creates all the directory 
       calls initMetaDB(exportDir) -> which internally calls createMetaDBFile(path)
```

So for all cmds except bulk, metainfo directory creation and then initializing MetaDB is taken by that function.

But for bulk command, we need to handle that separately, so i just update this function which is called from `root.go`

yb-voyager/cmd/root.go
priyanshi-yb: Right, this goroutine is for long running data commands, so we don't need this bulk command

priyanshi-yb (approved):
LGTM

sanyamsinghal:
Jenkins pipeline with all oracle and pg tests - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/3254/

yb-voyager/src/metadb/metadataDB.go
makalaaneesh: Ah I see. We are creating metadb ONLY because we want migrationUUID right? 
I guess we could just get a uuid instead of creating metadb, but that's okay. This also works..

sanyamsinghal: yes @makalaaneesh , but since `assess-migration-bulk` cmd is retryable, in case of retries/continue-on-error, it better to use same `mig_uuid`

makalaaneesh: Good point, my bad!

PR #1672, opened 2024-08-28
Fix: MVIEW with JSON_ARRAYAGG erroring out with nil pointer in analyze-schema

priyanshi-yb:
https://yugabyte.atlassian.net/browse/DB-12606

### DIFF
--- a/yb-voyager/cmd/analyzeSchema.go
+++ b/yb-voyager/cmd/analyzeSchema.go
@@ -820,7 +820,7 @@ func checkSql(sqlInfoArr []sqlInfo, fpath string) {
 }
 
 // Checks unsupported DDL statements
-func checkDDL(sqlInfoArr []sqlInfo, fpath string) {
+func checkDDL(sqlInfoArr []sqlInfo, fpath string, objType string) {
 
 	for _, sqlInfo := range sqlInfoArr {
 		if am := amRegex.FindStringSubmatch(sqlInfo.stmt); am != nil {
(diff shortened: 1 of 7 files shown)


yb-voyager/cmd/analyzeSchema.go
```diff
@@ -1004,8 +1004,8 @@ func checkDDL(sqlInfoArr []sqlInfo, fpath string) {
 			summaryMap["TABLE"].invalidCount[sqlInfo.objName] = true
 			reportCase(fpath, "URIType datatype doesn't have a mapping in YugabyteDB", "https://github.com/yugabyte/yb-voyager/issues/1541", `Remove the column with URIType datatype or change it to a relevant supported datatype`, "TABLE", regMatch[2], sqlInfo.formattedStmt, UNSUPPORTED_FEATURES, "")
 		} else if regMatch := jsonFuncRegex.FindStringSubmatch(sqlInfo.stmt); regMatch != nil {
-			summaryMap[strings.ToUpper(regMatch[2])].invalidCount[sqlInfo.objName] = true
-			reportCase(fpath, "JSON_ARRAYAGG() function is not available in YugabyteDB", "https://github.com/yugabyte/yb-voyager/issues/1542", `Rename the function to YugabyteDB's equivalent JSON_AGG()`, strings.ToUpper(regMatch[2]), regMatch[3], sqlInfo.formattedStmt, UNSUPPORTED_FEATURES, "")
+			summaryMap[objType].invalidCount[sqlInfo.objName] = true
```
makalaaneesh, line 1007: Ideally, we should be using objType for all the other regex matches also right? OK to skip for now, just asking..

makalaaneesh (approved):

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: Not sure completely, but I can think of a case if there is any DDL in some file of different objType then that might be correct completely (can come only in schema modification or manual intervention case).

PR #1674, opened 2024-08-29
FIx for unsupported types in export data from target

makalaaneesh:
Current query to fetch data types was incorrectly also checking for table names , not just data types. In one of the tests, there was a table name called "integer" which essentially cause all integer columns to be marked as unsupported. 

Changed the query to specifically look for columns of input table. 
THere is further scope for optimization, by querying all info of tables up front. WIll be taken up later.


ShivanshGahlot (approved):
LGTM

makalaaneesh:
https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/3272/pipeline/

PR #1677, opened 2024-08-30
Fix: datatype queries to remove the '::regtype' as its not  able to detetct the non-public schema types

priyanshi-yb:
* fixed the array type detection from `strings.Contains` to proper `strings.Equalfold` detection by removing the prefixed underscore (`_`), reference pg docs for the mention that array types are determined by having underscores with the type name
https://www.postgresql.org/docs/current/xtypes.html#:~:text=The%20array%20type%20typically%20has%20the%20same%20name%20as%20the%20base%20type%20with%20the%20underscore%20character%20(_)%20prepended

### DIFF
--- a/yb-voyager/src/srcdb/postgres.go
+++ b/yb-voyager/src/srcdb/postgres.go
@@ -66,7 +66,7 @@ AND NOT a.attisdropped
 AND t.relkind IN ('r', 'P')
 AND seq.relkind = 'S';`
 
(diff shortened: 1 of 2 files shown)


makalaaneesh (approved):
LGTM. Can you pls add the array type, array of enum, array of UDT to the non-public schema test ?

priyanshi-yb:
>  Can you pls add the array type, array of enum, array of UDT to the non-public schema test ?

As discussed, will take this up in separate PR.

PR #1678, opened 2024-08-30
[Test] run live tests on pg 10 

priyanshi-yb:


priyanshi-yb:
Live migration basic worked, the rest might be because of schema not supported in pg.

PR #1679, opened 2024-08-30
 Fix: datatype queries to remove the '::regtype' as its not  able to detetct the non-public schema types  (#1677)

priyanshi-yb:
Also, fixed the array type detection from `strings.Contains` to proper `strings.Equalfold` detection by removing the prefixed underscore (`_`), reference pg docs for the mention that array types are determined by having underscores with the type name https://www.postgresql.org/docs/current/xtypes.html#:~:text=The%20array%20type%20typically%20has%20the%20same%20name%20as%20the%20base%20type%20with%20the%20underscore%20character%20(_)%20prepended

* increase sleep in live-migraion-fallb-run-test.sh script for fixing the timing issue with yb debezium in export-data-from-target

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -2,9 +2,9 @@ name: Go
 
 on:
   push:
-    branches: [main]
+    branches: [main, '1.8-dev']
   pull_request:
-    branches: [main]
+    branches: [main, '1.8-dev']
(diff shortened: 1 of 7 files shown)


ShivanshGahlot (approved):
LGTM

PR #1683, opened 2024-09-02
Updated the links of known-issues from deploy preview to live docs link

priyanshi-yb:
https://yugabyte.atlassian.net/browse/DB-12607


sanyamsinghal (approved):
LGTM

Just a small suggestion - see if we can extra out the common part of the DocsLink in a variable, as much as possible, like:
https://docs.yugabyte.com/preview/yugabyte-voyager/known-issues
https://docs.yugabyte.com/preview/yugabyte-voyager/known-issues/postgresql
https://docs.yugabyte.com/preview/yugabyte-voyager/known-issues/oracle...

Its better to deal as much as possible in variables than simple texts, it will avoid bug possibilities in future.

PR #1684, opened 2024-09-02
Adding quotes for case sensitive tables in the table list used to create publication

ShivanshGahlot:


sanyamsinghal (approved):
lgtm

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -558,6 +558,16 @@ func createYBReplicationSlotAndPublication(finalTableList []string) error {
...
 	}()
+	var finalTableList []string
+	for _, table := range tableList {
+		_, ok := leafPartitions.Get(table)
+		if ok {
+			// tablelist should not have root and leaf both so not adding root table in table list
+			continue
+		}
+		// for case sensitive tables in yugabytedb, we need to use the quoted table name
+		finalTableList = append(finalTableList, table.ForOutput())
```
priyanshi-yb, line 569: we should use `table.ForKey()` to avoid any issues as `ForOutput` is just for output purposes it may change in the future based on UX changes or anything.

ShivanshGahlot: Done

PR #1691, opened 2024-09-04
Update default Debezium Server HTTP port from 8080 to 8181 in config file

sanyamsinghal:
- 8080 is a common port and likely to conflict with existing services

Testing: Verified with resume/restart in case of postgres export data.
The events were picked up from last left, and no errors.

### DIFF
--- a/yb-voyager/src/dbzm/config.go
+++ b/yb-voyager/src/dbzm/config.go
@@ -81,7 +81,7 @@ type Config struct {
 var baseConfigTemplate = `
 debezium.format.value=connect
 debezium.format.key=connect
-quarkus.http.port=8181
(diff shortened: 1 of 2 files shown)


yb-voyager/src/dbzm/config.go
```diff
@@ -80,6 +80,7 @@ type Config struct {
 var baseConfigTemplate = `
 debezium.format.value=connect
 debezium.format.key=connect
+quarkus.http.log=8181
```
makalaaneesh, line 83: `quarkus.http.port`

Also, let's use a different port for export data from source vs target

priyanshi-yb (approved):
LGTM

PR #1694, opened 2024-09-05
Fix: Use parser for ALTER TABLE CLUSTER ON instead of regex

priyanshi-yb:
https://yugabyte.atlassian.net/browse/DB-12707


makalaaneesh (approved):
LGTM

PR #1700, opened 2024-09-09
Removed the intermittently failing test

shubham-yb:
- Removed the intermittently failing test on GH actions (Added it on Jenkins)
- Removed the leading zeros from the datatypes validation due to difference in formats


priyanshi-yb (commented):
Instead of completely removing this add any other fallb test maybe as I think we have only have this fallb test.

priyanshi-yb (approved):
LGTM

PR #1701, opened 2024-09-10
[DB-12682] Unique Indexes are not being considered for conflict detection cache

sanyamsinghal:
- updated the queries(oracle, pg) to fetch unique index columns also for conflict detection cache

Bug fix: Conflicting evens were erroring out in case of Multiple columns with unique constraint/index in a table
- fix: check for all columns instead of returning false on first mismatch
 
 
Added a automation test pg-unique-key-conflicts-test for unique key conflicts handling which takes of these cases:
- Table with Unique Constraint (single column, multiple columns)
- Table with Unique Key (single column, multiple columns, where conditions)
- Table with Unique constraint and Unique index on same column
- Table with Unique constraint and Unique index on different set/subset of columns

Jenkins Pipeline PR - https://github.com/yugabyte/jenkins-helpers/pull/760


sanyamsinghal:
Jenkins pipeline - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline-test/813/

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -767,13 +767,48 @@ WHERE parent.relname='%s' AND nmsp_parent.nspname = '%s' `, tname, sname)
...
+        t.relname AS table_name,
+        a.attname AS column_name
+    FROM
+        pg_index ix
+    JOIN
+        pg_class t ON t.oid = ix.indrelid
+    JOIN
+        pg_namespace n ON n.oid = t.relnamespace
+    JOIN
+        pg_attribute a ON a.attrelid = t.oid AND a.attnum = ANY(ix.indkey)
```
makalaaneesh, line 799: `AND a.attnum = ANY(ix.indkey)` what does this mean? Can you pls add a comment?

migtests/lib/yb.py
```diff
@@ -208,6 +208,14 @@ def get_functions_count(self, schema_name="public") -> int:
 		cur.execute(f"SELECT count(routine_name) FROM  information_schema.routines WHERE  routine_type = 'FUNCTION' AND routine_schema = '{schema_name}' AND routine_name NOT IN ({self.EXPECTED_ORAFCE_FUNCTIONS});")
 		return cur.fetchone()[0]
 
+	def get_unique_values(self, schema_name, table_name, column_name):
```
makalaaneesh, line 211: We already have this function - `get_distinct_values__of_column_of_table`

migtests/tests/pg/unique-key-conflicts-test/validateAfterChanges
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python3
```
makalaaneesh, line 1: How about we also validate the contents of the unique columns details stored in metadb? I just want to be extra-careful here because it's a critical part. That would help us ensure that there are no false-positives in the unique columns list

.github/workflows/pg-migtests.yml
```diff
@@ -163,6 +163,9 @@ jobs:
 
       - name: "TEST: pg-basic-non-public-live-migration-test"
         run: migtests/scripts/live-migration-run-test.sh pg/basic-non-public-live-test
+      
+      - name: "TEST: pg-unique-key-conflicts-test"
```
makalaaneesh, line 167: Thanks for this test! Let's also run a fall-back OR fall-forward test? (for oracle as well). 
That will test the unique columns list for target YB.

sanyamsinghal: yeah planning to add test for fallf/fallb even for the conflicts detection also, in a next PR.

makalaaneesh: In general, since we are aiming for smaller release cycles, I was thinking of moving away from this practice of deferring tests to later PRs. I think it's okay if it takes longer(at worst, the feature will come in the next release), but for new features at least, we should ensure that they are all automation-tested. 
That is the only way we we will be able to sustain 2-week releases.
Let's discuss this in stand-up!

yb-voyager/src/srcdb/yugabytedb.go
sanyamsinghal: done.
It is basically matching the column number which are part of index with that number in pg_attribute table

makalaaneesh: Oh. Then why are we doing `ANY()` ? Shouldn't `a.attnum = ix.indkey` work?

.github/workflows/pg-migtests.yml
sanyamsinghal: for this specific case, fallf and fallforward, @shubham-yb is taking up, since i don't have enough bandwidth.
But adding fallf and fallb is on the TODO list before this bug fix goes out in release.

yb-voyager/src/srcdb/oracle.go
```diff
@@ -572,25 +572,47 @@ func (ora *Oracle) GetPartitions(tableName sqlname.NameTuple) []string {
...
+unique_indexes AS (
+    SELECT
+        indCols.TABLE_NAME,
+        indCols.COLUMN_NAME
+    FROM
+        ALL_IND_COLUMNS indCols
+    JOIN
+        ALL_INDEXES ind ON ind.INDEX_NAME = indCols.INDEX_NAME
+    WHERE
+        ind.UNIQUENESS = 'UNIQUE'
```
makalaaneesh, line 598: Will this exclude primary keys? (asking because pg required an explicit exclusion)

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -766,14 +766,56 @@ WHERE parent.relname='%s' AND nmsp_parent.nspname = '%s' `, tname, sname)
...
+	-- Join to get column information
+    JOIN
+        pg_attribute a ON a.attrelid = t.oid 
+		AND a.attnum = ANY(ix.indkey)  -- Match indexed columns
+	 -- Left join to ensure we exclude primary keys by checking associated constraints
+    LEFT JOIN
+        pg_constraint c ON ix.indexrelid = c.conindid AND c.contype = 'p'
+    WHERE
+        ix.indisunique = TRUE
+		AND c.contype IS NULL -- Ensure it's not a primary key
```
makalaaneesh, line 811: nit: An easier-to-understand way to write this would be to LEFT join based only on `ix.indexrelid = c.conindid` AND fileter out contype='p' in the WHERE

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -766,14 +766,56 @@ WHERE parent.relname='%s' AND nmsp_parent.nspname = '%s' `, tname, sname)
...
+        pg_index ix
+	-- Join to get table and schema information from the index
+    JOIN
+        pg_class t ON t.oid = ix.indrelid
+    JOIN
+        pg_namespace n ON n.oid = t.relnamespace
+	-- Join to get column information
+    JOIN
+        pg_attribute a ON a.attrelid = t.oid 
+		AND a.attnum = ANY(ix.indkey)  -- Match indexed columns
```
makalaaneesh, line 805: I'm still not sure why we need "ANY"? Is it to handle multi-column indexes?

sanyamsinghal: Basically we are joining the with pg_attribute to get the columns' name.
So this check only select the columns part of the index.

>  Is it to handle multi-column indexes?
yes

yb-voyager/src/srcdb/yugabytedb.go
sanyamsinghal: yeah what can be done here:
```
-- this ensure all rows of pg_index are there with the contype as 'p' or NULL I
 LEFT JOIN
        pg_constraint c ON ix.indexrelid = c.conindid
        
 -- filter based on contype to exclude PKs
 WHERE c.contype IS NULL; 
```

makalaaneesh: Shouldn't it be 

```
-- this ensure all rows of pg_index are there with the contype as 'p' or NULL I
 LEFT JOIN
        pg_constraint c ON ix.indexrelid = c.conindid
        
 -- filter based on contype to exclude PKs
 WHERE c.contype != 'p'; 
```

makalaaneesh: Maybe that will also filter out the rows where contype is NULL, something we don't want..

yb-voyager/src/srcdb/oracle.go
sanyamsinghal: yeah that needs to be added.
It might not be failing right now but that will cause false positives.

just wondering, currently we dont support changing of PK during live migration, right?
But once we start supporting that, we might need to handle it in a similar way for PK as well.

makalaaneesh: Yeah, and false positives will cause slowness (because we will be stopping the pipeline for every event essentially). Let's fix this. 

>currently we dont support changing of PK during live migration, right?
We do. If this happens, it comes as two events: a DELETE, followed by a CREATE. But they don't really need to be executed in order because their PKs will be different.

sanyamsinghal: got it. thanks for clarifying.

PR #1712, opened 2024-09-17
Bug Fix: remove from conflict cache when ignoring event 

makalaaneesh:
In live migration phase, for every event, we 
1. add to conflictDetectionCache if applicable
2. process event
3. remove from conflictDetectionCache if applicable

The issue: 
As part of step2, we can either add the event to batch and apply it on target DB OR we can ignore the event (because we had already applied it, for example). Step3 is only being executed if we add the event to the batch. In case the event is ignored, it remains in the conflict detection cache forever, causing import-data to wait hang and wait forever. 

The fix: 
Run step3 even in cases where we ignore the event.

### DIFF
--- a/yb-voyager/cmd/live_migration.go
+++ b/yb-voyager/cmd/live_migration.go
@@ -308,10 +308,12 @@ func processEvents(chanNo int, evChan chan *tgtdb.Event, lastAppliedVsn int64, d
 				}
 				if event.Vsn <= lastAppliedVsn {
 					log.Tracef("ignoring event %v because event vsn <= %v", event, lastAppliedVsn)
+					conflictDetectionCache.RemoveEvents([]*tgtdb.Event{event})
 					continue
 				}
 				if importerRole == SOURCE_DB_IMPORTER_ROLE && event.ExporterRole != TARGET_DB_EXPORTER_FB_ROLE {
 					log.Tracef("ignoring event %v because importer role is FB_DB_IMPORTER_ROLE and event exporter role is not TARGET_DB_EXPORTER_FB_ROLE.", event)
+					conflictDetectionCache.RemoveEvents([]*tgtdb.Event{event})
 					continue
 				}
 				batch = append(batch, event)
@@ -335,7 +337,6 @@ func processEvents(chanNo int, evChan chan *tgtdb.Event, lastAppliedVsn int64, d
 		for attempt := 0; attempt < EVENT_BATCH_MAX_RETRY_COUNT; attempt++ {
 			err = tdb.ExecuteBatch(migrationUUID, eventBatch)
 			if err == nil {
-				conflictDetectionCache.RemoveEvents(eventBatch)
 				break
 			} else if tdb.IsNonRetryableCopyError(err) {
 				break
@@ -368,6 +369,7 @@ func processEvents(chanNo int, evChan chan *tgtdb.Event, lastAppliedVsn int64, d
 		if err != nil {
 			utils.ErrExit("error executing batch on channel %v: %v", chanNo, err)
 		}
+		conflictDetectionCache.RemoveEvents(eventBatch.Events)
 		statsReporter.BatchImported(eventBatch.EventCounts.NumInserts, eventBatch.EventCounts.NumUpdates, eventBatch.EventCounts.NumDeletes)
 		log.Debugf("processEvents from channel %v: Executed Batch of size - %d successfully in time %s",
 			chanNo, len(batch), time.Since(start).String())
--- a/yb-voyager/cmd/live_migration_test.go
+++ b/yb-voyager/cmd/live_migration_test.go
@@ -35,8 +35,7 @@ func (myb *mockYugabyteDB) ExecuteBatch(migrationUUID uuid.UUID, batch *tgtdb.Ev
 
 func TestProcessEventsBasic(t *testing.T) {
 	evChan := make(chan *tgtdb.Event, EVENT_CHANNEL_SIZE)
-	var lastAppliedVsn int64
-	lastAppliedVsn = 0
+	lastAppliedVsn := int64(0)
 	doneChan := make(chan bool, 1)
 	statsReporter := &reporter.StreamImportStatsReporter{}
 	state := NewImportDataState(exportDir)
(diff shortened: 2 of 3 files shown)


priyanshi-yb (approved):
LGTM

yb-voyager/cmd/live_migration_test.go
```diff
@@ -0,0 +1,140 @@
...
+	if _, ok := conflictDetectionCache.m[e.Vsn]; ok {
+		t.Errorf("Event not removed from conflict detection cache")
+	}
+}
+
+// Even if event is ignored,
+// (because vsn is less than lastAppliedVsn or it is source_db_importer and event is not from target_db_importer_fb),
+// it should be removed from conflict detection cache
+func TestProcessEventsRemovesIgnoredEventFromConflicDetectionCache(t *testing.T) {
+	// to simulate the case where source db importer ignores
```
priyanshi-yb, line 94: Nice, thanks!

yb-voyager/cmd/live_migration.go
```diff
@@ -308,10 +308,12 @@ func processEvents(chanNo int, evChan chan *tgtdb.Event, lastAppliedVsn int64, d
 				}
 				if event.Vsn <= lastAppliedVsn {
 					log.Tracef("ignoring event %v because event vsn <= %v", event, lastAppliedVsn)
+					conflictDetectionCache.RemoveEvents([]*tgtdb.Event{event})
```
sanyamsinghal, line 311: We should change the function to a variadic function here:
```
func (cfdc *ConflictDetectionCache) RemoveEvents(events ...*tgtdb.Event) {

}
```

and here simply pass it as:
```
conflictDetectionCache.RemoveEvents(event)
```
and
```
conflictDetectionCache.RemoveEvents(eventBatch.Events...)
```

yb-voyager/cmd/live_migration_test.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func TestProcessEventsBasic(t *testing.T) {
+	evChan := make(chan *tgtdb.Event, EVENT_CHANNEL_SIZE)
+	lastAppliedVsn := int64(0)
+	doneChan := make(chan bool, 1)
+	statsReporter := &reporter.StreamImportStatsReporter{}
+	state := NewImportDataState(exportDir)
+	tdb = &mockYugabyteDB{}
+	conflictDetectionCache = NewConflictDetectionCache(utils.NewStructMap[sqlname.NameTuple, []string](), []chan *tgtdb.Event{evChan}, "postgresql")
```
sanyamsinghal, line 43: nit: you can use const POSTGRESQL here

yb-voyager/cmd/live_migration_test.go
```diff
@@ -0,0 +1,140 @@
...
+		Vsn: 1,
+		Op:  "c",
+		TableNameTup: sqlname.NameTuple{
+			CurrentName: oname,
+			TargetName:  oname,
+		},
+		ExporterRole: SOURCE_DB_EXPORTER_ROLE,
+	}
+	evChan <- END_OF_QUEUE_SEGMENT_EVENT
+	processEvents(1, evChan, lastAppliedVsn, doneChan, statsReporter, state)
```
sanyamsinghal, line 56: how about making `processEvents()` function to return err instead of ErrExit().

That we can add unit tests for the failure scenarios also, here for eg: `NonRetryableCopyError()`

yb-voyager/cmd/live_migration_test.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func TestProcessEventsRemovesEventFromConflicDetectionCache(t *testing.T) {
```
sanyamsinghal, line 59: can we add a comment, describing the steps/actions briefly in the test

yb-voyager/cmd/live_migration_test.go
```diff
@@ -0,0 +1,140 @@
...
+	evChan := make(chan *tgtdb.Event, EVENT_CHANNEL_SIZE)
+	lastAppliedVsn := int64(0)
+	doneChan := make(chan bool, 1)
+	statsReporter := &reporter.StreamImportStatsReporter{}
+	state := NewImportDataState(exportDir)
+	tdb = &mockYugabyteDB{}
+	conflictDetectionCache = NewConflictDetectionCache(utils.NewStructMap[sqlname.NameTuple, []string](), []chan *tgtdb.Event{evChan}, "postgresql")
```
sanyamsinghal, line 66: nit: POSTGRESQL

yb-voyager/cmd/live_migration_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	oname := sqlname.NewObjectName("yugabytedb", "public", "public", "users")
```
sanyamsinghal, line 68: nit: YUGABYTEDB

yb-voyager/cmd/live_migration_test.go
```diff
@@ -0,0 +1,140 @@
...
+	e := &tgtdb.Event{
+		Vsn: 1,
+		Op:  "c",
+		TableNameTup: sqlname.NameTuple{
+			CurrentName: oname,
+			TargetName:  oname,
+		},
+		ExporterRole: SOURCE_DB_EXPORTER_ROLE,
+	}
```
sanyamsinghal, line 77: just thinking out loud, can we some common abstracted code to generate object names/events in the code so that everyone can use that for adding tests, without thinking much about the sqlname or nametuple packages.

sanyamsinghal (commented):
Looks good, thanks for starting these unit test.

yb-voyager/cmd/live_migration_test.go
makalaaneesh: Ah good idea. The challenge is, since this is run in a separate goroutine, we ultimately need to ErrExit. We can write a wrapper function that calls processEvents, and on returning error, it ErrExits. That way we can test the processEvents function better. 

Added a TODO for now.

yb-voyager/cmd/live_migration_test.go
makalaaneesh: Definitely makes sense, and something to keep in mind. I think these patterns will unfold as we write more and more unit test cases.

yb-voyager/cmd/live_migration_test.go
sanyamsinghal: yeah, adding a wrapper function around it should be simple, and better choice than ErrExiting....

PR #1713, opened 2024-09-17
Reporting indexes on UDTs in analyze and assess reports

priyanshi-yb:
https://yugabyte.atlassian.net/browse/DB-12819


yb-voyager/cmd/analyzeSchema.go
```diff
@@ -417,6 +422,25 @@ func checkStmtsUsingParser(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+				e.g. CREATE TYPE non_public."Address_type" AS (
+						street VARCHAR(100),
+						city VARCHAR(50),
+						state VARCHAR(50),
+						zip_code VARCHAR(10)
+					);
+				stmt:{composite_type_stmt:{typevar:{schemaname:"non_public"  relname:"Address_type"  relpersistence:"p"  location:14}  coldeflist:{column_def:{colname:"street"
+				type_name:{names:{string:{sval:"pg_catalog"}}  names:{string:{sval:"varchar"}}  typmods:{a_const:{ival:{ival:100}  location:65}}  typemod:-1  location:57} ...
+
+				Here the type name is required which is available in typevar->relname (not considering the qualified and all as while checking type in create table we will get
```
makalaaneesh, line 438: Isn't this error-prone though? If in the CREATE TABLE definition, we get the unqualified type name? wouldn't that be an issue? HOw would we know which one it's referring to, if there are conflicting names?

priyanshi-yb: I don't think that can be error prone as we don't need to know the qualified name as we are just interested in the same udt type name exists or not and while checking the type of column of CREATE TABLE we just use the last name which unqualified.

makalaaneesh (approved):
LGTM

yb-voyager/cmd/analyzeSchema.go
makalaaneesh: As discussed, consider edge case: 
someone declares a type in public schema, and an enum in non-public schema with the same name. Then, creates an index on the enum in the non-public schema. We will incorrectly report it as unsupported because we are just comparing unqualified names, but indexes on enum types are supported.

PR #1715, opened 2024-09-19
Address review comments in https://github.com/yugabyte/yb-voyager/pull/1712

makalaaneesh:


yb-voyager/cmd/live_migration.go
```diff
@@ -289,6 +289,7 @@ func hashEvent(e *tgtdb.Event) int {
 	return int(hash.Sum64() % (uint64(NUM_EVENT_CHANNELS)))
 }
 
+// TODO: return err instead of ErrExit so that we can test better.
```
sanyamsinghal, line 292: is there any challenge in doing this right away?
Simple changing this:
```
go processEvents(i, evChans[i], chanLastAppliedVsn, processingDoneChans[i], statsReporter, state)
```
to something like this:
```
go func() {
       err = processEvents(i, evChans[i], chanLastAppliedVsn, processingDoneChans[i], statsReporter, state)
       if err != nil {
                utils.ErrExit(err)
       }
}() 
```
should work.

makalaaneesh: Didn't want to increase the scope of this PR more since it was a critical change, that's why I just added a TODO. But sure, I think this seems like a simple isolated change. let me pick it up.

sanyamsinghal: Yeah agree, you can take time to test and merge the change/PR, no hurries.

makalaaneesh: @sanyamsinghal can we take this up later? Thought this was simple, but this kind of go code has nuances, and is different for different go versions. https://go.dev/blog/loopvar-preview I would want to properly understand this before making such a change. Since this is not blocking any functionality, would like to defer this.

sanyamsinghal: 1. once we move to GO 1.22, this problem will go away.
2. With current go version, you have to make a copy of loop variables, and then pass it to the function.

Simplified by chatgpt - https://chatgpt.com/share/66ebe533-1fc0-8006-bccf-73af19132698

Feel free to decide, we can defer if you want.

sanyamsinghal (approved):
LGTM

PR #1726, opened 2024-09-27
Adaptive Parallelism

makalaaneesh:


yb-voyager/src/adaptiveparallelism/adaptive_parallelism.go
```diff
@@ -0,0 +1,108 @@
...
+
+type TargetYugabyteDBWithConnectionPool interface {
+	IsAdaptiveParallelismSupported() bool
+	GetClusterMetrics() (map[string]map[string]string, error) // node_uuid:metric_name:metric_value
+	GetNumConnectionsInPool() int
+	GetNumMaxConnectionsInPool() int
+	UpdateNumConnectionsInPool(int) error // (delta)
+}
+
+func AdaptParallelism(yb TargetYugabyteDBWithConnectionPool) error {
```
priyanshi-yb, line 42: I think this makes more sense in `tgtdb`?  similar to `conn_pool` and this function can be a `tdb` functionality that way we can call it using this `tdb.AdaptParallelism()` we will not implement it for other dbs.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1163,6 +1173,46 @@ func (yb *TargetYugabyteDB) isQueryResultNonEmpty(query string) bool {
 	return rows.Next()
 }
 
+func (yb *TargetYugabyteDB) IsAdaptiveParallelismSupported() bool {
```
priyanshi-yb, line 1176: Similarly here, Ideally we can have these functions as part of target interface.
like this function `IsAdaptiveParallelismSupported` if this will be a part of target interface we can return false for other dbs which will make it clear that Adaptive parallelism is not applicable for them.

Let me know, we can get on a call to discuss these if there are any challenges.

priyanshi-yb (commented):
Posting early comments regarding structure and high level implementation. Will continue review the internal implementation details.

yb-voyager/cmd/importData.go
```diff
@@ -483,6 +491,11 @@ func importData(importFileTasks []*ImportFileTask) {
 			utils.PrintAndLog("Tables to import: %v", importFileTasksToTableNames(pendingTasks))
 			prepareTableToColumns(pendingTasks) //prepare the tableToColumns map
 			poolSize := tconf.Parallelism * 2
+			if tconf.EnableAdaptiveParallelism {
+				// in case of adaptive parallelism, we need to use maxParalllelism * 2
+				yb := tdb.(*tgtdb.TargetYugabyteDB)
```
priyanshi-yb, line 496: assertion check similar to above should also be here right?

yb-voyager/src/adaptiveparallelism/adaptive_parallelism.go
priyanshi-yb: As we discussed, it is fine to not use `tdb` here and do type assertions, but see if its fine to move this to `tgtdb` package

priyanshi-yb (commented):
Overall, it looks good to me. But will let @sanyamsinghal also review the changes

yb-voyager/src/tgtdb/conn_pool.go
```diff
@@ -44,27 +45,94 @@ type ConnectionPool struct {
 	sync.Mutex
 	params                    *ConnectionParams
 	conns                     chan *pgx.Conn
+	idleConns                 chan *pgx.Conn
 	connIdToPreparedStmtCache map[uint32]map[string]bool // cache list of prepared statements per connection
 	nextUriIndex              int
 	disableThrottling         bool
+	size                      int // current size of the pool
+	pendingConnsToClose       int
+	pendingConnsToCloseLock   sync.Mutex
```
sanyamsinghal, line 54: Please add comments explaining `idleConns` and `pendingConnsToCloseLock`

sanyamsinghal (commented):
Few overall comments:
1. Please revisit the log statements, which are going to be printed every 10sec, remove if few of those seems redundant.
2. Can we add more top-level comments in the `conn_pool.go` file. The code is getting a bit more advanced now and detailed comments would help in future.

yb-voyager/src/tgtdb/conn_pool.go
```diff
@@ -44,27 +45,94 @@ type ConnectionPool struct {
...
+			if pool.pendingConnsToClose > delta {
+				log.Infof("adaptive: Decreasing pendingConnsToClose by %d", delta)
+				pool.pendingConnsToClose -= delta
+				delta = 0
+			} else {
+				log.Infof("adaptive: Decreasing pendingConnsToClose to 0 by %d", pool.pendingConnsToClose)
+				delta -= pool.pendingConnsToClose
+				pool.pendingConnsToClose = 0
+			}
+		}
```
sanyamsinghal, line 113:
```
		if pool.pendingConnsToClose > 0 {
			// since we are increasing, we can just reduce the pending close count
			connCount := Math.min(pool.pendingConnsToClose, delta)
			delta -= connCount
			pool.pendingConnsToClose -= connCount
		}
```

yb-voyager/src/tgtdb/conn_pool.go
```diff
@@ -44,27 +45,94 @@ type ConnectionPool struct {
...
+	} else {
+		// for decreases, process them asynchronously.
+		// The problem with processing them synchronously is that we may have to wait for the
+		// connections to be returned to the pool. Not only that, this goroutine that is trying to
+		// update the pool size by reading from the channel is also competing with multiple
+		// other goroutines that are trying to ingest data.
+		//  This might take significant time depending on the query execution time and the lock contention
+		//  So, instead, we just register the request to close the connections,
+		// and the connections are returned to the idle pool when the query execution is done.
+		pool.pendingConnsToClose += -delta
```
sanyamsinghal, line 129: Use `Math.Abs()`
https://pkg.go.dev/math#Abs

sanyamsinghal (commented):
Please test the new code in conn pool is not getting triggered if adaptive parallelism is not enabled.

Felt risky because new code is interleaved with current code.
Can we refactor `WithConn()` function and break it into smaller functions for simplicity and keep adaptive logic separate.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -219,12 +220,21 @@ func (yb *TargetYugabyteDB) InitConnPool() error {
 		log.Infof("Using %d parallel jobs by default. Use --parallel-jobs to specify a custom value", yb.tconf.Parallelism)
 	}
 
+	if yb.tconf.EnableAdaptiveParallelism {
+		if yb.tconf.MaxParallelism <= 0 {
+			yb.tconf.MaxParallelism = yb.tconf.Parallelism * 2
```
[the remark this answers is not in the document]
makalaaneesh: Just a heuristic. We go with 0.25x cores right now as default. Guess is that if we go higher than 0.5x, we're guaranteed to overload the cluster. 

When we get to testing, we can validate this.

yb-voyager/src/tgtdb/conn_pool.go
makalaaneesh: Unfortunately go doesn't seem to have an abs function for int. This is for float64 only.

makalaaneesh:
> 1. Please revisit the log statements, which are going to be printed every 10sec, remove if few of those seems redundant.
Deferring this, as adaptive parallelism is not enabled by default. Moreover, we'll get a better understanding of which logs to keep once we start testing.

yb-voyager/src/tgtdb/conn_pool.go
makalaaneesh: I find this a bit confusing to be honest. 😅 
Even though an if-else block is more verbose, I prefer it because it's simpler to reason.

yb-voyager/src/adaptiveparallelism/adaptive_parallelism.go
priyanshi-yb: Works

priyanshi-yb (approved):
LGTM

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1163,6 +1173,46 @@ func (yb *TargetYugabyteDB) isQueryResultNonEmpty(query string) bool {
...
+	result := make(map[string]map[string]string)
+
+	query := "select uuid, metrics, status, error from yb_servers_metrics();"
+	rows, err := yb.Query(query)
+	if err != nil {
+		return result, fmt.Errorf("querying yb_servers_metrics(): %w", err)
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var uuid, metrics, status, errorStr string
```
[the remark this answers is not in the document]
sanyamsinghal: I think using `sql.NullString` will still be safe and fullproof

yb-voyager/src/adaptiveparallelism/adaptive_parallelism.go
```diff
@@ -0,0 +1,108 @@
...
+		if err != nil {
+			return fmt.Errorf("updating parallelism with -1: %w", err)
+		}
+	} else {
+		log.Infof("adaptive: found CPU usage = %d <= %d, increasing parallelism to %d", maxCpuUsage, MAX_CPU_THRESHOLD, yb.GetNumConnectionsInPool()+1)
+		err := yb.UpdateNumConnectionsInPool(1)
+		if err != nil {
+			return fmt.Errorf("updating parallelism with +1 : %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
sanyamsinghal: Yes lets discuss, please keep this in your list and if required create a ticket, to not miss.

yb-voyager/src/tgtdb/conn_pool.go
```diff
@@ -44,27 +45,94 @@ type ConnectionPool struct {
...
 		pool.params.SessionInitScript = defaultSessionVars
 	}
 	return pool
 }
 
+func (pool *ConnectionPool) GetNumConnections() int {
+	return pool.size
+}
+
+func (pool *ConnectionPool) UpdateNumConnections(delta int) error {
```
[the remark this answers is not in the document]
sanyamsinghal: No, function has that support.
But to this function we always send `1` or `-1`, which can be decided on smart logic(based on cluster size etc)

yb-voyager/src/tgtdb/conn_pool.go
sanyamsinghal: 1. Zero equality in the condition is still of no use.
2. I dont think it will be confusing or hard to read. I feel it was quite verbose, You can add a comment saying - `take as much connections as possible from pool`

yb-voyager/src/tgtdb/conn_pool.go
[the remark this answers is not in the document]
makalaaneesh: Agreed! I was also thinking about this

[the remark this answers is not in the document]
sanyamsinghal: Please create a ticket/task to your list

yb-voyager/src/tgtdb/yugabytedb.go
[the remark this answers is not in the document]
makalaaneesh: makes sense. made the changes. While making this change, I also realised that GetClusterMetrics needs to return the Status/Error as well (it is only returning the metrics right now). Will take this up in a subsequent PR!

yb-voyager/src/tgtdb/conn_pool.go
makalaaneesh: I think this is subjective. However, on second thoughts, it seems alright. I've made changes!

sanyamsinghal (approved):
LGTM

PR #1735, opened 2024-10-07
Live migration Data Conversion fixes

makalaaneesh:
- UUID: When formatRequired=true(i.e. for UPDATEs), quote value
- Json, Enum: Quote value after escaping single quotes.


yb-voyager/src/tgtdb/suites/yugabytedbSuite_test.go
```diff
@@ -0,0 +1,93 @@
...
+*/
+// package for tgtdb value converter suite
+package tgtdbsuite
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
```
priyanshi-yb, line 24: Thanks for adding these tests, we can enhance this further for more tests.

priyanshi-yb (approved):
LGTM!

PR #1740, opened 2024-10-08
fix deadlock in WaitUntilNoConflict

makalaaneesh:


yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -152,7 +152,16 @@ retry:
...
-				c.evChans[i] <- FLUSH_BATCH_EVENT
+				// non-blocking send because blocking send can cause deadlock
+				// between main goroutine acquiring lock and blocking on sending to channel
+				// and processEvents goroutine waiting to acquire lock in RemoveEvents.
+				select {
+				case c.evChans[i] <- FLUSH_BATCH_EVENT:
+				default:
+					// channel is full, so it's okay not to send FLUSH_BATCH_EVENT
+					// because MAX_EVENTS_PER_BATCH would likely be reached in the next batch.
+					log.Infof("channel %d is full, not sending FLUSH_BATCH_EVENT", i)
```
sanyamsinghal, line 163: Is it confirmed that go routine are blocked only in case of size full.
If there are other possibilities, then we should print the channel size also.

sanyamsinghal (approved):

yb-voyager/cmd/conflictDetectionCache.go
makalaaneesh: don't think there are other possibilities, but added size to log nonetheless.

priyanshi-yb (approved):
Lgtm

PR #1745, opened 2024-10-10
[DB-13027] Implement reporting of XML functions as Unsupported Query Construct in migration assessment

sanyamsinghal:
- covers simple SELECT stmts with xml functions, in target_list, from clause or where clause of the statement

### DIFF
--- a/yb-voyager/src/queryparser/unsupportedConstructs.go
+++ b/yb-voyager/src/queryparser/unsupportedConstructs.go
@@ -366,6 +366,14 @@ func containsXmlFunctionsInTargetList(targetList []*pg_query.Node) bool {
 		if xmlExpr != nil {
 			return true
 		}
+
+		if funcCallNode, isFuncCall := resTarget.Val.Node.(*pg_query.Node_FuncCall); isFuncCall {
+			funcList := funcCallNode.FuncCall.Funcname
+			functionName := funcList[len(funcList)-1].GetString_().Sval
+			if slices.Contains(xmlFunctions, functionName) {
+				return true
+			}
+		}
 	}
 	return false
 }
--- a/yb-voyager/src/queryparser/unsupportedConstructs.go
+++ b/yb-voyager/src/queryparser/unsupportedConstructs.go
@@ -23,6 +23,19 @@ var unsupportedSysCols = []string{
 	"xmin", "xmax", "cmin", "cmax", "ctid",
 }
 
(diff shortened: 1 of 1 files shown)


yb-voyager/src/queryparser/unsupportedConstructs.go
```diff
@@ -315,8 +328,115 @@ func containsSystemColumnsInWhereClause(whereClause *pg_query.Node) bool {
...
+		if resTarget == nil {
+			return false
+		}
+
+		xmlExpr := resTarget.Val.GetXmlExpr()
+		if xmlExpr != nil {
+			return true
+		}
+	}
+	return false
```
priyanshi-yb, line 370: we can also have xml functions in the target list e.g. `SELECT id, xpath('/person/name/text()', data) AS name from xml_example;` ?

yb-voyager/src/queryparser/unsupportedConstructs.go
```diff
@@ -315,8 +328,115 @@ func containsSystemColumnsInWhereClause(whereClause *pg_query.Node) bool {
...
+	}
+
+	if funcCallNode := whereClause.GetFuncCall(); funcCallNode != nil {
+		funcList := funcCallNode.Funcname
+		functionName := funcList[len(funcList)-1].GetString_().Sval
+		if slices.Contains(xmlFunctions, functionName) {
+			return true
+		}
+	}
+
```
priyanshi-yb, line 439: Thinking loud, can there be any case in which where clause can have the `XmlExpr` way as in the above target and from clause?

sanyamsinghal: yes @priyanshi-yb , one case can be - WHERE clause having a subquery

sanyamsinghal: But let me cover those later, need to think it through.

yb-voyager/src/queryparser/unsupportedConstructs.go
sanyamsinghal: done

yb-voyager/src/queryparser/unsupportedConstructs.go
priyanshi-yb: okay

priyanshi-yb (approved):
LGTM

PR #1750, opened 2024-10-14
Revert version changes + output error logs if installer script fails

makalaaneesh:


ShivanshGahlot (commented):
Should we remove the cat?

ShivanshGahlot (approved):

PR #1752, opened 2024-10-15
Set the default value of run-guardrails-check flag to true

ShivanshGahlot:


sanyamsinghal (approved):
LGTM.
I was thinking should we remove this flag, i think its safer to have it, in case of any false or unnecessary errors user can skip it.

PR #1755, opened 2024-10-15
Using parser for Partition table cases in analyze-schema such as insufficient PK as per partition columns

priyanshi-yb:
Modifying the following cases of analyze-schema to be reported using pg-parser:
1. List partitioned table can't have multi-column partitions
2. Expression partitioned table can't have Primary key / Unique columns
3. The partition key on the partitioned table should have all the columns included in PK.
4. Alter table ADD PK on a partitioned table.

https://yugabyte.atlassian.net/browse/DB-13266

### DIFF
--- a/yb-voyager/cmd/analyzeSchema.go
+++ b/yb-voyager/cmd/analyzeSchema.go
@@ -122,7 +122,7 @@ var (
 	analyzeSchemaReportFormat string
 	sourceObjList             []string
 	schemaAnalysisReport      utils.SchemaReport
-	tblPartitions             = make(map[string]bool)
+	partitionTablesMap        = make(map[string]bool)
 	// key is partitioned table, value is filename where the ADD PRIMARY KEY statement resides
 	primaryCons      = make(map[string]string)
 	summaryMap       = make(map[string]*summaryInfo)
@@ -162,7 +162,6 @@ var (
 	currentOfRegex            = re("WHERE", "CURRENT", "OF")
 	amRegex                   = re("CREATE", "ACCESS", "METHOD", capture(ident))
 	idxConcRegex              = re("REINDEX", anything, capture(ident))
-	partitionColumnsRegex     = re("CREATE", "TABLE", ifNotExists, capture(ident), parenth(capture(optionalCommaSeperatedTokens)), "PARTITION BY", capture("[A-Za-z]+"), parenth(capture(optionalCommaSeperatedTokens)))
 	likeAllRegex              = re("CREATE", "TABLE", ifNotExists, capture(ident), anything, "LIKE", anything, "INCLUDING ALL")
 	likeRegex                 = re("CREATE", "TABLE", ifNotExists, capture(ident), anything, `\(LIKE`)
 	inheritRegex              = re("CREATE", opt(capture(unqualifiedIdent)), "TABLE", ifNotExists, capture(ident), anything, "INHERITS", "[ |(]")
@@ -196,8 +195,6 @@ var (
 	alterTypeRegex                  = re("ALTER", "TYPE", capture(ident))
 	alterTblSpcRegex                = re("ALTER", "TABLESPACE", capture(ident), "SET")
 
-	// table partition. partitioned table is the key in tblParts map
-	addPrimaryRegex = re("ALTER", "TABLE", opt("ONLY"), ifExists, capture(ident), "ADD CONSTRAINT", capture(ident), "PRIMARY KEY")
 	primRegex       = re("CREATE", "FOREIGN", "TABLE", capture(ident)+`\(`, anything, "PRIMARY KEY")
 	foreignKeyRegex = re("CREATE", "FOREIGN", "TABLE", capture(ident)+`\(`, anything, "REFERENCES", anything)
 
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/analyzeSchema.go
```diff
@@ -419,6 +413,138 @@ func checkStmtsUsingParser(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+		location:98  keys:{string:{sval:"id"}} keys:{string:{sval:"country_code"}}}}  partspec:{strategy:PARTITION_STRATEGY_RANGE
+		part_params:{partition_elem:{name:"country_code" location:150}}  part_params:{partition_elem:{name:"record_type"  ...
+	*/
+	if createTableNode.CreateStmt.GetPartspec() == nil {
+		//If not partition table then no need to proceed
+		return
+	}
+
+	if primaryConsInAlter[fullyQualifiedName] != nil {
+		//reporting the ALTER TABLE ADD PK on partition table here in case the order is different if ALTER is before the CREATE
```
makalaaneesh, line 476: Why can this happen?

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -419,6 +413,138 @@ func checkStmtsUsingParser(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+			}
+		}
+	}
+
+	partitionElements := createTableNode.CreateStmt.GetPartspec().GetPartParams()
+
+	for _, partElem := range partitionElements {
+		if partElem.GetPartitionElem().GetExpr() != nil {
+			//Expression partitions
+			fmt.Printf("expression par on %s, %v", fullyQualifiedName, primaryKeyColumns)
```
makalaaneesh, line 514: remove log?

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -419,6 +413,138 @@ func checkStmtsUsingParser(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+			"https://github.com/yugabyte/yb-voyager/issues/699", "Make it a single column partition by list or choose other supported Partitioning methods", "TABLE", fullyQualifiedName, sqlStmtInfo.formattedStmt, UNSUPPORTED_FEATURES, LIST_PARTIION_MULTI_COLUMN_DOC_LINK)
+		return
+	}
+
+	if len(primaryKeyColumns) == 0 { // no need to report in case of non-PK tables
+		return
+	}
+
+	for _, pc := range partitionColumns {
+		if !slices.Contains(primaryKeyColumns, pc) {
```
makalaaneesh, line 538: nit: let's do a set difference, and check if it's non-empty? 
Also, is there a way to add metadata to the unsupported feature? Would be ncie to report the missing columns as well.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -419,6 +413,138 @@ func checkStmtsUsingParser(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+
+	for _, partElem := range partitionElements {
+		if partElem.GetPartitionElem().GetExpr() != nil {
+			//Expression partitions
+			fmt.Printf("expression par on %s, %v", fullyQualifiedName, primaryKeyColumns)
+			if len(primaryKeyColumns) > 0 || len(uniqueKeyColumns) > 0 {
+				summaryMap["TABLE"].invalidCount[fullyQualifiedName] = true
+				reportCase(fpath, "Issue with Partition using Expression on a table which cannot contain Primary Key / Unique Key on any column",
+					"https://github.com/yugabyte/yb-voyager/issues/698", "Remove the Constriant from the table definition", "TABLE", fullyQualifiedName, sqlStmtInfo.formattedStmt, UNSUPPORTED_FEATURES, EXPRESSION_PARTIITON_DOC_LINK)
+				return
```
makalaaneesh, line 519: Is it okay to return here? Should we break instead? 
THere could be other issues as well, right (for example, the ones being detected below...) ?
It would be better to  NOT stop when we find an issue.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -419,6 +413,138 @@ func checkStmtsUsingParser(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+		if partitionTablesMap[fullyQualifiedName] {
+			reportCase(fpath, ADDING_PK_TO_PARTITIONED_TABLE_ISSUE_REASON,
+				"https://github.com/yugabyte/yugabyte-db/issues/10074", "", "TABLE", fullyQualifiedName, sqlStmtInfo.formattedStmt, MIGRATION_CAVEATS, ADDING_PK_TO_PARTITIONED_TABLE_DOC_LINK)
+		} else {
+			primaryConsInAlter[fullyQualifiedName] = &sqlStmtInfo
+		}
+	}
+}
+
+func reportPartitionsRelatedIssues(createTableNode *pg_query.Node_CreateStmt, sqlStmtInfo sqlInfo, fpath string) {
```
makalaaneesh, line 444: Since there are multiple issues here, can you add a comment above this function listing the types of issues we detect?

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: In manual intervention cases where the user adds the ALTER DDL before the CREATE DDL, we should also be reporting that case. We were doing that earlier, so made that change here as well.

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: Added columns in analyze schema issue's reason as of now, not sure of assessment's unsupported feature.

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: Yeah agreed, we can report other as well so that user can understand all the issues and modify the DDL accordingly.

yb-voyager/cmd/analyzeSchema.go
makalaaneesh: Okay, I guess that ends up working because of our deferred statements logic?

makalaaneesh (approved):
LGTM!

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: Yes, right

PR #1756, opened 2024-10-16
Add voyager version to assessment and analyze reports

hbhanawat:


yb-voyager/cmd/templates/schema_analysis_report.txt
```diff
@@ -2,9 +2,10 @@
 |               Database Migration Report         |
 +-------------------------------------------------+
 
-Database Information
+Migration Information
```
sanyamsinghal, line 5: I am not sure if we should be changing `Database Information` -> `Migration Information`

Most of the info in it are just Database info only, except Migration Complexity of the database...

sanyamsinghal (approved):

yb-voyager/cmd/templates/schema_analysis_report.txt
hbhanawat: Even Voyager version cant be put under database head. Keeping it as migration until we find a better head.

PR #1758, opened 2024-10-16
Added large dataset test for import data file and provision to enable adaptive parallelism

shubham-yb:
- Added a import data file test with a large dataset
- Added the provision to enable adaptive parallelism in the test pipelines

### DIFF
--- a/migtests/scripts/functions.sh
+++ b/migtests/scripts/functions.sh
@@ -451,20 +451,26 @@ import_data_to_source_replica() {
 }
 
 import_data_file() {
-	yb-voyager import data file --export-dir ${EXPORT_DIR} \
-		--target-db-host ${TARGET_DB_HOST} \
-		--target-db-port ${TARGET_DB_PORT} \
-		--target-db-user ${TARGET_DB_USER} \
-		--target-db-password ${TARGET_DB_PASSWORD:-''} \
-		--target-db-schema ${TARGET_DB_SCHEMA:-''} \
-		--target-db-name ${TARGET_DB_NAME} \
-		--disable-pb true \
-		--send-diagnostics=false \
-		$* || {
-			cat ${EXPORT_DIR}/metainfo/dataFileDescriptor.json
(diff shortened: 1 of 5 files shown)


migtests/scripts/functions.sh
```diff
@@ -451,22 +451,27 @@ import_data_to_source_replica() {
...
+    --target-db-user ${TARGET_DB_USER}
+    --target-db-password ${TARGET_DB_PASSWORD:-''}
+    --target-db-schema ${TARGET_DB_SCHEMA:-''}
+    --target-db-name ${TARGET_DB_NAME}
+    --disable-pb true
+    --send-diagnostics=false
+    "
+
+    # Check if RUN_USING_ADAPTIVE_PARALLELISM is true
+    if [ "${RUN_USING_ADAPTIVE_PARALLELISM}" = "true" ]; then
```
priyanshi-yb, line 467: Should we have this for `import-data` as well, so that we can test both the commands as and when require.

migtests/tests/import-file/run-import-file-test
```diff
@@ -239,6 +239,14 @@ main() {
 	step "Import data file from GCS (csv): t1_quote_escape_char1.csv -> gcs_quote_escape_char1"
 	import_data_file --data-dir "gs://voyager-automation-data" --format csv --delimiter '|' \
 			--file-table-map "t1_quote_escape_char1.csv:gcs_quote_escape_char1" --quote-char="'" --escape-char="'"
+	
+	
+	if [ "${RUN_LARGE_IMPORT_DATA_FILE_TEST}" = true ] ; then
```
priyanshi-yb, line 244: should we have time check for this to validate that ther eis no regression in performance, if nto check with a fixed expected value, maybe just print the time taken for this import. Whoever is running the pipeline can check it manually maybe and see if there is not any regression? 🤔

priyanshi-yb (approved):
LGTM

PR #1760, opened 2024-10-17
Added checks in import data/schema for YB and dependency check for export schema/data in PG

ShivanshGahlot:
- Checking whether the migration user is a superuser or not before running import schema and import data on YB.
- Checking whether the pg_dump and pg_restore dependencies are of the correct version before running export in PG
    - We have a minimum version check on both the dependencies of version 14
    - A second level check which ensures that the version of both the dependencies is greater than or equal to the source db version 
- Checking whether debezium server is installed and findable

### DIFF
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -125,7 +125,7 @@ func getMappingForTableNameVsTableFileName(dataDirPath string, noWait bool) map[
 		time.Sleep(time.Second * 1)
 	}
 
-	pgRestorePath, err := srcdb.GetAbsPathAndCheckVersionOfPGCommand("pg_restore", source.DBVersion)
+	pgRestorePath, err := srcdb.GetAbsPathOfPGCommandAboveVersion("pg_restore", source.DBVersion)
 	if err != nil {
 		utils.ErrExit("could not get absolute path of pg_restore command: %s", err)
 	}
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -125,7 +125,7 @@ func getMappingForTableNameVsTableFileName(dataDirPath string, noWait bool) map[
 		time.Sleep(time.Second * 1)
 	}
 
-	pgRestorePath, err := srcdb.GetAbsPathOfPGCommand("pg_restore")
+	pgRestorePath, err := srcdb.GetAbsPathAndCheckVersionOfPGCommand("pg_restore", source.DBVersion)
 	if err != nil {
 		utils.ErrExit("could not get absolute path of pg_restore command: %s", err)
 	}
--- a/yb-voyager/cmd/export.go
+++ b/yb-voyager/cmd/export.go
@@ -20,9 +20,11 @@ import (
 	"os"
 	"strings"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/slices"
 
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/dbzm"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/metadb"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/srcdb"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
@@ -38,6 +40,7 @@ var useDebezium bool
 var runId string
 var excludeTableListFilePath string
 var tableListFilePath string
+var pgExportDependencies = []string{"pg_dump", "pg_restore"}
 
 var exportCmd = &cobra.Command{
 	Use:   "export",
@@ -383,4 +386,34 @@ func saveExportTypeInMSR() {
 	if err != nil {
 		utils.ErrExit("error while updating export type in meta db: %v", err)
 	}
+}
+
+func checkDependenciesForExport() error {
+	var errs []string
+	if source.DBType == POSTGRESQL {
+		sourceDBVersion := source.DB().GetVersion()
+		for _, binary := range pgExportDependencies {
+			_, err := srcdb.GetAbsPathOfPGCommandAboveVersion(binary, sourceDBVersion)
+			if err != nil {
+				errs = append(errs, err.Error())
+			} else {
+				fmt.Printf("%s is compatible with %s\n", binary, sourceDBVersion)
+				log.Infof("%s is compatible", binary)
+			}
+		}
+	}
+
+	if exportType == SNAPSHOT_AND_CHANGES || exportType == CHANGES_ONLY {
+		// Check for debezium
+		err := dbzm.FindDebeziumDistribution(source.DBType, false)
+		if err != nil {
+			errs = append(errs, err.Error())
+		}
+	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("%s", strings.Join(errs, "\n"))
+	}
+
+	return nil
 }
--- a/yb-voyager/cmd/export.go
+++ b/yb-voyager/cmd/export.go
@@ -403,7 +403,7 @@ func checkDependenciesForExport() error {
 		}
 	}
 
-	if exportType == SNAPSHOT_AND_CHANGES || exportType == CHANGES_ONLY {
+	if changeStreamingIsEnabled(exportType) || useDebezium {
 		// Check for debezium
 		err := dbzm.FindDebeziumDistribution(source.DBType, false)
 		if err != nil {
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -217,9 +217,11 @@ func exportData() bool {
 			utils.ErrExit("Source DB version check failed: %s", err)
 		}
 
-		err = source.DB().CheckDependencies()
+		err = checkDependenciesForExport()
 		if err != nil {
-			utils.ErrExit("Source DB dependencies check failed: %s", err)
+			color.Red("\nSome dependencies required for export data are missing: ")
+			utils.PrintAndLog("%s", err.Error())
+			utils.ErrExit("Please install the required dependencies and try again.")
 		}
 	}
 
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -216,6 +216,11 @@ func exportData() bool {
 		if err != nil {
 			utils.ErrExit("Source DB version check failed: %s", err)
 		}
+
+		err = source.DB().CheckDependencies()
+		if err != nil {
+			utils.ErrExit("Source DB dependencies check failed: %s", err)
+		}
 	}
 
 	source.DBVersion = source.DB().GetVersion()
--- a/yb-voyager/cmd/importSchema.go
+++ b/yb-voyager/cmd/importSchema.go
@@ -98,7 +98,7 @@ func importSchema() error {
 			utils.ErrExit("Failed to initialize the target DB: %s", err)
 		}
 		targetDBDetails = tdb.GetCallhomeTargetDBInfo()
-		//Marking tdb as nil back to not allow others to use it as this is just dummy initialisation of tdb 
+		//Marking tdb as nil back to not allow others to use it as this is just dummy initialisation of tdb
 		//with public schema so Reintialise tdb if required with proper configs when it is available.
 		tdb = nil
 		tconf.Schema = tconfSchema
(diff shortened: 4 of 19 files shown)


yb-voyager/cmd/common.go
```diff
@@ -125,7 +125,7 @@ func getMappingForTableNameVsTableFileName(dataDirPath string, noWait bool) map[
 		time.Sleep(time.Second * 1)
 	}
 
-	pgRestorePath, err := srcdb.GetAbsPathOfPGCommand("pg_restore")
+	pgRestorePath, err := srcdb.GetAbsPathAndCheckVersionOfPGCommand("pg_restore", source.DBVersion)
```
makalaaneesh, line 128: How about `GetAbsPathOfPGCommandAboveVersion` ?

yb-voyager/cmd/exportData.go
```diff
@@ -216,6 +216,11 @@ func exportData() bool {
 		if err != nil {
 			utils.ErrExit("Source DB version check failed: %s", err)
 		}
+
+		err = source.DB().CheckDependencies()
```
makalaaneesh, line 220: This check won't be behind any flag?

yb-voyager/cmd/importSchema.go
```diff
@@ -86,24 +86,42 @@ func importSchema() error {
 
 	tconf.Schema = strings.ToLower(tconf.Schema)
 
+	tconfSchema := tconf.Schema
```
makalaaneesh, line 89: A little concerned about this change. We don't usualyl use tdb in import-schema. Even if you look at line 90, we are doing some workarounds, just to get tdb working somehow for callhome. Let me think about this and get back. cc: @priyanshi-yb

yb-voyager/src/srcdb/postgres.go
```diff
@@ -415,6 +416,12 @@ func GetAbsPathOfPGCommand(cmd string) (string, error) {
 		currVersion := strings.Fields(string(stdout))[2]
 
 		if version.CompareSimple(currVersion, PG_COMMAND_VERSION[cmd]) >= 0 {
+			// Check if the version of the command is less than the source DB version
+			if version.CompareSimple(currVersion, sourceDBVersion) < 0 {
+				err = fmt.Errorf("version of %v (%v) is less than the source DB version (%v)", cmd, currVersion, sourceDBVersion)
```
makalaaneesh, line 421: we should NOT error out here. It's possible we find another version of pg_dump that satisfies our conditions. We should `continue`.

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -57,6 +57,8 @@ type TargetDB interface {
 	QueryRow(query string) *sql.Row
 	Exec(query string) (int64, error)
 	WithTx(fn func(tx *sql.Tx) error) error
+	CheckTargetDBVersion() error
```
makalaaneesh, line 60: This has no implementation in any of the structs? Remove it then?

yb-voyager/src/srcdb/srcdb.go
```diff
@@ -59,6 +59,7 @@ type SourceDB interface {
 	CheckSourceDBVersion() error
 	GetMissingExportSchemaPermissions() ([]string, error)
 	GetMissingExportDataPermissions(exportType string) ([]string, error)
+	CheckDependencies() error
```
makalaaneesh, line 62: The more I think about it, the more I feel that this shouldn't be a method in sourceDB interface. 
RIght now, we have methods in sourceDB interface that do a lot more than just interact with the source DB. For example `ExportData` which essentially just calls pg_dump. Ideally, even `ExportData` shouldn't be part of SourceDB interface. It should be a method outside that takes a SourceDB as an input. 

Let me think about this more and get back.

yb-voyager/cmd/importSchema.go
```diff
@@ -86,24 +86,42 @@ func importSchema() error {
...
+		missingPermissions, err := tdb.GetMissingImportSchemaPermissions()
 		if err != nil {
-			utils.ErrExit("Failed to initialize the target DB: %s", err)
+			utils.ErrExit("Failed to get missing import schema permissions: %s", err)
+		}
+		if len(missingPermissions) > 0 {
+			utils.PrintAndLog(color.RedString("The target database is missing the following permissions required for importing schema:"))
+			output := strings.Join(missingPermissions, "\n")
+			utils.PrintAndLog(output)
+			utils.ErrExit("Please grant the required permissions and retry the import.")
```
makalaaneesh, line 114: Let's add a prompt here: "Are you sure you want to continue?" instead of exiting directly.

priyanshi-yb: From our internal code handling, yes, we can run import-schema without superuser, but I think as we are creating schemas here, which can be various object types, and I am not sure what object types might need a higher privileged user or superuser in that case. So maybe it's better to just ask for it?

yb-voyager/cmd/importSchema.go
```diff
@@ -86,24 +86,42 @@ func importSchema() error {
...
-		//Marking tdb as nil back to not allow others to use it as this is just dummy initialisation of tdb 
-		//with public schema so Reintialise tdb if required with proper configs when it is available.
-		tdb = nil
-		tconf.Schema = tconfSchema
 	}
 
+	//Marking tdb as nil back to not allow others to use it as this is just dummy initialisation of tdb
+	//with public schema so Reintialise tdb if required with proper configs when it is available.
+	tdb = nil
+	tconf.Schema = tconfSchema
```
sanyamsinghal, line 123: will setting it to nil drops the connection?
Please check shall we Close() it separately.

yb-voyager/cmd/importSchema.go
sanyamsinghal: +1

yb-voyager/src/srcdb/postgres.go
sanyamsinghal: Yeah, agree with Aneesh.
Check should be outside the for loop, to check and return err it couldn't find a suitable version.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1329,3 +1353,26 @@ type NodeMetrics struct {
 	Status  string
 	Error   string
 }
+
+func (yb *TargetYugabyteDB) checkIfCurrentUserIsSuperUser() (bool, error) {
+	// SELECT rolname, rolsuper FROM pg_roles WHERE rolname='yugabyte';
```
sanyamsinghal, line 1358: remove comment?

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1329,3 +1353,26 @@ type NodeMetrics struct {
...
+	query := "SELECT rolsuper FROM pg_roles WHERE rolname=current_user"
+	rows, err := yb.Query(query)
+	if err != nil {
+		return false, fmt.Errorf("querying if user is superuser: %w", err)
+	}
+	defer rows.Close()
```
sanyamsinghal, line 1364: nit: log the closeErr

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1329,3 +1353,26 @@ type NodeMetrics struct {
...
+
+	var isSuperUser bool
+	if rows.Next() {
+		err = rows.Scan(&isSuperUser)
+		if err != nil {
+			return false, fmt.Errorf("scanning row for superuser: %w", err)
+		}
+	}
```
sanyamsinghal, line 1372: Add a else case here, basically if we dont get a row then the error message should indicate that instead of return isSuperUser as false.

sanyamsinghal:
```
else {
        // Handle the case where no result is returned
        return false, fmt.Errorf("no current user found in pg_roles")
    }
```

yb-voyager/src/srcdb/postgres.go
```diff
@@ -1502,3 +1509,26 @@ func (pg *PostgreSQL) listSchemasMissingUsagePermission() ([]string, error) {
 
 	return schemasMissingUsagePermission, nil
 }
+
+func (pg *PostgreSQL) CheckDependencies() error {
+	var errs []string
+
+	// List of pg binaries to check
+	binaries := []string{"pg_dump", "pg_restore"}
```
sanyamsinghal, line 1517: Lets have a global variable for this with name: `pgExportDependecies` or something

yb-voyager/src/srcdb/postgres.go
```diff
@@ -1502,3 +1509,26 @@ func (pg *PostgreSQL) listSchemasMissingUsagePermission() ([]string, error) {
...
+			errs = append(errs, err.Error())
+		} else {
+			log.Infof("%s is compatible", binary)
+		}
+	}
+
+	if len(errs) > 0 {
+		return errors.New(strings.Join(errs, "\n"))
+	}
+
```
sanyamsinghal, line 1532: how is the output coming out for this.
I am just thinking if that is a good way to combine all the errors and make New().

how about simply returning
```
    if len(errs) > 0 {
        return fmt.Errorf("%w or %v", errs)
    }
```

yb-voyager/cmd/importSchema.go
sanyamsinghal: @makalaaneesh  I think superuser is required for setting some of the session variables in import data.
And since that requires superuser we decided to keep it as it is for import schema also

makalaaneesh: @sanyamsinghal Yes, but check this out - we recently removed this restriction. (we prompt user, and allow them to continue) https://github.com/yugabyte/yb-voyager/pull/1420

yb-voyager/src/srcdb/srcdb.go
makalaaneesh: I think this function is better implemented outside pf sourceDB interface. 
We should only keep methods in sourceDB that directly interacts with the DB. This function is merely checking if all dependencies are met. In future, I think even ExportData should be moved out of SourceDB, since we are essentially using pg_dump/ora2pg to get data, not directly interacting with DB. If you look at how live migration/dbzm is implemented, the logic is defined in dbzm/exportData.go and it depends on SourceDB for getting config/metadata/etc. 

For now, I think you can define it in cmd pkg. 
This will also solve the circular dependency issue you were facing with calling code from dbzm pkg. 
cc: @priyanshi-yb  @sanyamsinghal

yb-voyager/cmd/importSchema.go
makalaaneesh: Instead of introducing tdb here, how about just using raw connections directly (like how the rest of import-schema is implemented) to figure out if you have the right permissions.

Not defining `GetMissingImportSchemaPermissions` in target db interface also makes sense because we only import schema into YB now.

yb-voyager/src/srcdb/srcdb.go
ShivanshGahlot: Done. Added it to export.go. Added a check for DBZM too.

yb-voyager/src/srcdb/postgres.go
ShivanshGahlot: Done. But the location of the entire function has been changed. I have added it to export.go. So please have a look there.

yb-voyager/src/srcdb/postgres.go
ShivanshGahlot: Done

yb-voyager/src/tgtdb/target_db_interface.go
ShivanshGahlot: Yep removed

yb-voyager/src/tgtdb/yugabytedb.go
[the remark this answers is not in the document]
ShivanshGahlot: Done

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1319,7 +1319,31 @@ func (yb *TargetYugabyteDB) ClearMigrationState(migrationUUID uuid.UUID, exportD
...
+	// check if the user is a superuser
+	isSuperUser, err := yb.checkIfCurrentUserIsSuperUser()
+	if err != nil {
+		return nil, fmt.Errorf("checking if user is superuser: %w", err)
+	}
+	if !isSuperUser {
+		errorMsg := fmt.Sprintf("User %s is not a superuser.", yb.tconf.User)
+		return []string{errorMsg}, nil
+	}
+
```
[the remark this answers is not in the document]
ShivanshGahlot: Yeah I have removed the function GetMissingImportSchemaPermissions as discussed by Aneesh above since we anyways require it only for YB. I have added it to importSchema.go only. Can you have a look there?

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1329,3 +1353,26 @@ type NodeMetrics struct {
 	Status  string
 	Error   string
 }
+
+func (yb *TargetYugabyteDB) checkIfCurrentUserIsSuperUser() (bool, error) {
```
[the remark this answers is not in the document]
ShivanshGahlot: Done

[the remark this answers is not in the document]
ShivanshGahlot: Made it IsCurrentUserSuperUser since I will be calling it in importSchema.go as well

yb-voyager/src/tgtdb/yugabytedb.go
ShivanshGahlot: Done

yb-voyager/src/tgtdb/yugabytedb.go
ShivanshGahlot: Done

yb-voyager/src/srcdb/postgres.go
```diff
@@ -415,6 +415,11 @@ func GetAbsPathOfPGCommand(cmd string) (string, error) {
 		currVersion := strings.Fields(string(stdout))[2]
 
 		if version.CompareSimple(currVersion, PG_COMMAND_VERSION[cmd]) >= 0 {
```
makalaaneesh, line 417: nit: 
```
if version.CompareSimple(currVersion, PG_COMMAND_VERSION[cmd]) >= 0 && version.CompareSimple(currVersion, sourceDBVersion) >= 0 {
    return path, nil
}
```

makalaaneesh (approved):
LGTM! Pls let approval from @sanyamsinghal also before merging

yb-voyager/src/tgtdb/yugabytedb.go
ShivanshGahlot: > func (pgx.Rows) Close()
Close closes the rows, making the connection ready for use again. It is safe to call Close after rows is already closed.

It doesnt return an error

sanyamsinghal (approved):
LGTM.

yb-voyager/cmd/export.go
```diff
@@ -384,3 +387,33 @@ func saveExportTypeInMSR() {
...
+			if err != nil {
+				errs = append(errs, err.Error())
+			} else {
+				fmt.Printf("%s is compatible with %s\n", binary, sourceDBVersion)
+				log.Infof("%s is compatible", binary)
+			}
+		}
+	}
+
+	if exportType == SNAPSHOT_AND_CHANGES || exportType == CHANGES_ONLY {
```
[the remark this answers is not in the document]
ShivanshGahlot: Added check for useDebezium

PR #1761, opened 2024-10-17
Fix Table Inheritance handling and consider them as normal tables in the workflow

priyanshi-yb:
1. Added an end-to-end test for inherited tables, migrating them as normal tables 
005a188a6f01820b2551ed00f338d67d6106eaea this commit adds test which fails in import-data phase.
```
if required quote column names: quote attribute name for "state" in table [CurrentName=(cities) SourceName=(cities) TargetName=(cities)]: find best matching column name for "state" in table [CurrentName=(cities) SourceName=(cities) TargetName=(cities)]: column "state" not found amongst table columns [population elevation name]
```
2. Modifying the `GetPartitions` and `ParentTableOfPartition` function's queries to only considers partitioned tables, not inherited ones.

### DIFF
--- a/migtests/tests/pg/assessment-report-test/expectedAssessmentReport.json
+++ b/migtests/tests/pg/assessment-report-test/expectedAssessmentReport.json
@@ -36,8 +36,8 @@
 			},
 			{
 				"ObjectType": "TABLE",
-				"TotalCount": 63,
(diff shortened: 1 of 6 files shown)


sanyamsinghal:
@priyanshi-yb can we close this PR if not required anymore.

priyanshi-yb:
Closing this; will take this up separately.

PR #1764, opened 2024-10-21
Fix misleading error in batch for live migration

makalaaneesh:
When using pgx SendBatch, there can be two types of errors thrown:
1. Error while preparing the statement - this is preprocessing (parsing, preparinng statements, etc) that pgx will do before sending the batch. Examples - syntax error.
 No matter which statement in the batch has an issue, the error will be thrown on calling br.Exec() for the first time.
2. Error while executing the statement - this is the actual execution of the statement, and the error comes from the DB.
Examples - constraint violation, etc.
In this case, we get the error on the appropriate br.Exec() call associated with the statement that failed.

Therefore, if error is thrown on the first br.Exec() call, it could be either of the above cases. Modified the logs to indicate this as part of this PR. 

Reference - https://github.com/jackc/pgx/issues/872
This ideally needs to be fixed in pgx library, and has been partially fixed in https://github.com/jackc/pgx/pull/1441 (pgx v5)

### DIFF
--- a/yb-voyager/src/tgtdb/postgres.go
+++ b/yb-voyager/src/tgtdb/postgres.go
@@ -535,9 +535,9 @@ func (pg *TargetPostgreSQL) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
 		for i := 0; i < len(batch.Events); i++ {
 			res, err := br.Exec()
 			if err != nil {
-				log.Errorf("error executing stmt for event in batch(%s): %v. Event VSNs in batch:%v", batch.ID(), err, batch.GetAllVsns())
+				log.Errorf("error executing stmt for event with vsn(%d) in batch(%s): %v", batch.Events[i].Vsn, batch.ID(), err)
 				closeBatch()
-				return false, fmt.Errorf("error executing stmt for event in batch(%s): %v", batch.ID(), err)
+				return false, fmt.Errorf("error executing stmt for event with vsn(%d): %v", batch.Events[i].Vsn, err)
 			}
 			switch true {
 			case res.Insert():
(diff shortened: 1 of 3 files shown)


yb-voyager/src/tgtdb/postgres.go
```diff
@@ -535,9 +535,25 @@ func (pg *TargetPostgreSQL) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
...
+				//  No matter which statement in the batch has an issue, the error will be thrown on calling br.Exec() for the first time.
+				// 2. Error while executing the statement - this is the actual execution of the statement, and the error comes from the DB.
+				// 	Examples - constraint violation, etc.
+				//  In this case, we get the error on the appropriate br.Exec() call associated with the statement that failed.
+				// Therefore, if error is thrown on the first br.Exec() call, it could be either of the above cases.
+				// Reference - https://github.com/jackc/pgx/issues/872
+				// This ideally needs to be fixed in pgx library.
+				errorMsg := ""
+				if i == 0 {
+					errorMsg = fmt.Sprintf("error preparing statements for events in batch (%s) or when executing event with vsn(%d)", batch.ID(), batch.Events[i].Vsn)
```
sanyamsinghal, line 550: is printing start and end of the batch enough here? or should we print the VSNs of all events in that batch.

sanyamsinghal (approved):

yb-voyager/src/tgtdb/postgres.go
makalaaneesh: Yeah, thought of this @sanyamsinghal , but I plan to add debug logs for every statement in a subsequent PR, which I thought would be more valuable than simply listing all the VSNs here. 

Let me know if you think I should still add it here as well!

sanyamsinghal: Yeah, i think we can add it here also, since it won't flood the logs, just the last line in case of failures.
We can use debug logs for more level of details(like logging particular events)

makalaaneesh: Done.

PR #1770, opened 2024-10-21
Introduced UnsupportedConstructsDetector interface and parse tree Traversal function to cover all cases for the unsupported constructs

sanyamsinghal:
- FuncCallDetector, ColumnRefDetector and XmlExprDetector are the implementations of the interface
- TraverseParseTree() covers the whole parse tree(each node and field) for a given SQL stmt
- added unit tests in unsupported_constructs_test.go for all three features detection(advisory locks, sys cols, and xml)

https://yugabyte.atlassian.net/browse/DB-13168

Sample HTML report
image

Test Pipeline: https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/3752/


yb-voyager/src/queryparser/query_parser.go
```diff
@@ -0,0 +1,87 @@
...
+func (qp *QueryParser) Parse() error {
+	tree, err := pg_query.Parse(qp.QueryString)
+	if err != nil {
+		return err
+	}
+	qp.ParseTree = tree
+	return nil
+}
+
+func (qp *QueryParser) GetUnsupportedQueryConstructs() ([]utils.UnsupportedQueryConstruct, error) {
```
makalaaneesh, line 41: Let's call parse if not already?

yb-voyager/src/queryparser/unsupported_constructs.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+const (
+	ADVISORY_LOCKS = "Advisory Locks"
+	SYSTEM_COLUMNS = "System Columns"
+	XML_FUNCTIONS  = "XML Functions"
+)
+
+// To Add a new unsupported query construct implement this interface
+type UnsupportedConstructDetector interface {
```
makalaaneesh, line 16: I like your general idea of structuring it as detectors! But, do we really need an interface with structs implementing them? 
How about we just work with functions? https://go.dev/play/p/zNiENQOAmVI

Reasoning is same as in my earlier PR. None of these detectors really need to maintain any state over the lifetime of the struct; it just performs some operation, and returns some values. A function is better-suited. For example, look at the `YBValueConverterSuite`

makalaaneesh (commented):
Went over it at a very high-level. Nicely structured! Will go over in detail again and post more comments

yb-voyager/src/queryparser/unsupported_constructs.go
```diff
@@ -0,0 +1,110 @@
...
+	Detect(msg protoreflect.Message) ([]string, error)
```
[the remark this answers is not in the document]
sanyamsinghal: yeah right, func/col names, any Notes/Details field can be also there.
I didn't add a provision for that right now, lets do it later as and when needed

yb-voyager/src/queryparser/unsupported_constructs.go
sanyamsinghal: Yeah, i agree generally if we don't need to maintain any state we should avoid interface/structs

But couple of points here to choose interface:
1. Current design enforces the implementation of Detect() for every new detector.
2. Due to struct, in future if we requirements like, for example
- tell how many times xmlparse()/xpath() function has occured
- any niche check for a particular detector
- or any other specific stats based on the workload queries

overall i think interface approach is less simpler than function based, but in future for this codepath as we enhance this might become more useful and easier to implement complex checks.

yb-voyager/src/queryparser/query_parser.go
sanyamsinghal: didn't get you

yb-voyager/src/queryparser/unsupported_constructs.go
makalaaneesh: >Current design enforces the implementation of Detect() for every new detector.

This will be true even if we use funcs because every detector will have to be defined to meet that same function signature. 

>Due to struct, in future if we requirements like, for example
tell how many times xmlparse()/xpath() function has occured
any niche check for a particular detector
or any other specific stats based on the workload queries

Since your Detect() is defined to operate on the `protoreflect.Message` , things like stats can be computed in your processor function. 

However, there is one particular case that comes to mind: What if the detection logic requires to look at multiple nodes? In that case you can't just detect just by looking at a single message, you might only be able to detect after parsing the entire tree.  In which case I would define the interface like this instead: 

```
type UnsupportedConstructDetector interface {
    ProcessMessage(msg protoreflect.Message) (error)
    GetUnsupportedConstructs() ([]string)
```
Does that make sense? We can also discuss this on call if it's easier.

sanyamsinghal: Thinking out loud 
> What if the detection logic requires to look at multiple nodes?
Even in this case, their will be parent node(may be root node) on which we need to look for 2 nodes under it. Which can be done with the current approach also.

yb-voyager/src/queryparser/traversal.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+		log.Debugf("Field: %s, Type: %s\n", fieldDesc.Name(), fieldDesc.Kind())
+		value := msg.Get(fieldDesc)
+		switch {
+		case fieldDesc.IsList():
+			list := value.List()
+			for j := 0; j < list.Len(); j++ {
+				elem := list.Get(j)
+				// every elem in the list have same Kind
+				if fieldDesc.Kind() == protoreflect.MessageKind {
```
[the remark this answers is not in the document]
priyanshi-yb: Or if this check is required on field, then it should be in case right?

yb-voyager/src/queryparser/traversal.go
```diff
@@ -0,0 +1,173 @@
...
+			log.Debugf("Scalar field of Type: %s, value: %v\n", fieldDesc.Kind(), value.Interface())
+		default:
+			log.Infof("field kind case not covered: %s\n", fieldDesc.Kind())
+		}
+	}
+
+	return nil
+}
+
+// ScalarKind is all datatypes including enums
```
priyanshi-yb, line 163: Is this the kind for column datatypes? If so, are we covering all the types in the list?

priyanshi-yb (commented):
A few comments, overall structure looks Nice!

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -974,7 +969,11 @@ func fetchUnsupportedQueryConstructs() ([]utils.UnsupportedQueryConstruct, error
 		}
 	}
 
-	// TODO: sort the slice for better readability
+	// sort the slice to group same constructType in html and json reports
+	log.Infof("sorting the result slice based on construct type")
```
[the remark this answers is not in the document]
sanyamsinghal: In HTML it will a consolidated report. Each type vs the queries.
But in json, yes it will be multiple queries.

Makes sense to use a map here.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -950,21 +953,13 @@ func fetchUnsupportedQueryConstructs() ([]utils.UnsupportedQueryConstruct, error
...
-		if strings.HasPrefix(upperQuery, "CREATE") || strings.HasPrefix(upperQuery, "INSERT") ||
-			strings.HasPrefix(upperQuery, "UPDATE") || strings.HasPrefix(upperQuery, "DELETE") {
-			continue
-		}
-
+		log.Debugf("fetched unsupported query constructs for query-%s", query)
 		queryParser := queryparser.New(query)
 		err := queryParser.Parse()
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse query-%s: %w", query, err)
```
priyanshi-yb, line 960: Just thinking: should we log the error here as well and not return, just in case there is some query in pgss that parser failed to parse? Not sure if there can be any case.

yb-voyager/cmd/assessMigrationCommand.go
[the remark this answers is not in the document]
sanyamsinghal: good idea i can change it from slice to map in AssessmentReport struct

yb-voyager/src/queryparser/traversal.go
sanyamsinghal: Yes it will be column mostly, not sure of any special cases yet.
About coverage, yes i checked the code, apart from these only two message kinds are `Message` and `Group` for protobufs

yb-voyager/src/queryparser/traversal.go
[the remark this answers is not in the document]
sanyamsinghal: you are right.
The thing is every elem will have same kind as list, also there is no function like `elem.Kind()`
Makes sense to move it to switch case

yb-voyager/src/queryparser/traversal.go
```diff
@@ -0,0 +1,173 @@
...
+	fields := msg.Descriptor().Fields()
+	if fields == nil || fields.Len() == 0 {
+		return nil
+	}
+
+	for i := 0; i < fields.Len(); i++ {
+		fieldDesc := fields.Get(i)
+
+		// check if the field is set or not (for example: fields with DEFAULT values(set) won't be NULL)
+		if !msg.Has(fieldDesc) {
```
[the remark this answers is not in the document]
sanyamsinghal: As per documentation
```
// Fields is a list of nested field declarations.
Fields() FieldDescriptors
// Oneofs is a list of nested oneof declarations.
Oneofs() OneofDescriptors
```

It is fetching all declarations of the message from [pg_query.proto file](https://github.com/pganalyze/libpg_query/blob/16-latest/protobuf/pg_query.proto)

In my experiment also unnecessary/undefined fields were coming. so this check is required.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -723,13 +724,12 @@ func generateAssessmentReport() (err error) {
 	}
 	assessmentReport.UnsupportedFeatures = append(assessmentReport.UnsupportedFeatures, unsupportedFeatures...)
 
-	if os.Getenv(REPORT_UNSUPPORTED_QUERY_CONSTRUCTS) == "true" && source.DBType == POSTGRESQL {
```
makalaaneesh, line 726: As discussed, add this back?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -950,21 +953,13 @@ func fetchUnsupportedQueryConstructs() ([]utils.UnsupportedQueryConstruct, error
...
 		query := executedQueries[i]
-
-		// Check if the query starts with CREATE, INSERT, UPDATE, or DELETE
-		upperQuery := strings.ToUpper(strings.TrimSpace(query))
-		if strings.HasPrefix(upperQuery, "CREATE") || strings.HasPrefix(upperQuery, "INSERT") ||
-			strings.HasPrefix(upperQuery, "UPDATE") || strings.HasPrefix(upperQuery, "DELETE") {
-			continue
-		}
-
+		log.Debugf("fetched unsupported query constructs for query-%s", query)
```
makalaaneesh, line 956: "fetching"
"query - [%s]" (just so it's easier to read)

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: Yeah, log.Error and continue and makes sense to me!

yb-voyager/cmd/assessMigrationCommand.go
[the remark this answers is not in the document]
makalaaneesh: Agreed. Similar to unsupported features, where we have one entry for a feature, and a list of object names that we detected

Sorting seems like a hacky way of doing it..

yb-voyager/src/queryparser/query_parser.go
```diff
@@ -0,0 +1,90 @@
...
+
+	log.Debugf("Query: %s\n", qp.QueryString)
+	log.Debugf("ParseTree: %+v\n", qp.ParseTree)
+	detectors := []UnsupportedConstructDetector{
+		NewFuncCallDetector(),
+		NewColumnRefDetector(),
+		NewXmlExprDetector(),
+	}
+
+	compositeDetector := &CompositeDetector{detectors: detectors}
```
makalaaneesh, line 55: As discussed, this seems fancy and nice, but IMO, unnecessary complication 😄 
Would suggest just looping over the detectors in your processor function

yb-voyager/src/queryparser/query_parser.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"fmt"
+
+	pg_query "github.com/pganalyze/pg_query_go/v5"
+	log "github.com/sirupsen/logrus"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+const (
```
makalaaneesh, line 12: Define these in traversal.go? it is only used there.

yb-voyager/src/queryparser/unsupported_constructs.go
```diff
@@ -0,0 +1,121 @@
...
+
+// Detect checks if a FuncCall node uses an unsupported function.
+func (d *FuncCallDetector) Detect(msg protoreflect.Message) ([]string, error) {
+	if getMsgFullName(msg) != PG_QUERY_FUNCCALL_NODE {
+		return nil, nil
+	}
+
+	funcName := getFuncNameFromFuncCall(msg)
+	log.Debugf("fetched function name from %s node: %q", PG_QUERY_FUNCCALL_NODE, funcName)
+	if constructType, ok := d.unsupportedFuncs[funcName]; ok {
```
makalaaneesh, line 49: nit for readability : 
`if constructType, isUnsupported := d.unsupportedFuncs[funcName]; isUnsupported {`

yb-voyager/src/queryparser/unsupported_constructs.go
```diff
@@ -0,0 +1,121 @@
...
+	Detect(msg protoreflect.Message) ([]string, error)
+}
+
+type FuncCallDetector struct {
+	// right now it covers Advisory Locks and XML functions
+	unsupportedFuncs map[string]string
+}
+
+func NewFuncCallDetector() *FuncCallDetector {
+	fcd := &FuncCallDetector{
```
makalaaneesh, line 27: nit: prepare the map, before passing it to the constructor?

yb-voyager/src/queryparser/unsupported_constructs.go
```diff
@@ -0,0 +1,121 @@
...
+// Detect checks if a FuncCall node uses an unsupported function.
+func (d *FuncCallDetector) Detect(msg protoreflect.Message) ([]string, error) {
+	if getMsgFullName(msg) != PG_QUERY_FUNCCALL_NODE {
+		return nil, nil
+	}
+
+	funcName := getFuncNameFromFuncCall(msg)
+	log.Debugf("fetched function name from %s node: %q", PG_QUERY_FUNCCALL_NODE, funcName)
+	if constructType, ok := d.unsupportedFuncs[funcName]; ok {
+		log.Debugf("detected unsupported function: %s", funcName)
```
makalaaneesh, line 50: Consider also logging the msg? . Otherwise, it might not be useful. Same for the other types of detection logs

yb-voyager/src/queryparser/unsupported_constructs.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return nil, nil
+}
+
+type ColumnRefDetector struct {
+	unsupportedColumns map[string]string
+}
+
+func NewColumnRefDetector() *ColumnRefDetector {
+	crd := &ColumnRefDetector{
```
makalaaneesh, line 61: same comment: prepare the inputs before the constructor.

yb-voyager/src/queryparser/unsupported_constructs.go
```diff
@@ -0,0 +1,121 @@
...
+		unsupportedColumns: make(map[string]string),
+	}
+
+	for _, colName := range unsupportedSysCols {
+		crd.unsupportedColumns[colName] = SYSTEM_COLUMNS
+	}
+	log.Debug("initialized ColumnRefDetector with unsupported columns")
```
makalaaneesh, line 68: not really useful..

yb-voyager/src/queryparser/unsupported_constructs.go
```diff
@@ -0,0 +1,121 @@
...
+		log.Debugf("detected unsupported system column: %s", colName)
+		return []string{constructType}, nil
+	}
+	return nil, nil
+}
+
+type XmlExprDetector struct{}
+
+func NewXmlExprDetector() *XmlExprDetector {
+	log.Debug("initialized XmlExprDetector")
```
makalaaneesh, line 90: not useful

yb-voyager/src/queryparser/unsupported_constructs_test.go
```diff
@@ -0,0 +1,231 @@
...
+import (
+	"testing"
+
+	pg_query "github.com/pganalyze/pg_query_go/v5"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+// TestFuncCallDetector tests the Advisory Lock Detector.
+func TestFuncCallDetector(t *testing.T) {
```
makalaaneesh, line 12: nice tests! How about one with all three present, just to ensure that multiple constructs are returned.

yb-voyager/src/queryparser/traversal.go
```diff
@@ -0,0 +1,173 @@
...
+	log.Debugf("Traversing NodeType: %s\n", nodeType)
+	// applying the processor to the current node
+	if err := processor(msg); err != nil {
+		log.Debugf("error processing node %s: %v", nodeType, err)
+		return fmt.Errorf("error processing node %s: %w", nodeType, err)
+	}
+
+	// Reference Oneof - https://protobuf.dev/programming-guides/proto3/#oneof
+	if nodeType == PG_QUERY_NODE_NODE {
+		nodeField := msg.WhichOneof(msg.Descriptor().Oneofs().ByName("node"))
```
makalaaneesh, line 91: call this `childNodeField` ?

yb-voyager/src/queryparser/traversal.go
```diff
@@ -0,0 +1,173 @@
...
+		if nodeField != nil {
+			value := msg.Get(nodeField)
+			if value.IsValid() {
```
makalaaneesh, line 94: Do we know when it won't be valid? Should we log ?

yb-voyager/src/queryparser/traversal.go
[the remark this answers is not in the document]
makalaaneesh: So `fields` is just a list of "declarations", not the actual field values? And you call `msg.Get(fieldDesc)` to actually get the field? 
Pls add your above msg as a comment in this function!

yb-voyager/src/queryparser/traversal.go
```diff
@@ -0,0 +1,173 @@
...
+			err := TraverseParseTree(value.Message(), visited, processor)
+			if err != nil {
+				log.Debugf("error traversing field %s: %v", fieldDesc.Name(), err)
+				return fmt.Errorf("error traversing field %s: %w", fieldDesc.Name(), err)
+			}
+
+		case IsScalarKind(fieldDesc.Kind()):
+			log.Debugf("Scalar field of Type: %s, value: %v\n", fieldDesc.Kind(), value.Interface())
+		default:
+			log.Infof("field kind case not covered: %s\n", fieldDesc.Kind())
```
makalaaneesh, line 156: Would debug be better? I fear it might flood the logs

yb-voyager/src/queryparser/traversal.go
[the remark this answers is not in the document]
makalaaneesh: I didn't understand why "Makes sense to move it to switch case".

yb-voyager/src/queryparser/traversal.go
```diff
@@ -0,0 +1,173 @@
...
+		case fieldDesc.IsList():
+			list := value.List()
+			for j := 0; j < list.Len(); j++ {
+				elem := list.Get(j)
+				// every elem in the list have same Kind
+				if fieldDesc.Kind() == protoreflect.MessageKind {
+					err := TraverseParseTree(elem.Message(), visited, processor)
+					if err != nil {
+						log.Debugf("error traversing field %s: %v", fieldDesc.Name(), err)
+						return fmt.Errorf("error traversing field %s: %w", fieldDesc.Name(), err)
```
makalaaneesh, line 141: Here, as well, if there is an error whlie processing some node, do we want to fail altogether? I would think we logError and move on?

yb-voyager/src/queryparser/traversal.go
```diff
@@ -0,0 +1,173 @@
...
+					}
+				}
+			}
+
+		case fieldDesc.Kind() == protoreflect.MessageKind:
+			err := TraverseParseTree(value.Message(), visited, processor)
+			if err != nil {
+				log.Debugf("error traversing field %s: %v", fieldDesc.Name(), err)
+				return fmt.Errorf("error traversing field %s: %w", fieldDesc.Name(), err)
```
makalaaneesh, line 150: same comment as above

yb-voyager/src/queryparser/helpers.go
```diff
@@ -0,0 +1,100 @@
...
+	"xmlagg", "xmlcomment", "xmlconcat2", // XML Aggregation and Construction
+	"xmlexists", "xmlvalidate", // XML Existence and Validation
+	"xpath", "xpath_exists", // XPath Functions
+	"xml_in", "xml_out", "xml_recv", "xml_send", // System XML I/O
+	"xml", // Data Type Conversion
+}
+
+// Sample example: {func_call:{funcname:{string:{sval:"pg_advisory_lock"}}
+func getFuncNameFromFuncCall(funcCallNode protoreflect.Message) string {
+	funcnameField := funcCallNode.Get(funcCallNode.Descriptor().Fields().ByName("funcname"))
```
makalaaneesh, line 32: You should check here in this function that the node type is matching what you expect. (a PG_QUERY_FUNCCALL_NODE)

yb-voyager/src/queryparser/helpers.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	nodeType := node.Descriptor().FullName()
+	switch nodeType {
+	case PG_QUERY_STRING_NODE:
+		strField := node.Descriptor().Fields().ByName("sval")
+		strValue := node.Get(strField)
+		return strValue.String()
+	// example: SELECT * FROM employees;
+	case PG_QUERY_ASTAR_NODE:
```
makalaaneesh, line 91: remove this case, if we're not doing anything?

yb-voyager/src/queryparser/helpers.go
```diff
@@ -0,0 +1,100 @@
...
+			names = append(names, name)
+		}
+	}
+	if len(names) == 0 {
+		return ""
+	}
+	return names[len(names)-1] // ignoring schema_name
+}
+
+// Sample example:: {column_ref:{fields:{string:{sval:"s"}}  fields:{string:{sval:"tableoid"}}  location:7}
```
makalaaneesh, line 66: @priyanshi-yb can you pls take a look at this func, looks similar to what you were dealing with? I will skip this.

yb-voyager/src/queryparser/helpers.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	if len(names) == 0 {
+		return ""
+	}
+	return names[len(names)-1] // ignoring schema_name
+}
+
+// Sample example:: {column_ref:{fields:{string:{sval:"xmax"}}
+func getColNameFromColumnRef(columnRefNode protoreflect.Message) string {
+	fields := columnRefNode.Get(columnRefNode.Descriptor().Fields().ByName("fields"))
```
makalaaneesh, line 50: Same; you should validate that the node type is matching what you expect.

yb-voyager/src/queryparser/helpers.go
```diff
@@ -0,0 +1,100 @@
...
+		item := funcnameList.Get(i)
+		name := getStringValueFromNode(item.Message())
+		if name != "" {
+			names = append(names, name)
+		}
+	}
+	if len(names) == 0 {
+		return ""
+	}
+	return names[len(names)-1] // ignoring schema_name
```
makalaaneesh, line 45: Would it be easier if we just process the last item in funcnameList ?  (to ignore schema_name)

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: yes

yb-voyager/src/queryparser/query_parser.go
sanyamsinghal: ok will remove it. It can be complicated to understand..

yb-voyager/src/queryparser/query_parser.go
sanyamsinghal: Also used in `unsupported_constructs.go`

yb-voyager/src/queryparser/traversal.go
priyanshi-yb: > If so, are we covering all the types in the list?

With this I meant the datatypes in this list

yb-voyager/src/queryparser/unsupported_constructs.go
sanyamsinghal: `msg` variable directly wont be of any use since it has pointers also.
I was thinking `msg.Descriptor()` but check it for a case and didn't see anything helpful in that either

```
func call name "pg_advisory_lock" in msg MessageDescriptor{
	Syntax:   proto3
	FullName: pg_query.FuncCall
	Fields: [{
		Name:        funcname
		Number:      1
		Cardinality: repeated
		Kind:        message
		HasJSONName: true
		JSONName:    "funcname"
		IsList:      true
		Message:     pg_query.Node
	}, {
		Name:        args
		Number:      2
		Cardinality: repeated
		Kind:        message
		HasJSONName: true
		JSONName:    "args"
		IsList:      true
		Message:     pg_query.Node
	}, {
		Name:        agg_order
		Number:      3
		Cardinality: repeated
		Kind:        message
		HasJSONName: true
		JSONName:    "agg_order"
		IsList:      true
		Message:     pg_query.Node
	}, {
		Name:        agg_filter
		Number:      4
		Cardinality: optional
		Kind:        message
		HasJSONName: true
		JSONName:    "agg_filter"
		HasPresence: true
		Message:     pg_query.Node
	}, {
		Name:        over
		Number:      5
		Cardinality: optional
		Kind:        message
		HasJSONName: true
		JSONName:    "over"
		HasPresence: true
		Message:     pg_query.WindowDef
	}, {
		Name:        agg_within_group
		Number:      6
		Cardinality: optional
		Kind:        bool
		HasJSONName: true
		JSONName:    "agg_within_group"
	}, {
		Name:        agg_star
		Number:      7
		Cardinality: optional
		Kind:        bool
		HasJSONName: true
		JSONName:    "agg_star"
	}, {
		Name:        agg_distinct
		Number:      8
		Cardinality: optional
		Kind:        bool
		HasJSONName: true
		JSONName:    "agg_distinct"
	}, {
		Name:        func_variadic
		Number:      9
		Cardinality: optional
		Kind:        bool
		HasJSONName: true
		JSONName:    "func_variadic"
	}, {
		Name:        funcformat
		Number:      10
		Cardinality: optional
		Kind:        enum
		HasJSONName: true
		JSONName:    "funcformat"
		Enum:        pg_query.CoercionForm
	}, {
		Name:        location
		Number:      11
		Cardinality: optional
		Kind:        int32
		HasJSONName: true
		JSONName:    "location"
	}]
}
```

yb-voyager/src/queryparser/unsupported_constructs_test.go
sanyamsinghal: done

yb-voyager/src/queryparser/traversal.go
sanyamsinghal: It's the same node. Just the field from OneOfs is being find out to use.

yb-voyager/src/queryparser/traversal.go
sanyamsinghal: discussed

yb-voyager/src/queryparser/traversal.go
sanyamsinghal: done

yb-voyager/src/queryparser/traversal.go
sanyamsinghal: Top level, inside assessMigrationCommand.go, we won't fail entire assessment but for that particular query.

But here i think our assumption is that our traversal algo is full proof. Ideally should not fail to traverse the fields.
In case it does, i think we would need to revisit the algo

yb-voyager/src/queryparser/traversal.go
sanyamsinghal: responded there

yb-voyager/src/queryparser/helpers.go
sanyamsinghal: this helper func is being called from FuncCallDetector.Detect(). where we have check for the nodeType.
If i add here it will be redundant.

yb-voyager/src/queryparser/helpers.go
sanyamsinghal: yeah i thought to keep it as a placeholder thinking of advanced usecase
Maybe in future we want the list of columns used for the table in case `*`

yb-voyager/src/queryparser/helpers.go
sanyamsinghal: responded above, i can add if it seems important?

yb-voyager/src/queryparser/traversal.go
sanyamsinghal: @priyanshi-yb not sure if i get you completely.
Take a look at this - https://github.com/protocolbuffers/protobuf-go/blob/v1.35.1/reflect/protoreflect/proto.go#L248
Let me know.

yb-voyager/src/queryparser/helpers.go
makalaaneesh: Right, but still, as a standalone function that lets you `getFuncNameFromFuncCall`, I think it makes sense for the function to validate. 

Worst case you might hit a panic.
Which just made me realise - I guess it's possible for `funcnameField` to be NULL, right? We should add that check before accessing it. (same for the columnref func)

yb-voyager/src/queryparser/traversal.go
makalaaneesh: Got it. Even in this case, you might want to ignore if traversing some particular node fails, but yeah I get your point. Seems okay

yb-voyager/src/queryparser/helpers.go
makalaaneesh: Here as well, pls check if `fields` can be nil before accessing it?

yb-voyager/src/queryparser/helpers.go
sanyamsinghal: cc @priyanshi-yb

yb-voyager/src/queryparser/helpers.go
sanyamsinghal: It cant be null
```
func (protoreflect.Message) Get(protoreflect.FieldDescriptor) protoreflect.Value
Get retrieves the value for a field.

For unpopulated scalars, it returns the default value, where the default value of a bytes scalar is guaranteed to be a copy. For unpopulated composite types, it returns an empty, read-only view of the value; to obtain a mutable reference, use Mutable.
```

yb-voyager/src/queryparser/helpers.go
sanyamsinghal: responded above

yb-voyager/src/queryparser/traversal.go
priyanshi-yb: Oh okay, got it! 
The list in this functions + `Message` + `Group`  are the only  `Kind` types

yb-voyager/src/queryparser/helpers.go
priyanshi-yb: yeah looks okay to me

makalaaneesh (approved):
LGTM!

sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/3762/

PR #1772, opened 2024-10-22
Sort the import and export table list based on row count

hbhanawat:


yb-voyager/cmd/common.go
```diff
@@ -272,8 +272,14 @@ func displayExportedRowCountSnapshot(snapshotViaDebezium bool) {
 		} else {
 			addHeader(uitable, "DATABASE", "TABLE", "ROW COUNT")
 		}
-		keys := lo.Keys(exportedRowCount)
-		sort.Strings(keys)
+		keys := make([]string, 0, len(exportedRowCount))
+		for key := range exportedRowCount {
+			keys = append(keys, key)
+		}
```
sanyamsinghal, line 278: lets use `lo.Keys()` here https://github.com/samber/lo?tab=readme-ov-file#keys

```
keys := lo.Keys(exportedRowCount)
```

yb-voyager/cmd/common.go
```diff
@@ -394,13 +400,23 @@ func displayImportedRowCountSnapshot(state *ImportDataState, tasks []*ImportFile
 			utils.ErrExit("failed to get imported snapshot rows map: %v", err)
 		}
 	}
+	keys := make([]sqlname.NameTuple, 0, len(snapshotRowCount.Keys()))
```
sanyamsinghal, line 403: isn't just this enough to get the keys
```
keys := snapshotRowCount.Keys()
```

yb-voyager/cmd/common.go
hbhanawat: done.

yb-voyager/cmd/common.go
hbhanawat: snapshotRowCount.Keys returns string[]. I cannot use the string object to look up in the snapshotRowCount in the slice function.

sanyamsinghal (approved):

PR #1773, opened 2024-10-22
Made changes in the error statements for checking dependencies and added a check for psql

ShivanshGahlot:


yb-voyager/src/srcdb/postgres.go
```diff
@@ -394,19 +394,19 @@ func (pg *PostgreSQL) getExportedColumnsListForTable(exportDir, tableName string
 func GetAbsPathOfPGCommandAboveVersion(cmd string, sourceDBVersion string) (string, error) {
 	paths, err := findAllExecutablesInPath(cmd)
 	if err != nil {
-		err = fmt.Errorf("error in finding executables: %w", err)
+		err = fmt.Errorf("error in finding executables in PATH for %v: %w", cmd, err)
 		return "", err
 	}
 	if len(paths) == 0 {
-		err = fmt.Errorf("the command %v is not installed", cmd)
+		err = fmt.Errorf("could not find %v with version greater than or equal to %v in the PATH", cmd, max(PG_COMMAND_VERSION[cmd], sourceDBVersion))
```
priyanshi-yb, line 401: We are merging the error msgs and actual problem in checking dependency, e.g. the first `error while finding the executables in PATH` is errr msg, but `could not find  path in PATH` is actual issue. So we should some how differentiate them for better user experience like error msg should be returned as normal errors in the workflow but the actual issue should go in `Some dependencies are missing....`
We can add a separate return variable for the issue.
What do you think? cc: @makalaaneesh @sanyamsinghal

ShivanshGahlot: Seems fine to me. Let me try to do this.

yb-voyager/cmd/export.go
```diff
@@ -388,32 +387,28 @@ func saveExportTypeInMSR() {
 	}
 }
 
-func checkDependenciesForExport() error {
-	var errs []string
+func checkDependenciesForExport() (problems []string, err error) {
```
priyanshi-yb, line 390: rename the `problems` to `binaryCheckIssues`  and `problem` to `binaryCheckIssue` or similar to `issue`

priyanshi-yb (approved):
LGTM

PR #1777, opened 2024-10-23
Enable adaptive parallelism default to true

makalaaneesh:
- Changed prompt(when adaptive parallelism is not supported) to just a note.
- Change default value of adaptive parallelism to true
- Modify CLI output of parallel jobs based on if adaptive parallelism is used or not. 
- Print if envs are overriding default configurations

### DIFF
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -348,13 +348,11 @@ func registerFlagsForTarget(cmd *cobra.Command) {
 			"number of cores N and use N/4 as parallel jobs. "+
 			"Otherwise, it fall back to using twice the number of nodes in the cluster. "+
 			"Any value less than 1 reverts to the default calculation.")
-	BoolVar(cmd.Flags(), &tconf.EnableYBAdaptiveParallelism, "enable-adaptive-parallelism", false,
-		"Adapt parallelism based on the resource usage (CPU, memory) of the target YugabyteDB cluster")
-	cmd.Flags().MarkHidden("enable-adaptive-parallelism") // not officially released
+	BoolVar(cmd.Flags(), &tconf.EnableYBAdaptiveParallelism, "enable-adaptive-parallelism", true,
+		"Adapt parallelism based on the resource usage (CPU, memory) of the target YugabyteDB cluster.")
 	cmd.Flags().IntVar(&tconf.MaxParallelism, "adaptive-parallelism-max", 0,
 		"number of max parallel jobs to use while importing data when adaptive parallelism is enabled."+
 			"By default, voyager will try if it can determine the total number of cores N and use N/2 as the max parallel jobs. ")
-	cmd.Flags().MarkHidden("adaptive-parallelism-max") // not officially released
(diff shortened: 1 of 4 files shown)


yb-voyager/cmd/import.go
```diff
@@ -399,14 +397,19 @@ func validateFFDBSchemaFlag() {
...
-			utils.ErrExit("Error: --adaptive-parallelism-max flag can only be used with --enable-adaptive-parallelism true")
+		if tconf.MaxParallelism > 0 {
+			if !tconf.EnableYBAdaptiveParallelism {
+				utils.ErrExit("Error: --adaptive-parallelism-max flag can only be used with --enable-adaptive-parallelism true")
+			}
 		}
+	} else {
+		// not relevant for source-replica and source
+		tconf.EnableYBAdaptiveParallelism = false
 	}
```
sanyamsinghal, line 414: should we add a console message saying not relevant.

yb-voyager/cmd/importData.go
```diff
@@ -453,27 +453,39 @@ func importData(importFileTasks []*ImportFileTask) {
...
+	// 	} else {
+	// 		go func() {
+	// 			err := adaptiveparallelism.AdaptParallelism(yb)
+	// 			if err != nil {
+	// 				log.Errorf("adaptive parallelism error: %v", err)
+	// 			}
+	// 		}()
+	// 	}
+
+	// }
```
sanyamsinghal, line 488: remove

yb-voyager/cmd/importData.go
```diff
@@ -652,6 +664,25 @@ func importData(importFileTasks []*ImportFileTask) {
...
+	if !yb.IsAdaptiveParallelismSupported() {
+		utils.PrintAndLog(color.YellowString("Note: Continuing without adaptive parallelism as it is not supported in this version of YugabyteDB."))
+		return false, nil
+	}
+
+	go func() {
+		err := adaptiveparallelism.AdaptParallelism(yb)
+		if err != nil {
+			log.Errorf("adaptive parallelism error: %v", err)
+		}
```
sanyamsinghal, line 681: Just thinking, shall we take some if err is not null for more than x number of times(let's say)
Similar comment about the err condition inside `adaptiveparallelism.AdaptParallelism()`

sanyamsinghal (approved):
LGTM

yb-voyager/cmd/import.go
makalaaneesh: Ah, so the thing is that the flag `enable-adaptive-parallelsim` isn't available for import-data-to-source-replica/source. It's just because the default is now `true`, i'm just setting it to false in those scenarios. Since it's the same flag, there isn't a nice way to choose different defaults based on mode.

yb-voyager/cmd/importData.go
makalaaneesh: I agree with the general idea, but this needs more thought. It'll probably be a bad idea to do this for all types of errors.

Issue #1791, opened 2024-10-24
[Voyager] [Assess Migration] Unique constraint error in case queryid is empty / duplicate in pg_stat_statements


shubham-yb:
Jira Link: [DB-13651](https://yugabyte.atlassian.net/browse/DB-13651)
### Description

In case we have a few rows in the PG SS which are not accessible to the migration user, their query and queryid come up as empty. This causes a Unique constraint error during the insertion of data in the meta db.
We need to handle this case. 
Even for privileged users, the query id can be same for multiple rows and thus this check will fail.

```
       queryid       |                             query                             | calls 
---------------------+---------------------------------------------------------------+-------
 -778708154092909096 | select query from pg_stat_statements where queryid=$1         |     1
 -778708154092909096 | select query from pg_stat_statements where queryid=$1         |     3
 5787495921082967521 | select queryid from pg_stat_statements order by queryid       |     5
 5787495921082967521 | select queryid from pg_stat_statements order by queryid       |     4
 6856587723325601129 | select * from pg_stat_statements                              |     2
 6856587723325601129 | select * from pg_stat_statements                              |     1
 9029586359568942362 | select queryid,query from pg_stat_statements order by queryid |     2
 9029586359568942362 | select queryid,query from pg_stat_statements order by queryid |     2
(8 rows)
```

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

sanyamsinghal:
Also need to explore -
1. What were the missing permissions on user in this case
2. does the voyager assess-migration user(source) as per the docs, does have enough permissions to read the all the required queries from `pgss`

sanyamsinghal:
Fixed in https://github.com/yugabyte/yb-voyager/pull/1792 and https://github.com/yugabyte/yb-voyager/pull/1800

PR #1792, opened 2024-10-24
Removed PK on db_queries_summary table and added a null check

sanyamsinghal:
- source `pg_stat_statements` table also doesn't have a PK

This PR fixes the errors faced in https://github.com/yugabyte/yb-voyager/issues/1791

### DIFF
--- a/yb-voyager/src/migassessment/assessmentDB.go
+++ b/yb-voyager/src/migassessment/assessmentDB.go
@@ -141,8 +141,7 @@ func InitAssessmentDB() error {
 			mean_exec_time REAL,
 			min_exec_time REAL,
 			max_exec_time REAL,
-			rows		BIGINT,
-			PRIMARY KEY(queryid));`, DB_QUERIES_SUMMARY),
+			rows		BIGINT);`, DB_QUERIES_SUMMARY),
 	}
 
 	for _, cmd := range cmds {
--- a/yb-voyager/src/queryparser/query_parser.go
+++ b/yb-voyager/src/queryparser/query_parser.go
@@ -41,6 +41,11 @@ func (qp *QueryParser) Parse() error {
 }
 
 func (qp *QueryParser) GetUnsupportedQueryConstructs() ([]utils.UnsupportedQueryConstruct, error) {
+	if qp.ParseTree == nil {
+		log.Infof("parse tree of the query is null for some reason, skipping unsupported construct check")
+		return nil, nil
+	}
+
 	var result []utils.UnsupportedQueryConstruct = nil
 	visited := make(map[protoreflect.Message]bool)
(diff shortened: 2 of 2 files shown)


sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/3782/

yb-voyager/src/queryparser/query_parser.go
```diff
@@ -41,6 +41,11 @@ func (qp *QueryParser) Parse() error {
 }
 
 func (qp *QueryParser) GetUnsupportedQueryConstructs() ([]utils.UnsupportedQueryConstruct, error) {
+	if qp.ParseTree == nil {
```
makalaaneesh, line 44: Looks like if parse fails, we're moving on, therefore we run into this. 

I'd suggest: return error. GetUnsupportedQueryConstructs() should NOT silently return no error in such a scenario. (Anyway the caller logs the error and moves on for now)

yb-voyager/src/migassessment/assessmentDB.go
```diff
@@ -141,8 +141,7 @@ func InitAssessmentDB() error {
 			mean_exec_time REAL,
 			min_exec_time REAL,
 			max_exec_time REAL,
-			rows		BIGINT,
-			PRIMARY KEY(queryid));`, DB_QUERIES_SUMMARY),
+			rows		BIGINT);`, DB_QUERIES_SUMMARY),
```
makalaaneesh, line 144: Looks like there can be duplicates then. We should somehow detect and skip duplicates, otherwise they will come up in assessment report?

sanyamsinghal: Yes while queries the sqlite assessmentDB we have `DISTINCT` clause in the query
```
query := fmt.Sprintf("SELECT DISTINCT query from %s", migassessment.DB_QUERIES_SUMMARY)
```

yb-voyager/src/queryparser/query_parser.go
sanyamsinghal: done

yb-voyager/src/queryparser/query_parser.go
```diff
@@ -41,6 +41,10 @@ func (qp *QueryParser) Parse() error {
...
+		return nil, fmt.Errorf("query's parse tree is null for some reason, skipping unsupported construct check")
```
makalaaneesh, line 45: Let's just say "query's parse tree is nil" ? 
Since it's an error, we don't have to say "skipping ... "

makalaaneesh (approved):
LGTM with minor comment

Issue #1786, opened 2024-10-24
[Voyager] Export data from target fails with log-level debug


makalaaneesh:
Jira Link: [DB-13648](https://yugabyte.atlassian.net/browse/DB-13648)
### Description

```
[2024-10-24T09:02:28.405Z] 2024-10-24 09:02:21.073858 ERROR exportData.go:392 Export Data using debezium failed: failed to check if snapshot is complete: failed to poll for message in log file: error scanning file /home/ubuntu/workspace/users/yb-voyager-testing/yb-voyager-testing-pipeline/code/yb-voyager/migtests/tests/pg/basic-non-public-live-test/export-dir/logs/debezium-target_db_exporter_ff.log: bufio.Scanner: token too long
```

Possible solution - https://stackoverflow.com/a/21124415

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

makalaaneesh:
Culprit line with a huge token: 

[...]

PR #1793, opened 2024-10-24
Fix token too long in export-data-from-target when log-level is debug

makalaaneesh:
Fix for https://github.com/yugabyte/yb-voyager/issues/1786

Using [Reader](https://pkg.go.dev/bufio#Reader) instead of [Scanner](https://pkg.go.dev/bufio#Scanner) which does not suffer from the "token too long" errors.

### DIFF
--- a/yb-voyager/src/utils/utils.go
+++ b/yb-voyager/src/utils/utils.go
@@ -19,6 +19,7 @@ import (
 	"bufio"
 	"database/sql"
 	"fmt"
+	"io"
 	"math/rand"
 	"net"
 	"net/url"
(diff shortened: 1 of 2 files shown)


makalaaneesh:
https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/3814/pipeline/

ShivanshGahlot (approved):
LGTM

yb-voyager/src/utils/utils.go
```diff
@@ -388,16 +389,23 @@ func WaitForLineInLogFile(filePath string, message string, timeoutDuration time.
...
 
 	for {
-		scanner := bufio.NewScanner(file)
-		for scanner.Scan() {
-			line := scanner.Text()
-			if strings.Contains(line, message) {
-				return nil
+		reader := bufio.NewReader(file)
+		for {
+			line, err := reader.ReadString('\n')
```
sanyamsinghal, line 394: @makalaaneesh can you explain why that error would come with scanner but not reader.

Both of these packages having something like default buffer size.
I am not sure if this will also not break in case of large string.

Also refer `reader.ReadLine()` function once. That seems more full proof to me.
We also have `utils.Readline()` in our code which is not used anywhere anymore

makalaaneesh: @sanyamsinghal from the docs of [Scanner](https://pkg.go.dev/bufio#Scanner): 
`
Scanning stops unrecoverably at EOF, the first I/O error, or a token too large to fit in the [Scanner.Buffer](https://pkg.go.dev/bufio#Scanner.Buffer). When a scan stops, the reader may have advanced arbitrarily far past the last token. Programs that need more control over error handling or large tokens, or must run sequential scans on a reader, should use [bufio.Reader](https://pkg.go.dev/bufio#Reader) instead.
` 
The error we were running into was because of large tokens, hence why I started exploring the `Reader`. 

wrt [Reader.Readline](https://pkg.go.dev/bufio#Reader.ReadLine)  : 
from docs: `ReadLine is a low-level line-reading primitive. Most callers should use Reader.ReadBytes('\n') or Reader.ReadString('\n') instead or use a [Scanner](https://pkg.go.dev/bufio#Scanner).`

Moreover, `If the line was too long for the buffer then isPrefix is set and the beginning of the line is returned. The rest of the line will be returned from future calls. isPrefix will be false when returning the last fragment of the line. ` which basically means we need extra handling to process a line. (which is what `utils.Readline` in our code is doing). 

It seems like like `Reader.ReadString` abstracts all this away, and serves our purpose at least.

PR #1807, opened 2024-10-27
Added provision to run the tests without adaptive parallelism

shubham-yb:
Since Adaptive Parallelism is switched on by default, reversed the logic in the tests to disable it whenever required.


priyanshi-yb (approved):
LGTM

PR #1809, opened 2024-10-28
hide log level flag

makalaaneesh:
Hiding flag because of known issue - https://github.com/yugabyte/yb-voyager/issues/1786

### DIFF
--- a/yb-voyager/cmd/getDataMigrationReportCommand.go
+++ b/yb-voyager/cmd/getDataMigrationReportCommand.go
@@ -418,6 +418,7 @@ func init() {
 	registerExportDirFlag(getDataMigrationReportCmd)
 	getDataMigrationReportCmd.PersistentFlags().StringVarP(&config.LogLevel, "log-level", "l", "info",
 		"log level for yb-voyager. Accepted values: (trace, debug, info, warn, error, fatal, panic)")
+	getDataMigrationReportCmd.PersistentFlags().MarkHidden("log-level")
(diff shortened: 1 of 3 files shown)


priyanshi-yb (approved):
LGTM

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -418,6 +418,7 @@ func init() {
 	registerExportDirFlag(getDataMigrationReportCmd)
 	getDataMigrationReportCmd.PersistentFlags().StringVarP(&config.LogLevel, "log-level", "l", "info",
 		"log level for yb-voyager. Accepted values: (trace, debug, info, warn, error, fatal, panic)")
+	getDataMigrationReportCmd.PersistentFlags().MarkHidden("log-level")
```
sanyamsinghal, line 421: just curious, if we have defined it on root why do we need on bulk and get command here.

sanyamsinghal (approved):

yb-voyager/cmd/getDataMigrationReportCommand.go
makalaaneesh: Something I wanted to also discuss! - these commands do not call the `registerCommonGlobalFlags` function. Not ideal; we should pick it up sometime.

priyanshi-yb: I think it was intentional that some global flags are not required for these commands e.g. get datamigration one doesn't need the yes or send-diagnostics etc..

PR #1819, opened 2024-10-30
Make Object Names cells scrollable in analyze schema html report

sanyamsinghal:
- previously the html page become too long in case large number of objects


priyanshi-yb (approved):
LGTM, thanks!

PR #1821, opened 2024-10-30
Enable GH automation tests workflows to be triggered in case of release branches also

sanyamsinghal:
- didn't enable on dev branches PR since their will be cherry-picks only. Rare chances that it will be required(can be enabled in that PR itself if required)


ShivanshGahlot (approved):
LGTM, I guess just remove your branch from the arrays before merging

PR #1834, opened 2024-11-06
Adaptive parallelism: do not attempt to increase parallelism beyond limits

makalaaneesh:
Previously, the adaptive parallelism layer would blindly try to increase/decrease beyond limits, and the YB conn pool layer would throw errors, which would translate to WARN logs. 

Now, adaptive parallelism layer is aware of limits, and does not try to increase/decrease beyond those limits.


priyanshi-yb (approved):
LGTM

PR #1853, opened 2024-11-07
Fixed regex parser for sql file to remove the extra spaces on right for a line

priyanshi-yb:
Fixes this issue https://yugabyte.atlassian.net/browse/DB-13861, where extra spaces were there before the PROCEDURE DDL and because of that this PROCEDURE was getting considered as normal DDL and hence the PLPGSQL parsing logic wasn't kicking in for that.


yb-voyager/cmd/sqlfile_parser_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func setupFile(objType string, sqlContent string) (*os.File, error) {
+	tmpFile, err := os.CreateTemp("", fmt.Sprintf("%s.sql", strings.ToLower(objType)))
```
sanyamsinghal, line 28: how does os.CreateTemp works, providing `""` as dir will pickup which directory?

sanyamsinghal (approved):

yb-voyager/cmd/sqlfile_parser_test.go
priyanshi-yb: It picks some temp kind of folder automatically e.g. `/var/folders/hz/jwtd4ll10515m40x2bb1kx4h0000gn/T/table.sql75800918`

PR #1858, opened 2024-11-07
Added new steps to install the cpan modules in the airgapped script itself

ShivanshGahlot:
We used to ask the users to install the cpan modules themselves in an airgapped environment. We only checked whether the packages with the required versions were installed or not.
However, manually installing cpan modules can be a big task for the users. They have to first get the tar zip and then perform lengthy `make` steps.
So to automate this process for the users, we are now including the cpan tars in the airgapped  bundle and running build/installation from these tars ourselves inside the airgapped script.

### DIFF
--- a/installer_scripts/install-voyager-airgapped.sh
+++ b/installer_scripts/install-voyager-airgapped.sh
@@ -37,12 +37,13 @@ ubuntu_apt_package_requirements=(
   "oracle-instantclient-sqlplus|exact|21.5.0.0.0"
 )
 
+# Array with format "Module::Name|requirement_type|required_version|tarball_name"
 cpan_modules_requirements=(
-  "DBD::mysql|min|5.005"
-  "Test::NoWarnings|min|1.06"
-  "DBD::Oracle|min|1.83"
-  "String::Random|min|0"
-  "IO::Compress::Base|min|0"
+  "DBD::mysql|min|5.005|DBD-mysql-5.005.tar.gz"
+  "Test::NoWarnings|min|1.06|Test-NoWarnings-1.06.tar.gz"
+  "DBD::Oracle|min|1.83|DBD-Oracle-1.83.tar.gz"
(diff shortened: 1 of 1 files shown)


installer_scripts/install-voyager-airgapped.sh
```diff
@@ -132,57 +141,72 @@ output() {
...
-        echo ""
-        echo -e "\e[31mERROR: the following CPAN modules are missing or do not meet the version requirements:\e[0m"
-        for missing in "${missing_cpan_modules[@]}"; do
-            echo "$missing"
-        done
+    # Version comparison based on requirement type
+    if [[ "$requirement_type" == "min" ]]; then
+        # Check if installed version is at least the required version
+        if [[ $(echo -e "$installed_version\n$required_version" | sort -V | head -n1) != "$required_version" ]]; then
+            echo "Error: Installed version of $module_name ($installed_version) does not meet the minimum required version ($required_version)."
```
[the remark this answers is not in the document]
ShivanshGahlot: Yeah I have added them just for an extra layer of security. So that any pre existing lesser version package doesnt get picked up by chance.

installer_scripts/install-voyager-airgapped.sh
```diff
@@ -132,57 +141,72 @@ output() {
...
+            exit 1
+        fi
+    elif [[ "$requirement_type" == "exact" ]]; then
```
priyanshi-yb, line 198: we don't have any `exact` requirement type right?

priyanshi-yb (approved):
LGTM

installer_scripts/install-voyager-airgapped.sh
ShivanshGahlot: Yeah not rn. But added in case we add any dependency with exact requirement in the future and also for consistency.

PR #1873, opened 2024-11-11
Create new packages issue, queryissue

makalaaneesh:


yb-voyager/src/issue/dml.go
```diff
@@ -0,0 +1,56 @@
...
+
+package issue
+
+var advisoryLocksIssue = Issue{
+	Type:            ADVISORY_LOCKS,
+	TypeName:        "Advisory Locks",
+	TypeDescription: "",
+	Suggestion:      "",
+	GH:              "",
+	DocsLink:        "https://docs.yugabyte.com/preview/yugabyte-voyager/known-issues/postgresql/#advisory-locks",
```
priyanshi-yb, line 25: I think we should keep the links in constants still. Also, links need to be changed.

yb-voyager/src/issue/constants.go
```diff
@@ -0,0 +1,31 @@
...
+// Types
+const (
+	ADVISORY_LOCKS = "advisory_locks"
+	SYSTEM_COLUMNS = "system_columns"
+	XML_FUNCTIONS  = "xml_functions"
+)
+
+// Object types
+const (
+	TABLE_OBJECT_TYPE     = "table"
```
priyanshi-yb, line 28: value should be all caps as we use it currently like that - `TABLE`, `FUNCTION`.. so similar for `DML QUERY`

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1021,20 +1022,23 @@ func fetchUnsupportedQueryConstructs() ([]utils.UnsupportedQueryConstruct, error
...
+				query, err)
 		}
-		if unsupportedConstructs != nil {
-			result = append(result, unsupportedConstructs...)
+
+		for _, issue := range issues {
+			uqc := utils.UnsupportedQueryConstruct{
+				Query:             issue.SqlStatement,
+				ConstructType:     issue.Type,
+				ConstructTypeName: issue.TypeName,
```
priyanshi-yb, line 1036: why do we need to have these two `ConstructType`, and `ConstructTypeName` separate fields?

makalaaneesh: Thought process was:
`type` fields should be like enums (constants, no spaces, underscores). Currently `Advisory Locks` was a value being passed to `type` field, which seems wrong. 

I guess we need two things:
- `type` - a constant /enum determining which type of issue it is
- `typeName` - display name/title of the issue. 

So, I put both of these in the main `Issue` struct. And here in `UnsupportedQueryConstruct`, I figured it'd be good to put both (we use the `typeName` when displaying in html` ), but it's probably not required for json so i omitted it from that. I kept the `type` field also because it will be useful in json and also when we send to yugabyted. (the `typeName` is also probably useful for yugabyted)

Does that make sense?

yb-voyager/src/issue/constants.go
makalaaneesh: Should we do this for even the types ? (ADVISORY_LOCKS, SYSTEM_COLUMNS) ?

yb-voyager/src/issue/dml.go
makalaaneesh: WIll change links, but why constants? 
I only put the `type` in constants, in case we want to refer to the issue from some other package, but DocsLink should only be accessed via the issue instance, so why constant?

yb-voyager/src/issue/issue.go
```diff
@@ -0,0 +1,44 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package issue
+
+type Issue struct {
+	Type            string // (advisory_locks, index_not_supported, etc)
```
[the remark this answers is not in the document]
makalaaneesh: So, what I realised when adding the issues was that this higher-level type/category like `UNSUPPORTED_FEATURES` / `UNSUPPORTED_QUERY_CONSTRUCTS` is best decided by the consumer, not the issue package. 

For example, consider "Advisory Locks". It can be both under `UNSUPPORTED_QUERY_CONSTRUCTS` or `UNSUPPORTED_STORED_PROCEDURES` (or whatever new type you will assign in your upcoming PR).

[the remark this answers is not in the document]
priyanshi-yb: Hmm okay, makes sense

yb-voyager/src/issue/dml.go
priyanshi-yb: Hmm, sounds right it's not a problem as this place is for just keeping issue structs and these links will not interfere directly with any other code so these are in some way constant data only.

yb-voyager/src/issue/constants.go
priyanshi-yb: yeah should be fine

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: Yes, Agreed we should have both `Type` and `TypeName` in `Issue` but for the `Unsupported Query Construct` I am not sure if we should put both as it's not required as we can keep either of them, so as`Advisory Locks` is making more sense for `ConstructTypeName` field, so we can rename the `ConstructType` to that and keep only that which can be used by both json and html.

yb-voyager/src/issue/dml.go
makalaaneesh: Yeah exactly! I'm looking at these issue structs essentially as constants!

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: To be consistent I think I agree it makes sense for us right now to just have the `ConstructTypeName` field in `UnsupportedQueryConstruct` . 

But this is something we should think about long term. IMO, in a json report, we should have `Type` for sure, and `TypeName` is optional.

yb-voyager/src/queryissue/queryissue.go
```diff
@@ -0,0 +1,87 @@
...
+type ParserIssueDetector struct {
+	// TODO: Add fields here
+	// e.g. store composite types, etc. for future processing.
+}
+
+func NewParserIssueDetector() *ParserIssueDetector {
+	return &ParserIssueDetector{}
+}
+
+func (p *ParserIssueDetector) GetIssues(query string) ([]issue.IssueInstance, error) {
```
priyanshi-yb, line 39: Maybe this is for later when we try to add other issue-type reporting here like DDL, PLPGSQL..: but just bringing up here in the context: do we need a parameter in this function to return the Issues of specific type only, otherwise the upstream caller needs to filter the issues.

priyanshi-yb (approved):
LGTM with a comment.

yb-voyager/src/queryissue/queryissue.go
makalaaneesh: Interesting. What are cases where you might want to filter? 
The way I was thinking about it was: You just pass a query and get back all the Issues in that query. Why would we want to filter by type?

yb-voyager/src/issue/dml.go
```diff
@@ -0,0 +1,56 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package issue
+
+var advisoryLocksIssue = Issue{
+	Type:            ADVISORY_LOCKS,
+	TypeName:        "Advisory Locks",
```
sanyamsinghal, line 21: why do we need Type and TypeName both?

yb-voyager/src/issue/dml.go
```diff
@@ -0,0 +1,56 @@
...
+	Suggestion:      "",
+	GH:              "",
+	DocsLink:        "https://docs.yugabyte.com/preview/yugabyte-voyager/known-issues/postgresql/#advisory-locks-is-not-yet-implemented",
+}
+
+func NewAdvisoryLocksIssue(objectType string, objectName string, sqlStatement string) IssueInstance {
+	return newIssueInstance(advisoryLocksIssue, objectType, objectName, sqlStatement, map[string]interface{}{})
+}
+
+var systemColumnsIssue = Issue{
```
sanyamsinghal, line 32: why do we need a global variable for this? It can go inside `NewSystemColumnsIssue` constructor.

yb-voyager/src/issue/issue.go
```diff
@@ -0,0 +1,44 @@
...
+	GH              string
+	DocsLink        string
+}
+
+type IssueInstance struct {
+	Issue
+	ObjectType   string // TABLE, FUNCTION, DML_QUERY?
+	ObjectName   string // table name/function name/etc
+	SqlStatement string
+	Details      map[string]interface{} // additional details about the issue
```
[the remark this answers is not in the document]
sanyamsinghal: hmm, regarding point 2, using constructor to put the metadata.
Metadata can be difference for each issue right? Different function name or other detail.

Although it should be achievable, but can you highlight - how, with the new refactored code.

yb-voyager/src/issue/issue.go
sanyamsinghal, line 1: Do we really need two abstraction here - Issue and IssueInstance. I would advise to keep it in one, adding complexity to the code.

On a side note, in the final json report generated, this will introduce one more level due to embedding of Issue in IssueInstance

sanyamsinghal (commented):
1. The packages - issue, queryissue, queryparser
These are all related, can we put them inside a same top level directory, instead of putting directly under `src`


2. I didn't get the advantage of dividing it into two separate packages - queryissue and queryparser. Can you please explain the reasoning and usecase
To me it seems a bit of over engineering to me and it will better to keep the code in same package.

yb-voyager/src/issue/dml.go
makalaaneesh: https://github.com/yugabyte/yb-voyager/pull/1873#discussion_r1837759390

yb-voyager/src/issue/issue.go
makalaaneesh: The thought process is: 
`Issue` - a particular problem/limitation. This is static. It just stores information about the type of issue, suggestions, workarounds, GH links, fixed in which versions, etc. 

`IssueInstance` - an instance of an issue that was found in a certain context. It has the Issue details + context details (ObjectName, SQLStatement, any other metadata/details). 

`Issue` is more like a constant, which is why I kept it separate and also declared the various issue structs  as global variables. We are adding a lot of issues day-by-day, so it would be good to define them as though they were constants, as it will make it easier to manage.

makalaaneesh: >On a side note, in the final json report generated, this will introduce one more level due to embedding of Issue in IssueInstance

I don't think these Issue structs will/should be directly converted to json, they will first be converted into `AnalyzeIssue` / `AssessmentIssue` structs. 
That being said, embedding does not create nesting in json! Check this out - https://go.dev/play/p/urTAWQsFUh_m

yb-voyager/src/issue/issue.go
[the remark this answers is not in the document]
makalaaneesh: So, for example, right now, the constructor for advisory locks is as follows: 
`
func NewAdvisoryLocksIssue(objectType string, objectName string, sqlStatement string) IssueInstance {
`

If we want to also include the functions that were used in advisory locks in the `Details` field, you would modify the constructor as follows: 
`func NewAdvisoryLocksIssue(objectType string, objectName string, sqlStatement string, functions []string) IssueInstance {`

- `functions` is properly typed(which is our intention), ensuring that only []string is passed. 
- only constructor is exported, not the `advisoryLocksIssue` struct, so you are forced to use the constructor to create the issueInstance.

makalaaneesh:
>The packages - issue, queryissue, queryparser
These are all related, can we put them inside a same top level directory, instead of putting directly under src
I didn't get the advantage of dividing it into two separate packages - queryissue and queryparser. Can you please explain the reasoning and usecase
To me it seems a bit of over engineering to me and it will better to keep the code in same package.

I like the idea of a top level directory. That makes sense. 
I definitely agree that I'm creating packages left and right, probably the reverse effect of having all code in `cmd`  😄. But we definitely thought about it ! Here was the motivation:  

Initially we wanted to keep just one package with code of both `queryissue` and `queryparser`. Then @priyanshi-yb pointed out that we may need to use parser logic in situations that are not related to `queryissue` . For instance, we already do this today - in `export-data`, we use parser to add the `WITH COLOCATION false` clause to `CREATE TABLE`. Therefore, we thought we can put all the parser related logic (the parsing, the parsing of plpgsql to json, the traversal through the proto messages, the traversal through the nested plpgsql json, adding clauses to statements, etc) into a separate package - `queryparser`.

yb-voyager/src/issue/dml.go
sanyamsinghal: yeah went through that thread, but still didn't what are the usecases where we need both.
Is this the only difference - "replacing _ in Type with whitespaces for TypeName" @makalaaneesh

makalaaneesh:
One potential change that makes sense to me is to move all the issue definitions right now from `issue` pkg into `queryissues` package 🤔  (because that's where we would be detecting and returning those issues. 
Even `IssueInstance` is something that is specific for `queryissues` , so that can also be moved into `queryissues` pkg. What do you think @sanyamsinghal @priyanshi-yb

yb-voyager/src/issue/dml.go
makalaaneesh: Just separation of concerns. Type is the enum. TypeName is for display purposes.
In `cmd` layer, when we need to filter issues based on their types, it feels wrong to me to do them by their names (i.e. strings with spaces). Even in json, for the type, we should be sending the type (enum, without white spaces), and not the type name.

sanyamsinghal: type name is just for html report?

makalaaneesh:
>but can't understand what queryissue is for, on top of issue.

So, I wanted to keep the issue definition separate to capture all types of issues that we may want to report in voyager. Today, we only report query issues (i.e. issues in some query construct/syntax). We have plans to also report data issues (for example, rows in table x are too large, YB does not support this). There can be other types of issues as well. 
So, the `Issue` struct itself felt like it deserved an independent package.  

As per my previous comment though, I think it might be good to move  `IssueInstance` and all the issue definitions(advisoryLocks, systemIssues, etc) into `queryissues` .

yb-voyager/src/issue/dml.go
makalaaneesh: Yep. Good point, I should probably remove `TypeName` from here then, no? This can be the headache of the assessmentReport layer 🤔

sanyamsinghal: yeah, extra handling might be a headache. I think we can live with `_` in html report, or do some hardcoding where required.

sanyamsinghal:
> all the issue definitions(advisoryLocks, systemIssues, etc) into queryissues .
agree

> As per my previous comment though, I think it might be good to move IssueInstance
IssueInstance is not just for queryissue, it can be used for dataissue(in future) also.

makalaaneesh:
>IssueInstance is not just for queryissue, it can be used for dataissue(in future) also.

if you look at IssueInstance right now, it mostly seems like a QueryIssueInstance. (sqlstatement, objectName, etc). So I'm thinking let's just move that into queryissue for now, and then we will deal with DataIssues in future when it comes.

```
type IssueInstance struct {
	Issue
	ObjectType   string // TABLE, FUNCTION, DML_QUERY?
	ObjectName   string // table name/function name/etc
	SqlStatement string
	Details      map[string]interface{} // additional details about the issue
}
```

yb-voyager/src/issue/dml.go
priyanshi-yb: I am not sure if it can always be the case that `TypeName` will be the string replacing underscores with whitespace from `Type.` Maybe in some cases, we want to keep the Type very minimal and TypeName a bit descriptive, e.g., `unsupported_indexes`, `Indexes on complex datatypes`?

sanyamsinghal:
> if you look at IssueInstance right now, it mostly seems like a QueryIssueInstance. (sqlstatement, objectName, etc). So I'm thinking let's just move that into queryissue for now, and then we will deal with DataIssues in future when it comes.

@makalaaneesh then we can rename it to QueryIssue? rather than generic name as IssueInstance.

PR #1874, opened 2024-11-11
Handle export data status panics

priyanshi-yb:
Handling nil pointer and the index out-of-range issues while running export data status before, during, and after export data.

https://yugabyte.atlassian.net/browse/DB-13971

### DIFF
--- a/migtests/scripts/run-test.sh
+++ b/migtests/scripts/run-test.sh
@@ -168,6 +168,11 @@ main() {
 	expected_file="${TEST_DIR}/export_data_status-report.json"
 	actual_file="${EXPORT_DIR}/reports/export-data-status-report.json"
 
+	if [ "${EXPORT_TABLE_LIST}" != "" ]
+	then
+		expected_file="${TEST_DIR}/export-data-status-with-table-list-report.json"
+	fi
+
 	step "Verify export-data-status report"
 	verify_report ${expected_file} ${actual_file}
 
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -900,11 +900,6 @@ func exportDataOffline(ctx context.Context, cancel context.CancelFunc, finalTabl
 	if source.DBType == POSTGRESQL {
 		//Make leaf partitions data files entry under the name of root table
 		renameDatafileDescriptor(exportDir)
-		//Similarly for the export snapshot status file
-		err = renameExportSnapshotStatus(exportSnapshotStatusFile)
-		if err != nil {
-			return fmt.Errorf("rename export snapshot status: %w", err)
-		}
 	}
 	displayExportedRowCountSnapshot(false)
 
--- a/yb-voyager/cmd/exportDataStatusCommand.go
+++ b/yb-voyager/cmd/exportDataStatusCommand.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io/fs"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 
@@ -183,6 +184,7 @@ func runExportDataStatusCmd() ([]*exportTableMigStatusOutputRow, error) {
 		partitions := leafPartitions[finalFullTableName.ForOutput()]
 		//Changing the display of the partition tables in case table-list is set because there can be case where user has passed a subset of leaft tables in the list
 		if source.DBType == POSTGRESQL && partitions != nil && msr.IsExportTableListSet {
+			slices.Sort(partitions)
 			partitions := strings.Join(partitions, ", ")
 			displayTableName = fmt.Sprintf("%s (%s)", displayTableName, partitions)
 		}
@@ -207,6 +209,8 @@ func runExportDataStatusCmd() ([]*exportTableMigStatusOutputRow, error) {
 			if exportingLeaf > 0 {
 				finalStatus = "EXPORTING"
 			} else if not_started == len(snapshotStatus) {
+				//In case of partition tables in PG, we are clubbing the status of all leafs and then returning the status
+				//For root table we are sending NOT_STARTED and if only all leaf partitions will have NOT_STARTED else EXPORTING/DONE
 				finalStatus = "NOT-STARTED"
 			} else {
 				finalStatus = "DONE"
--- a/yb-voyager/cmd/exportDataStatusCommand.go
+++ b/yb-voyager/cmd/exportDataStatusCommand.go
@@ -124,7 +124,7 @@ func runExportDataStatusCmdDbzm(streamChanges bool, leafPartitions map[string][]
 		utils.ErrExit("Failed to read export status file %s: %v", exportStatusFilePath, err)
 	}
 	if status == nil {
-		utils.ErrExit("Export data has not started yet. Try running after export has started.")
+		return nil, fmt.Errorf("export data has not started yet. Try running after export has started")
 	}
 	InProgressTableSno = status.InProgressTableSno()
 	var rows []*exportTableMigStatusOutputRow
@@ -179,7 +179,7 @@ func runExportDataStatusCmd(msr *metadb.MigrationStatusRecord, leafPartitions ma
 	exportStatusSnapshot, err := exportSnapshotStatusFile.Read()
 	if err != nil {
 		if errors.Is(err, fs.ErrNotExist) {
-			utils.ErrExit("Export data has not started yet. Try running after export has started.")
+			return nil, fmt.Errorf("export data has not started yet. Try running after export has started")
 		}
 		utils.ErrExit("Failed to read export status file %s: %v", exportSnapshotStatusFilePath, err)
 	}
--- a/yb-voyager/cmd/exportDataStatusCommand.go
+++ b/yb-voyager/cmd/exportDataStatusCommand.go
@@ -181,6 +181,7 @@ func runExportDataStatusCmd() ([]*exportTableMigStatusOutputRow, error) {
 		}
 		displayTableName := finalFullTableName.ForMinOutput()
 		partitions := leafPartitions[finalFullTableName.ForOutput()]
+		//Changing the display of the partition tables in case table-list is set because there can be case where user has passed a subset of leaft tables in the list
 		if source.DBType == POSTGRESQL && partitions != nil && msr.IsExportTableListSet {
 			partitions := strings.Join(partitions, ", ")
 			displayTableName = fmt.Sprintf("%s (%s)", displayTableName, partitions)
--- a/yb-voyager/cmd/exportDataStatusCommand.go
+++ b/yb-voyager/cmd/exportDataStatusCommand.go
@@ -150,6 +150,9 @@ func runExportDataStatusCmd() ([]*exportTableMigStatusOutputRow, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error while getting migration status record: %v", err)
 	}
+	if msr == nil {
+		return nil, nil
+	}
 	tableList := msr.TableListExportedFromSource
 	source = *msr.SourceDBConf
 	sqlname.SourceDBType = source.DBType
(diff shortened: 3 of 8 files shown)


yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -182,7 +185,10 @@ func runExportDataStatusCmd() ([]*exportTableMigStatusOutputRow, error) {
 			partitions := strings.Join(partitions, ", ")
 			displayTableName = fmt.Sprintf("%s (%s)", displayTableName, partitions)
 		}
-		snapshotStatus, _ := exportedSnapshotStatus.Get(finalFullTableName)
+		snapshotStatus, ok := exportedSnapshotStatus.Get(finalFullTableName)
+		if !ok {
+			continue
```
makalaaneesh, line 190: Should we display partial data? (i.e. status available only for a subset of rows?). It seems bad IMO.

I would be inclined to do the following: 
explicitly check in the beginning if we are able to get status (exportSnapshotStatus/export_status.json file is available). If not, then just return without printing anything.

priyanshi-yb: That check is already present above. When we read the files, we return from there if the file doesn't exist. This check I added is only applicable to the window in which the file is created but not completely populated with content, e.g., FilePath.

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -199,13 +205,16 @@ func runExportDataStatusCmd() ([]*exportTableMigStatusOutputRow, error) {
 			}
 			if exportingLeaf > 0 {
 				finalStatus = "EXPORTING"
-			} else if doneLeaf == len(snapshotStatus) {
-				finalStatus = "DONE"
 			} else if not_started == len(snapshotStatus) {
-				finalStatus = "NOT_STARTED"
+				finalStatus = "NOT-STARTED"
+			} else {
```
makalaaneesh, line 210: I'm assuming this is just a re-order, not relevant?

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -178,11 +181,14 @@ func runExportDataStatusCmd() ([]*exportTableMigStatusOutputRow, error) {
 		}
 		displayTableName := finalFullTableName.ForMinOutput()
 		partitions := leafPartitions[finalFullTableName.ForOutput()]
-		if source.DBType == POSTGRESQL && partitions != nil {
+		if source.DBType == POSTGRESQL && partitions != nil && msr.IsExportTableListSet {
```
makalaaneesh, line 184: Can you add a comment here explaining why we want to change the display name only if table-list is specified?

yb-voyager/cmd/exportData.go
```diff
@@ -900,11 +900,6 @@ func exportDataOffline(ctx context.Context, cancel context.CancelFunc, finalTabl
 	if source.DBType == POSTGRESQL {
 		//Make leaf partitions data files entry under the name of root table
 		renameDatafileDescriptor(exportDir)
-		//Similarly for the export snapshot status file
-		err = renameExportSnapshotStatus(exportSnapshotStatusFile)
```
makalaaneesh, line 904: It seems like previously this rename was done much later (after snapshot was complete), and now we seem to be doing it before we actually start export-data (i.e. in initializeExportTableMetadata). 

Is the behavior not impacted because of this change? If this is the correct behaviour, how was it working before?

yb-voyager/cmd/exportDataStatusCommand.go
priyanshi-yb: yeah mostly, just that this reorder helps now for the partition tables as we are also sending the root table's status which is `NOT-STARTED`.

yb-voyager/cmd/exportDataStatusCommand.go
priyanshi-yb: Yeah, I will add it. Reason: This is just for display in case the user has provided a table list and in case he has subset of leaf tables so let them know that we are only exporting those not all.

yb-voyager/cmd/exportData.go
priyanshi-yb: Yeah, so I also realized this thing yesterday. while I was testing so this is the correct behavior to rename this File in the starting itself because in status we are depending upon this to club the status of all leaf to root. So, yeah before this change we were not properly showing the status of partitions during the export data is happening (we were showing <root_table>-<NOT_STARTED>).

yb-voyager/cmd/exportDataStatusCommand.go
makalaaneesh: Ah okay. Pls add a comment here to explain this special case.

yb-voyager/cmd/exportData.go
makalaaneesh: Got it, thanks! 
Pls verify this manually for partitioned tables (in pg specifically) since we don't have tests for this. 
cc: @shubham-yb

makalaaneesh: @priyanshi-yb 
I imagine that this logic also kicks in for get-data-migration-report, right? 
Does it make sense to then add a test-case for the display name of partitioned tables in the report? (i.e. assert that it shows `<roottable>(<p1>, <p2>)` and not just `<roottable>` ?

priyanshi-yb: > I imagine that this logic also kicks in for get-data-migration-report, right?

Yeah I will verify that as well.


> Does it make sense to then add a test-case for the display name of partitioned tables in the report? (i.e. assert that it shows <roottable>(<p1>, <p2>) and not just <roottable> ?


I think this is in reference to table list above. Yeah, I think we can add. Let me see.

priyanshi-yb: Verified `get data-migration-report` and added export-data-status report case for table-list and partition handling.

migtests/scripts/run-test.sh
```diff
@@ -168,6 +168,11 @@ main() {
 	expected_file="${TEST_DIR}/export_data_status-report.json"
```
makalaaneesh, line 168: oh nice, didn't realise we already validate this!

migtests/scripts/run-test.sh
```diff
@@ -168,6 +168,11 @@ main() {
...
 	actual_file="${EXPORT_DIR}/reports/export-data-status-report.json"
 
+	if [ "${EXPORT_TABLE_LIST}" != "" ]
```
makalaaneesh, line 171: Where are we changing the invocation of this test where we actually pass a table-list?

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -119,19 +131,27 @@ func runExportDataStatusCmdDbzm(streamChanges bool) ([]*exportTableMigStatusOutp
...
 }
 
-func getSnapshotExportStatusRow(tableStatus *dbzm.TableExportStatus) *exportTableMigStatusOutputRow {
+func getSnapshotExportStatusRow(tableStatus *dbzm.TableExportStatus, leafPartitions map[string][]string, msr *metadb.MigrationStatusRecord) *exportTableMigStatusOutputRow {
 	nt, err := namereg.NameReg.LookupTableName(fmt.Sprintf("%s.%s", tableStatus.SchemaName, tableStatus.TableName))
 	if err != nil {
 		utils.ErrExit("lookup %s in name registry: %v", tableStatus.TableName, err)
 	}
+	displayTableName := nt.ForMinOutput()
+	partitions := leafPartitions[nt.ForOutput()]
```
makalaaneesh, line 146: In `getLeafPartitionsFromRootTable` , you are setting the key as `root = rootTable.Qualified.MinQuoted` and logic for `ForOutput` is `t.CurrentName.Qualified.MinQuoted`, that's why this works. 

IMO, this is a bit fragile. Either try to have leafPartitions return a structmap[nametuple] OR let's document this as comments clearly here and in `getLeafPartitionsFromRootTable` as a TODO.

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -169,20 +183,23 @@ func runExportDataStatusCmd() ([]*exportTableMigStatusOutputRow, error) {
...
 		finalFullTableName, err := namereg.NameReg.LookupTableName(tableName)
 		if err != nil {
 			return nil, fmt.Errorf("lookup %s in name registry: %v", tableName, err)
 		}
 		displayTableName := finalFullTableName.ForMinOutput()
 		partitions := leafPartitions[finalFullTableName.ForOutput()]
-		if source.DBType == POSTGRESQL && partitions != nil {
+		//Changing the display of the partition tables in case table-list is set because there can be case where user has passed a subset of leaft tables in the list
+		if source.DBType == POSTGRESQL && partitions != nil && msr.IsExportTableListSet {
+			slices.Sort(partitions)
```
makalaaneesh, line 195: This logic is re-rewritten above (for dbzm case). Move it to a function?

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -150,6 +155,9 @@ func getDataMigrationReportCmdFn(msr *metadb.MigrationStatusRecord) {
 	if source.DBType == POSTGRESQL {
 		exportSnapshotStatus, err = exportSnapshotStatusFile.Read()
 		if err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				return
```
makalaaneesh, line 159: Shall we exit with a msg here along the lines of "export has not started yet. Try running after export has started." (same for export-data-status as well)

migtests/scripts/run-test.sh
priyanshi-yb: Yes! We validate the status reports but do not exhaustively do that for all tests.

migtests/scripts/run-test.sh
priyanshi-yb: In the `pg-migtests.yml`

makalaaneesh: Ah okay, this is already the case that we use table-list for this test-case. got it!

yb-voyager/cmd/exportDataStatusCommand.go
priyanshi-yb: > Either try to have leafPartitions return a structmap[nametuple] 

I remember this struct map wasn't helpful in this case as I have to get the NameTuples and create manually in leaf partition cases so to avoid I was using the sqlName in that function. I can add a comment properly for this.

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -124,7 +124,7 @@ func runExportDataStatusCmdDbzm(streamChanges bool, leafPartitions map[string][]
 		utils.ErrExit("Failed to read export status file %s: %v", exportStatusFilePath, err)
 	}
 	if status == nil {
-		return nil, nil
+		utils.ErrExit("Export data has not started yet. Try running after export has started.")
```
makalaaneesh, line 127: return error instead of ErrExit?

yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -173,7 +179,7 @@ func runExportDataStatusCmd(msr *metadb.MigrationStatusRecord, leafPartitions ma
 	exportStatusSnapshot, err := exportSnapshotStatusFile.Read()
 	if err != nil {
 		if errors.Is(err, fs.ErrNotExist) {
-			return nil, nil
+			utils.ErrExit("Export data has not started yet. Try running after export has started.")
```
makalaaneesh, line 182: return error instead of ErrExit?

makalaaneesh (approved):
LGTM! minor comment on errors

PR #1897, opened 2024-11-14
Implement reporting of Unsupported PLpg/SQL objects in assessment and analyze

priyanshi-yb:
Description:
1. Implement the framework to parse the PLPGSQL blocks of objects like FUNCTION/PROCEDURE to extract the individual queries and then the issues of those queries and reporting them in analyze/assess report as `Unsupported Plpg/SQL objects`. https://yugabyte.atlassian.net/browse/DB-13982
2. Added an ENV var to control the reporting of this feature with `REPORT_UNSUPPORTED_PLPGSQL_OBJECTS`.
3. Enabled this feature for analyze-schema in this PR.
4. Added tests for it in analyze-schema tests.

Sample assessment report - 
Screenshot 2024-11-15 at 4 58 45 PM

Sample analyze report (issues)- 
Screenshot 2024-11-15 at 2 47 16 PM

TODO: add VIEW / MVIEW in detection logic to parse its select stmts


yb-voyager/src/queryparser/traversal_plpgsql.go
```diff
@@ -0,0 +1,171 @@
...
+	          {
+	            "PLpgSQL_stmt_return": {}
+	          }
+	        ]
+	      }
+	    }
+	  }
+	}
+*/
+func TraversePlPgSQLActions(action interface{}, plPgSqlStatements *[]string) {
```
makalaaneesh, line 123: how about calling the argument `fieldValue` instead of actions, because it's only actions for the first time, after that it receursively traverses through all the fields, correct?

yb-voyager/src/queryparser/traversal_plpgsql.go
```diff
@@ -0,0 +1,171 @@
...
+	actionMap, ok := action.(map[string]interface{})
```
makalaaneesh, line 124: Can we write this as a switch case? 
case 1 : it's a map : then iterate through all its values and either process or recursive call
case 2: it's a list: iterate and traverse.

yb-voyager/src/queryparser/traversal_plpgsql.go
```diff
@@ -0,0 +1,171 @@
...
+				}
+			}
+		default:
+			TraversePlPgSQLActions(v, plPgSqlStatements)
+		}
+	}
+}
+
+// Function to format the PLPGSQL EXPR query from the json string
+func formatExprQuery(q string) string {
```
makalaaneesh, line 156: If I remember, parser has a format option also, right? The reason we don't do that here is because of the limitations, parsing itself can fail, correct?

yb-voyager/src/queryissue/queryissue.go
```diff
@@ -41,9 +48,64 @@ func (p *ParserIssueDetector) GetIssues(query string) ([]issue.IssueInstance, er
...
+		plpgsqlQueries, err := p.GetAllPLPGSQLStatements(query)
+		if err != nil {
+			return nil, fmt.Errorf("error getting all the queries from query: %w", err)
+		}
+		var issues []issue.IssueInstance
+		for _, plpgsqlQuery := range plpgsqlQueries {
+			issuesInQuery, err := p.GetIssues(plpgsqlQuery)
+			if err != nil {
+				//there can be plpgsql expr queries no parseable via parser e.g. "withdrawal > balance" 
+				log.Infof("error getting issues in query: %v", err)
```
makalaaneesh, line 62: 1. Errorf
2. log the query as well.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -857,6 +857,10 @@ func getAssessmentReportContentFromAnalyzeSchema() error {
 	}
 	assessmentReport.UnsupportedFeatures = append(assessmentReport.UnsupportedFeatures, unsupportedFeatures...)
 	assessmentReport.UnsupportedFeaturesDesc = FEATURE_ISSUE_TYPE_DESCRIPTION
+	if utils.GetEnvAsBool("REPORT_UNSUPPORTED_PLPGSQL_OBJECTS", false) {
```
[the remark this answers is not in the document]
priyanshi-yb: yeah will be doing it in the second PR as I have to modify the tests as well.

yb-voyager/src/queryissue/queryissue.go
```diff
@@ -41,9 +48,64 @@ func (p *ParserIssueDetector) GetIssues(query string) ([]issue.IssueInstance, er
...
+			issues = append(issues, issuesInQuery...)
+		}
+		return issues, nil
+	}
+	//TODO: add handling for VIEW/MVIEW to parse select 
 	return p.getDMLIssues(query, parseTree)
 }
 
+
+func(p *ParserIssueDetector) GetAllPLPGSQLStatements(query string) ([]string, error) {
```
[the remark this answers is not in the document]
priyanshi-yb: Yeah right it makes more sense there, I wasn't sure much so kept it like this. Will move

yb-voyager/src/queryparser/traversal_plpgsql.go
priyanshi-yb: I am not sure about the format option from parser, I don't think there is any but will check. Anyways yes we can't use that here as you mentioned because of limitations.

yb-voyager/cmd/templates/migration_assessment_report.template
```diff
@@ -282,6 +292,61 @@
...
+                    <th>Details</th>
+                </tr>
+               {{ range  .UnsupportedPlPgSqlObjects }}
+                <!-- Feature Row with rowspan for grouping -->
+                <tr>
+                    {{ $objectsGroupByObjectType := groupByObjectType .Objects}}
+                    {{ $numUniqueObjectNamesOfAllTypes := totalUniqueObjectNamesOfAllTypes $objectsGroupByObjectType}}
+                    {{ $docsLink := .DocsLink }}
+                    <td rowspan={{$numUniqueObjectNamesOfAllTypes}}><strong>{{ .FeatureName }}</strong></td>
+                    {{ $isNextRowRequiredForObjectType := false }}
```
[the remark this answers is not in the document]
priyanshi-yb: This is for merging things where we have a couple of columns as rowspan based on number of types and object names etc.. I couldn't find a better way to do this as we need the dynamic number of types and objects from the struct I have to add this logic here.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -1579,9 +1581,40 @@ func checker(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+			issue := convertIssueInstanceToAnalyzeIssue(issueInstance, sqlInfoStmt, objType)
+			schemaAnalysisReport.Issues = append(schemaAnalysisReport.Issues, issue)
+		}
+	}
+
+}
+
+func convertIssueInstanceToAnalyzeIssue(issueInstance issue.IssueInstance, sqlInfoStmt sqlInfo, objType string) utils.Issue {
+	return utils.Issue{
+		ObjectType:   objType,
```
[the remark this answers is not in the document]
priyanshi-yb: > is that correct?
yes correct.

Initially I thought it might make more sense for the analyze-schema to modify the issue as per its requirement, but now when I think that analyze is sending the function query to `GetIssues`, it will expect that IssueInstance should have the proper type of Object and name I believe so for PLPGSQL case I think it makes sense to modify the type/name in queryissue layer.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -1579,9 +1581,40 @@ func checker(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
 	checkRemaining(sqlInfoArr, fpath)
+	if utils.GetEnvAsBool("REPORT_UNSUPPORTED_PLPGSQL_OBJECTS", true) {
+		checkPlPgSQLStmtsUsingParser(sqlInfoArr, fpath, objType)
+	}
 	checkStmtsUsingParser(sqlInfoArr, fpath, objType)
 }
 
+func checkPlPgSQLStmtsUsingParser(sqlInfoArr []sqlInfo, fpath string, objType string) {
+	for _, sqlInfoStmt := range sqlInfoArr {
+		parserIssueDetector := queryissue.NewParserIssueDetector()
```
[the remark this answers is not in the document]
priyanshi-yb: yeah.

yb-voyager/src/queryparser/traversal_plpgsql.go
priyanshi-yb: yes

yb-voyager/src/queryissue/queryissue.go
```diff
@@ -48,64 +42,32 @@ func (p *ParserIssueDetector) GetIssues(query string) ([]issue.IssueInstance, er
...
 			issuesInQuery, err := p.GetIssues(plpgsqlQuery)
 			if err != nil {
-				//there can be plpgsql expr queries no parseable via parser e.g. "withdrawal > balance" 
-				log.Infof("error getting issues in query: %v", err)
+				//there can be plpgsql expr queries no parseable via parser e.g. "withdrawal > balance"
+				log.Errorf("error getting issues in query-%s: %v", query, err)
 				continue
 			}
-			issues = append(issues, issuesInQuery...)
+			for _, i := range issuesInQuery {
```
makalaaneesh, line 59: let's add a comment here?

yb-voyager/src/queryparser/query_parser.go
```diff
@@ -42,3 +45,45 @@ func ParsePLPGSQLToJson(query string) (string, error) {
...
 }
+
+func IsPLPGSQLObject(parseTree *pg_query.ParseResult) bool {
+	// CREATE FUNCTION is same parser NODE for FUNCTION/PROCEDURE
+	_, isPlPgSQLObject := parseTree.Stmts[0].Stmt.Node.(*pg_query.Node_CreateFunctionStmt)
+	return isPlPgSQLObject
+}
+
+func GetObjectTypeAndObjectName(parseTree *pg_query.ParseResult) (string, string) {
+	createFuncNode, isCreateFunc := parseTree.Stmts[0].Stmt.Node.(*pg_query.Node_CreateFunctionStmt)
```
makalaaneesh, line 56: If this is a generic `GetObjectTypeAndObjectName` function for any input of `pg_query.ParseResult`, you probably should write this differently. 

Maybe a switch (type of parseTree.Stmts[0].Stmt.Node) and handle it for function now?

Alternatively, you could just define `GetObjectTypeAndObjectNameForFunction` and take in  input `pg_query.Node_CreateFunctionStmt`

yb-voyager/src/queryparser/traversal_plpgsql.go
```diff
@@ -120,47 +180,46 @@ parsed json -
...
-	}
-
-	for k, v := range actionMap {
-		switch k {
-		// base case of recursive calls to reach this PLPGSQL_EXPR field in json which will have "query" field with statement
-		case PLPGSQL_EXPR: 
-			expr, ok := v.(map[string]interface{})
-			if ok {
-				query, ok := expr[QUERY]
+func TraversePlPgSQLActions(fieldValue interface{}, plPgSqlStatements *[]string) {
```
makalaaneesh, line 183: `TraversePlPgSQLJson` ?

yb-voyager/src/queryparser/traversal_plpgsql.go
```diff
@@ -120,47 +180,46 @@ parsed json -
...
+	actionMap, isMap := fieldValue.(map[string]interface{})
```
makalaaneesh, line 184: Instead of `actionMap`, `lists` :
How about `fieldMap` , `fieldList` ?

yb-voyager/src/queryparser/query_parser.go
priyanshi-yb: I kept it generic for any input based on type of stmt, I will be adding a few more to this in the second PR #1911

yb-voyager/src/queryparser/traversal_plpgsql.go
priyanshi-yb: `TraversePlPgSQLJsonField`?

makalaaneesh: both are okay. just because `fieldValue` is already an argument, imo `TraversePlPgSQLJson` suffices

makalaaneesh (approved):

makalaaneesh:
LGTM!

yb-voyager/src/queryparser/traversal_plpgsql.go
```diff
@@ -0,0 +1,230 @@
...
+	            "PLpgSQL_stmt_return": {}
+	          }
+	        ]
+	      }
+	    }
+	  }
+	}
+*/
+func TraversePlPgSQLJson(fieldValue interface{}, plPgSqlStatements *[]string) {
+	fieldMap, isMap := fieldValue.(map[string]interface{})
```
sanyamsinghal, line 184: for now the traverse function seems straightforward. If gets to handle a lot of cases and would need generalisation, consider exploring the implementation using `reflect` pacakge.

yb-voyager/src/queryparser/traversal_plpgsql.go
```diff
@@ -0,0 +1,230 @@
...
+			default:
+				TraversePlPgSQLJson(v, plPgSqlStatements)
+			}
+		}
+	case isList:
+		//In case the value of a field is not a <key , val> but a list of <key, val> e.g. "body"
+		for _, l := range fieldList {
+			TraversePlPgSQLJson(l, plPgSqlStatements)
+		}
+	}
```
sanyamsinghal, line 210: consider using https://github.com/tidwall/gjson module as it simplifies the traversal of json strings

For example traversal plpgsql code can be converted to this::

```
func ExtractQueriesUsingGjson(parsedJson string) []string {
    var queries []string

    // Navigate to the action block
    actions := gjson.Get(parsedJson, "PLpgSQL_function.action")

    // Iterate over each statement in the body
    actions.Get("PLpgSQL_stmt_block.body").ForEach(func(key, value gjson.Result) bool {
        stmtType := key.String()

        switch stmtType {
        case "PLpgSQL_stmt_execsql":
            // Extract query from 'sqlstmt.PLpgSQL_expr.query'
            query := value.Get("sqlstmt.PLpgSQL_expr.query").String()
            formattedQuery := formatExprQuery(query)
            queries = append(queries, formattedQuery)

        case "PLpgSQL_stmt_perform":
            // Extract query from 'expr.PLpgSQL_expr.query'
            query := value.Get("expr.PLpgSQL_expr.query").String()
            formattedQuery := formatExprQuery(query)
            queries = append(queries, formattedQuery)

        // Add more cases for other statement types as needed
        // For example: PLpgSQL_stmt_if, PLpgSQL_stmt_loop, etc.

        default:
            // Handle or ignore other statement types
        }

        return true // Continue iteration
    })

    return queries
}
```

sanyamsinghal (commented):
Thanks for adding multiple cases in end to end tests. 
But please add a placeholder for unit tests also in case of plpgsql, you can start with simple cases for now.

PR #1901, opened 2024-11-15
YBVersion to be able to store and compare yb versions

makalaaneesh:
YBVersion is a wrapper around hashicorp/go-version.Version that adds some Yugabyte-specific
functionality.
 1. It only supports versions with 2, 3, or 4 segments (A.B, A.B.C, A.B.C.D).
 2. It only accepts one of supported Yugabyte version series.
 3. It provides a method to compare versions based on common prefix of segments. This is useful in cases
    where the version is not fully specified (e.g. 2.20 vs 2.20.7.0).
    If the user specifies 2.20, a reasonable assumption is that they would be on the latest 2.20.x.y version.
    Therefore, we want only want to compare the prefix 2.20 between versions and ignore the rest.

### DIFF
--- a/yb-voyager/src/version/yb_version.go
+++ b/yb-voyager/src/version/yb_version.go
@@ -82,7 +82,7 @@ func (ybv *YBVersion) originalSegmentsLen() int {
 	return len(segments)
 }
 
-func (ybv *YBVersion) ComparePrefix(other *YBVersion) (int, error) {
+func (ybv *YBVersion) CompareCommonPrefix(other *YBVersion) (int, error) {
 	if ybv.Series() != other.Series() {
 		return 0, fmt.Errorf("Cannot compare versions with different series: %s and %s", ybv.Series(), other.Series())
 	}
@@ -101,16 +101,16 @@ func (ybv *YBVersion) ComparePrefix(other *YBVersion) (int, error) {
 	return ybvMin.Compare(otherMin), nil
 }
 
-func (ybv *YBVersion) PrefixGreaterThanOrEqual(other *YBVersion) (bool, error) {
-	res, err := ybv.ComparePrefix(other)
+func (ybv *YBVersion) CommonPrefixGreaterThanOrEqual(other *YBVersion) (bool, error) {
+	res, err := ybv.CompareCommonPrefix(other)
 	if err != nil {
 		return false, err
 	}
 	return res >= 0, nil
 }
 
-func (ybv *YBVersion) PrefixLessThan(other *YBVersion) (bool, error) {
-	res, err := ybv.ComparePrefix(other)
+func (ybv *YBVersion) CommonPrefixLessThan(other *YBVersion) (bool, error) {
+	res, err := ybv.CompareCommonPrefix(other)
 	if err != nil {
 		return false, err
 	}
--- a/yb-voyager/src/version/yb_version.go
+++ b/yb-voyager/src/version/yb_version.go
@@ -38,7 +38,7 @@ const (
 )
 
 type YBVersion struct {
-	V *version.Version
+	*version.Version
 }
 
 func NewYBVersion(v string) (*YBVersion, error) {
@@ -60,7 +60,7 @@ func NewYBVersion(v string) (*YBVersion, error) {
 }
 
 func (ybv *YBVersion) Series() string {
-	return joinIntsWith(ybv.V.Segments()[:2], ".")
+	return joinIntsWith(ybv.Segments()[:2], ".")
 }
 
 func (ybv *YBVersion) ReleaseType() string {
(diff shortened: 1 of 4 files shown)


yb-voyager/src/version/yb_version.go
```diff
@@ -0,0 +1,150 @@
...
+	if ybv.Series() != other.Series() {
+		return 0, fmt.Errorf("cannot compare versions with different series: %s and %s", ybv.Series(), other.Series())
+	}
+	myOriginalSegLen := ybv.originalSegmentsLen()
+	otherOriginalSegLen := other.originalSegmentsLen()
+	minSegLen := min(myOriginalSegLen, otherOriginalSegLen)
+
+	ybvMin, err := version.NewVersion(joinIntsWith(ybv.Segments()[:minSegLen], "."))
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
makalaaneesh: oops, fixed!

yb-voyager/src/version/yb_version.go
```diff
@@ -0,0 +1,150 @@
...
+
+	"github.com/hashicorp/go-version"
+	"golang.org/x/exp/slices"
+)
+
+var supportedYBVersionStableSeriesOld = []string{"2.20"}
+var supportedYBVersionStableSeries = []string{"2024.1"}
+var supportedYBVersionPreviewSeries = []string{"2.21"}
+
+var allSupportedYBVersionSeries = append(append(supportedYBVersionStableSeries, supportedYBVersionPreviewSeries...), supportedYBVersionStableSeriesOld...)
```
sanyamsinghal, line 32: nit: 
```
allSupportedYBVersionSeries = lo.Flatten([][]string{supportedYBVersionStableSeries,supportedYBVersionPreviewSeries,supportedYBVersionStableSeriesOld})
```

https://github.com/samber/lo?tab=readme-ov-file#flatten

yb-voyager/src/version/yb_version.go
```diff
@@ -0,0 +1,150 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/go-version"
+	"golang.org/x/exp/slices"
+)
+
+var supportedYBVersionStableSeriesOld = []string{"2.20"}
+var supportedYBVersionStableSeries = []string{"2024.1"}
+var supportedYBVersionPreviewSeries = []string{"2.21"}
```
sanyamsinghal, line 30: 1. what about the version apart from these?
2. Also if tomorrow 2025 comes, then voyager code will start breaking.. Can we have some generic but smart check, instead of comparing with fixed constants.

yb-voyager/src/version/yb_version.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, fmt.Errorf("invalid YB version: %s. Version should have between min 2 and max 4 segments (A.B.C.D). Version %s has ybv.originalSegmentsLen()", v, v)
+	}
+
+	if !slices.Contains(allSupportedYBVersionSeries, ybv.Series()) {
+		return nil, fmt.Errorf("unsupported YB version series: %s. Supported YB version series = %v", ybv.Series(), allSupportedYBVersionSeries)
+	}
+	return ybv, nil
+}
+
+func (ybv *YBVersion) Series() string {
```
sanyamsinghal, line 72: some comment/example will help to understand here.

yb-voyager/src/version/yb_version.go
```diff
@@ -0,0 +1,150 @@
...
+		return STABLE_OLD
+	} else {
+		panic("unknown release type for series: " + series)
+	}
+}
+
+func (ybv *YBVersion) originalSegmentsLen() int {
+	orig := ybv.Original()
+	segments := strings.Split(orig, ".")
+	return len(segments)
```
sanyamsinghal, line 92: why not use version.Segments() here instaed of splitting?

yb-voyager/src/version/yb_version.go
makalaaneesh: As discussed, added 2.18, 2.14, 2.23 to this list.

yb-voyager/src/version/yb_version.go
makalaaneesh: nice, done!

yb-voyager/src/version/yb_version.go
makalaaneesh: Added comment here as well. Essentially: 

This returns the len of the segments in the original input. For instance if input is 2024.1,
- go-version.Version.Segments() will return [2024, 1, 0, 0] (assume 2024.1.0.0)
- original = 2024.1
- originalSegmentsLen  = 2 ([2024,1])

sanyamsinghal: got it

sanyamsinghal (approved):

PR #1919, opened 2024-11-19
Added check for java as a dependency in guardrails

ShivanshGahlot:
Sample output:
image

### DIFF
--- a/debezium-server-voyager/debezium-server-voyager-dist/src/main/resources/distro/run.sh
+++ b/debezium-server-voyager/debezium-server-voyager-dist/src/main/resources/distro/run.sh
@@ -36,24 +36,6 @@ fi
 # Add * to the end of the connector path
 YB_OR_PG_CONNECTOR_PATH="$YB_OR_PG_CONNECTOR_PATH/*"
 
-
-# Resolve Java
-if [ -z "$JAVA_HOME" ]; then
-  JAVA_BINARY="java"
-else
-  JAVA_BINARY="$JAVA_HOME/bin/java"
-fi
-MIN_REQUIRED_MAJOR_VERSION='17'
-JAVA_MAJOR_VER=$(${JAVA_BINARY} -version 2>&1 | awk -F '"' '/version/ {print $2}' | awk -F. '{print $1}')
-if ([ -n "$JAVA_MAJOR_VER" ] && (( 10#${JAVA_MAJOR_VER} >= 10#${MIN_REQUIRED_MAJOR_VERSION} )) ) #integer compare of versions.
-then
-    echo "Found sufficient java version = ${JAVA_MAJOR_VER}"
-else
-    echo "ERROR: Java not found or insufficient version ${JAVA_MAJOR_VER}. Please install java>=${MIN_REQUIRED_MAJOR_VERSION}"
-    exit 1;
-fi
-
-
 # Run
 if [ "$OSTYPE" = "msys" ] || [ "$OSTYPE" = "cygwin" ]; then
   PATH_SEP=";"
--- a/debezium-server-voyager/debezium-server-voyager-dist/src/main/resources/distro/run.sh
+++ b/debezium-server-voyager/debezium-server-voyager-dist/src/main/resources/distro/run.sh
@@ -36,6 +36,24 @@ fi
 # Add * to the end of the connector path
 YB_OR_PG_CONNECTOR_PATH="$YB_OR_PG_CONNECTOR_PATH/*"
 
+
+# Resolve Java
+if [ -z "$JAVA_HOME" ]; then
+  JAVA_BINARY="java"
+else
+  JAVA_BINARY="$JAVA_HOME/bin/java"
+fi
+MIN_REQUIRED_MAJOR_VERSION='17'
+JAVA_MAJOR_VER=$(${JAVA_BINARY} -version 2>&1 | awk -F '"' '/version/ {print $2}' | awk -F. '{print $1}')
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/export.go
```diff
@@ -395,20 +399,27 @@ func checkDependenciesForExport() (binaryCheckIssues []string, err error) {
...
 				binaryCheckIssues = append(binaryCheckIssues, binaryCheckIssue)
 			}
 		}
-		if len(binaryCheckIssues) > 0 {
-			binaryCheckIssues = append(binaryCheckIssues, "Install or Add the required dependencies to PATH and try again\n")
-		}
 	}
 
 	if changeStreamingIsEnabled(exportType) || useDebezium {
+		// Check for java
```
makalaaneesh, line 408: It's not ideal to be adding user messages ("install or add ... ") to the list of `binaryCheckIssues`. 
Nevertheless, for now: (we can pick that up later)
the else-ifs make it difficult to read IMO. 
```
javaIssue, err := checkJavaVersion()
if err != nil {
	return nil, err
} 
if javaIssue {
    		binaryCheckIssues = append(binaryCheckIssues, javaIssue)
 }
 if len(binaryCheckIssues) > 0 {
			binaryCheckIssues = append(binaryCheckIssues, "Install or Add the required dependencies to PATH and try again\n")

```
This should work, right?

yb-voyager/cmd/export.go
```diff
@@ -417,3 +428,44 @@ func checkDependenciesForExport() (binaryCheckIssues []string, err error) {
...
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Sprintf("java: required version >= %d", MIN_REQUIRED_JAVA_VERSION), nil
+	}
+
+	// Example output
+	// openjdk version "17" 2021-09-14
+	// OpenJDK Runtime Environment (build 17+35-2724)
+	// OpenJDK 64-Bit Server VM (build 17+35-2724, mixed mode, sharing)
+	versionOutput := string(output)
```
makalaaneesh, line 448: pls test this with a few java versions on all OS (mac, ubuntu, centos) just to make sure our assumptions here(version is the 3rd part, etc) are valid on all OS

yb-voyager/cmd/export.go
ShivanshGahlot: Yeah agreed. But in order to get the desired output format, this seemed like the easiest way. Also I missed the else condition on the external if:
```
else {
		if len(binaryCheckIssues) > 0 {
			binaryCheckIssues = append(binaryCheckIssues, "Install or Add the required dependencies to PATH and try again\n")
		}
	}
```

yb-voyager/cmd/export.go
ShivanshGahlot: I have now modified this to basically mimic the steps we had in run.sh to extract the version.
We search for a line that contains the string 'version' and get that and over there we have the actual java version enclosed in ""

debezium-server-voyager/debezium-server-voyager-dist/src/main/resources/distro/run.sh
```diff
@@ -36,24 +36,6 @@ fi
 # Add * to the end of the connector path
 YB_OR_PG_CONNECTOR_PATH="$YB_OR_PG_CONNECTOR_PATH/*"
 
-
-# Resolve Java
```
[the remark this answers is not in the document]
ShivanshGahlot: Done

yb-voyager/cmd/export.go
makalaaneesh: Ah, this is getting a bit confusing to read. 

How about structure it this way : 
```
do pg checks, append to binaryCheckIssues
do java check, append to binaryCheckIssues (`if changeStreamingIsEnabled(exportType) || useDebezium {` block)
if binarchCheckIssues.len > 0: add that "Install or Add" message. 

Finally, do check for dbzm (again inside a `if changeStreamingIsEnabled(exportType) || useDebezium {` block) 
there you can directly append the "Please check your Voyager installation and try again" 
```

yb-voyager/cmd/export.go
makalaaneesh: Got it, would still recommend doing a sanity test on ubuntu, rhel atleast. Because failing voyager because we weren't able to properly figure out the version would be bad. (especially since we don't even have a prompt for these dependency checks).

yb-voyager/cmd/export.go
ShivanshGahlot: Done

yb-voyager/cmd/export.go
ShivanshGahlot: Yeah I checked in on ubuntu and rhel8

ShivanshGahlot: And macos too

yb-voyager/cmd/export.go
```diff
@@ -395,25 +399,99 @@ func checkDependenciesForExport() (binaryCheckIssues []string, err error) {
...
 		// Check for debezium
 		// FindDebeziumDistribution returns an error only if the debezium distribution is not found
 		// So its error mesage will be added to problems
-		err := dbzm.FindDebeziumDistribution(source.DBType, false)
+		err = dbzm.FindDebeziumDistribution(source.DBType, false)
 		if err != nil {
 			binaryCheckIssues = append(binaryCheckIssues, strings.ToUpper(err.Error()[:1])+err.Error()[1:])
 			binaryCheckIssues = append(binaryCheckIssues, "Please check your Voyager installation and try again")
 		}
+	} else {
```
makalaaneesh, line 431: This else should not be redundant.

makalaaneesh (approved):
LGTM. Pls address the comment before merging

makalaaneesh (approved):
LGTM

Issue #1920, opened 2024-11-19
[Voyager] export-data errors out if background metadata queries (count*) are still running when pg_dump completes.


makalaaneesh:
### Description

Sample logs: 
```
2024-11-16 18:16:47.240719 ERROR logging.go:30 Failed to query "select count(*) from public.\"table_1\"" for row count of "[CurrentName=(table_1) SourceName=(table_1) TargetName=()]": sql: database is closed

2024-11-16 18:16:47.240989 ERROR logging.go:30 Failed to query "select count(*) from public.\"table_2\"" for row count of "[CurrentName=(table_2) SourceName=(table_2) TargetName=()]": sql: database is closed
```

This happens when : 
1. goroutine A is executing count(*) on table x
2. goroutines B,C,D,... are waiting to get access to the connection in order to execute count(*) on tables y,z,a,... 
3. pg_dump completes, therefore as part of clean up in the main goroutine, we call `source.DB().Disconnect()` . which closes the connections that goroutines A,B,C,D are using/waiting to use. 
4. goroutines A,B,C,D encounter error and exit the application. 

Mitigation: 
1. Even though there are multiple tables being exported at the same time, there is only one connection in the pool for goroutines A,B,C,D. This is the reason there is a backlog of goroutines waiting to execute count(*). The pool size should be equal to parallel-jobs
2. The goroutine running count(*) on table X should be shut down after export of table X is complete. This does not happen at this time. 
3. When goroutines A,B,C,D encounter an error, they should not bring down the process, they should silently log the error and continue, as this is not a critical process. 

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

PR #1932, opened 2024-11-20
Partial fix for [export-data errors out if background metadata queries (count*) are still running when pg_dump completes.]

makalaaneesh:
Implements fixes 1,3 in https://github.com/yugabyte/yb-voyager/issues/1920 : 
When goroutines A,B,C,D encounter an error, they should not bring down the process, they should silently log the error and continue, as this is not a critical process.

### DIFF
--- a/yb-voyager/src/srcdb/yugabytedb.go
+++ b/yb-voyager/src/srcdb/yugabytedb.go
@@ -74,16 +74,16 @@ func (yb *YugabyteDB) CheckRequiredToolsAreInstalled() {
 	checkTools("strings")
 }
 
-func (yb *YugabyteDB) GetTableRowCount(tableName sqlname.NameTuple) int64 {
+func (yb *YugabyteDB) GetTableRowCount(tableName sqlname.NameTuple) (int64, error) {
 	var rowCount int64
 	query := fmt.Sprintf("select count(*) from %s", tableName.ForUserQuery())
 	log.Infof("Querying row count of table %q", tableName)
 	err := yb.db.QueryRow(query).Scan(&rowCount)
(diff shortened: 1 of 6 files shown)


makalaaneesh:
> @makalaaneesh can we do this change also
> 
> ```
> SetMaxIdleConns(paralleljobs)
> ```
> 
> To me this seems harmless and should fix the issue for large databases to a go extend, what do you think.

Yeaah, I was thinking about it. For the most part, no concerns. Just about total connections to source db being parallel-jobs x 2 (pg_dump/ora2pg using parallel-jobs and voyager layer using parallel-jobs). But since it's capped, I don't think it should be a big problem. Let me go ahead and do it. cc: @hbhanawat

priyanshi-yb (approved):
LGTM

yb-voyager/src/srcdb/yugabytedb.go
sanyamsinghal, line 1: same for SetMaxOpenConns in yugabytedb.go? Just for consistency i guess.

sanyamsinghal (approved):

yb-voyager/src/srcdb/yugabytedb.go
makalaaneesh: Ah, yeah, I intentionally avoided it because in export-data-from-target, we don't support parallel-jobs, so was unsure. but looks like we take the default 4, so made the change for yugabytedb as well.

makalaaneesh:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/3997/

PR #1937, opened 2024-11-20
Introduced RangeTableFunc Detector for detecting XMLTABLE() case

sanyamsinghal:
Added end to end test for XMLTABLE() case

https://yugabyte.atlassian.net/browse/DB-13675

### DIFF
--- a/yb-voyager/src/queryissue/unsupported_dml_constructs.go
+++ b/yb-voyager/src/queryissue/unsupported_dml_constructs.go
@@ -155,6 +155,14 @@ SQL Query:
 			author TEXT PATH 'author'
 	) AS x;
 
+Parsetree: stmt:{select_stmt:{target_list:{res_target:{val:{column_ref:{fields:{string:{sval:"x"}}  fields:{a_star:{}}  location:7}}  location:7}}
+from_clause:{range_table_func:
+	{docexpr:{column_ref:{fields:{string:{sval:"xml_column"}}  location:57}}
+	rowexpr:{a_const:{sval:{sval:"/bookstore/book"}  location:29}}
+	columns:{range_table_func_col:{colname:"title"  type_name:{names:{string:{sval:"text"}}  typemod:-1  location:87}  colexpr:{a_const:{sval:{sval:"title"}  location:97}}  location:81}}
+	columns:{range_table_func_col:{colname:"author"  type_name:{names:{string:{sval:"text"}}  typemod:-1  location:116}  colexpr:{a_const:{sval:{sval:"author"}  location:126}}  location:109}}
+alias:{aliasname:"x"}  location:17}}  limit_option:LIMIT_OPTION_DEFAULT  op:SETOP_NONE}}  stmt_len:142
+
 Here, 'docexpr' points to 'xml_column' containing XML data, 'rowexpr' selects each 'book' node, and 'columns' extract 'title' and 'author' from each book.
 Hence Presence of XPath in 'rowexpr' and structured 'columns' typically indicates XMLTABLE usage.
 */
--- a/yb-voyager/src/queryissue/unsupported_dml_constructs.go
+++ b/yb-voyager/src/queryissue/unsupported_dml_constructs.go
@@ -185,10 +185,12 @@ func isXMLTable(rangeTableFunc protoreflect.Message) bool {
 	if columnsField == nil {
 		return false
 	}
+
 	columnsList := rangeTableFunc.Get(columnsField).List()
 	if columnsList.Len() == 0 {
 		return false
 	}
 
+	// this means all the required fields of RangeTableFunc node for being a XMLTABLE() are present
 	return true
 }
--- a/yb-voyager/src/queryissue/unsupported_dml_constructs.go
+++ b/yb-voyager/src/queryissue/unsupported_dml_constructs.go
@@ -111,3 +111,84 @@ func (d *XmlExprDetector) Detect(msg protoreflect.Message) ([]string, error) {
 	}
 	return nil, nil
 }
+
+type RangeTableFunDetector struct{}
+
+func NewRangeTableFuncDetector() *RangeTableFunDetector {
+	return &RangeTableFunDetector{}
+}
+
+// Detect checks if a RangeTableFunc node is present for a XMLTABLE() function
+func (d *RangeTableFunDetector) Detect(msg protoreflect.Message) ([]string, error) {
+	if queryparser.GetMsgFullName(msg) == queryparser.PG_QUERY_RANGE_TABLE_FUNC_NODE {
+		if isXMLTable(msg) {
+			return []string{XML_FUNCTIONS}, nil
+		}
+	}
+	return nil, nil
+}
+
+/*
+XMLTABLE transforms XML data into relational table format, making it easier to query XML structures.
+Detection in RangeTableFunc Node:
+- docexpr: Refers to the XML data source, usually a column storing XML.
+- rowexpr: XPath expression (starting with '/' or '//') defining the rows in the XML.
+- columns: Specifies the data extraction from XML into relational columns.
+
+Example: Converting XML data about books into a table:
+SQL Query:
+
+	SELECT x.*
+	FROM XMLTABLE(
+		'/bookstore/book'
+		PASSING xml_column
+		COLUMNS
+			title TEXT PATH 'title',
+			author TEXT PATH 'author'
+	) AS x;
+
+Here, 'docexpr' points to 'xml_column' containing XML data, 'rowexpr' selects each 'book' node, and 'columns' extract 'title' and 'author' from each book.
+Hence Presence of XPath in 'rowexpr' and structured 'columns' typically indicates XMLTABLE usage.
+
+Function to detect if a RangeTableFunc node represents XMLTABLE()
+*/
+func isXMLTable(rangeTableFunc protoreflect.Message) bool {
+	log.Infof("checking if range table func node is for XMLTABLE()")
+	// Check for 'docexpr' field
+	docexprField := rangeTableFunc.Descriptor().Fields().ByName("docexpr")
+	if docexprField == nil {
+		return false
+	}
+	docexprNode := rangeTableFunc.Get(docexprField).Message()
+	if docexprNode == nil {
+		return false
+	}
+
+	// Check for 'rowexpr' field
+	rowexprField := rangeTableFunc.Descriptor().Fields().ByName("rowexpr")
+	if rowexprField == nil {
+		return false
+	}
+	rowexprNode := rangeTableFunc.Get(rowexprField).Message()
+	if rowexprNode == nil {
+		return false
+	}
+
+	xpath := queryparser.GetStringValueFromNode(rowexprNode)
+	log.Debugf("xpath value found: %s\n", xpath)
+	if xpath == "" || !queryparser.IsXPathExprForXmlTable(xpath) {
+		return false
+	}
+
+	// Check for 'columns' field
+	columnsField := rangeTableFunc.Descriptor().Fields().ByName("columns")
+	if columnsField == nil {
+		return false
+	}
+	columnsList := rangeTableFunc.Get(columnsField).List()
+	if columnsList.Len() == 0 {
+		return false
+	}
+
+	return true
+}
--- a/yb-voyager/src/queryissue/unsupported_dml_constructs.go
+++ b/yb-voyager/src/queryissue/unsupported_dml_constructs.go
@@ -134,94 +134,9 @@ func NewRangeTableFuncDetector() *RangeTableFuncDetector {
 // Detect checks if a RangeTableFunc node is present for a XMLTABLE() function
 func (d *RangeTableFuncDetector) Detect(msg protoreflect.Message) ([]string, error) {
 	if queryparser.GetMsgFullName(msg) == queryparser.PG_QUERY_RANGETABLEFUNC_NODE {
-		if isXMLTable(msg) {
+		if queryparser.IsXMLTable(msg) {
 			return []string{XML_FUNCTIONS}, nil
 		}
 	}
 	return nil, nil
-}
-
-/*
-Note: XMLTABLE() is not a simple function(stored in FuncCall node), its a table function
-which generates set of rows using the info(about rows, columns, content) provided to it
-Hence its requires a more complex node structure(RangeTableFunc node) to represent.
-
-XMLTABLE transforms XML data into relational table format, making it easier to query XML structures.
-Detection in RangeTableFunc Node:
-- docexpr: Refers to the XML data source, usually a column storing XML.
-- rowexpr: XPath expression (starting with '/' or '//') defining the rows in the XML.
-- columns: Specifies the data extraction from XML into relational columns.
-
-Example: Converting XML data about books into a table:
-SQL Query:
-
-	SELECT x.*
-	FROM XMLTABLE(
-		'/bookstore/book'
-		PASSING xml_column
-		COLUMNS
-			title TEXT PATH 'title',
-			author TEXT PATH 'author'
-	) AS x;
-
-Parsetree: stmt:{select_stmt:{target_list:{res_target:{val:{column_ref:{fields:{string:{sval:"x"}}  fields:{a_star:{}}  location:7}}  location:7}}
-from_clause:{range_table_func:
-	{docexpr:{column_ref:{fields:{string:{sval:"xml_column"}}  location:57}}
-	rowexpr:{a_const:{sval:{sval:"/bookstore/book"}  location:29}}
-	columns:{range_table_func_col:{colname:"title"  type_name:{names:{string:{sval:"text"}}  typemod:-1  location:87}  colexpr:{a_const:{sval:{sval:"title"}  location:97}}  location:81}}
-	columns:{range_table_func_col:{colname:"author"  type_name:{names:{string:{sval:"text"}}  typemod:-1  location:116}  colexpr:{a_const:{sval:{sval:"author"}  location:126}}  location:109}}
-alias:{aliasname:"x"}  location:17}}  limit_option:LIMIT_OPTION_DEFAULT  op:SETOP_NONE}}  stmt_len:142
-
-Here, 'docexpr' points to 'xml_column' containing XML data, 'rowexpr' selects each 'book' node, and 'columns' extract 'title' and 'author' from each book.
-Hence Presence of XPath in 'rowexpr' and structured 'columns' typically indicates XMLTABLE usage.
-*/
-// Function to detect if a RangeTableFunc node represents XMLTABLE()
-func isXMLTable(rangeTableFunc protoreflect.Message) bool {
-	log.Infof("checking if range table func node is for XMLTABLE()")
-	// Check for 'docexpr' field
-	docexprField := rangeTableFunc.Descriptor().Fields().ByName("docexpr")
-	if docexprField == nil {
-		return false
-	}
-	docexprNode := rangeTableFunc.Get(docexprField).Message()
-	if docexprNode == nil {
-		return false
-	}
-
-	// Check for 'rowexpr' field
-	rowexprField := rangeTableFunc.Descriptor().Fields().ByName("rowexpr")
-	if rowexprField == nil {
-		return false
-	}
-	rowexprNode := rangeTableFunc.Get(rowexprField).Message()
-	if rowexprNode == nil {
-		return false
-	}
-
-	xpath := queryparser.GetStringValueFromNode(rowexprNode)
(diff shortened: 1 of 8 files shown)


sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/3989/

yb-voyager/src/queryissue/unsupported_dml_constructs.go
```diff
@@ -111,3 +111,93 @@ func (d *XmlExprDetector) Detect(msg protoreflect.Message) ([]string, error) {
 	}
 	return nil, nil
 }
+
+type RangeTableFunDetector struct{}
```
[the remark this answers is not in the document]
makalaaneesh: Can you add a comment to describe what a rangeTableFunc node is?

yb-voyager/src/queryissue/unsupported_dml_constructs.go
```diff
@@ -111,3 +111,93 @@ func (d *XmlExprDetector) Detect(msg protoreflect.Message) ([]string, error) {
...
+
+func NewRangeTableFuncDetector() *RangeTableFunDetector {
+	return &RangeTableFunDetector{}
+}
+
+// Detect checks if a RangeTableFunc node is present for a XMLTABLE() function
+func (d *RangeTableFunDetector) Detect(msg protoreflect.Message) ([]string, error) {
+	if queryparser.GetMsgFullName(msg) == queryparser.PG_QUERY_RANGETABLEFUNC_NODE {
+		if isXMLTable(msg) {
+			return []string{XML_FUNCTIONS}, nil
```
makalaaneesh, line 125: Discussion for later: 
I;m wondering if it might be a good idea for `Detect` to not return anything, just store the detection information in it's struct; We declare another func for all detectors `GetIssues() []IssueInstance` that will generate the issues directly and give it. `Detector.GetIssues()` will be called at the end after the full proto message has been traverserd.   

Reason being: What we do now is combine is just get the constructTypes from each Detect() call, store it in one list, and then create `IssueInstances` out of it. This flow can't work when we want to add more details/metadata (like the function names, etc) to each specific issue.

Thoughts?

yb-voyager/src/queryissue/unsupported_dml_constructs.go
```diff
@@ -111,3 +111,93 @@ func (d *XmlExprDetector) Detect(msg protoreflect.Message) ([]string, error) {
...
+	xpath := queryparser.GetStringValueFromNode(rowexprNode)
+	log.Debugf("xpath expression in the node: %s\n", xpath)
+	// Keep both cases check(param placeholder and absolute check)
+	if xpath == "" {
+		// Attempt to check if 'rowexpr' is a parameter placeholder like '$1'
+		isPlaceholder := queryparser.IsParameterPlaceholder(rowexprNode)
+		if !isPlaceholder {
+			return false
+		}
+	} else if !queryparser.IsXPathExprForXmlTable(xpath) {
```
makalaaneesh, line 186: Why do we need to validate the XPath? Won't it suffice to check that it is non-empty? We only want to detect if XMLTable function is being used, right? The fact that the query showed up in pg_stat_statements means that it will be a valid query, and therefore, a valid XPath?

yb-voyager/src/queryissue/unsupported_dml_constructs.go
```diff
@@ -111,3 +111,93 @@ func (d *XmlExprDetector) Detect(msg protoreflect.Message) ([]string, error) {
...
+Function to detect if a RangeTableFunc node represents XMLTABLE()
+*/
+func isXMLTable(rangeTableFunc protoreflect.Message) bool {
+	log.Infof("checking if range table func node is for XMLTABLE()")
+	// Check for 'docexpr' field
+	docexprField := rangeTableFunc.Descriptor().Fields().ByName("docexpr")
+	if docexprField == nil {
+		return false
+	}
+	docexprNode := rangeTableFunc.Get(docexprField).Message()
```
makalaaneesh, line 162: I'm not entirely sure, but from the docs it seems like this docexpr node can be NULL? 
https://www.postgresql.org/docs/current/functions-xml.html#:~:text=No%20rows%20will%20be%20produced%20if%20the%20document_expression%20is%20null%2C%20nor%20if%20the%20row_expression%20produces%20an%20empty%20node%2Dset%20or%20any%20value%20other%20than%20a%20node%2Dset.

yb-voyager/src/queryissue/unsupported_dml_constructs.go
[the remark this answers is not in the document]
makalaaneesh: Can you also document why we can't use the `FuncCallDetector` to detect XMLTABLE?

yb-voyager/src/queryparser/helpers.go
```diff
@@ -94,9 +105,11 @@ func GetStringValueFromNode(nodeMsg protoreflect.Message) string {
 	nodeType := node.Descriptor().FullName()
 	switch nodeType {
 	case PG_QUERY_STRING_NODE:
-		strField := node.Descriptor().Fields().ByName("sval")
-		strValue := node.Get(strField)
-		return strValue.String()
+		return extractStringField(node, "sval")
```
makalaaneesh, line 108: can you add one-liner comments here explaining what these node types are?

yb-voyager/src/queryparser/helpers.go
```diff
@@ -105,6 +118,118 @@ func GetStringValueFromNode(nodeMsg protoreflect.Message) string {
...
 }
+
+// IsParameterPlaceholder checks if the given node represents a parameter placeholder like $1
+func IsParameterPlaceholder(nodeMsg protoreflect.Message) bool {
+	if nodeMsg == nil || !nodeMsg.IsValid() {
+		return false
+	}
+
+	// Retrieve the 'node' oneof descriptor
+	nodeOneof := nodeMsg.Descriptor().Oneofs().ByName("node")
```
makalaaneesh, line 209: Consider taking this logic (getting nodeField out of nodeMsg) out to a different function if it's being done in multiple places.

makalaaneesh (commented):
Thanks for the code comments! Was helpful!

yb-voyager/src/queryissue/unsupported_dml_constructs.go
sanyamsinghal: My understanding:
- RangeTableFunc node is a generic node, can be used apart from XMLTABLE() also
- It has three main components - rowexpr, docexpr, and columns
- These components can also be present for any other RangeTableFunc node, so we should also look at the content of components here. Hence check the value/str which we expect it to be xpath.


Also in most of the cases, pgss will replace with a placeholder like $1...
But it was not doing that in case of `CREATE VIEW... SELECT XMLTABLE()....` also the check should be fullproof so i separately handling if its placeholder then always true(chances of false positives are very low)

Jfyi, i couldn't find any other examples for RangeTableFunc node, apart from XMLTABLE(), but since it is a node, there are chances of more cases.

yb-voyager/src/queryissue/unsupported_dml_constructs.go
sanyamsinghal: yeah good idea, we can write getter functions of the Detectors.

yb-voyager/src/queryissue/unsupported_dml_constructs.go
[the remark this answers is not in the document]
sanyamsinghal: added both.

yb-voyager/src/queryparser/helpers.go
sanyamsinghal: done

yb-voyager/src/queryissue/unsupported_dml_constructs.go
makalaaneesh: Ah, and there's no way to figure out that directly that the `RangeTableFunc` node is an XMLTABLE node? That seems strange!

makalaaneesh: Shooting in the dark here but from postgres code 
```static RangeTableFunc *
_copyRangeTableFunc(const RangeTableFunc *from)
{
	RangeTableFunc *newnode = makeNode(RangeTableFunc);

	COPY_SCALAR_FIELD(lateral);
	COPY_NODE_FIELD(docexpr);
	COPY_NODE_FIELD(rowexpr);
	COPY_NODE_FIELD(namespaces);
	COPY_NODE_FIELD(columns);
	COPY_NODE_FIELD(alias);
	COPY_LOCATION_FIELD(location);

	return newnode;
}
```

Could `alias` be helpful??

yb-voyager/src/queryparser/helpers.go
sanyamsinghal: done

yb-voyager/staticcheck.conf
```diff
@@ -0,0 +1,2 @@
+# Disable the S1008 check to allow more explicit boolean returns
```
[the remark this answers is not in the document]
sanyamsinghal: staticcheck module of golang was enforcing some simplifications which would reduce the readability of the code(in my case) and also enforces unnecessary restrictions to the developer.

So i added this to not check that case via staticchecker module.


Example::
```
func isEven(number int) bool {
    if number%2 == 0 {
        return true
    }
    return false
}
```
staticcheck would force to write this as:
```
func isEven(number int) bool {
    return number%2 == 0
}
```

Although in this case it seems to be correct simplication, my code has multiple if check return false, then the last one was being enforced, which doesn't make sense.
Refer `isXMLTable()` function in code.

yb-voyager/src/queryissue/unsupported_dml_constructs.go
```diff
@@ -111,3 +111,93 @@ func (d *XmlExprDetector) Detect(msg protoreflect.Message) ([]string, error) {
...
+	if queryparser.GetMsgFullName(msg) == queryparser.PG_QUERY_RANGETABLEFUNC_NODE {
+		if isXMLTable(msg) {
+			return []string{XML_FUNCTIONS}, nil
+		}
+	}
+	return nil, nil
+}
+
+/*
+XMLTABLE transforms XML data into relational table format, making it easier to query XML structures.
```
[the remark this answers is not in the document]
sanyamsinghal: > If so can we put a reference docs link here.
What reference docs you are mentioning here?

> Are we sure only XMLTABLE is the only one that needs a different kind of detection logic?
Yes ofc, XMLTABLE() needs a separate detection logic..?

yb-voyager/src/queryissue/unsupported_dml_constructs.go
```diff
@@ -111,3 +111,93 @@ func (d *XmlExprDetector) Detect(msg protoreflect.Message) ([]string, error) {
...
+		'/bookstore/book'
+		PASSING xml_column
+		COLUMNS
+			title TEXT PATH 'title',
+			author TEXT PATH 'author'
+	) AS x;
+
+Here, 'docexpr' points to 'xml_column' containing XML data, 'rowexpr' selects each 'book' node, and 'columns' extract 'title' and 'author' from each book.
+Hence Presence of XPath in 'rowexpr' and structured 'columns' typically indicates XMLTABLE usage.
+
```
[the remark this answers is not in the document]
sanyamsinghal: done

yb-voyager/src/queryissue/unsupported_dml_constructs.go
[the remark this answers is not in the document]
priyanshi-yb: from coverage point of view of these XML functions, like are we are sure if this XMLTABLE one is only which needs to be detected via `RangeFunNode` and we cover all others with other detectors, basically if we have referred any PG docs we can put that in the comments.

yb-voyager/src/queryissue/unsupported_dml_constructs.go
sanyamsinghal: hmm, code mentions `json_table` which is introduced in PG 17.
I think the only way is to check the xpath expression for xml table.

Also i am not sure if there are a lot of cases for table function apart from xmltable() and json_table()

yb-voyager/src/queryissue/unsupported_dml_constructs.go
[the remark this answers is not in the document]
sanyamsinghal: yes we had gone through the docs plus catalog tables to generate the list of xml functions.
Refer this in code
```
// Refer: https://www.postgresql.org/docs/17/functions-xml.html#FUNCTIONS-XML-PROCESSING
var unsupportedXmlFunctions = []string{
	// 1. Producing XML content
	"xmltext", "xmlcomment", "xmlconcat", "xmlelement", "xmlforest",
	"xmlpi", "xmlroot", "xmlagg",
	// 2. XML predicates
	"xml", "xmlexists", "xml_is_well_formed", "xml_is_well_formed_document",
	"xml_is_well_formed_content",
	// 3. Processing XML
	"xpath", "xpath_exists", "xmltable",
	// 4. Mapping Table to XML
	"table_to_xml", "table_to_xmlschema", "table_to_xml_and_xmlschema",
	"cursor_to_xmlschema", "cursor_to_xml",
	"query_to_xmlschema", "query_to_xml", "query_to_xml_and_xmlschema",
	"schema_to_xml", "schema_to_xmlschema", "schema_to_xml_and_xmlschema",
	"database_to_xml", "database_to_xmlschema", "database_to_xml_and_xmlschema",

	/*
		5. extras - not in ref doc but exists
		SELECT proname FROM pg_proc
		WHERE prorettype = 'xml'::regtype;
	*/
	"xmlconcat2", "xmlvalidate", "xml_in", "xml_out", "xml_recv", "xml_send", // System XML I/O
}
```

xmltable was the only one not being detected with the existing code, rest all are.

yb-voyager/src/queryissue/unsupported_dml_constructs.go
sanyamsinghal: Jfyi, golang pg parser fails for parsing a SQL with json_table. 
SQL stmt: https://www.postgresql.org/docs/current/functions-json.html#:~:text=The%20following%20query%20shows%20how%20to%20use%20JSON_TABLE

yb-voyager/src/queryissue/unsupported_dml_constructs.go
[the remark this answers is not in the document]
priyanshi-yb: okay thanks

yb-voyager/src/queryissue/unsupported_dml_constructs.go
priyanshi-yb: > hmm, code mentions json_table which is introduced in PG 17.

Pg-parser is on PG16 only yet. So It won't be able to detect further features

makalaaneesh: I think if you look at the documentation in postgres - https://github.com/postgres/postgres/blob/ea792bfd93ab8ad4ef4e3d1a741b8595db143677/src/include/nodes/parsenodes.h#L651 

It says 
```
 * RangeTableFunc - raw form of "table functions" such as XMLTABLE
 *
 * Note: JSON_TABLE is also a "table function", but it uses JsonTable node,
 * not RangeTableFunc.
 ```
 
 So, as of now, XML_TABLE is the only case that uses RangeTableFunc. 
 Therefore, It's okay to go ahead with this for now IMO, but let's document in code this assumption  + link to postgres code above @sanyamsinghal

File: yb-voyager/src/queryissue/unsupported_dml_constructs.go (hunk 4 above)
Comment (makalaaneesh) [reply yb-voyager/src/queryissue/unsupported_dml_constructs.go:177]: Ah, understood!

File: yb-voyager/src/queryissue/unsupported_dml_constructs.go (hunk 3 above)
Comment (sanyamsinghal) [reply yb-voyager/src/queryissue/unsupported_dml_constructs.go:201]: oh ok, missed `but it uses JsonTable node`

Yeah makes sense, in future there can be more cases to consider under RangeTableFunc node but as not as of now.
Will do comments.

Comment (sanyamsinghal) [reply yb-voyager/src/queryissue/unsupported_dml_constructs.go:210]: done

File: yb-voyager/src/queryissue/unsupported_dml_constructs.go [hunk 10]
```diff
@@ -111,3 +111,117 @@ func (d *XmlExprDetector) Detect(msg protoreflect.Message) ([]string, error) {
...
+	rowexpr:{a_const:{sval:{sval:"/bookstore/book"}  location:29}}
+	columns:{range_table_func_col:{colname:"title"  type_name:{names:{string:{sval:"text"}}  typemod:-1  location:87}  colexpr:{a_const:{sval:{sval:"title"}  location:97}}  location:81}}
+	columns:{range_table_func_col:{colname:"author"  type_name:{names:{string:{sval:"text"}}  typemod:-1  location:116}  colexpr:{a_const:{sval:{sval:"author"}  location:126}}  location:109}}
+alias:{aliasname:"x"}  location:17}}  limit_option:LIMIT_OPTION_DEFAULT  op:SETOP_NONE}}  stmt_len:142
+
+Here, 'docexpr' points to 'xml_column' containing XML data, 'rowexpr' selects each 'book' node, and 'columns' extract 'title' and 'author' from each book.
+Hence Presence of XPath in 'rowexpr' and structured 'columns' typically indicates XMLTABLE usage.
+*/
+// Function to detect if a RangeTableFunc node represents XMLTABLE()
+func isXMLTable(rangeTableFunc protoreflect.Message) bool {
```
makalaaneesh, line 179: Shall we put this function into `queryparser` ? because there is nothing queryissue specific that we're doing..

makalaaneesh, line 179: If you pass in a message that is NOT a rangeTableFunc to this function, ensure that it returns false. An explicit validation that msg is a rangeTableFunc might also help..

yb-voyager/src/queryissue/unsupported_dml_constructs.go
```diff
@@ -111,3 +111,117 @@ func (d *XmlExprDetector) Detect(msg protoreflect.Message) ([]string, error) {
...
+	log.Infof("checking if range table func node is for XMLTABLE()")
```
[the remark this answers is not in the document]
sanyamsinghal: yes, that was the intention, got changed during testing.

yb-voyager/src/queryissue/unsupported_dml_constructs.go
sanyamsinghal: currently the detector has that check. 
```
// Detect checks if a RangeTableFunc node is present for a XMLTABLE() function
func (d *RangeTableFuncDetector) Detect(msg protoreflect.Message) ([]string, error) {
	if queryparser.GetMsgFullName(msg) == queryparser.PG_QUERY_RANGETABLEFUNC_NODE {
		if isXMLTable(msg) {
			return []string{XML_FUNCTIONS}, nil
		}
	}
	return nil, nil
}
```

Do you still feel the need to add a check here, considering func arg name is `rangeTableFunc protoreflect.Message`?

sanyamsinghal: I was thinking about that. But kept it here deliberately, because Detection logic is also present in this function, along with some parsing details.

makalaaneesh: Ah, I see. 
The way I see it is that if you're defining a function called `IsXMLTable(protoreflect.Message)`, it should be able to handle any input of protoreflect.Message and return the proper result. 

If you define it as `isXMLTable(node rangeFuncNode)`, then the type is fixed to be rangeFuncNode, so that's okay. 
But currently, it's defined as `isXMLTable(rangeTableFunc protoreflect.Message)` and you're just making an assumption that the message is of type rangeTableFunc, which is not ideal. 
I would suggest: Either 
1. Don't make that assumption. Move the check of `GetMsgFullName(msg) == queryparser.PG_QUERY_RANGETABLEFUNC_NODE` inside `isXMLTable` 
OR 
2. define it as `isXMLTable(node rangeFuncNode)`. 

I don't think 2 is possible, given that you're dealing in proto messages, so maybe 1?

makalaaneesh: The detection logic is just "is the msg xmltable or not", which is in your detector, right?
The actual implementation of `isXMLTable` seems to be pure parser/pg logic?

sanyamsinghal: added check in IsXMLTable() to make it independent also.

makalaaneesh (commented):
LGTM!

makalaaneesh (approved):

PR #1949, opened 2024-11-20
[DB-14143] [Unsupported Query Constructs] [callhome] Part of assessment payload for UQC should have construct type and query count separated.

sanyamsinghal:


yb-voyager/src/callhome/diagnostics.go
```diff
@@ -102,6 +102,11 @@ type UnsupportedFeature struct {
 	TotalOccurrences int      `json:"TotalOccurrences"`
 }
 
+type UnsupportedQueryConstruct struct {
```
makalaaneesh, line 105: Let's discuss this in standup. I think we may need to think a little more and standardize the callhome json structure for assessment

sanyamsinghal: We can flatten it as we do for ybd. Not sure if it would break anything or not.

sanyamsinghal:
closing this for now - updated the jira ticket to solve it at a higher level
https://yugabyte.atlassian.net/browse/DB-14143

PR #1953, opened 2024-11-21
Add truncate-tables flag to import-data-to-target

makalaaneesh:
- `--truncate-tables` allows the user to truncate the tables before starting-clean. 
- Only applicable along with `--start-clean true`.
- only implemented for import-data-to-target, not import-data-to-source-replica, import-data-file
- caveats: 
1. In case CDC is enabled for those tables, it may fail as such: 
`
failed to truncate tables: truncate tables with query "TRUNCATE TABLE public.\"u\", public.\"fkt\", public.\"tfkt\", public.\"test_simple\", public.\"te\"": run query "TRUNCATE TABLE public.\"u\", public.\"fkt\", public.\"tfkt\", public.\"test_simple\", public.\"te\"" on target "127.0.0.1": ERROR: Invalid table definition: Error creating table test.u on the master: cannot rewrite a table that is a part of CDC or XCluster replication. See https://github.com/yugabyte/yugabyte-db/issues/16625. (SQLSTATE XX000)
`

When some tables are not empty: 
image

When passing --truncate-tables true
image

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -924,12 +924,21 @@ func cleanImportState(state *ImportDataState, tasks []*ImportFileTask) {
 		nonEmptyTableNames := lo.Map(nonEmptyNts, func(nt sqlname.NameTuple, _ int) string {
 			return nt.ForOutput()
 		})
-		utils.PrintAndLog("Non-Empty tables: [%s]", strings.Join(nonEmptyTableNames, ", "))
-		utils.PrintAndLog("The above list of tables on target DB are not empty. ")
-		yes := utils.AskPrompt("Are you sure you want to start afresh without truncating tables")
-		if !yes {
-			utils.ErrExit("Aborting import. Manually truncate the tables on target DB before continuing.")
+		if truncateTables {
+			utils.PrintAndLog("Truncating non-empty tables on target DB : %v", nonEmptyTableNames)
+			err := tdb.TruncateTables(nonEmptyNts)
+			if err != nil {
+				utils.ErrExit("failed to truncate tables: %s", err)
+			}
+		} else {
+			utils.PrintAndLog("Non-Empty tables: [%s]", strings.Join(nonEmptyTableNames, ", "))
+			utils.PrintAndLog("The above list of tables on target DB are not empty. If you wish to truncate them, re-run the import command with --truncate-tables flag.")
+			yes := utils.AskPrompt("Do you want to start afresh without truncating tables")
+			if !yes {
+				utils.ErrExit("Aborting import.")
+			}
 		}
+
 	}
 
 	for _, task := range tasks {
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -924,7 +924,8 @@ func cleanImportState(state *ImportDataState, tasks []*ImportFileTask) {
 		nonEmptyTableNames := lo.Map(nonEmptyNts, func(nt sqlname.NameTuple, _ int) string {
 			return nt.ForOutput()
 		})
-		if truncateTables {
+		if importerRole == TARGET_DB_IMPORTER_ROLE && truncateTables {
+			// truncate tables only supported for import-data-to-target.
 			utils.PrintAndLog("Truncating non-empty tables on target DB : %v", nonEmptyTableNames)
 			err := tdb.TruncateTables(nonEmptyNts)
 			if err != nil {
@@ -932,7 +933,10 @@ func cleanImportState(state *ImportDataState, tasks []*ImportFileTask) {
 			}
 		} else {
 			utils.PrintAndLog("Non-Empty tables: [%s]", strings.Join(nonEmptyTableNames, ", "))
-			utils.PrintAndLog("The above list of tables on target DB are not empty. If you wish to truncate them, re-run the import command with --truncate-tables flag.")
+			utils.PrintAndLog("The above list of tables on target DB are not empty.")
+			if importerRole == TARGET_DB_IMPORTER_ROLE {
+				utils.PrintAndLog("If you wish to truncate them, re-run the import command with --truncate-tables true")
+			}
 			yes := utils.AskPrompt("Do you want to start afresh without truncating tables")
 			if !yes {
 				utils.ErrExit("Aborting import.")
@@ -1431,8 +1435,8 @@ func init() {
 	registerTargetDBConnFlags(importDataToTargetCmd)
 	registerImportDataCommonFlags(importDataCmd)
 	registerImportDataCommonFlags(importDataToTargetCmd)
-	registerImportDataFlags(importDataCmd)
-	registerImportDataFlags(importDataToTargetCmd)
+	registerImportDataToTargetFlags(importDataCmd)
+	registerImportDataToTargetFlags(importDataToTargetCmd)
 }
 
 func createSnapshotImportStartedEvent() cp.SnapshotImportStartedEvent {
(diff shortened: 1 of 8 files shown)


yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -397,6 +398,19 @@ func (yb *TargetYugabyteDB) GetNonEmptyTables(tables []sqlname.NameTuple) []sqln
 	return result
 }
 
+func (yb *TargetYugabyteDB) TruncateTables(tables []sqlname.NameTuple) error {
+	tableNames := lo.Map(tables, func(nt sqlname.NameTuple, _ int) string {
+		return nt.ForUserQuery()
+	})
+	commaSeparatedTableNames := strings.Join(tableNames, ", ")
+	query := fmt.Sprintf("TRUNCATE TABLE %s", commaSeparatedTableNames)
```
priyanshi-yb, line 406: What should be our behavior with dependent tables (FK constraints) and table list?
For ex - if some tables are there in the table-list for which truncate and start-clean is happening but their dependent tables are already imported, this will error out to do cascade but I know we can't do `CASCADE` on our own so what should we expect from users?

makalaaneesh: Yeah, was thinking about it; I think we should use defaults, which is RESTRICT, and so if it fails, it fails, the onus is on the user to truncate manually.  
What do you think?

priyanshi-yb: Right, but if they truncate it with cascade by them, they will have to pass the dependent table in the list as well right so that will also get imported, so should we mention that somewhere maybe if we can detect it with error msg else in docs only?

makalaaneesh: IMO, that sort of falls outside the purview of voyager because that is something manually done by the user. If they truncate with cascade, they should know to also pass those extra tables in our table-list. I can't really see how we can communicate this in voyager docs 🤔

priyanshi-yb: Yeah I was also just thinking, if it is something we can tell them somehow, so probably one thing we can do is enhance the error msg to also include the db side Details (just to help user as we are doing the truncate for all tables at once.)
```
postgres=> truncate table t_a;
ERROR:  cannot truncate a table referenced in a foreign key constraint
DETAIL:  Table "t_b" references "t_a".
HINT:  Truncate table "t_b" at the same time, or use TRUNCATE ... CASCADE.
postgres=> 

```

makalaaneesh: Ah, that's a good idea, works. 
`
Truncating non-empty tables on target DB : [public.u public.fkt public.tfkt public.test_simple public.te]
failed to truncate tables: run query "TRUNCATE TABLE public.\"u\", public.\"fkt\", public.\"tfkt\", public.\"test_simple\", public.\"te\"" on target "127.0.0.1": ERROR: cannot truncate a table referenced in a foreign key constraint (SQLSTATE 0A000) HINT:"Truncate table \"test_simple_fk\" at the same time, or use TRUNCATE ... CASCADE." DETAIL:"Table \"test_simple_fk\" references \"test_simple\"."`

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -78,6 +79,12 @@ func (yb *TargetYugabyteDB) Exec(query string) (int64, error) {
 
 	res, err := yb.db.Exec(query)
 	if err != nil {
+		var pgErr *pgconn5.PgError
+		if errors.As(err, &pgErr) {
+			if pgErr.Hint != "" || pgErr.Detail != "" {
+				return rowsAffected, fmt.Errorf("run query %q on target %q: %w HINT:%q DETAIL:%q", query, yb.tconf.Host, err, pgErr.Hint, pgErr.Detail)
```
priyanshi-yb, line 85: nit: next lines for HINT and DETAIL - `fmt.Errorf("run query %q on target %q: %w\n HINT:%q\n DETAIL:%q", query, yb.tconf.Host, err, pgErr.Hint, pgErr.Detail)`

priyanshi-yb (approved):
LGTM!

yb-voyager/cmd/importData.go
```diff
@@ -924,12 +924,25 @@ func cleanImportState(state *ImportDataState, tasks []*ImportFileTask) {
...
+			}
+		} else {
+			utils.PrintAndLog("Non-Empty tables: [%s]", strings.Join(nonEmptyTableNames, ", "))
+			utils.PrintAndLog("The above list of tables on target DB are not empty.")
+			if importerRole == TARGET_DB_IMPORTER_ROLE {
+				utils.PrintAndLog("If you wish to truncate them, re-run the import command with --truncate-tables true")
+			}
+			yes := utils.AskPrompt("Do you want to start afresh without truncating tables")
+			if !yes {
+				utils.ErrExit("Aborting import.")
```
sanyamsinghal, line 942: In case of import data to source repliace, manual truncation is the only option right?
So should we add this - `Manually truncate the tables on target DB before continuing.`

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -331,6 +332,19 @@ func (pg *TargetPostgreSQL) GetNonEmptyTables(tables []sqlname.NameTuple) []sqln
 	return result
 }
 
+func (pg *TargetPostgreSQL) TruncateTables(tables []sqlname.NameTuple) error {
+	tableNames := lo.Map(tables, func(nt sqlname.NameTuple, _ int) string {
+		return nt.ForUserQuery()
+	})
+	commaSeparatedTableNames := strings.Join(tableNames, ", ")
+	query := fmt.Sprintf("TRUNCATE TABLE %s", commaSeparatedTableNames)
```
sanyamsinghal, line 340: Q: If we have mentioned a list with truncate, then if there are dependencies among tables, we won't need to `CASCADE` in this case?
I guess `CASCADE` is required in case of single table or if the list doesn't contain the dependent tables?

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -77,6 +79,12 @@ func (yb *TargetYugabyteDB) Exec(query string) (int64, error) {
 
 	res, err := yb.db.Exec(query)
 	if err != nil {
+		var pgErr *pgconn5.PgError
+		if errors.As(err, &pgErr) {
+			if pgErr.Hint != "" || pgErr.Detail != "" {
+				return rowsAffected, fmt.Errorf("run query %q on target %q: %w \nHINT:%q\nDETAIL:%q", query, yb.tconf.Host, err, pgErr.Hint, pgErr.Detail)
```
sanyamsinghal, line 85: space
`\nHINT: %q\n`

Also do we need %q here, or %s is sufficient. In case the Hint/Detail contains tabs/newline they will be quoted.

sanyamsinghal (changes requested):

yb-voyager/src/tgtdb/postgres.go
makalaaneesh: yup, https://www.postgresql.org/docs/current/sql-truncate.html#:~:text=is%20the%20default.-,CASCADE,that%20are%20not%20listed%20in%20the%20command.%20This%20is%20the%20default.,-Notes

sanyamsinghal:
Q: Why not implemented/enabled the same for `import data file`, understandable for source replica but import-data-file is for target only?

Q: The truncate table flag will work for snapshot only or both(live and snapshot) in case of import data to target

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: looks good, nice idea.

makalaaneesh:
> Q: Why not implemented/enabled the same for `import data file`, understandable for source replica but import-data-file is for target only?
> 
> Q: The truncate table flag will work for snapshot only or both(live and snapshot) in case of import data to target

1. Just to reduce (testing) scope since we're close to code freeze. 
2. Yes, both snapshot and live.

sanyamsinghal (approved):
LGTM, some nit comments.

PR #1963, opened 2024-11-22
Github actions workflow optimisations

sanyamsinghal:
Improvements:
1. Total time taken by GH actions: `57min`([run](https://github.com/yugabyte/yb-voyager/actions/runs/11967458025/job/33364722401?pr=1963)) -> `26min`([run](https://github.com/yugabyte/yb-voyager/actions/runs/11969727538/job/33371047249?pr=1963)) (**~2x** improvement)
    - divided the pg-migtests workflow into multiple parallel jobs(independent) to achieve this.
    
    
2. Refactoring: 
   - Removed old way of spawning yugabytedb cluster using docker -> Updated to use `yugabyted` CLI

3. Added GHA machine caching for Go modules(dependencies voyager) based on `go.sum` as hash/caching key.
    
Q: Replace `if: always()` with `continue-on-error: true` but it might result in nothing reported in GHA UI as failed.
   - Tried `!cancelled()` in place of `always()` since we don't want tests to run if workflow got cancelled. But once that job completed it called cancel for the incomplete jobs causing error out.
   - Issue with `always()`: It does let all the tests run but you can't stop a workflow once started manually also, because of always()
   - Going with `!cancelled()` considering that there is still caveat(mentioned above) but would suffice for most of the cases and `always()` should be used for special cases only line cleanup/teardown at end of tests.


makalaaneesh:
> Q: Replace if: always() with continue-on-error: true but it might result in nothing reported in GHA UI as failed. Reverify.

I believe @shubham-yb had already evaluated this at some point

sanyamsinghal:
> I believe @shubham-yb had already evaluated this at some point
I was seeing that behaviour with `continue-on-error` so thought to confirm.
cc @shubham-yb

shubham-yb:
Yes. When I tried it, if we put `continue-on-error` as true, the remaining tests will execute post failure but the overall build is set as passed.
https://stackoverflow.com/questions/58858429/how-to-run-a-github-actions-step-even-if-the-previous-step-fails-while-still-f?rq=3

Not sure if the behaviour has changed since then.

.github/workflows/misc-migtests.yml
```diff
@@ -85,15 +92,15 @@ jobs:
...
         run: |
           versions=("2.20.5.0-b72" "2.21.1.0-b271" "2024.1.1.0-b137")
           for version in "${versions[@]}"; do
             echo "Running tests on version $version"
 
             echo "Start YugabyteDB cluster"
-            docker pull yugabytedb/yugabyte:$version
-            VERSION=$version docker compose -f migtests/setup/yb-docker-compose.yaml up -d
+            docker run -d --name yugabytedb-$version -p7000:7000 -p9000:9000 -p15433:15433 -p5433:5433 -p9042:9042 \
+            yugabytedb/yugabyte:$version bin/yugabyted start --background=false --ui=false
```
shubham-yb, line 103: nit: Can we format this as well like the other files.

.github/workflows/misc-migtests.yml
```diff
@@ -43,6 +43,14 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
 
+      - name: Cache Go modules
+        uses: actions/cache@v3
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
```
shubham-yb, line 52: Just curious, how much benefit does this provide?
Since we might want to replicate what our users face while installing Voyager.

sanyamsinghal: It doesn't seem to be providing any speedup. I thought to just keep it thinking it won't harm anything.
Are you suggesting to remove it?

shubham-yb: Yes if there's no benefit, we should remove it. Voyager installation should handle it to replicate the user's environment as closely as possible.

sanyamsinghal: Will remove it. 
Btw it shouldn't make an difference wrt to replicating user's environment. If the `go.sum`(checksum) files for dependent modules is same then it would cache them and not install again. 
Voyager will be build irrespective of that.

.github/workflows/misc-migtests.yml
sanyamsinghal: done

.github/workflows/misc-migtests.yml
shubham-yb: Ah right! thanks :)

sanyamsinghal:
Seems like `yb-docker-compose.yaml` was used by jenkins also. Need to update the pipelines too.

sanyamsinghal:
Jenkins side changes - https://github.com/yugabyte/jenkins-helpers/pull/865

sanyamsinghal:
Pipeline run - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline-test/1040/

PR #1964, opened 2024-11-22
Fix: export data status command should not write in name registry

priyanshi-yb:
https://yugabyte.atlassian.net/browse/DB-14187

### DIFF
--- a/yb-voyager/cmd/exportDataStatusCommand.go
+++ b/yb-voyager/cmd/exportDataStatusCommand.go
@@ -55,7 +55,7 @@ var exportDataStatusCmd = &cobra.Command{
 			utils.ErrExit("\nNote: Run the following command to get the current report of live migration:\n" +
 				color.CyanString("yb-voyager get data-migration-report --export-dir %q\n", exportDir))
 		}
-		err = InitNameRegistry(exportDir, SOURCE_DB_EXPORTER_ROLE, nil, nil, nil, nil, false)
+		err = InitNameRegistry(exportDir, namereg.SOURCE_DB_EXPORT_STATUS_ROLE, nil, nil, nil, nil, false)
 		if err != nil {
 			utils.ErrExit("initializing name registry: %v", err)
 		}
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/exportDataStatusCommand.go
```diff
@@ -69,7 +69,9 @@ var exportDataStatusCmd = &cobra.Command{
 		}
 		useDebezium = msr.IsSnapshotExportedViaDebezium()
 
-
+		if msr.SourceDBConf == nil {
```
makalaaneesh, line 72: There is a small window after sourceDBConf is created but before export data actually starts exporting tables. If export-data-status is called in that window what happens? 
Just confirming that we also have a check later to print the same message if status files are not yet created, right?

yb-voyager/src/namereg/namereg.go
```diff
@@ -25,6 +25,7 @@ const (
 	SOURCE_DB_IMPORTER_ROLE         = "source_db_importer"         // Fallback.
 	SOURCE_REPLICA_DB_IMPORTER_ROLE = "source_replica_db_importer" // Fall-forward.
 	SOURCE_DB_EXPORTER_ROLE         = "source_db_exporter"
+	SOURCE_DB_EXPORT_STATUS_ROLE    = "source_db_export_status"
```
makalaaneesh, line 28: nit: keep it source_db_exporter_status just for consistency?

yb-voyager/cmd/exportDataStatusCommand.go
priyanshi-yb: Yes yes, that is already there

yb-voyager/src/namereg/namereg.go
```diff
@@ -25,6 +25,7 @@ const (
 	SOURCE_DB_IMPORTER_ROLE         = "source_db_importer"         // Fallback.
 	SOURCE_REPLICA_DB_IMPORTER_ROLE = "source_replica_db_importer" // Fall-forward.
 	SOURCE_DB_EXPORTER_ROLE         = "source_db_exporter"
+	SOURCE_DB_EXPORT_STATUS_ROLE    = "source_db_exporter_status"
```
makalaaneesh, line 28: nit: if you decided to change the value, change the variable name also? 😄

makalaaneesh (approved):
LGTM

PR #1967, opened 2024-11-22
Bug fix[DB-14190]: Assess Migration guardrails checks always detecting(false positives) for pgss extension is not loaded via shared libraries

sanyamsinghal:
- check only if able to fetch the libraries list

### DIFF
--- a/yb-voyager/src/srcdb/postgres.go
+++ b/yb-voyager/src/srcdb/postgres.go
@@ -1195,27 +1195,27 @@ func (pg *PostgreSQL) checkPgStatStatementsSetup() (string, error) {
 		}
 	}
 
-	// 2. check if its properly installed/loaded
+	// 2. User has permission to read from pg_stat_statements table
+	var hasReadAllStats bool
+	err = pg.db.QueryRow(queryHasReadStatsPermission).Scan(&hasReadAllStats)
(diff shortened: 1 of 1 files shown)


yb-voyager/src/srcdb/postgres.go
```diff
@@ -1195,29 +1195,29 @@ func (pg *PostgreSQL) checkPgStatStatementsSetup() (string, error) {
...
 	err = pg.db.QueryRow(queryHasReadStatsPermission).Scan(&hasReadAllStats)
 	if err != nil {
 		return "", fmt.Errorf("failed to check pg_read_all_stats grant on migration user: %w", err)
 	}
-
 	if !hasReadAllStats {
 		return "User doesn't have permissions to read pg_stat_statements view, unsupported query constructs won't be detected/reported", nil
 	}
 
+	// 3. check if its properly installed/loaded
```
makalaaneesh, line 1208: Seems like you moved around the validations, but it's not very clear why. Can you elaborate?

sanyamsinghal: yeah 3 one is like a extra check right. Not strict. Because user might not have required privileges to check shared preload libraries list. So just moved it to later after main permissions check.

PR #1968, opened 2024-11-22
Fix: export data status command should not write in name registry (#1964)

priyanshi-yb:
…

https://yugabyte.atlassian.net/browse/DB-14187


ShivanshGahlot (approved):
LGTM

PR #1971, opened 2024-11-25
Filter Issues by Version [2]

makalaaneesh:
- New flag: target-db-version flag in analyze-schema/assess-migration 
user has to pass in the full version (A.B.C.D)
only versions from a list of supported series can be passed. This list will be all the YB series that are not EOL. (Although officially we document that voyager only supports the two latest stable releases and one latest preview, we would NOT like to fail here if user passes 2.18, for example.)
If user passes a version from a series that is not supported by voyager (say 2026.1.x.x), voyager will print the error, prompt the user if they want to continue with default (latest stable).
Defaults to latest stable version (2024.1.3.0, for example)
- Every Issue that we report will have the following details
For every series (from the supported list), if it is fixed in any version of the series, we store that information in "minVersionsFixedIn" which is a map[series]version
2.20: fixed in 2.20.8.0
2024.1: fixed in 2024.1.1.0
2.23: fixed in 2.23.5.1
2024.2: fixed in 2024.2.1.0
- Checking if issue is fixed in target-db-version (specified by user)
We compare target-db-version with Issue.minVersionsFixedIn[target-db-version.Series]
- Getting list of versions that issue is fixed in (to display in reports):
Option1: return all values from "minVersionsFixedIn"
Option2: return values from "minVersionsFixedIn" whose series >= target-db-version.Series. Caveat: We will not be able to compare between stable and preview releases (no way to determine if 2.23 > 2024.1 or not). Therefore for target-db-version=2024.2.1.1(stable), we will omit the older stable series (2024.1 series) but include all preview series - (2.21, 2.23).


yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1400,3 +1410,27 @@ func validateAssessmentMetadataDirFlag() {
 		}
 	}
 }
+
+func validateAndSetTargetDbVersionFlag() error {
```
sanyamsinghal, line 1414: do we need to return error from this function? since we are not adding any context to the err message from the caller func

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1400,3 +1410,27 @@ func validateAssessmentMetadataDirFlag() {
...
+		targetDbVersion = version.LatestStable
+		utils.PrintAndLog("Defaulting to latest stable YugabyteDB version: %s", targetDbVersion)
+		return nil
+	}
+	var err error
+	targetDbVersion, err = version.NewYBVersion(targetDbVersionStrFlag)
+
+	if !errors.Is(err, version.ErrUnsupportedSeries) {
+		return err
+	}
```
sanyamsinghal, line 1425: wondering, should we explicitly add condition here for `err == nil` being handled also? Just to make the code more readable.

makalaaneesh:
> Let's say there was a bug which got fixed in `2024.3.1.0`
> And user is using `2026.2` series, then are we going maintain anything in our `minVersionsFixedIn` map for `2026.2` series, since it got completely fixed in some previous stable release?
> If not then how are we gonna compare correctly for `2026.x`

Yes, the expectation is that we will maintain an item in `minVersionsFixedIn` for 2026.2 as well. (because of backports, we can't directly assume that the issue will also be fixed in 2026.2)

sanyamsinghal:
> Yes, the expectation is that we will maintain an item in `minVersionsFixedIn` for 2026.2 as well

Just to be sure, lets say something is fixed in `2026` series completely. 
Then we will store something like `2026.1.0.0` to conclude that it is fixed across `2026` versions

makalaaneesh:
> > Yes, the expectation is that we will maintain an item in `minVersionsFixedIn` for 2026.2 as well
> 
> Just to be sure, lets say something is fixed in `2026` series completely. Then we will store something like `2026.1.0.0` to conclude that it is fixed across `2026` versions

Yes, exactly, we will have to do that!

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: Yeah, alternative could be to just ErrExit directly. 
I just thought it would be good to follow the pattern of returning errors everywhere; only utils.ErrExit in the cobra command's PreRun/Run functions (and ofc prompts)

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: Makes sense, done!

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: ok, yeah the pattern makes sense. at somepoint we can refactor other validates() also....

sanyamsinghal (approved):

PR #1972, opened 2024-11-25
Do not detect the DML issues on MVIEW/VIEW object in GetDMLIssues()

priyanshi-yb:
https://yugabyte.atlassian.net/browse/DB-14204

We should not detect issues on CREATE MVIEw/VIEw queries coming from pgss to `GetDMLIssues` as we are already detecting them in PL/pgSQL section.

### DIFF
--- a/yb-voyager/src/queryissue/queryissue.go
+++ b/yb-voyager/src/queryissue/queryissue.go
@@ -95,6 +95,9 @@ func (p *ParserIssueDetector) GetDMLIssues(query string) ([]issue.IssueInstance,
 	if err != nil {
 		return nil, fmt.Errorf("error parsing query: %w", err)
 	}
+	if queryparser.IsViewObject(parseTree) || queryparser.IsMviewObject(parseTree) {
+		return nil, nil
+	}
 	var result []issue.IssueInstance
 	var unsupportedConstructs []string
 	visited := make(map[protoreflect.Message]bool)


yb-voyager/src/queryissue/queryissue.go
```diff
@@ -95,6 +95,9 @@ func (p *ParserIssueDetector) GetDMLIssues(query string) ([]issue.IssueInstance,
 	if err != nil {
 		return nil, fmt.Errorf("error parsing query: %w", err)
 	}
+	if queryparser.IsViewObject(parseTree) || queryparser.IsMviewObject(parseTree) {
```
makalaaneesh, line 98: I think we need to think about this a bit more. What about all the other types of DDLs? 
What if we have a CREATE TABLE with the default of some column value being an xml function. If that is present in the pg_stat_statements, that will come up... 

The only thing that comes to mind right now is that we have a function `isDDL` that detects all types of DDLs, but ofc that is also not fool-proof.

yb-voyager/src/queryissue/queryissue.go
```diff
@@ -95,6 +95,9 @@ func (p *ParserIssueDetector) GetDMLIssues(query string) ([]issue.IssueInstance,
...
+		return nil, nil
+	}
```
sanyamsinghal, line 100: @priyanshi-yb, can we have some higher level check to find out if its a `ddl` or `dml` as the name of the function also, rather than doing it based on the object type in here.

PR #2004, opened 2024-11-26
Reporting %TYPE syntax for type names in the Function and Procedure objects

priyanshi-yb:


yb-voyager/src/queryissue/queryissue.go
```diff
@@ -57,6 +58,13 @@ func (p *ParserIssueDetector) GetAllIssues(query string) ([]issue.IssueInstance,
 			}
 			issues = append(issues, issuesInQuery...)
 		}
+
+		percentTypeSyntaxIssues, err := p.GetPercentTypeSyntaxIssues(query)
```
makalaaneesh, line 62: is `%TYPE` valid syntax ONLY in functions/procedures, or is it possible to use it in other contexts?

migtests/tests/pg/assessment-report-test/pg_assessment_report.sql
```diff
@@ -294,3 +294,56 @@ END;
 $$ LANGUAGE plpgsql;
 
 select process_order(1);
+
+-- In PG migration from pg_dump the function parameters and return will never have the %TYPE syntax, instead they have the actual type in the DDLs
+-- e.g. for the below function this will be the export one `CREATE FUNCTION public.process_combined_tbl(p_id integer, p_c cidr, p_bitt bit, p_inds3 interval) RETURNS macaddr`
+CREATE OR REPLACE FUNCTION public.process_combined_tbl(
```
makalaaneesh, line 300: In end-to-end tests, let's only add a very basic case (with one instance of `%TYPE` in a function and procedure. This will help us avoid changing these expected files frequently (which can be a pain). 

Let's write unit tests (we can test `GetAllIssues` function) for all other cases (func params, func return, etc).

makalaaneesh: same for the other analyze-schema test.

yb-voyager/src/queryissue/queryissue.go
priyanshi-yb: As per this [docs](https://www.postgresql.org/docs/current/sql-createfunction.html#:~:text=The%20type%20of%20a%20column%20is%20referenced%20by%20writing%20table_name.column_name%25TYPE.%20Using%20this%20feature%20can%20sometimes%20help%20make%20a%20function%20independent%20of%20changes%20to%20the%20definition%20of%20a%20table), it looks like the feature is for Functions and Procedures. I even tried with other DDLs like TABLE and TYPE, it didn't work - 
```
schema_changes=> create type test_type as (id int, d combined_tbl.d%TYPE);
ERROR:  syntax error at or near "%"
LINE 1: create type test_type as (id int, d combined_tbl.d%TYPE);
                                                      ^
schema_changes=> create type test_type as (id int, d daterange);
CREATE TYPE
schema_changes=> create table tt(id int, d combined_tbl.d%TYPE);
ERROR:  syntax error at or near "%"
LINE 1: create table tt(id int, d combined_tbl.d%TYPE);
                                                ^
schema_changes=> create table tt(id int, d daterange);
CREATE TABLE
schema_changes=> 
```

migtests/tests/pg/assessment-report-test/pg_assessment_report.sql
priyanshi-yb: For assessment, its mainly a basic case only getting tested, as pg_dump doesn't give `%TYPE` syntax in parameters and return types, so those are not getting tested but I thought its better to have it here in case pg_dump behavior changes anytime.

For analyze yes, we can do that I can take that up later when we will add unit tests for `GetAllIssues` to remove them from here.

makalaaneesh: Even for assessment, I think end-to-end tests should only test the most basic thing: Are `% TYPE` issues getting reported or not? 
The other cases can be unit tested with `GetAllIssues` . 
That way, less things to change in the expectedFiles as well! 
Okay if you want to take it up later 👍

yb-voyager/src/issue/ddl.go
```diff
@@ -0,0 +1,31 @@
+/*
```
makalaaneesh, line 1: Let's move this file to `queryissue` pkg as we had thought of doing? It makes sense for all the definitions to be in `queryissue` rather than in `issue` pkg

yb-voyager/src/queryissue/queryissue.go
```diff
@@ -120,6 +126,31 @@ func (p *ParserIssueDetector) getAllIssues(query string) ([]issue.IssueInstance,
 	return issues, nil
 }
 
+func (p *ParserIssueDetector) GetPercentTypeSyntaxIssues(query string) ([]issue.IssueInstance, error) {
```
makalaaneesh, line 129: maybe take the parseTree as an input instead of the query? we already ran parse above.

yb-voyager/src/queryparser/query_parser.go
```diff
@@ -163,4 +163,74 @@ func getCreateViewNode(parseTree *pg_query.ParseResult) (*pg_query.Node_ViewStmt
...
+
+func getQualifiedTypeName(typeNames []*pg_query.Node) string {
+	var typeNameStrings []string
+	for _, n := range typeNames {
+		typeNameStrings = append(typeNameStrings, n.GetString_().Sval)
+	}
+	return strings.Join(typeNameStrings, ".")
+}
+
+func getParserTypeInString(typeVar *pg_query.TypeName) string {
```
makalaaneesh, line 199: What do you mean by `ParserType` ?  isn't this function more about getting fully qualified type name?

makalaaneesh (commented):
First round of comments. Remaining: `traversal_plpgsql.go`

yb-voyager/src/queryparser/traversal_plpgsql.go
```diff
@@ -228,3 +216,183 @@ func formatExprQuery(q string) string {
...
+*/
+func GetAllTypeNamesInPlpgSQLStmt(query string) ([]string, error) {
+	parsedJson, parsedJsonMap, err := getParsedJsonMap(query)
+	if err != nil {
+		return []string{}, nil
+	}
+	function := parsedJsonMap[PLPGSQL_FUNCTION]
+	parsedFunctionMap, ok := function.(map[string]interface{})
+	if !ok {
+		return []string{}, fmt.Errorf("error getting the PlPgSQL_Function field in parsed json-%s", parsedJson)
```
makalaaneesh, line 323: nit: remove the word "error" in the error msg

yb-voyager/src/queryparser/traversal_plpgsql.go
```diff
@@ -228,3 +216,183 @@ func formatExprQuery(q string) string {
...
+	datums := parsedFunctionMap[DATUMS]
+	datumList, isList := datums.([]interface{})
+	if !isList {
+		return []string{}, fmt.Errorf("error getting type names datums field is not list in parsed json-%s", parsedJson)
+	}
+
+	var typeNames []string
+	for _, datum := range datumList {
+		datumMap, ok := datum.(map[string]interface{})
+		if !ok {
```
makalaaneesh, line 335: do we know in what situations it would NOT be a map[string]interface. If not, should we error out? 
If we do not want to error and just move on, atleast, let's log.ERROR?

yb-voyager/src/queryissue/queryissue.go
```diff
@@ -120,6 +126,31 @@ func (p *ParserIssueDetector) getAllIssues(query string) ([]issue.IssueInstance,
...
+		return nil, fmt.Errorf("error parsing the query-%s: %v", query, err)
+	}
+
+	objType, objName := queryparser.GetObjectTypeAndObjectName(parseTree)
+	typeNames, err := queryparser.GetAllTypeNamesInPlpgSQLStmt(query)
+	if err != nil {
+		return nil, fmt.Errorf("error getting type names in PLPGSQL: %v", err)
+	}
+
+	if queryparser.IsFunctionObject(parseTree) {
```
makalaaneesh, line 141: add the caveats comment here as well, so we know why we're making additional function calls after calling `GetAllTypeNamesInPlpgSQLStmt`

yb-voyager/src/queryparser/query_parser.go
```diff
@@ -163,4 +163,74 @@ func getCreateViewNode(parseTree *pg_query.ParseResult) (*pg_query.Node_ViewStmt
...
+)
+parseTree-
+parameters:{function_parameter:{name:"p_id"  arg_type:{names:{string:{sval:"pg_catalog"}}  names:{string:{sval:"int4"}}  typemod:-1  location:66} 
+mode:FUNC_PARAM_DEFAULT}}  parameters:{function_parameter:{name:"p_c"  arg_type:{names:{string:{sval:"public"}}  names:{string:{sval:"combined_tbl"}}
+names:{string:{sval:"c"}}  pct_type:true  typemod:-1  location:87}  mode:FUNC_PARAM_DEFAULT}}  parameters:{function_parameter:{name:"p_bitt"  
+arg_type:{names:{string:{sval:"public"}}  names:{string:{sval:"combined_tbl"}}  names:{string:{sval:"bitt"}}  pct_type:true  typemod:-1  
+location:136}  mode:FUNC_PARAM_DEFAULT}} 
+
+*/
+func GetFuncParametersTypeNames(parseTree *pg_query.ParseResult) []string {
```
makalaaneesh, line 224: (you can pick this up later since you are already refactoring...)
In-line with how you are refactoring currently, you could 
 - parse the function into a function DDLObject, 
- store all the types in parameters, return, etc, along with additional field `isPctType`. 

Then your detection can just go over and filter that list. Adding `%TYPE` suffix in `getParserTypeInString` isn't the cleanest way of doing it...

yb-voyager/src/queryparser/traversal_plpgsql.go
```diff
@@ -228,3 +216,183 @@ func formatExprQuery(q string) string {
...
+		"PLpgSQL_type": {
+			"typname": "employees.tax_rate%TYPE"
+		}
+	}
+}
+*/
+func getTypeNameFromPlpgSQLVar(plpgsqlVar interface{}) string {
+	//getting the map of <key,val > of PLpgSQL_Var json
+	valueMap, ok := plpgsqlVar.(map[string]interface{})
+	if !ok {
```
makalaaneesh, line 366: In this function, for all the `!ok`, how about we just return an error. The function should either successfully get the type name from the plpgSQLVar or return error. Why return empty string and expect the caller to check if it's empty or not?

In this case, I guess the caller wouldn't want to fail, so we can always log the error, and continue.

yb-voyager/src/queryparser/query_parser.go
priyanshi-yb: It's more of getting a string representation for fully qualified type name with `%TYPE` suffix from parser `TypeName`

yb-voyager/src/issue/ddl.go
priyanshi-yb: I will have to move everything then `constants.go`, `issue.go` to `queryissue` , is that fine?

makalaaneesh: You know what, let's not club this change with this PR. I will take it up later.

yb-voyager/src/queryparser/traversal_plpgsql.go
priyanshi-yb: okay makes sense

yb-voyager/src/queryissue/queryissue.go
priyanshi-yb: If I did that, I would expose the `pg_query` library in the `queryissue,` which, i think, doesn't make sense as we are using `queryparser`.

makalaaneesh: Hmm got it. Hopefully parsing multiple times is not an perf issue 👍

priyanshi-yb: I don't think it will. I remember Sanyam doing a perf test for Parse of pg_query; it was in microseconds.

yb-voyager/src/queryparser/traversal_plpgsql.go
```diff
@@ -320,28 +327,31 @@ func GetAllTypeNamesInPlpgSQLStmt(query string) ([]string, error) {
...
 		}
 		for key, val := range datumMap {
 			switch key {
 			case PLPGSQL_VAR:
-				typeName := getTypeNameFromPlpgSQLVar(val)
-				if typeName != "" {
-					typeNames = append(typeNames, typeName)
+				typeName, err := getTypeNameFromPlpgSQLVar(val)
+				if err != nil {
+					log.Errorf("error in getting typename from PLPGSQL_VAR(%v): %v", val, err)
```
makalaaneesh, line 351: nit: remove error in error msg

makalaaneesh (approved):
LGTM

PR #2006, opened 2024-11-26
[DB-13912]: Filter database name from YUGABYTED_DB_CONN_STRING string if present

sanyamsinghal:
- if provided other than yugabyte then ybd-ui won't be able to pickup the payload

Fixes - https://yugabyte.atlassian.net/browse/DB-13912

Q. With the usage of RedactedURL what was the goal:
 - Just avoid password from being sent to services like callhome
 - Or Also avoid password from getting logged in voyager log files also.
 
 I assume it's later, but in that we need to handle case like `url.Parse()` itself fails with some `err`(contains conn string)
 Then we use ErrExit() which ends logging the password

### DIFF
--- a/yb-voyager/cmd/root.go
+++ b/yb-voyager/cmd/root.go
@@ -19,8 +19,10 @@ import (
 	"fmt"
 	"net/http"
 	_ "net/http/pprof"
+	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


sanyamsinghal:
we decided to document this, and move on since its voyager issue

PR #2007, opened 2024-11-27
Report target-db-version and supported versions in assessment/analyze-schema reports

makalaaneesh:
Analyze Schema: 
image
image

Assessment: 
image
image


yb-voyager/cmd/analyzeSchema.go
```diff
@@ -265,6 +265,12 @@ func reportCase(filePath string, reason string, ghIssue string, suggestion strin
 		issue.DocsLink = docsLink
 	}
 
+	// just for testing
```
priyanshi-yb, line 268: please remove them before merging.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -964,10 +998,15 @@ func fetchUnsupportedPGFeaturesFromSchemaReport(schemaAnalysisReport utils.Schem
 }
 
 func getIndexesOnComplexTypeUnsupportedFeature(schemaAnalysisiReport utils.SchemaReport, unsupportedIndexDatatypes []string) UnsupportedFeature {
+	// TODO: include MinimumVersionsFixedIn
 	indexesOnComplexTypesFeature := UnsupportedFeature{
 		FeatureName: "Index on complex datatypes",
 		DisplayDDL:  false,
 		Objects:     []ObjectInfo{},
+		MinimumVersionsFixedIn: map[string]*version.YBVersion{
```
priyanshi-yb, line 1006: testing right?
please make sure to remove these.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -924,9 +951,16 @@ func getUnsupportedFeaturesFromSchemaAnalysisReport(featureName string, issueRea
 			}
 			link = issue.DocsLink
 			objects = append(objects, objectInfo)
+			if !minVersionsFixedInSet {
```
priyanshi-yb, line 954: nit: we can use the `minVersionsFixedIn` to check if its set or not.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1057,12 +1096,22 @@ func fetchUnsupportedPlPgSQLObjects(schemaAnalysisReport utils.SchemaReport) []U
...
+		var minVersionsFixedIn map[string]*version.YBVersion
+		var minVersionsFixedInSet bool
+
 		for _, issue := range issues {
 			objects = append(objects, ObjectInfo{
 				ObjectType:   issue.ObjectType,
 				ObjectName:   issue.ObjectName,
 				SqlStatement: issue.SqlStatement,
 			})
+			if !minVersionsFixedInSet {
```
priyanshi-yb, line 1108: same here as above

yb-voyager/src/version/constants.go
```diff
@@ -29,12 +27,22 @@ const (
...
 var V2024_1_3_1 *YBVersion
+var V2024_1_4_0 *YBVersion
+var V2_23_5_0 *YBVersion
 
 func init() {
 	var err error
 	V2024_1_3_1, err = NewYBVersion("2024.1.3.1")
 	if err != nil {
-		utils.ErrExit("could not create version 2024.1")
+		panic("could not create version 2024.1.3.1")
```
priyanshi-yb, line 37: can't we use utils.ErrExit?

yb-voyager/src/utils/commonVariables.go
```diff
@@ -124,9 +135,10 @@ type TableColumnsDataTypes struct {
 }
 
 type UnsupportedQueryConstruct struct {
-	ConstructTypeName string
-	Query             string
-	DocsLink          string
+	ConstructTypeName      string
+	Query                  string
+	DocsLink               string
+	MinimumVersionsFixedIn map[string]*version.YBVersion // key: series (2024.1, 2.21, etc)
```
priyanshi-yb, line 141: we don't need `isFixedIn` here as we will get the filtered ones from `queryissue` pkg, right?

yb-voyager/src/version/constants.go
makalaaneesh: utils has some structs whose fields are of type `YBVersion`. so there needs to be a dependency on utils -> version. adding version->utils dependency causes circular dependency, so i thought we can just panic (As this is in the init function, it will definitely be caught in dev testing)

yb-voyager/src/utils/commonVariables.go
makalaaneesh: We still need because we need to include in the report the (higher) versions that it might be supported in. So, for both UnsupportedFeature, UnsupportedQueryConstruct, we'd need to store `MinimumVersionsFixedIn`.

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: Ah, so it would be wrong to check if `minVersionsFixedIn` is not nil. Consider the case where out of 2 issues, first one has `minVersionsFixedIn= nil` and second one has `minVersionsFixedIn as some not nil value`. This is a case where we want to catch and throw an error, because we want all issues in an unsupportedFeature to have the same `minVersionsFixedIn` (at least for now. Maybe when we change the report structure into a table, this might change). But if we just check if `minVersionsFixedIn` is nil or not, we will fail to catch this case.

yb-voyager/src/utils/commonVariables.go
priyanshi-yb: yeah, for map I got it, I was asking about the `IsFixedIn()` function for the struct, to check the construct is fixed or not.

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: Ah okay got it.

yb-voyager/src/utils/commonVariables.go
makalaaneesh: I added it based on our previous discussion:
Ideally all filtering of issues should happen in queryissue pkg layer, so we should not need to filter here. But until all code is moved out, we should also filter in analyze-schema layer. 

Although now you're moving a majority of the code from analyze-schema into queryissues, there will still be some regex-based code in analyze-schema layer, I assume?

priyanshi-yb: > Although now you're moving a majority of the code from analyze-schema into queryissues, there will still be some regex-based code in analyze-schema layer, I assume?

yeah right

yb-voyager/src/ybversion/constants.go
```diff
@@ -29,12 +27,22 @@ const (
...
+		panic("could not create version 2024.1.3.1")
+	}
+	V2024_1_4_0, err = NewYBVersion("2024.1.4.0")
+	if err != nil {
+		panic("could not create version 2024.1.4.0")
+	}
+	V2_23_5_0, err = NewYBVersion("2.23.5.0")
+	if err != nil {
+		panic("could not create version 2.23.5.0")
 	}
```
sanyamsinghal, line 46: were these two version added for testing ?

yb-voyager/src/issue/dml.go
```diff
@@ -49,6 +49,10 @@ var xmlFunctionsIssue = Issue{
 	Suggestion:      "",
 	GH:              "",
 	DocsLink:        "https://docs.yugabyte.com/preview/yugabyte-voyager/known-issues/postgresql/#xml-functions-is-not-yet-supported",
+	// MinimumVersionsFixedIn: map[string]*ybversion.YBVersion{
```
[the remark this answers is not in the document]
makalaaneesh: My bad, I thought I had cleaned up the PR. Will ping again

yb-voyager/src/utils/commonVariables.go
```diff
@@ -97,15 +99,49 @@ type DBObject struct {
...
+	IssueType              string                          `json:"IssueType"`
+	ObjectType             string                          `json:"ObjectType"`
+	ObjectName             string                          `json:"ObjectName"`
+	Reason                 string                          `json:"Reason"`
+	SqlStatement           string                          `json:"SqlStatement,omitempty"`
+	FilePath               string                          `json:"FilePath"`
+	Suggestion             string                          `json:"Suggestion"`
+	GH                     string                          `json:"GH"`
+	DocsLink               string                          `json:"DocsLink,omitempty"`
+	MinimumVersionsFixedIn map[string]*ybversion.YBVersion `json:"MinimumVersionsFixedIn" xml:"-"` // key: series (2024.1, 2.21, etc)
```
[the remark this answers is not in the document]
makalaaneesh: Initially I wanted to ignore xml, but I'm having second thoughts. I don't want to block this PR because of that. Can I take it up in a subsequent PR?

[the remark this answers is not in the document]
priyanshi-yb: Yes sure

makalaaneesh:
https://jenkins.dev.yugabyte.com/blue/organizations/jenkins/users%2Fyb-voyager-testing%2Fyb-voyager-testing-pipeline/detail/yb-voyager-testing-pipeline/4125/pipeline

priyanshi-yb (approved):
LGTM

PR #2020, opened 2024-12-02
Introduced DDLParser and DDLIssueDetector for parsing DDLs and detecting their issues.

priyanshi-yb:


yb-voyager/src/queryparser/query_parser.go
```diff
@@ -163,4 +213,79 @@ func getCreateViewNode(parseTree *pg_query.ParseResult) (*pg_query.Node_ViewStmt
...
+func IsCreateType(parseTree *pg_query.ParseResult) bool {
+	_, isComposite := getCompositeTypeStmtNode(parseTree)
+	_, isEnum := getEnumTypeStmtNode(parseTree)
+	return isComposite || isEnum
+}
+
+func IsCreateForeign(parseTree *pg_query.ParseResult) bool {
+	_, isForeignTable := getForeignTableStmtNode(parseTree)
+	return isForeignTable
+}
```
sanyamsinghal, line 291: Here we were dealing with protobuf messages instead of structs. 
The above code can be generalised and simplified like this(can be ::
```
// IsNodeType checks if the provided message is of the specified type.
func IsNodeType(msg protoreflect.Message, targetType string) bool {
	if msg == nil || !msg.IsValid() {
		return false
	}
	fmt.Printf("msgType: %s\n", GetMsgFullName(msg))
	return GetMsgFullName(msg) == targetType
}

// Predicate function to check for specific statement types
func IsStatementType(msg protoreflect.Message, nodeType string) bool {
	nodeMsg := getOneOfActiveField(msg, "node")
	if nodeMsg == nil {
		return false
	}
	return IsNodeType(nodeMsg, nodeType)
}

func IsCreateTable(msg protoreflect.Message) bool {
	return IsStatementType(msg, NodeCreateStmt)
}

func IsCreateView(msg protoreflect.Message) bool {
	return IsStatementType(msg, NodeViewStmt)
}

func IsCreateFunction(msg protoreflect.Message) bool {
	return IsStatementType(msg, NodeCreateFunctionStmt)
}

func IsCreateIndex(msg protoreflect.Message) bool {
	return IsStatementType(msg, NodeIndexStmt)
}

func IsAlterTable(msg protoreflect.Message) bool {
	return IsStatementType(msg, NodeAlterTableStmt)
}

func IsCreateTrigger(msg protoreflect.Message) bool {
	return IsStatementType(msg, NodeCreateTrigStmt)
}

func IsCreatePolicy(msg protoreflect.Message) bool {
	return IsStatementType(msg, NodeCreatePolicyStmt)
}

func IsCreateCompositeType(msg protoreflect.Message) bool {
	return IsStatementType(msg, NodeCompositeTypeStmt)
}

func IsCreateEnumType(msg protoreflect.Message) bool {
	return IsStatementType(msg, NodeCreateEnumStmt)
}

func IsCreateForeignTable(msg protoreflect.Message) bool {
	return IsStatementType(msg, NodeCreateForeignTableStmt)
}
```

sanyamsinghal: All the above added helper function can be further reduced to just have one helper function:
```
// Predicate function to check for specific statement types
func GetStatementType(msg protoreflect.Message) bool {
	nodeMsg := getOneOfActiveField(msg, "node")
	if nodeMsg == nil {
		return false
	}
	return GetMsgFullName(nodeMsg)
}
```

The main caller function can directly use that::
```
func GetDDLParser(parseTree *pg_query.ParseResult) (DDLParser, error) {
	switch GetStatementType(parseTree..Stmts[0].Stmt.ProtoReflect()) {
	case NODE_CreateTable:
		return NewTableParser(), nil
	case NODE_CreateIndex:
		return NewIndexParser(), nil
	case NODE_AlterTable:
		return NewAlterTableParser(), nil
	case NODE_CreatePolicy:
		return NewPolicyParser(), nil
	case NODE_CreateTrigger:
		return NewTriggerParser(), nil
	case NODE_CreateType:
		return NewTypeParser(), nil
	case NODE_CreateForeign:
		return NewForeignTableParser(), nil
	default:
		return NewNoOpParser(), nil
	}
}
```

sanyamsinghal (commented):
Consider converting some pieces of query_parser code to protobuf message approach.

yb-voyager/src/queryissue/unsupported_ddls.go
```diff
@@ -0,0 +1,617 @@
...
+
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/issue"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/queryparser"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/srcdb"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
+)
+
+// DDLIssueDetector interface defines methods for detecting issues in DDL objects
+type DDLIssueDetector interface {
+	DetectIssues(queryparser.DDLObject) ([]issue.IssueInstance, error)
```
makalaaneesh, line 33: Do we really need to do 1. Parse into DDLObject, 2. pass DDLObject to Detector? 

Can we just pass query directly to each detector, let each detector parse accordingly however it wants ? 

You wouldn't need a `GetDDLParser` function because each detector is doing it's own parsing. 
You just need a`GetDDLDetector` that can just be based on the nodeType, rather than the DDLObject type.

makalaaneesh: An interface like `DDLObject` is useful when you are operating on that interface type directly (like if a caller calls `DDLObject().GetSchemaName(), but here as far as I can see, all uses of DDLObject first typecast it to the concrete type. So, it doesn't really make sense then to have a DDLObject interface. 

Correct me if my assumption is wrong..

yb-voyager/src/queryparser/ddl_parser.go
```diff
@@ -0,0 +1,861 @@
...
+part_params:{partition_elem:{name:"country_code"  location:310}}  part_params:{partition_elem:{name:"record_type"  location:324}}
+location:290}  oncommit:ONCOMMIT_NOOP}}  stmt_location:178  stmt_len:159}
+
+In case if PRIMARY KEY in column list CREATE TABLE example1 (..., PRIMARY KEY(id,country_code) ) PARTITION BY RANGE (country_code, record_type);
+stmts:{stmt:{create_stmt:{relation:{relname:"example1"  inh:true  relpersistence:"p"  location:15}  table_elts:{column_def:{colname:"id"
+type_name:{names:{string:{sval:"pg_catalog"}}  names:{string:{sval:"numeric"}}  ... table_elts:{constraint:{contype:CONSTR_PRIMARY
+location:98  keys:{string:{sval:"id"}} keys:{string:{sval:"country_code"}}}}  partspec:{strategy:PARTITION_STRATEGY_RANGE
+part_params:{partition_elem:{name:"country_code" location:150}}  part_params:{partition_elem:{name:"record_type"  ...
+*/
+func (p *TableParser) Parse(parseTree *pg_query.ParseResult) (DDLObject, error) {
```
makalaaneesh: General thumb rule is: you wouldn't want to create a struct _just_ because you want to group a bunch of functions together. Only if the operations you do on the struct maintain some state over time, then it makes sense to create a struct. 

IMO, all these can be simple functions  `ParseTable` , `ParseForeignTable` , etc. 
Even the ddl detectors tbh, can be simple functions, but you want them to follow a certain function signature, so it's okay to create an interface, and a struct that follows that function signature. 

Similar to my other comment, do we really need `DDLParser`? If the detectors directly do the parsing, would we still need this?

sanyamsinghal: Agree to Aneesh's point. Having simple functions makes the code simpler.
But interfaces does help in enforcing the functions signatures, and provide some generalisation in code.

I was also thinking to suggest to avoid interface for DDLParser, maybe keep it for DDLIssueDetector.
See if that is easily possible to simplify that, otherwise interface is also fine.

priyanshi-yb: My Idea was to have the `DDLParser` interface and not simple functions for each ddl type is that we can utilize the DDLParser interface in other areas of code for example `export-schema/ import-schema` where if we want to parse a particular ddl type and get the object with information to modify it and also add some `Deparse()` function to get modified DDL. 
So with that, I feel we can keep the logic of figuring out the type of parser based on node type in `queryparser` only

yb-voyager/src/queryissue/unsupported_ddls.go
priyanshi-yb: Yes currently, not using it like that when I am calling the `DDLObject.GetObjectName()` because wherever I am using the DDLObject I have to typecast it anyway to the specific type for other information of that type, but as I said we can utilize this in other areas where we could utilize this and get the object names directly from the DDLObject for various DDL types. 
One such instance, I added one function for getting objectName from the parser for INDEX, TRIGGER, and POLICY for the schema summary of analyze/assess so instead of that function we could directly get the objectName with this interface.
```
obj := ParseAndProcessDDL(ddl)
obj.GetObjectName()
```

yb-voyager/src/queryissue/unsupported_ddls.go
```diff
@@ -0,0 +1,617 @@
...
+//=============TABLE ISSUE DETECTOR ===========================
+
+// TableIssueDetector handles detection of table-related issues
+type TableIssueDetector struct {
+	primaryConsInAlter                   map[string]*queryparser.AlterTable
+	columnsWithUnsupportedIndexDatatypes map[string]map[string]string
+	compositeTypes                       []string
+	enumTypes                            []string
+}
+
```
sanyamsinghal, line 45:
```

type TableIssueDetector struct {
	ParserIssueDetector
}
```
embed the ParseIssueDetector struct here

yb-voyager/src/queryparser/helpers_protomsg.go
```diff
@@ -334,3 +334,11 @@ func getOneofActiveField(msg protoreflect.Message, oneofName string) protoreflec
 	// Determine which field within the oneof is set
 	return msg.WhichOneof(oneofDescriptor)
 }
+
+func GetStatementType(msg protoreflect.Message) string {
+	nodeMsg := getOneofActiveField(msg, "node")
+	if nodeMsg == nil {
+		return ""
+	}
+	return string(nodeMsg.Message().FullName())
```
sanyamsinghal, line 343: refer other code at other place on how to fetch the message out of a field

yb-voyager/src/queryparser/query_parser.go
```diff
@@ -43,197 +40,21 @@ func ParsePLPGSQLToJson(query string) (string, error) {
...
-	viewNode, isViewStmt := getCreateViewNode(parseTree)
-	createAsNode, _ := getCreateTableAsStmtNode(parseTree) //For MVIEW case
-	var selectStmt *pg_query.SelectStmt
-	if isViewStmt {
-		selectStmt = viewNode.ViewStmt.GetQuery().GetSelectStmt()
-	} else {
-		selectStmt = createAsNode.CreateTableAsStmt.GetQuery().GetSelectStmt()
-	}
-	selectStmtQuery, err := DeparseSelectStmt(selectStmt)
+func ParseAndProcessDDL(query string) (DDLObject, error) {
```
sanyamsinghal, line 43: break it into ProcessDDL(), call Parse() separately in caller function

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -378,1090 +334,21 @@ func checkStmtsUsingParser(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
-		here role_spec of each roles is managing the roles related information in a POLICY DDL if any, so we can just check if there is
-		a role name available in it which means there is a role associated with this DDL. Hence report it.
-
-	*/
-	for _, role := range roles {
-		roleName := role.GetRoleSpec().GetRolename() // only in case there is role associated with a policy it will error out in schema migration
-		if roleName != "" {
-			//this means there is some role or grants used in this Policy, so detecting it
-			roleNames = append(roleNames, roleName)
+		err = parserIssueDetector.ParseRequiredDDLs(sqlStmtInfo.formattedStmt)
```
sanyamsinghal, line 337: What you can do here is:
1. Store the parse tree from line 328
2. Pass the parse tree directly to ParseRequiredDDLs()
3. Internally, Change ParseAndProcessDDL() -> ProcessDDL()
4. Also rename ParseRequiredDDLs() -> ProcessDDLs() or ProcessRequiredDDLs()

sanyamsinghal (approved):
LGTM, thanks for this big refactoring work, code is in much better state from before.

Please take care of the new review comments before merge.

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: I would not like to introduce the `pg_query` library in the `queryissue` pkg because we already have the `queryparser`, so will call the Parse in the `ParseRequiredDDLs`.

PR #2021, opened 2024-12-02
Modifications in cpan modules installation in the airgapped installer script

ShivanshGahlot:
- In the installation of cpan modules, we now check if the package is already installed with the required version. If so, we skip the installation using our packages. Moreover, if we are installing the cpan module, we do a final sanity check to ensure that the module was correctly installed.
- Removed the individual package names from cpan modules installation output. Only a single line is printed now:
`Installing cpan modules...`
- Removed cpan modules as a prerequisite in the help message of the airgapped installer script.

### DIFF
--- a/installer_scripts/install-voyager-airgapped.sh
+++ b/installer_scripts/install-voyager-airgapped.sh
@@ -150,8 +150,12 @@ install_perl_module() {
     local requirement_type="$2"
     local required_version="$3"
     local package="$4"
-    
-    echo "Installing module $module_name..."
+
+    # Check if the module is already installed and meets the version requirements
+    check_perl_module_version "$module_name" "$requirement_type" "$required_version" "true"
+    if [[ $? -eq 0 ]]; then
+        return
+    fi
     
     # Extract the package
     tar -xzvf "$package" 1>&2 || { echo "Error: Failed to extract $package"; exit 1; }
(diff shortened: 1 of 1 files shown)


installer_scripts/install-voyager-airgapped.sh
```diff
@@ -150,8 +150,12 @@ install_perl_module() {
...
-    echo "Installing module $module_name..."
+
+    # Check if the module is already installed and meets the version requirements
+    check_perl_module_version "$module_name" "$requirement_type" "$required_version" "true"
+    if [[ $? -eq 0 ]]; then
+        return
+    fi
     
     # Extract the package
     tar -xzvf "$package" 1>&2 || { echo "Error: Failed to extract $package"; exit 1; }
```
priyanshi-yb, line 161: What happens if a module is already installed with a non-required version? How does our installation work? Will it update it to the required version?

installer_scripts/install-voyager-airgapped.sh
```diff
@@ -183,35 +187,56 @@ install_perl_module() {
     # Return to the original directory
     cd ..
     
-    # Verification and version check
+    # Verification of the installed module
+    check_perl_module_version "$module_name" "$requirement_type" "$required_version" "false"
```
priyanshi-yb, line 191: do we have such verification of other packages we install such as voyager, ora2pg etc..?

priyanshi-yb (approved):
LGTM with a question

installer_scripts/install-voyager-airgapped.sh
ShivanshGahlot: Hey so it will overwrite any packages that are installed at the expected locations i.e. the paths related to perl etc. However, there might be an edge case where the user has installed modules at an unusual location. Then these commands won't replace the already installed packages. This is an expected edge case in any form of installation I believe.

PR #2022, opened 2024-12-02
Added check to ensure archive changes command is not run before the streaming phase of export data has started

ShivanshGahlot:
Before running archives changes command that following checks are done:
- Code firstly checks if meta.db is present or not
- If the the first check passes then it checks whether `msr.ExportDataSourceDebeziumStarted` is true or not

If either of the two checks fail, we exit with an appropriate message.

### DIFF
--- a/yb-voyager/cmd/archiveChangesCommand.go
+++ b/yb-voyager/cmd/archiveChangesCommand.go
@@ -23,6 +23,7 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


sanyamsinghal (commented):
Code seems a bit redundant.

- How about having a separate function - `IsExportDirInitialised()`
- And call it from `root.go` based on the command names(prepare a list).

PR #2033, opened 2024-12-04
Sample Issue Tests against YB Versions

makalaaneesh:
Adding unit tests for issues.

Workflow that will help us detect when issues are fixed in YB:
1. Whenever we introduce a new issue, write a unit test asserting that it fails.
2. The unit test will be run against all the latest YB versions in every series that we support (2.18,2.20,2.23,2024.1,2024.2 for example)
3. YB fixes an issue in 2025.1.0.0. 
4. As part of qualifying voyager for 2025.1 release, all unit tests are also run against 2025.1. Since the issue was fixed, "assert that it fails" will fail. 
5. This prompts us to a. Add `MinimumVersionsFixedIn[2025.1] = 2025.1.0.0` to the issue. b. Write a unit test that asserts that  `parserDetector.GetAllIssues(..., 2025.1)` does not return an issue of that type. 
6. The failing unit test will now start passing because it takes into account `MinimumVersionsFixedIn[2025.1]`. (this part is not implemented yet)

N/A

N/A. Adding unit tests.


yb-voyager/src/queryissue/ddl_issues_test.go
```diff
@@ -0,0 +1,112 @@
...
+	success = t.Run(fmt.Sprintf("%s-%s", "xml functions", ybVersion), testXMLFunctionIssue)
+	assert.True(t, success)
+
+	success = t.Run(fmt.Sprintf("%s-%s", "stored generated functions", ybVersion), testStoredGeneratedFunctionsIssue)
+	assert.True(t, success)
+
+	success = t.Run(fmt.Sprintf("%s-%s", "unlogged table", ybVersion), testUnloggedTableIssue)
+	assert.True(t, success)
+
+	yugabytedbContainer.Terminate(context.Background())
```
sanyamsinghal, line 110: you already have defer terminate

sanyamsinghal (commented):
There are two approaches to figure the supporting versions for an issue:
1. For every issue check the support on all version till date.
2. For every issue check the support on the latest version, if it is supported then we can look at some other versions(or last few series)

Is that correct? 

Our code can have some stale issues also, which be fixed long time back or not very recently.

So how do we plan to ensure that for an issue if it works on latest, then what was the version yugabytedb started supporting it.

makalaaneesh:
> Our code can have some stale issues also, which be fixed long time back or not very recently.
So how do we plan to ensure that for an issue if it works on latest, then what was the version yugabytedb started supporting it.

(Not sure I understand your question completely, we can discuss on call)
These unit tests will only help in detecting cases where issues are fixed **going forward**. For the ones that are already supported, we will have to do the manual work of figuring out which version was support added in . (Although I don't think there are a lot of cases)

.github/workflows/issue-tests.yml
```diff
@@ -0,0 +1,50 @@
...
+        # using s3 release instead of docker image to allow testing against un-released versions
+        wget https://s3.us-west-2.amazonaws.com/releases.yugabyte.com/${YB_VERSION}/yugabyte-${YB_VERSION}-centos-x86_64.tar.gz
+        mkdir -p yugabyte-${YB_VERSION}
+        tar -xvzf yugabyte-${YB_VERSION}-centos-x86_64.tar.gz -C yugabyte-${YB_VERSION} --strip-components=1
+        yugabyte-${YB_VERSION}/bin/yugabyted start --advertise_address 127.0.0.1
+        sleep 20
+    
+    - name: Test YugabyteDB connection
+      run: |
+        psql "postgresql://yugabyte:@127.0.0.1:5433/yugabyte" -c "SELECT version();"
```
priyanshi-yb, line 39: nit: `psql $YB_CONN_STR -c "SELECT version();"`

priyanshi-yb (approved):
LGTM

PR #2036, opened 2024-12-04
Create NOT VALID constraints only in post-snapshot-import mode with import-schema

priyanshi-yb:
Fixing an issue where presence of NOT VALID constraints in schema can `constraint violation` errors in import-data.
https://yugabyte.atlassian.net/browse/DB-13881

Fix: Creating the NOT VALID constraints during post-snapshot-import mode when the data is already loaded which will not led to any `constraint violation` errors.

No, its like deferring the execution of a DDL after import-data

Tested PG and Oracle both. Added tests for both
Jenkins for Oracle - 
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/4144


yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -90,6 +113,20 @@ func executeSqlFile(file string, objType string, skipFn func(string, string) boo
 	return nil
 }
 
+func shouldSkipDDL(stmt string) (bool, error) {
+	skipReplicaIdentity := strings.Contains(stmt, "ALTER TABLE") && strings.Contains(stmt, "REPLICA IDENTITY")
```
sanyamsinghal, line 117: you can return here if this comes out to be true. No need for further check, which is a bit heavy involves parsing and all...

yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -90,6 +113,20 @@ func executeSqlFile(file string, objType string, skipFn func(string, string) boo
...
+	isNotValid, err := isNotValidConstraint(stmt)
+	if err != nil {
+		return false, fmt.Errorf("error checking whether stmt is to add not valid constraint: %v", err)
+	}
+	if (isNotValid && !bool(flagPostSnapshotImport)) || // Skipping NOT VALID CONSTRAINT in import schema without post-snapshot-mode
+		(bool(flagPostSnapshotImport) && !isNotValid) || // Skipping other TABLE DDLs in post-snapshot-import mode
+		skipReplicaIdentity { // Skip ALTER REPLICA IDENTITY always
+		return true, nil
+	}
```
sanyamsinghal, line 126: lets define boolean vars here for readability
```
skipNotValidConstraint := (isNotValid && !bool(flagPostSnapshotImport))
// similarly for other

if skipNotValidConstraint || ... {
        return true, nil...
}


```

sanyamsinghal: I also think the scenario based skip logic should go to the skipFunc() passed here.

In this function executeSql() - we were simply either using skipFn() to check  the sql, or outright ignoring if SQL is of specific syntax like `SET` or `REPLICA IDENTITY`

priyanshi-yb: That SkipFn is currently only for skipping DDLs in single execution to later, because we run this function twice once with executeSQL(skipFn) and another for executeSQL(!skipFn) 
but for these REPLICA IDENTIY we want to ignore them all the time

sanyamsinghal (approved):

PR #2053, opened 2024-12-10
Introduce TestContainer interface to streamline test setup of supported databases

sanyamsinghal:
- Any package in voyager code, for eg `srcdb`, `tgtdb`, `connpool`, `yugabyted` etc.. can spin up the required source db(oracle, pg, mysql) with a specific version to write unit tests.

-  Every call to `TestContainer.Start()` check if a required database(dbtype + version) is already run then reuse it, otherwise start a new one

- In every package add a `TestMain()` function, to setup the environment needed specifically for that test.

- Added a new Github actions workflow to run the integration tests separately(required some oracle instance client libraries). 
Moving forward it will be good to have separation between go tests as `unit` or `integration`. 

Challenges/Limitations:
1. Currently we are not calling `TestContainer.TerminateAllContainers()`, right now all the containers gets terminated at the end, but we should also call that. Challenge was i couldn't find the right point in code where i can call that. 

2. In Golang, for each packages separate new process is spawned to run its tests, hence the containerRegistry is not shared across all the packages due to separate processes, hence with the current approach it is **not possible to reuse the same containers across all packages but across tests in a single package.**
- One solution is maintain the containerRegistry info in some json file(act as a registry) which can be reused by all processes

NA

- All the existing unit tests are working as expected.


yb-voyager/src/srcdb/main_test.go
```diff
@@ -0,0 +1,165 @@
...
+			Port:      port,
+			SSLMode:   "disable",
+		},
+	}
+
+	err = testMySQLSource.DB().Connect()
+	if err != nil {
+		utils.ErrExit("Failed to connect to mysql database: %w", err)
+	}
+	defer testMySQLSource.DB().Disconnect()
```
makalaaneesh, line 128: nit: since you're only connecting to test that the connections are working, you don't really need the "defer", you can disconnect at this point itself.

yb-voyager/src/srcdb/mysql_test.go
```diff
@@ -0,0 +1,52 @@
...
+	testutils "github.com/yugabyte/yb-voyager/yb-voyager/test/utils"
+	"gotest.tools/assert"
+)
+
+func TestMysqlGetAllTableNames(t *testing.T) {
+	sqlname.SourceDBType = "mysql"
+
+	// Test GetAllTableNames
+	actualTables := testMySQLSource.DB().GetAllTableNames()
+	expectedTables := []*sqlname.SourceName{
```
makalaaneesh, line 31: Reading this test-case in isolation, it's confusing where these tables come from. I would be in favour of NOT having a default schema that is assumed with testcontainer startup. Don't really see a point of it. 
I think it would be better for each testcase to be self-contained. clearly set up it's schema objects at the beginning of the test, run its tests, delete/drop the schema objects. 

Another reason to avoid a single common schema is that over time, we will end up writing all the DDLs for our tests into that schema file itself, which beats the point of having self-container "unit" tests.

yb-voyager/test/containers/testcontainers.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+	case MYSQL:
+		testContainer = &MysqlContainer{
+			ContainerConfig: *containerConfig,
+		}
+	default:
+		panic(fmt.Sprintf("unsupported db type '%q' for creating test container\n", dbType))
+	}
+
+	return testContainer
```
makalaaneesh, line 82: we should add to the containerRegistry here, before returning, right?

makalaaneesh (commented):
Thanks for this! left a few comments

yb-voyager/test/containers/testcontainers.go
sanyamsinghal: oh right, my bad.

yb-voyager/src/srcdb/mysql_test.go
sanyamsinghal: Done @makalaaneesh

yb-voyager/src/srcdb/main_test.go
makalaaneesh: Hmm, good point. How is it working now then for each test that uses the test containers?

yb-voyager/src/srcdb/mysql_test.go
makalaaneesh: Nice! `ExecuteSqls` is nice, and using it with defer for cleanup is very useful

yb-voyager/src/srcdb/main_test.go
sanyamsinghal: > How is it working now then for each test that uses the test containers?

didn't get you exactly, i think all the tests(`srcdb` or `tgtdb` packages) i have added is following this. 
Are there any existing tests which don't follow this but still works?

makalaaneesh: My understanding is that `TestMain` runs before any other test in the package. 
And if `TestMain` has a defer to disconnect the DB, then how is it that in `TestPostgresGetAllTableNames`, you are able to use that container to execute SQLs? It should have been disconnected by then, right?

.github/workflows/issue-tests.yml
```diff
@@ -47,5 +47,4 @@ jobs:
     - name: Test Issues Against YB Version
       run: |
         cd yb-voyager
-        go test -v -run '^(TestDDLIssuesInYBVersion|TestDMLIssuesInYBVersion)$' ./...
-        
+        go test -v -run '^(TestDDLIssuesInYBVersion|TestDMLIssuesInYBVersion)$' ./... -tags 'integration'
```
makalaaneesh, line 50: we should skip `integration` here, right?

yb-voyager/src/srcdb/main_test.go
sanyamsinghal: @makalaaneesh in `TestMain` we have to call `m.Run()` which executes all the tests of that package.
So here defer will be invoked after all the tests for that package has run.
like you can i added `terminateAllContainers` after that.`

I will add one comment in the code for this.

.github/workflows/issue-tests.yml
sanyamsinghal: yes, will change this.

makalaaneesh (approved):
LGTM!

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,53 @@
...
+      with:
+        go-version: "1.23.1"
+
+    - name: Build
+      run: |
+        cd yb-voyager
+        go build -v ./...
+
+    # required by godror driver used in the tests
+    - name: Install Oracle Instant Clients
```
priyanshi-yb, line 32: Q: any specific reason we are not using the installer script here?

yb-voyager/src/srcdb/postgres.go
```diff
@@ -940,6 +940,7 @@ var PG_QUERY_TO_CHECK_IF_TABLE_HAS_PK = `SELECT nspname AS schema_name, relname
 FROM pg_class c
 LEFT JOIN pg_namespace n ON n.oid = c.relnamespace
 LEFT JOIN pg_constraint con ON con.conrelid = c.oid AND con.contype = 'p'
+WHERE c.relkind = 'r'  -- Only consider ordinary tables
```
priyanshi-yb, line 943: why do we need to do this, we should report partitioned tables not having pk as well right?

.github/workflows/integration-tests.yml
sanyamsinghal: we can add installer script later if required. also modified the installer script to skip debezium in a separate PR.

yb-voyager/src/srcdb/postgres.go
sanyamsinghal: updated to have `p` and `r` both. Previously it was also considering objects like sequences since they also don't have PK on them.

PR #2061, opened 2024-12-10
Reporting the Advisory locks, XML Functions and System columns on the DDLs in the unsupported features in assess/analyze

priyanshi-yb:

- Added three more features to `Unsupported features` in the reports Advisory locks, XML functions and System columns in the DDL.
- Moved the MVIEW/VIEW issues from Unsupported PL/pgSQL objects to Unsupported Features.
- Minor format change of constraint reporting for PK/Unique on complex datatypes issue
- Added unit tests for `GetDDLIssues()` in the `parser_issue_detector_test.go`

Fixes https://github.com/yugabyte/yb-voyager/issues/2025

Sample Assessment report - 
Screenshot 2024-12-11 at 12 21 39 PM

Sample Analyze report - 
Screenshot 2024-12-11 at 12 23 24 PM

Screenshot 2024-12-11 at 12 23 49 PM

Added new features section under Unsupported Features 

Existing tests were enough.
Added unit tests


yb-voyager/src/query/queryissue/detectors_ddl.go
```diff
@@ -50,7 +50,7 @@ func (d *TableIssueDetector) DetectIssues(obj queryparser.DDLObject) ([]QueryIss
 	// Check for generated columns
 	if len(table.GeneratedColumns) > 0 {
 		issues = append(issues, NewGeneratedColumnsIssue(
-			TABLE_OBJECT_TYPE,
+			obj.GetObjectType(),
```
makalaaneesh, line 53: a `TableIssueDetector` will always return issues for Table object types, right? 
Is this just a change to not use a constant and use the function, or is there a certain bug/case because of which you had to do this?

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -289,6 +269,22 @@ func (p *ParserIssueDetector) getDDLIssues(query string) ([]QueryIssue, error) {
 			issues[i].SqlStatement = query
 		}
 	}
+
+	if _, ok := ddlObj.(*queryparser.Object); ok { // In case the DDL doesn't have any processor skip checking generic issues
+		return issues, nil
+	}
+
+	genericIssues, err := p.genericIssues(query)
```
makalaaneesh, line 277: Add a comment here explaining why we're making this call for DDLs. Give an example.

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -289,6 +269,22 @@ func (p *ParserIssueDetector) getDDLIssues(query string) ([]QueryIssue, error) {
 			issues[i].SqlStatement = query
 		}
 	}
+
+	if _, ok := ddlObj.(*queryparser.Object); ok { // In case the DDL doesn't have any processor skip checking generic issues
```
makalaaneesh, line 273: a ddlobject is an interface that already adheres to `GetObjectName` and `GetSchemaName`. So why do we need to additionally check if you can type cast it to a `queryparser.Object`. As long as ddlObject is of type `DDLObject interface`, you should be good, no?

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -323,16 +319,6 @@ func (p *ParserIssueDetector) GetPercentTypeSyntaxIssues(query string) ([]QueryI
...
-	issues, err := p.getDMLIssues(query)
-	if err != nil {
-		return issues, err
-	}
-
-	return p.getIssuesNotFixedInTargetDbVersion(issues, targetDbVersion)
-}
-
-// TODO: in future when we will DDL issues detection here we need `GetDDLIssues`
-func (p *ParserIssueDetector) getDMLIssues(query string) ([]QueryIssue, error) {
```
makalaaneesh, line 335: would recommend still keeping the `getDMLIssues` and `GetDMLIssues` both. The idea was to not touch the `GetDMLIssues` often and keep the `getIssuesNotFixedInTargetDbVersion` in it so that it does not get missed. 

If we have just one function, there is a high possibility that we accidentally just write some return from the middle of that function which will lead to  issues not being filtered out..

yb-voyager/src/query/queryissue/parser_issue_detector_test.go
```diff
@@ -198,3 +232,54 @@ func TestAllDDLIssues(t *testing.T) {
...
+		err := parserIssueDetector.ParseRequiredDDLs(stmt)
+		assert.NoError(t, err, "Error parsing required ddl: %s", stmt)
+	}
+	for stmt, expectedIssues := range stmtsWithExpectedIssues {
+		issues, err := parserIssueDetector.GetDDLIssues(stmt, ybversion.LatestStable)
+		assert.NoError(t, err, "Error detecting issues for statement: %s", stmt)
+
+		assert.Equal(t, len(expectedIssues), len(issues), "Mismatch in issue count for statement: %s", stmt)
+		for _, expectedIssue := range expectedIssues {
+			found := slices.ContainsFunc(issues, func(QueryIssue QueryIssue) bool {
```
makalaaneesh, line 275: As discussed, see if we can use cmp.Equal for this.

yb-voyager/src/query/queryparser/helpers_struct.go
```diff
@@ -58,22 +39,6 @@ func IsMviewObject(parseTree *pg_query.ParseResult) bool {
 	return isCreateAsStmt && createAsNode.CreateTableAsStmt.Objtype == pg_query.ObjectType_OBJECT_MATVIEW
 }
 
-func GetSelectStmtQueryFromViewOrMView(parseTree *pg_query.ParseResult) (string, error) {
```
makalaaneesh, line 61: Ah, so we don't really need these anymore because parser can directly parse MVIEW/VIEW, correct?

yb-voyager/src/query/queryissue/detectors_ddl.go
priyanshi-yb: > Is this just a change to not use a constant and use the function,

Yes right, that way I was able to directly use `ddlObj.GetObjectType()` to get the respective type.

yb-voyager/src/query/queryparser/helpers_struct.go
priyanshi-yb: yeah

makalaaneesh:
@priyanshi-yb It would be good to also include details for those issues in the report. 
- advisory locks: which functions were used? 
- XML functions - which functions?
- system columns - which system columns. 

I believe these details are not available right now, am i right?

priyanshi-yb:
> I believe these details are not available right now, am i right?

Yes, currently it is not available we might have to enhance the traversal logic to return all the variety of functions  we found in the statement

yb-voyager/src/query/queryissue/parser_issue_detector.go
priyanshi-yb: So there were reasons to add this check here-
1. To handle the scenario where the stmt coming to getDDLIssues is DML stmt which can be in case of PLPGQueries where we call internal getAllIssues(), so with that generic function is getting called for the select stmt twice giving the duplicate issues. - This can be solved by just unique issues by doing proper cmp.Equal()
2. To handle the scenario where the ObjectType is not known in this code path (the case where DDLObject is not implemented for some DDL types yet) so it might return. the objectType - `OBJECT` (NoOpProcessor). For such case, the analyze code to convertIssue toAnalyzeIssue() where we set invalidCount of objectname based on ObjectTYpe will fail with nil pointer as it doesn't know about it. - This can also be solved by handling it properly in that function.

priyanshi-yb: As discussed,  changedthis condition with isDDL in the starting of getDDLIssues()

yb-voyager/src/query/queryissue/parser_issue_detector_test.go
```diff
@@ -210,6 +219,11 @@ func TestAllIssues(t *testing.T) {
 		},
 	}
 
+	//Should modify it in similar way we do it actual code as the particular DDL issue in plpgsql can have different Details map on the basis of objectType
```
makalaaneesh, line 222: I'm confused. What was wrong with the previous approach where you define issues directly with "FUNCTION", "list_high_earners" ?

priyanshi-yb: example for this cluster on -
```
func NewClusterONIssue(objectType string, objectName string, sqlStatement string) QueryIssue {
	details := map[string]interface{}{}
	//for ALTER AND INDEX both  same struct now how to differentiate which one to not
	if objectType == "TABLE" {
		details["INCREASE_INVALID_COUNT"] = false
	}
	return newQueryIssue(clusterOnIssue, objectType, objectName, sqlStatement, details)
}
``` 
earlier I was using the New function for getting queryissue - 
```
NewClusterONIssue("FUNCTION", "list_high_earners", "ALTER TABLE employees CLUSTER ON idx;")
```
giving this QueryIssue, with details as empty as - 
```
{{CLUSTER_ON ALTER TABLE CLUSTER not supported yet.  Remove it from the exported schema. https://github.com/YugaByte/yugabyte-db/issues/1124 https://docs.yugabyte.com/preview/yugabyte-voyager/known-issues/postgresql/#unsupported-alter-table-ddl-variants-in-source-schema map[]} FUNCTION list_high_earners ALTER TABLE employees CLUSTER ON idx; map[]}
```
But the actual query issue will look like this where details will be populated based no TABLE object type- 
```
{{CLUSTER_ON ALTER TABLE CLUSTER not supported yet.  Remove it from the exported schema. https://github.com/YugaByte/yugabyte-db/issues/1124 https://docs.yugabyte.com/preview/yugabyte-voyager/known-issues/postgresql/#unsupported-alter-table-ddl-variants-in-source-schema map[]} FUNCTION list_high_earners ALTER TABLE employees CLUSTER ON idx; map[INCREASE_INVALID_COUNT:false]}
```

So, to detect these issues in PLPGSQL, we modify the issues later (in getAllPLPGSQLIssues()) to change the objectType and objectName to the actual object name and type of the PLPGSQL object.
So now with cmp.Equal, it also compares `details` map I had to change the way I generate the expected issues to be able to check properly.

makalaaneesh: Orthogonal point: 
`INCREASE_INVALID_COUNT` should not be in queryissue layer. 
It is an analyze-schema detail, we should have some logic of figuring it out at that layer.

makalaaneesh (approved):
LGTM

yb-voyager/src/query/queryissue/parser_issue_detector_test.go
priyanshi-yb: Yeah, I agree. I removed it in this PR with handling all cases for invalid count #2073

PR #2075, opened 2024-12-13
Printing a note if the export dir version is higher than the previous breaking change version but does not match the current version

ShivanshGahlot:

- Printing a note in the scenario when the export dir version(msr version) is greater than the previous  breaking change version of the new voyager binary working on the export dir but the export dir version is not the same as the current version of the new voyager binary.
- Moreover, moved the logic to check the version compatibility from initMetaDB to root.go where we are checking whether metaDB is already present or nor. With the previous implementation, the not was printing twice - every time initMetaDB was called.

Just a note will get printed in the scenario mentioned above which will look like:
image

The PR was tested manually, I sent various values of msrVersion to the function and also modified the current and previous breaking change version

### DIFF
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -480,14 +480,6 @@ func initMetaDB(migrationExportDir string) *metadb.MetaDB {
 		utils.ErrExit("could not init migration status record: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


priyanshi-yb (approved):
LGTM

PR #2094, opened 2024-12-16
Minor refactor to proto detectors

makalaaneesh:
1. Change interface of `UnsupportedConstructDetector` to make each detector responsible for constructing and returning issues. Currently, this was being done in ParserIssueDetector.getGenericIssues, which seems like the wrong place for it. With pg17 features, all the issue construction will be added to this function itself, which I want to avoid.  Detectors do the detection, and they should also construct the issues since they have all the information necessary to do so. 
2. Store functions being detected in `FuncCallDetector` . Will be useful for later reporting.
3. This refactor led to one side-effect related to XML issues. Worked around that for now. 

Existing unit tests suffice. 

### DIFF
--- a/yb-voyager/src/query/queryissue/detectors.go
+++ b/yb-voyager/src/query/queryissue/detectors.go
@@ -16,10 +16,10 @@ limitations under the License.
 package queryissue
 
 import (
+	mapset "github.com/deckarep/golang-set/v2"
 	log "github.com/sirupsen/logrus"
-	"google.golang.org/protobuf/reflect/protoreflect"
-
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/query/queryparser"
+	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


yb-voyager/src/query/queryissue/detectors_test.go
```diff
@@ -141,27 +143,26 @@ func TestColumnRefDetector(t *testing.T) {
...
 			}
-			unsupportedConstructs = append(unsupportedConstructs, constructs...)
 			return nil
 		}
 
 		parseTreeMsg := queryparser.GetProtoMessageFromParseTree(parseResult)
 		err = queryparser.TraverseParseTree(parseTreeMsg, visited, processor)
 		assert.NoError(t, err)
-		assert.Contains(t, unsupportedConstructs, SYSTEM_COLUMNS_NAME, "System Columns not detected in SQL: %s", sql)
+		assert.Equal(t, 1, len(detector.GetIssues()), "Expected 1 issue for SQL: %s", sql)
```
priyanshi-yb, line 164: nit: `issues := detector.GetIssues()` and use issues further no need to call GetIssues() twice

yb-voyager/src/query/queryissue/detectors_test.go
```diff
@@ -613,30 +633,29 @@ func TestCombinationOfDetectors1WithObjectCollector(t *testing.T) {
 		},
 	}
 
-	expectedConstructs := []string{ADVISORY_LOCKS_NAME, SYSTEM_COLUMNS_NAME, XML_FUNCTIONS_NAME}
+	// expectedConstructs := []string{ADVISORY_LOCKS_NAME, SYSTEM_COLUMNS_NAME, XML_FUNCTIONS_NAME}
```
priyanshi-yb, line 636: remove comment

yb-voyager/src/query/queryissue/detectors_test.go
```diff
@@ -530,37 +542,45 @@ RETURNING id,
...
+		var allIssues []QueryIssue
+		for _, detector := range detectors {
+			allIssues = append(allIssues, detector.GetIssues()...)
+		}
+		issueTypesDetected := mapset.NewThreadUnsafeSet[string]()
+		for _, issue := range allIssues {
+			issueTypesDetected.Add(issue.Type)
+		}
+
+		assert.True(t, expectedIssueTypes.Equal(issueTypesDetected), "Expected issue types do not match the detected issue types. Expected: %v, Actual: %v", expectedIssueTypes, issueTypesDetected)
```
priyanshi-yb, line 583: Can we enhance this assert to count of issues for each issue Type?

yb-voyager/src/query/queryissue/detectors_test.go
```diff
@@ -645,7 +664,17 @@ func TestCombinationOfDetectors1WithObjectCollector(t *testing.T) {
...
```
priyanshi-yb, line 677: same as above

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -397,15 +395,26 @@ func (p *ParserIssueDetector) genericIssues(query string) ([]QueryIssue, error)
...
+					// currently, both FuncCallDetector and XmlExprDetector can detect XMLFunctionsIssue
+					// but we want to only return one XMLFunctionsIssue.
+					// TODO: refactor to avoid this
+					// Possible Solutions:
+					// 1. Have a dedicated detector for XMLFunctions and Expressions so that a single issue is returned
+					// 2. Separate issue types for XML Functions and XML expressions.
+					continue
+				} else {
+					xmlIssueAdded = true
+				}
```
priyanshi-yb, line 413: maybe instead you can directly use the `lo.UniqBy` on the `result` based on `issue.Type`

yb-voyager/src/query/queryissue/detectors_test.go
makalaaneesh: So, the thing is: 
Ideally, there should only be 1 issue of each type.
The code to return a single issue for XML functiosn is in parserIssueDetector. So instead of checking the counts here, i added a separate test case `TestSingleXMLIssueIsDetected` to ensure that only one is finally returned. 

Moving forward, I think we should majorly focus on writing tests at the parserIssueDetector.GetIssues() rather than at the detector directly (unless ofcourse there is some corner case we want to specifically test).

yb-voyager/src/query/queryissue/parser_issue_detector.go
makalaaneesh: Don't want to blindly do this. This is just a workaround for xmlfunctions right now. 
If there are multiple issues of the same type, we can't blindly pick 1, we will have to essentially merge the issues (along with the details field).
 
If we encounter more such cases(other than XML) in the future, we need to figure out one of those solutions listed above. Did not want to do that work now as part of this refactor.

priyanshi-yb: > If there are multiple issues of the same type, we can't blindly pick 1, we will have to essentially merge the issues (along with the details field).

Hmm makes sense when we start adding the function names or actual constructs in detail from detector layer.

yb-voyager/src/query/queryissue/parser_issue_detector_test.go
```diff
@@ -304,3 +304,19 @@ func TestUnloggedTableIssueReportedInOlderVersion(t *testing.T) {
 	assert.Equal(t, 1, len(issues))
 	assert.True(t, cmp.Equal(issues[0], NewUnloggedTableIssue("TABLE", "tbl_unlog", stmt)))
 }
+
+func TestSingleXMLIssueIsDetected(t *testing.T) {
```
sanyamsinghal, line 308: nit: small comment explaining the test that the stmt has two xml function/expression but should be reported only once.

yb-voyager/src/query/queryissue/detectors_test.go
```diff
@@ -76,27 +76,31 @@ func TestFuncCallDetector(t *testing.T) {
...
-	detector := NewFuncCallDetector()
 	for _, sql := range advisoryLockSqls {
+		detector := NewFuncCallDetector(sql)
+
 		parseResult, err := queryparser.Parse(sql)
 		assert.NoError(t, err, "Failed to parse SQL: %s", sql)
 
 		visited := make(map[protoreflect.Message]bool)
-		unsupportedConstructs := []string{}
+		// unsupportedConstructs := []string{}
```
sanyamsinghal, line 86: remove

yb-voyager/src/query/queryissue/detectors.go
```diff
@@ -31,86 +31,115 @@ const (
...
+	Detect(msg protoreflect.Message) error
+	GetIssues() []QueryIssue
 }
 
 type FuncCallDetector struct {
+	query string
 	// right now it covers Advisory Locks and XML functions
-	unsupportedFuncs map[string]string
+	advisoryLocksFuncsDetected mapset.Set[string]
+	xmlFuncsDetected           mapset.Set[string]
```
sanyamsinghal, line 42: great, now we are storing info about what all functions of advisory lock or xml are detected, can be used in future.

sanyamsinghal (approved):
LGTM, thanks for this refactoring

PR #2098, opened 2024-12-17
Misc changes to target-db-version flag

makalaaneesh:
- Unhide the flag
- Print the version being used in analyze-schema/ assess-migration
- Prevent users from passing build number in target-db-version. 
   The problem is that 2024.2.0.0-b145 would be considered to be > 2024.2.0.0. Which means we will have two options:
   a. prevent users from passing build number
   b. Always ask users to pass build number, so that it can be compared against the issues' MinimumVersionsFixedIn(which should also have build number.)
   Going with option a. 

Printing to stdout in case of analyze-schema/assess-migration

- added a unit test 
- manual testing

### DIFF
--- a/yb-voyager/cmd/analyzeSchema.go
+++ b/yb-voyager/cmd/analyzeSchema.go
@@ -1261,7 +1261,6 @@ func init() {
 
 	analyzeSchemaCmd.Flags().StringVar(&targetDbVersionStrFlag, "target-db-version", "",
 		fmt.Sprintf("Target YugabyteDB version to analyze schema for. Defaults to latest stable version (%s)", ybversion.LatestStable.String()))
-	analyzeSchemaCmd.Flags().MarkHidden("target-db-version")
 }
 
 func validateReportOutputFormat(validOutputFormats []string, format string) {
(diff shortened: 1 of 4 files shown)


yb-voyager/cmd/analyzeSchema.go
```diff
@@ -1261,7 +1262,6 @@ func init() {
 
 	analyzeSchemaCmd.Flags().StringVar(&targetDbVersionStrFlag, "target-db-version", "",
 		fmt.Sprintf("Target YugabyteDB version to analyze schema for. Defaults to latest stable version (%s)", ybversion.LatestStable.String()))
```
priyanshi-yb, line 1264: > It only supports versions with 4 segments (A.B.C.D). No build number is allowed.


Should we mention this format of version to be provided in the flag?

priyanshi-yb (approved):
LGTM with a comment

yb-voyager/cmd/analyzeSchema.go
makalaaneesh: Makes sense, thanks! done.

PR #2108, opened 2024-12-20
[DB-14231] Report Security Invoker View in analyze and assess migration commands generated reports

sanyamsinghal:

Implement ViewProcessor's  `Process()` method using protomsg approach.
Made use of existing `TraverseParseTree()` function to write a helper func `TraverseAndExtractDefNameFromDefElem()` to extract all defnames in any type of DDL.

Analyze Schema Report
image

Assess Migration Report
image

No

Added Unit tests
Tested manually by generating the reports.
Added end-to-end but disabled.


yb-voyager/src/query/queryparser/ddl_processor.go
```diff
@@ -849,16 +851,52 @@ func (v *ViewProcessor) Process(parseTree *pg_query.ParseResult) (DDLObject, err
...
+		view_stmt:{view:{schemaname:"public" relname:"invoker_view" inh:true relpersistence:"p" location:12}
+		query:{select_stmt:{target_list:{res_target:{val:{column_ref:{fields:{string:{sval:"id"}} location:95}} location:95}}
+		from_clause:{...}
+		where_clause:{...}
+		options:{def_elem:{defname:"security_invoker" arg:{string:{sval:"true"}} defaction:DEFELEM_UNSPEC location:32}}
+		options:{def_elem:{defname:"security_barrier" arg:{string:{sval:"false"}} defaction:DEFELEM_UNSPEC location:57}}
+		with_check_option:NO_CHECK_OPTION}
+	*/
+
+	log.Infof("checking the view '%s' is security invoker view", qualifiedViewName)
```
makalaaneesh, line 868: As per my understanding, the benefit of proto detection comes from writing generic code :
- without worrying about what type of statement it is. For example, be it a view/mview/index/table, you can get the defName of the defElemNodes.
- without worrying about which fields do I have to look at to get the "with options". In case of a view, it seems like `ViewStmt.Options`. (In some others that I've checked it also seems to be the `Options` field)

Therefore, I see the following approaches:
1. Generic proto detection approach for defElemNodes: How about writing a generic proto detector of defElem nodes? DefElem seems like the generic option definition node. The current logic what you've written can be re-used for other statement types as well in future. (for example, to detect a particular option being unsupported in CREATE TABLE).
 https://github.com/postgres/postgres/blob/578a7fe7b6f8484f6d7caa2fda288abb3fe87aa0/src/include/nodes/parsenodes.h#L795 
 

3. Write a generic proto ViewStmtNodeDetector. All issues pertaining to views can live there. This would be in parallel to the ViewProcessor that we already have today.

4. In the current ddl processor, use the viewStmt node directly. I personally see this also as a viable option because this is exactly what we would have had to do if we wrote a ViewStmtNodeDetector (option 2 above);
```
	securityInvokerView := false
	for _, node := range viewNode.ViewStmt.Options {
		defElemNode := node.GetDefElem()
		if defElemNode != nil {
			if defElemNode.GetDefname() == "security_invoker" {
				securityInvokerView = true
				break
			}
		}
	}
```


I guess my main point is that we aren't gaining much by writing a view-specific proto detection logic. Either go with option 1(write generic proto detection logic), or option 2/3 (view-specific logic on ViewStmt node.)

yb-voyager/src/query/queryparser/helpers_protomsg.go
```diff
@@ -328,6 +328,21 @@ func GetStatementType(msg protoreflect.Message) string {
 	return GetMsgFullName(node)
 }
 
+func getOneofActiveNode(msg protoreflect.Message) protoreflect.Message {
```
makalaaneesh, line 331: not really using this?

yb-voyager/src/query/queryparser/ddl_processor.go
sanyamsinghal: Couple of points:
1. I think moving forward we wanted to have mix of both(protomsg and struct approach), where something is directly available in the struct use it, and if it requires some lookup or wherever generic code is required then use protomsg approach.
So here going with the option 2/3 does not to me aligning with that, it again pulls us to forcing struct approach.

2. I didn't get exactly why do we need to introduce something like `ViewStmtNodeDetector`? What i understood from implementation in `detectors_ddl.go` that we will specific detectors for each object type like we have `ViewIssueDetector`, and enhance that to detect all issues related to views.

3. Regarding - `Generic proto detection approach for defElemNodes` , i did think for adding this piece of code so that its reusable, since the `DefElem` node can be present at other places like `CREATE TABLE ... WITH (fillfactor=70)...` but couldn't come up how/where in existing code.
Seems like you are suggesting it placing parallel to `FuncCallDetector`, `ColumnDef` detectors. 
The problem is that the `security invoker` is just a parameter, i am not sure if it can be present with other nodes also, or just specific to VIEWs. (For example: [functions definition have security invoker](https://www.postgresql.org/docs/current/sql-createfunction.html#:~:text=INPUT%20%7C%20STRICT%20%7D%0A%20%20%20%20%7C%20%7B%20%5B%20EXTERNAL%20%5D-,SECURITY%20INVOKER,-%7C%20%5B%20EXTERNAL%20%5D%20SECURITY%20DEFINER))

sanyamsinghal: going with `Generic proto detection approach for defElemNodes` as discussed

yb-voyager/src/query/queryparser/helpers_protomsg.go
```diff
@@ -370,3 +385,21 @@ func GetSchemaAndObjectName(nameList protoreflect.List) (string, string) {
...
+			return nil
+		}
+
+		fieldVal := GetStringField(msg, "defname")
+		defNames = append(defNames, fieldVal)
+		return nil
+	}
+	visited := make(map[protoreflect.Message]bool)
+	TraverseParseTree(msg, visited, collectorFunc)
+
```
priyanshi-yb, line 403: You can also write a detector similar to FuncCallDetector, etc., for this. Instead of storing all the defNames, report unsupported def names, e.g., `security_invoker`, there itself?
This helps to keep the generic feature detection with Traverse in a single place.

sanyamsinghal: Discussed that with Aneesh, detection can't be just based on DefElem, since there are chances of same parameters like `security_invoker`, `on_error` etc being available with multiple DDLs.
So for making this into detector, we will also need to check objectType.
Refer - https://github.com/yugabyte/yb-voyager/pull/2108#:~:text=Regarding%20%2D%20Generic,security%20invoker)

So lets defer it for now, we can add if later we find it worth doing that.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1014,6 +1015,7 @@ func fetchUnsupportedPGFeaturesFromSchemaReport(schemaAnalysisReport utils.Schem
 	unsupportedFeatures = append(unsupportedFeatures, getUnsupportedFeaturesFromSchemaAnalysisReport(queryissue.SYSTEM_COLUMNS_NAME, queryissue.SYSTEM_COLUMNS_NAME, "", schemaAnalysisReport, false, ""))
 	unsupportedFeatures = append(unsupportedFeatures, getUnsupportedFeaturesFromSchemaAnalysisReport(queryissue.LARGE_OBJECT_FUNCTIONS_NAME, queryissue.LARGE_OBJECT_FUNCTIONS_NAME, "", schemaAnalysisReport, false, ""))
 	unsupportedFeatures = append(unsupportedFeatures, getUnsupportedFeaturesFromSchemaAnalysisReport(REGEX_FUNCTIONS_FEATURE, "", queryissue.REGEX_FUNCTIONS, schemaAnalysisReport, false, ""))
+	unsupportedFeatures = append(unsupportedFeatures, getUnsupportedFeaturesFromSchemaAnalysisReport(queryissue.SECURITY_INVOKER_VIEWS_NAME, SECURITY_INVOKER_VIEWS_ISSUE, "", schemaAnalysisReport, false, ""))
```
makalaaneesh, line 1018: Let's filter by the type, not the type name? It's not great to be filtering on text that is used for display purposes, we should be checking an enum value.

(See the above call for regex functions as an example)

yb-voyager/src/query/queryparser/ddl_processor.go
```diff
@@ -849,16 +850,40 @@ func (v *ViewProcessor) Process(parseTree *pg_query.ParseResult) (DDLObject, err
...
+		from_clause:{...}
+		where_clause:{...}
+		options:{def_elem:{defname:"security_invoker" arg:{string:{sval:"true"}} defaction:DEFELEM_UNSPEC location:32}}
+		options:{def_elem:{defname:"security_barrier" arg:{string:{sval:"false"}} defaction:DEFELEM_UNSPEC location:57}}
+		with_check_option:NO_CHECK_OPTION}
+	*/
+	log.Infof("checking the view '%s' is security invoker view", qualifiedViewName)
+	msg := GetProtoMessageFromParseTree(parseTree)
+	defNames := TraverseAndExtractDefNamesFromDefElem(msg)
+	securityInvokerView := false
```
makalaaneesh, line 870: nit: `securityInvokerView := slices.Contains(defNames, "security_invoker")`

yb-voyager/src/query/queryparser/helpers_protomsg.go
```diff
@@ -370,3 +385,21 @@ func GetSchemaAndObjectName(nameList protoreflect.List) (string, string) {
 	}
 	return schemaName, objectName
 }
+
+func TraverseAndExtractDefNamesFromDefElem(msg protoreflect.Message) []string {
+	var defNames []string
```
makalaaneesh, line 390: just a note for future: we might need the values as well.

makalaaneesh (approved):
LGTM with a couple of comments. Pls address the filter on type one before merging.

yb-voyager/src/query/queryparser/helpers_protomsg.go
sanyamsinghal: added that comment in code

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: i agree but if i use that facing issue that the issue is not getting reported.
Also i see there are other places also in this function which needs this change, lets do them together.

sanyamsinghal: done

PR #2109, opened 2024-12-20
Add --skip-debezium flag in installer script to avoid installing debezium when only offline setup is required

sanyamsinghal:
Added a new flag to installer script `--skip-debezium` to avoid installing debezium for cases like when only offline setup is required

```
./installer_scripts/install-yb-voyager  --install-from-local-source --skip-debezium
```

If user doesn't provide this new flag(old way of invoking installler script) there won't be any change in behaviour for them.

Existing tests ensure that nothing is getting affected with this new flag, unless its specified

### DIFF
--- a/installer_scripts/install-yb-voyager
+++ b/installer_scripts/install-yb-voyager
@@ -543,9 +543,9 @@ rebuild_voyager_local() {
 get_passed_options() {
 	if [ "$1" == "linux" ]
 	then
-		OPTS=$(getopt -o "lpnvV", --long install-from-local-source,only-pg-support,skip-debezium,rebuild-voyager-local,version: --name 'install-yb-voyager' -- $ARGS_LINUX)
+		OPTS=$(getopt -o "lpdvV", --long install-from-local-source,only-pg-support,skip-debezium,rebuild-voyager-local,version: --name 'install-yb-voyager' -- $ARGS_LINUX)
 	else
-		OPTS=$(getopt  lpvV  $ARGS_MACOS)
+		OPTS=$(getopt  lpdvV  $ARGS_MACOS)
 	fi
 
 	eval set -- "$OPTS"
@@ -560,11 +560,11 @@ get_passed_options() {
 				ONLY_PG="true"; 
 				shift 
 				;;
-			--skip-debezium )
+			-d | --skip-debezium )
 				SKIP_DEBEZIUM="true"
 				shift
 				;;
-			--rebuild-voyager-local )
+			-v | --rebuild-voyager-local )
 				REBUILD_VOYAGER_LOCAL="true";
 				shift
 				;;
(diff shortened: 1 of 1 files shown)


installer_scripts/install-yb-voyager
```diff
@@ -555,6 +560,10 @@ get_passed_options() {
 				ONLY_PG="true"; 
 				shift 
 				;;
+			-n | --no-debezium )
```
makalaaneesh, line 563: This is a subjective thing but how about "skip-debezium" or "exclude-debezium" instead? "no-" prefix feels slightly weird imo in the context of installation.   
Also `-n` is a bit confusing. (Same with `-v` that we currently have as well... ). What if we want to add "no-ora2pg" or "skip-ora2pg" in the future?
Can you explore if we can have two-letter short-forms?

sanyamsinghal: `skip-debezium` sounds good. let me explore `two-letter short-forms`

sanyamsinghal: Doesn't look like that is possible.
```
For example, -ab is interpreted as two short options: -a and -b.
```

But i think what we can do is just have long forms for some of the options.

makalaaneesh: I think on mac, long-form options don't work. Let's just choose some letter then in that case. maybe "-d"  for skip-debeizum?

sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/4299/

installer_scripts/install-yb-voyager
```diff
@@ -555,7 +560,11 @@ get_passed_options() {
 				ONLY_PG="true"; 
 				shift 
 				;;
-			-v | --rebuild-voyager-local )
+			--skip-debezium )
```
makalaaneesh, line 563: let's add -v back and -d for skip-debezium.

installer_scripts/install-yb-voyager
```diff
@@ -538,7 +543,7 @@ rebuild_voyager_local() {
 get_passed_options() {
 	if [ "$1" == "linux" ]
 	then
-		OPTS=$(getopt -o "lpvV", --long install-from-local-source,only-pg-support,rebuild-voyager-local,version: --name 'install-yb-voyager' -- $ARGS_LINUX)
+		OPTS=$(getopt -o "lpnvV", --long install-from-local-source,only-pg-support,skip-debezium,rebuild-voyager-local,version: --name 'install-yb-voyager' -- $ARGS_LINUX)
```
makalaaneesh, line 546: lpvsV?

makalaaneesh (approved):
LGTM with couple of comments

installer_scripts/install-yb-voyager
sanyamsinghal: yes, `lpdvV` where -d is for skip debezium

PR #2110, opened 2024-12-20
Do not add the feature in UnsupportedFeatures field if there are no objects for it in the JSON report

priyanshi-yb:
Modified the code to not put the Feature in the UnsupportedFeatures field of the struct in case there are no objects for making JSON report cleaner. This will also help developers not to update the expected files every time we add new feature.

Json report will not have feature if there are no objects

Existing tests
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/4300/ - 1 intermittent failure
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/4301


yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -947,7 +947,7 @@ func areMinVersionsFixedInEqual(m1 map[string]*ybversion.YBVersion, m2 map[strin
 	return true
 }
 
-func getUnsupportedFeaturesFromSchemaAnalysisReport(featureName string, issueReason string, schemaAnalysisReport utils.SchemaReport, displayDDLInHTML bool, description string) UnsupportedFeature {
+func getUnsupportedFeaturesFromSchemaAnalysisReport(featureName string, issueReason string, schemaAnalysisReport utils.SchemaReport, displayDDLInHTML bool, description string) *UnsupportedFeature {
```
sanyamsinghal, line 950: Instead of changing it to pointer, how about we keep the return type same.
And define EMPTY_UNSUPPORTED_FEATURE and return that, and later filter that out of the final array.

```
var EMPTY_UNSUPPORTED_FEATURE = UnsupportedFeature{}
```

priyanshi-yb: Changed the way to make pointer changes, but just filter at the end if there are objects

migtests/tests/oracle/bulk-assessment-test/expected_reports/expectedChild2AssessmentReport.json
sanyamsinghal, line 1: Just thinking out loud after looking at the expected test results.
In case of UnsupportedFeatures it is null if there is nothing,
Whereas in case of migration caveats its a empty list - `[]`

can we have the same behaviour in both cases

sanyamsinghal (approved):

PR #2130, opened 2024-12-30
Make voyager import-data metadata tables columns as TEXT

makalaaneesh:
Currently, voyager metadata tables have columns of type VARCHAR(250). In one of our tests, the table name + export-dir turned out to be longer than 250, thereby leading to `data_file_name` field exceeding the 250 limit and therefore preventing import-data from running successfully. There's no real reason to limit any of the fields to 250. 
There isn't any documented difference in performance between the VARCHAR(n) and TEXT, [even in postgres](https://www.postgresql.org/docs/current/datatype-character.html#:~:text=There%20is%20no,be%20used%20instead.)

Going with ALTER TABLE ALTER COLUMN TYPE so as to ensure that the type is changed even when a user is using voyager on a cluster where voyager was previously used. 

ALTER table from VARCHAR to TEXT  is a safe operation. It does not lead to a table-rewrite and is an idempotent operation. 

Existing tests suffice.

### DIFF
--- a/yb-voyager/src/tgtdb/yugabytedb.go
+++ b/yb-voyager/src/tgtdb/yugabytedb.go
@@ -343,6 +343,9 @@ func (yb *TargetYugabyteDB) CreateVoyagerSchema() error {
 			rows_imported BIGINT,
 			PRIMARY KEY (migration_uuid, data_file_name, batch_number, schema_name, table_name)
 		);`, BATCH_METADATA_TABLE_NAME),
+		fmt.Sprintf(`ALTER TABLE %s ALTER COLUMN data_file_name TYPE TEXT;`, BATCH_METADATA_TABLE_NAME),
+		fmt.Sprintf(`ALTER TABLE %s ALTER COLUMN schema_name TYPE TEXT;`, BATCH_METADATA_TABLE_NAME),
+		fmt.Sprintf(`ALTER TABLE %s ALTER COLUMN table_name TYPE TEXT;`, BATCH_METADATA_TABLE_NAME),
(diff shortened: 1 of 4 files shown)


yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -343,6 +343,9 @@ func (yb *TargetYugabyteDB) CreateVoyagerSchema() error {
 			rows_imported BIGINT,
 			PRIMARY KEY (migration_uuid, data_file_name, batch_number, schema_name, table_name)
 		);`, BATCH_METADATA_TABLE_NAME),
+		fmt.Sprintf(`ALTER TABLE %s ALTER COLUMN data_file_name TYPE TEXT;`, BATCH_METADATA_TABLE_NAME),
```
priyanshi-yb, line 346: Should we do it for PG fall-forward/back scenario
And for Oracle, though we don't those creation steps in code but for docs as well https://docs.yugabyte.com/preview/yugabyte-voyager/migrate/live-fall-forward/#prepare-source-replica-database

makalaaneesh: Completely forgot about source-replica, thanks :) 
Made the change for PG. Noted that we need to change docs for oracle. 
Remind me- why don't we do this automatically in the case of oracle?

priyanshi-yb: > Remind me- why don't we do this automatically in the case of oracle?

Because of permissions, oracle requires ADMIN privilege; hence, we ask users to create manually with privileged users.

priyanshi-yb (approved):
LGTM

PR #2156, opened 2025-01-06
Fix: MAP value converter for data via debezium

priyanshi-yb:
1. Fix `hstore` datatype value converter (MAP) to convert values properly from debezium in debezium exporter.
2. Report the store datatype in `Unsupported datatype for live migration with fall-forward/fall-back` as gRPC connector doesn't support it yet. 
https://yugabyte.atlassian.net/browse/DB-14753

No

end-to-end tests


yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -173,17 +173,31 @@ var YBValueConverterSuite = map[string]ConverterFn{
...
+
+		// Create a map to store the parsed key-value pairs
 		var transformedMapValue string
-		for key, value := range mapValue {
-			transformedMapValue = transformedMapValue + fmt.Sprintf("\"%s\"=>\"%s\",", key, value)
+
+		// Iterate over each pair
+		for _, pair := range pairs {
+			// Split the pair by '=' to separate key and value
+			kv := strings.SplitN(pair, "=", 2)
```
makalaaneesh, line 190: What happens if "=" is part of the key/value string? Is that even possible?

yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -173,17 +173,31 @@ var YBValueConverterSuite = map[string]ConverterFn{
...
+			if len(kv) == 2 {
```
makalaaneesh, line 191: It'd be bad to silently ignore in case len is not = 2. (for whatever reason). If that is the expectation, we should fail otherwise.

yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -173,17 +173,31 @@ var YBValueConverterSuite = map[string]ConverterFn{
...
+				key := kv[0]
+				value := kv[1]
+				transformedMapValue = transformedMapValue + fmt.Sprintf("\"%s\"=>\"%s\",", key, value)
```
makalaaneesh, line 194: can we use "`" ? instead of having to escape double quotes?

yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -173,17 +173,31 @@ var YBValueConverterSuite = map[string]ConverterFn{
...
+			}
+		}
+		if len(transformedMapValue) > 1 {
```
makalaaneesh, line 197: is it possible for len to be exactly equal to 1? (perhaps one key, value pair, and key is empty?

makalaaneesh: maybe check len(pairs) instead? i.e. len(pairs) > 0

yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -173,17 +173,31 @@ var YBValueConverterSuite = map[string]ConverterFn{
...
+			transformedMapValue = transformedMapValue[:len(transformedMapValue)-1]//remove last comma and
 		}
-		return fmt.Sprintf("'%s'", transformedMapValue[:len(transformedMapValue)-1]), nil //remove last comma and add quotes
+		return quoteValueIfRequired(transformedMapValue, formatIfRequired, colDbzmSchema) // add quotes if required
```
makalaaneesh, line 200: do we need to also escape? i.e. use `quoteValueIfRequiredWithEscaping` ? What if key or value has double quotes. Is that possible?

yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -173,17 +173,31 @@ var YBValueConverterSuite = map[string]ConverterFn{
...
 		return string(hexValue), nil
 	},
-	"MAP": func(columnValue string, _ bool, _ *schemareg.ColumnSchema) (string, error) {
-		mapValue := make(map[string]interface{})
-		err := json.Unmarshal([]byte(columnValue), &mapValue)
-		if err != nil {
-			return columnValue, fmt.Errorf("parsing map: %v", err)
-		}
+	"MAP": func(columnValue string, formatIfRequired bool, colDbzmSchema *schemareg.ColumnSchema) (string, error) {
+		//e.g. val -  {key1=value1, key2=value2}
```
makalaaneesh, line 177: @priyanshi-yb I think we might be better off doing all this in java layer. In java, we directly have to deal with a hashmap. if we let it come to golang layer, then we need to generate that map (i.e. split on ",", split on "="). More room for error here. 

in java it's basically what we had earlier; just need to fix that bug in case of empty value. 
```
            case MAP:
                StringBuilder mapString = new StringBuilder();
                for (Map.Entry<String, String> entry : ((HashMap<String, String>) fieldValue).entrySet()) {
                    String key = entry.getKey();
                    String val = entry.getValue();
                    mapString.append("\"");
                    mapString.append(key);
                    mapString.append("\"");
                    mapString.append(" => ");
                    mapString.append("\"");
                    mapString.append(val);
                    mapString.append("\"");
                    mapString.append(",");
                }
                return mapString.toString().substring(0, mapString.length() - 1);
```

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/DebeziumRecordTransformer.java
```diff
@@ -70,6 +74,36 @@ private String makeFieldValueSerializable(Object fieldValue, Field field){
             case BYTES:
             case STRUCT:
                 return toKafkaConnectJsonConverted(fieldValue, field);
+            case MAP:
+	            StringBuilder mapString = new StringBuilder();
+                for (Map.Entry<String, String> entry : ((HashMap<String, String>) fieldValue).entrySet()) {
+                    String key = entry.getKey();
```
makalaaneesh, line 80: let's add debug logs before and after our transformation for the key and value.

migtests/scripts/run-test.sh
```diff
@@ -136,6 +136,14 @@ main() {
 	    fi
 	fi
 
+	if [ "${TEST_DIR}" = "${TESTS_DIR}/pg/datatypes" ]; then
+		cat ${EXPORT_DIR}/data/hstore_example_data.sql
```
makalaaneesh, line 140: this is just for debugging? remove?

migtests/tests/pg/datatypes/validate
```diff
@@ -26,6 +26,7 @@ EXPECTED_ROW_COUNT = {
 	'datetime_type2': 2, 
 	'null_and_default' :2,
 	'decimal_types': 3,
+	'hstore_example': 13,
```
makalaaneesh, line 29: let's also add data validations (for each row), for source-target, source-source-replica, target-source-replica(ignore if this is not supported). 

You can use `get_distinct_values_of_column_of_table`

yb-voyager/src/tgtdb/suites/yugabytedbSuite.go
```diff
@@ -173,17 +173,8 @@ var YBValueConverterSuite = map[string]ConverterFn{
...
-		err := json.Unmarshal([]byte(columnValue), &mapValue)
-		if err != nil {
-			return columnValue, fmt.Errorf("parsing map: %v", err)
-		}
-		var transformedMapValue string
-		for key, value := range mapValue {
-			transformedMapValue = transformedMapValue + fmt.Sprintf("\"%s\"=>\"%s\",", key, value)
-		}
-		return fmt.Sprintf("'%s'", transformedMapValue[:len(transformedMapValue)-1]), nil //remove last comma and add quotes
+	"MAP": func(columnValue string, formatIfRequired bool, dbzmSchema *schemareg.ColumnSchema) (string, error) {
```
makalaaneesh, line 176: - what about postgressuite? Do we re-use yugabytedbSuite? 
- we have some implementation for "MAP" in oracleSuite as well. I don't imagine that is useful, since we won't have MAP coming from oracle source at all, so remove it?

priyanshi-yb: > what about postgressuite? Do we re-use yugabytedbSuite?

yes we use the same for both

> we have some implementation for "MAP" in oracleSuite as well. I don't imagine that is useful, since we won't have MAP coming from oracle source at all, so remove it?

yeah ok

migtests/scripts/run-test.sh
priyanshi-yb: yes

migtests/tests/pg/datatypes/validateAfterChanges
```diff
@@ -130,6 +130,14 @@ def migration_completed_checks(tgt):
 	expected_distinct_values = EXPECTED_DISNTICT_VALUES['v5']
 	tgt.assert_distinct_values_of_col("datatypes2", "v5", "public", 
 		None, expected_distinct_values = expected_distinct_values)
+	
+	print("hstore_example:")
```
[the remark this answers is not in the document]
priyanshi-yb: yes it runs on all target, source, and source-replica.
For the `fall-forward/fall-back` workflows we still run with the `hstore_example` table and snapshot data is still present in all the DBs but the CDC events are not added for the target_delta i.e. after cutover to target. So only CDC events for hstore are not getting tested.

makalaaneesh (approved):
LGTM

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/DebeziumRecordTransformer.java
```diff
@@ -70,6 +74,42 @@ private String makeFieldValueSerializable(Object fieldValue, Field field){
...
+                    LOGGER.debug("[MAP] after transforming value - {}", val);
+                    
+                    mapString.append("\"");
+                    mapString.append(key);
+                    mapString.append("\"");
+                    mapString.append(" => ");
+                    mapString.append("\"");
+                    mapString.append(val);
+                    mapString.append("\"");
+                    mapString.append(",");
```
sanyamsinghal, line 106: @priyanshi-yb, do we have something like `fmt.Sprintf()` in Java to use here?

makalaaneesh: +1 this would be a better. No idea why I wrote it like this initially 😀

priyanshi-yb: There is this `String.format()` which looks similar to `fmt.Sprintf()` can fix that in a follow up PR.

PR #2162, opened 2025-01-08
Fix: retrieving Migration UUID as early as possible in the commands

priyanshi-yb:
Fixes retrieving the migration uuid as early as possible to be able send the diagnosis to callhome with proper uuid.

already available tests
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/4426


yb-voyager/cmd/importData.go
```diff
@@ -81,7 +81,11 @@ var importDataCmd = &cobra.Command{
 			importerRole = TARGET_DB_IMPORTER_ROLE
 		}
 		sourceDBType = GetSourceDBTypeFromMSR()
-		err := validateImportFlags(cmd, importerRole)
+		err := retrieveMigrationUUID()
```
makalaaneesh, line 84: can we do this above `GetSourceDBTypeFromMSR` ? that also ErrExits.

yb-voyager/cmd/importSchema.go
```diff
@@ -47,8 +47,12 @@ var importSchemaCmd = &cobra.Command{
 		if tconf.TargetDBType == "" {
 			tconf.TargetDBType = YUGABYTEDB
 		}
+		err := retrieveMigrationUUID()
```
makalaaneesh, line 50: Might be slightly unrelated question, but at this point what if export-dir is empty and user runs import-schema? What happens then?
Same question for import-data,  analyze-schema as well..

priyanshi-yb: In case of empty dirs and these commands we already have other checks just before this that checks if export-dir is initiated. And we can't do much in those cases we will still get the 00000... uuid in callhome

makalaaneesh: got it, but i guess now we will get the error message, so that should be helpful.

makalaaneesh (approved):

PR #2166, opened 2025-01-09
Implement Migration Complexity Explanation providing the summary and rationale for reported complexity

sanyamsinghal:
Different explanation in case of html and json. Html also includes the summary about categories and issues count.

https://yugabyte.atlassian.net/browse/DB-14717

- Analyze Schema report has stopped reporting migration complexity
- Assessment Report will have one more field - `MigrationComplexityExplanation`

**Sample HTML Report**

1. Non zero issues
image

2. Zero Issues
image

**Sample JSON Report**

image

Manually. 
This is a new field in the assessment report which we plan to add to test later with other UI changes.
Also have to modified the upgrade unit test for the new field.

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -425,6 +425,7 @@ func assessMigration() (err error) {
 	}
 
 	log.Infof("number of assessment issues detected: %d\n", len(assessmentReport.Issues))
+
 	utils.PrintAndLog("Migration assessment completed successfully.")
 	completedEvent := createMigrationAssessmentCompletedEvent()
 	controlPlane.MigrationAssessmentCompleted(completedEvent)
@@ -1604,11 +1605,11 @@ func generateAssessmentReportJson(reportDir string) error {
 	log.Infof("writing assessment report to file: %s", jsonReportFilePath)
 
 	var err error
-	assessmentReport.MigrationComplexityExplainability, err = buildMigrationComplexityExplainability(assessmentReport, "text")
+	assessmentReport.MigrationComplexityExplaination, err = buildMigrationComplexityExplaination(source.DBType, assessmentReport, "")
 	if err != nil {
-		utils.PrintAndLog("ERROR: unable to build migration complexity explainability for json report: %v", err)
+		utils.PrintAndLog("ERROR: unable to build migration complexity explanation for json report: %v", err)
 	}
-	log.Info(assessmentReport.MigrationComplexityExplainability)
+	log.Info(assessmentReport.MigrationComplexityExplaination)
 
 	strReport, err := json.MarshalIndent(assessmentReport, "", "\t")
 	if err != nil {
@@ -1629,9 +1630,9 @@ func generateAssessmentReportHtml(reportDir string) error {
 	log.Infof("writing assessment report to file: %s", htmlReportFilePath)
 
 	var err error
-	assessmentReport.MigrationComplexityExplainability, err = buildMigrationComplexityExplainability(assessmentReport, "html")
+	assessmentReport.MigrationComplexityExplaination, err = buildMigrationComplexityExplaination(source.DBType, assessmentReport, "html")
 	if err != nil {
-		utils.PrintAndLog("ERROR: unable to build migration complexity explainability for html report: %v", err)
+		utils.PrintAndLog("ERROR: unable to build migration complexity explanation for html report: %v", err)
 	}
 
 	file, err := os.Create(htmlReportFilePath)
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -1602,6 +1602,14 @@ func postProcessingOfAssessmentReport() {
 func generateAssessmentReportJson(reportDir string) error {
 	jsonReportFilePath := filepath.Join(reportDir, fmt.Sprintf("%s%s", ASSESSMENT_FILE_NAME, JSON_EXTENSION))
 	log.Infof("writing assessment report to file: %s", jsonReportFilePath)
+
+	var err error
+	assessmentReport.MigrationComplexityExplainability, err = buildMigrationComplexityExplainability(assessmentReport, "text")
+	if err != nil {
+		utils.PrintAndLog("ERROR: unable to build migration complexity explainability for json report: %v", err)
+	}
+	log.Info(assessmentReport.MigrationComplexityExplainability)
+
 	strReport, err := json.MarshalIndent(assessmentReport, "", "\t")
 	if err != nil {
 		return fmt.Errorf("failed to marshal the assessment report: %w", err)
@@ -1619,6 +1627,12 @@ func generateAssessmentReportJson(reportDir string) error {
 func generateAssessmentReportHtml(reportDir string) error {
 	htmlReportFilePath := filepath.Join(reportDir, fmt.Sprintf("%s%s", ASSESSMENT_FILE_NAME, HTML_EXTENSION))
 	log.Infof("writing assessment report to file: %s", htmlReportFilePath)
+
+	var err error
+	assessmentReport.MigrationComplexityExplainability, err = buildMigrationComplexityExplainability(assessmentReport, "html")
+	if err != nil {
+		utils.PrintAndLog("ERROR: unable to build migration complexity explainability for html report: %v", err)
+	}
 
 	file, err := os.Create(htmlReportFilePath)
 	if err != nil {
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -1607,7 +1607,7 @@ func generateAssessmentReportJson(reportDir string) error {
 	var err error
 	assessmentReport.MigrationComplexityExplanation, err = buildMigrationComplexityExplanation(source.DBType, assessmentReport, "")
 	if err != nil {
-		utils.PrintAndLog("ERROR: unable to build migration complexity explanation for json report: %v", err)
+		return fmt.Errorf("unable to build migration complexity explanation for json report: %w", err)
 	}
 	log.Info(assessmentReport.MigrationComplexityExplanation)
 
@@ -1632,7 +1632,7 @@ func generateAssessmentReportHtml(reportDir string) error {
 	var err error
 	assessmentReport.MigrationComplexityExplanation, err = buildMigrationComplexityExplanation(source.DBType, assessmentReport, "html")
 	if err != nil {
-		utils.PrintAndLog("ERROR: unable to build migration complexity explanation for html report: %v", err)
+		return fmt.Errorf("unable to build migration complexity explanation for html report: %w", err)
 	}
 
 	file, err := os.Create(htmlReportFilePath)
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -1605,11 +1605,11 @@ func generateAssessmentReportJson(reportDir string) error {
 	log.Infof("writing assessment report to file: %s", jsonReportFilePath)
 
 	var err error
-	assessmentReport.MigrationComplexityExplaination, err = buildMigrationComplexityExplaination(source.DBType, assessmentReport, "")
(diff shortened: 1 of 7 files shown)


yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1602,6 +1603,14 @@ func postProcessingOfAssessmentReport() {
 func generateAssessmentReportJson(reportDir string) error {
 	jsonReportFilePath := filepath.Join(reportDir, fmt.Sprintf("%s%s", ASSESSMENT_FILE_NAME, JSON_EXTENSION))
 	log.Infof("writing assessment report to file: %s", jsonReportFilePath)
+
+	var err error
+	assessmentReport.MigrationComplexityExplanation, err = buildMigrationComplexityExplanation(source.DBType, assessmentReport, "")
+	if err != nil {
+		utils.PrintAndLog("ERROR: unable to build migration complexity explanation for json report: %v", err)
```
priyanshi-yb, line 1610: why `utils.PrintAndLog`, use return `fmt.Errorf(..)` ?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1620,6 +1629,12 @@ func generateAssessmentReportHtml(reportDir string) error {
 	htmlReportFilePath := filepath.Join(reportDir, fmt.Sprintf("%s%s", ASSESSMENT_FILE_NAME, HTML_EXTENSION))
 	log.Infof("writing assessment report to file: %s", htmlReportFilePath)
 
+	var err error
+	assessmentReport.MigrationComplexityExplanation, err = buildMigrationComplexityExplanation(source.DBType, assessmentReport, "html")
+	if err != nil {
+		utils.PrintAndLog("ERROR: unable to build migration complexity explanation for html report: %v", err)
```
priyanshi-yb, line 1635: same here

yb-voyager/cmd/common.go
```diff
@@ -1051,14 +1051,15 @@ func storeTableListInMSR(tableList []sqlname.NameTuple) error {
...
-	MigrationComplexity string                                `json:"MigrationComplexity"`
-	SchemaSummary       utils.SchemaSummary                   `json:"SchemaSummary"`
-	Sizing              *migassessment.SizingAssessmentReport `json:"Sizing"`
-	Issues              []AssessmentIssue                     `json:"-"` // disabled in reports till corresponding UI changes are done(json and html reports)
-	TableIndexStats     *[]migassessment.TableIndexStats      `json:"TableIndexStats"`
-	Notes               []string                              `json:"Notes"`
+	VoyagerVersion                 string                                `json:"VoyagerVersion"`
+	TargetDBVersion                *ybversion.YBVersion                  `json:"TargetDBVersion"`
+	MigrationComplexity            string                                `json:"MigrationComplexity"`
+	MigrationComplexityExplanation string                                `json:"MigrationComplexityExplanation"`
```
priyanshi-yb, line 1057: `MigrationComplexityExplaination`

yb-voyager/cmd/migration_complexity.go
```diff
@@ -200,3 +210,121 @@ func getComplexityForLevel(level string, count int) string {
 		panic(fmt.Sprintf("unknown impact level %s for determining complexity", level))
 	}
 }
+
+// ======================================= Migration Complexity Explanation ==========================================
+
+const explainTemplateHTML = `
```
makalaaneesh, line 216: Would like to avoid having HTML in multiple places. It should all ideally be in the template file. Morevoer, field `AssessmentReport.MigrationComplexityExplanation` having HTML would be weird. 
But I get why you did this. (because json and HTML have different requirements). 

Here's one way to do this: 
1. in `AssessmentReport`, instead of storing just the final string in `MigrationComplexityExplanation`, store the actual `ExplanationData` struct. 
2. in HTML template, you have all the necessary information to build the table there. 
3. for json, just exclude the summaries from it. (in fact there's no harm in even keeping it tbh).

yb-voyager/cmd/migration_complexity.go
```diff
@@ -200,3 +210,121 @@ func getComplexityForLevel(level string, count int) string {
...
+</p>
+
+<p>
+	<strong>Reasoning:</strong> {{ .ComplexityRationale }}
+</p>
+`
+
+const explainTemplateText = `Reasoning: {{ .ComplexityRationale }}`
+
+type ExplanationData struct {
```
makalaaneesh, line 253: `MigrationComplexityExplanationData` ? if we were inside a `migraitonComplexity` pkg, it would be okay, but we're in `cmd` pkg, so it would help to be specific. 

Same for `CategorySummary` I suppose.

yb-voyager/src/utils/utils.go
```diff
@@ -744,3 +744,13 @@ func CheckTools(tools ...string) []string {
 func BuildObjectName(schemaName, objName string) string {
 	return lo.Ternary(schemaName != "", schemaName+"."+objName, objName)
 }
+
+// ToTitleCase converts a snake_case string to a title case string with spaces.
+func ToTitleCase(snake string) string {
```
makalaaneesh, line 749: would be good to mention `snake case` in the func name. So `SnakeCaseToTitleCase` ?

yb-voyager/cmd/migration_complexity.go
```diff
@@ -200,3 +210,121 @@ func getComplexityForLevel(level string, count int) string {
...
+			}
+		}
+
+		summaryMap[issue.Category].TotalIssueCount++
+		summaryMap[issue.Category].ImpactCounts[issue.Impact]++
+	}
+
+	var result []CategorySummary
+	for _, summary := range summaryMap {
+		summary.Category = utils.ToTitleCase(summary.Category)
```
makalaaneesh, line 326: In golang layer, it would be good to still retain the default category values. This is a UI requirement, so it makes sense to do it at the HTML layer. You can export the `ToTitleCase` func to the template, and call it.

yb-voyager/cmd/migration_complexity.go
```diff
@@ -200,3 +210,121 @@ func getComplexityForLevel(level string, count int) string {
...
+	explanation.Summaries = buildCategorySummary(assessmentReport.Issues)
+
+	var tmpl *template.Template
+	var err error
+	if reportFormat == "html" {
+		tmpl, err = template.New("Explain").Parse(explainTemplateHTML)
+	} else {
+		tmpl, err = template.New("Explain").Parse(explainTemplateText)
+	}
+
```
priyanshi-yb, line 281: why are doing this here and not in HTML assessment template?

yb-voyager/cmd/migration_complexity.go
priyanshi-yb: Agree to this, we should keep all the HTML code in that template 

> for json, just exclude the summaries from it. (in fact there's no harm in even keeping it tbh).
yes we can keep it in JSON as well an, in case of tests and all you are anyways ignoring it

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: I was thinking of showing something on console also for users to know that explanation is not there for xyz reason

yb-voyager/cmd/migration_complexity.go
```diff
@@ -200,3 +210,121 @@ func getComplexityForLevel(level string, count int) string {
...
+            <td>{{ .TotalIssueCount }}</td>
+        </tr>
+    {{- end }}
+    </tbody>
+</table>
+
+<p>
+	Level-1 Impact: Resolutions are available with minimal effort.<br/>
+	Level-2 Impact: Resolutions are available requiring moderate effort.<br/>
+	Level-3 Impact: Resolutions may not be available or are highly complex.
```
hbhanawat, line 243: Resolutions may not be available or are highly complex -> Resolutions may not be available or are complex ->

sanyamsinghal: yes, thanks.

yb-voyager/cmd/migration_complexity.go
sanyamsinghal: we will have to add the explanation structure in the assessment report struct also.
This is also a low level detail which can reside separately...

yb-voyager/cmd/common.go
sanyamsinghal: `Explanation` is the correct spelling  :)

priyanshi-yb: Oh right please update the PR description, The html report screenshot doens't reflect the current code's html

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: The top level functions which is calling the generateAssessmentReportJson must be printing the error on console for users so it will come eventually right?

sanyamsinghal: will this not look good? any suggestions?

yb-voyager/cmd/migration_complexity.go
sanyamsinghal: done

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: > The top level functions which is calling the generateAssessmentReportJson must be printing the error on console for users so it will come eventually right?

No, what i did was to not error out the cmd if building explanation fails...

yb-voyager/cmd/common.go
sanyamsinghal: yeah will do, those are old.

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: Oh, I see. Ideally, we should return the error,and mostly the errors occur while `buildMigrationComplexityExplanation ` is generating HTML string, etc.
For the case, we should fail the command for this new addition, we can use env var approach and user faces any error we can always ask them to disable it.

sanyamsinghal: hmm makes sense, this is the standard code - creating template and executing it.
Should not be error here ideally.

sanyamsinghal: done

yb-voyager/cmd/migration_complexity.go
sanyamsinghal: added a TODO comment

yb-voyager/cmd/migration_complexity.go
```diff
@@ -200,3 +210,127 @@ func getComplexityForLevel(level string, count int) string {
...
+			}
+		}
+
+		summaryMap[issue.Category].TotalIssueCount++
+		summaryMap[issue.Category].ImpactCounts[issue.Impact]++
+	}
+
+	var result []MigrationComplexityCategorySummary
+	for _, summary := range summaryMap {
+		summary.Category = utils.SnakeCaseToTitleCase(summary.Category)
```
priyanshi-yb, line 339: Instead of using this function we can get the Feature heading for that Enum to be consistent with the heading in report. 
but given that we are anyways changing the UI, we can revisit it later so okay for now.

yb-voyager/cmd/migration_complexity.go
```diff
@@ -200,3 +210,127 @@ func getComplexityForLevel(level string, count int) string {
 		panic(fmt.Sprintf("unknown impact level %s for determining complexity", level))
 	}
 }
+
+// ======================================= Migration Complexity Explanation ==========================================
+
+// TODO: discuss if the html should be in main report or here
+const explainTemplateHTML = `
+<p>Below is a detailed breakdown of issues by category, showing the count for each impact level.</p>
```
makalaaneesh, line 218: How about `Below is a breakdown of the issues detected in different categories for each impact level` ?

sanyamsinghal: yeah better.

makalaaneesh (approved):
LGTM

priyanshi-yb (approved):
LGTM

yb-voyager/cmd/migration_complexity.go
sanyamsinghal: yeah lets revisit this. The category here is being fetched from the analyze schema issue.

I see the same issues of formatting in schema reports. We need to fix the root.

PR #2165, opened 2025-01-08
Reporting CTE Materialized clause and Sql body in Function 

priyanshi-yb:
### Describe the changes in this pull request
Reports the CTE statements with `Materialized` clause as not supported. https://yugabyte.atlassian.net/browse/DB-14227
Reports the presence of SQL body in Functions - PG15 [docs](https://www.postgresql.org/docs/15/sql-createfunction.html#:~:text=a%20new%20session.-,sql_body,-The%20body%20of) 
https://yugabyte.atlassian.net/browse/DB-14234
### Describe if there are any user-facing changes

Added new feature `SQL body in Function` to report functions and `Modifying the Materialized option with CTE ` in assessment.

### How was this pull request tested?

unit tests and end-to-end-tests

### Does your PR have changes that can cause upgrade issues? 
| Component        | Breaking changes? |
| :----------------------------------------------: | :-----------: |
| MetaDB                                                    | No |
| Name registry json                                        | No |
| Data File Descriptor Json                                 | No |
| Export Snapshot Status Json                               | o |
| Import Data State                                         | No |
| Export Status Json                                        | No |
| Data .sql files of tables                                 |No |
| Export and import data queue                              | No |
| Schema Dump                                               | No |
| AssessmentDB                                              | No |
| Sizing DB                                                 | No |
| Migration Assessment Report Json                          | No |
| Callhome Json                                             | No |
| YugabyteD Tables                                          | No |
| TargetDB Metadata Tables                                  | No |

PR #2201, opened 2025-01-16
Fix regex parser for parsing functions having SQL body with language sql (PG 15 feature)

priyanshi-yb:
SQL body syntax - https://www.postgresql.org/docs/15/sql-createfunction.html#:~:text=a%20new%20session.-,sql_body,-The%20body%20of, where with language SQL
```
BEGIN ATOMIC;
....
END;
```
or 
```
language sql
RETURN ...;
```

N/A

unit tests added and end-to end tests added in other PR - #2165 

### DIFF
--- a/yb-voyager/cmd/analyzeSchema.go
+++ b/yb-voyager/cmd/analyzeSchema.go
@@ -921,6 +921,9 @@ sqlParsingLoop:
 			} else if matches := dollarQuoteRegex.FindStringSubmatch(currLine); matches != nil {
 				dollarQuoteFlag = 1 //denotes start of the code/body part
 				codeBlockDelimiter = matches[0]
+			} else if strings.Contains(currLine, "BEGIN ATOMIC") {
+				dollarQuoteFlag = 1 //denotes start of the sql body part https://www.postgresql.org/docs/15/sql-createfunction.html#:~:text=a%20new%20session.-,sql_body,-The%20body%20of
+				codeBlockDelimiter = "END"
 			}
 		case CODE_BLOCK_STARTED:
 			if strings.Contains(currLine, codeBlockDelimiter) {
--- a/yb-voyager/cmd/analyzeSchema.go
+++ b/yb-voyager/cmd/analyzeSchema.go
@@ -926,14 +926,25 @@ sqlParsingLoop:
 				codeBlockDelimiter = "END" //SQL body to determine the end of BEGIN ATOMIC ... END; sql body
 			}
 		case CODE_BLOCK_STARTED:
-			if strings.Contains(currLine, codeBlockDelimiter) ||
-				strings.Contains(currLine, strings.ToLower(codeBlockDelimiter)) {
-				//TODO: anyways we should be using pg-parser: but for now for the END sql body delimiter checking the UPPER and LOWER both
-				dollarQuoteFlag = 2 //denotes end of code/body part
-				if isEndOfSqlStmt(currLine) {
-					break sqlParsingLoop
+			switch codeBlockDelimiter {
+			case "END":
+				if strings.Contains(currLine, codeBlockDelimiter) ||
+					strings.Contains(currLine, strings.ToLower(codeBlockDelimiter)) {
+					//TODO: anyways we should be using pg-parser: but for now for the END sql body delimiter checking the UPPER and LOWER both
+					dollarQuoteFlag = 2 //denotes end of code/body part
+					if isEndOfSqlStmt(currLine) {
+						break sqlParsingLoop
+					}
+				}
+			default:
+				if strings.Contains(currLine, codeBlockDelimiter) {
+					dollarQuoteFlag = 2 //denotes end of code/body part
+					if isEndOfSqlStmt(currLine) {
+						break sqlParsingLoop
+					}
 				}
 			}
+
 		case CODE_BLOCK_COMPLETED:
 			if isEndOfSqlStmt(currLine) {
 				break sqlParsingLoop
--- a/yb-voyager/cmd/analyzeSchema.go
+++ b/yb-voyager/cmd/analyzeSchema.go
@@ -134,6 +134,7 @@ var (
 	parserIssueDetector = queryissue.NewParserIssueDetector()
 	multiRegex          = regexp.MustCompile(`([a-zA-Z0-9_\.]+[,|;])`)
 	dollarQuoteRegex    = regexp.MustCompile(`(\$.*\$)`)
+	sqlBodyBeginRegex   = re("BEGIN", "ATOMIC")
 	//TODO: optional but replace every possible space or new line char with [\s\n]+ in all regexs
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/analyzeSchema.go
```diff
@@ -921,6 +921,9 @@ sqlParsingLoop:
 			} else if matches := dollarQuoteRegex.FindStringSubmatch(currLine); matches != nil {
 				dollarQuoteFlag = 1 //denotes start of the code/body part
 				codeBlockDelimiter = matches[0]
+			} else if strings.Contains(currLine, "BEGIN ATOMIC") {
```
makalaaneesh, line 924: seems okay to me but would like @sanyamsinghal also to take a look

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -921,6 +921,9 @@ sqlParsingLoop:
...
+				dollarQuoteFlag = 1 //denotes start of the sql body part https://www.postgresql.org/docs/15/sql-createfunction.html#:~:text=a%20new%20session.-,sql_body,-The%20body%20of
+				codeBlockDelimiter = "END"
```
sanyamsinghal, line 926: 1. There are chances of `BEGIN ATOMIC` or `END` being on same line, different casing(smallcase, mixed case)
```
CREATE OR REPLACE FUNCTION subtract_and_log(a INT, b INT)
RETURNS INT
LANGUAGE SQL
BEGIN ATOMIC INSERT INTO test_log(action) VALUES ('Subtracting numbers'); RETURN a - b; END;
```

2. Using regexp here would be right way to go here, given there are N number of posibilities.

sanyamsinghal (changes requested):

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: > There are chances of BEGIN ATOMIC or END being on same line, different line, something on same line and rest on the next line.

This automatically works with the current code, added unit test case for that.

>Posibility of different casing(smallcase, mixed case) used for BEGIN ATOMIC ... END;

Fixed the casing by using the BEGIN ATOMIC with regexp but END with string lower upper both

Not trying to cover a lot cases given that we should use the pg-parser anyways but having this meanwhile to atleast able to detect the SQL body case in this PR #2165

sanyamsinghal: I doubt we are going to have pg-parser soon for breaking the sqlfile into sql statements.
Also for oracle cases where syntax is off, (not fully converted to PG), there the parser will fail, so we could either use current approach or  mix of both.

I would say lets cover most of it if we can, with small efforts.
Since this SQL body thing is a part of DDL, and our import schema can also break.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -134,6 +134,7 @@ var (
 	parserIssueDetector = queryissue.NewParserIssueDetector()
 	multiRegex          = regexp.MustCompile(`([a-zA-Z0-9_\.]+[,|;])`)
 	dollarQuoteRegex    = regexp.MustCompile(`(\$.*\$)`)
+	sqlBodyBeginRegex   = re("BEGIN", "ATOMIC")
```
sanyamsinghal, line 137: Should we put a `^` in the regex to ensure that `BEGIN ATOMIC` is at the start of line?
Do we need to ensure that?

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -921,9 +921,14 @@ sqlParsingLoop:
...
 				dollarQuoteFlag = 1 //denotes start of the code/body part
 				codeBlockDelimiter = matches[0]
+			} else if matches := sqlBodyBeginRegex.FindStringSubmatch(currLine); matches != nil {
+				dollarQuoteFlag = 1        //denotes start of the sql body part https://www.postgresql.org/docs/15/sql-createfunction.html#:~:text=a%20new%20session.-,sql_body,-The%20body%20of
+				codeBlockDelimiter = "END" //SQL body to determine the end of BEGIN ATOMIC ... END; sql body
 			}
 		case CODE_BLOCK_STARTED:
-			if strings.Contains(currLine, codeBlockDelimiter) {
+			if strings.Contains(currLine, codeBlockDelimiter) ||
+				strings.Contains(currLine, strings.ToLower(codeBlockDelimiter)) {
```
sanyamsinghal, line 930: Concern here is that we don't know the variety of cases possible with plpgsql, since it can almost have anything inside the body. Here the condition is getting changed to match with either same or lowercase variant of it.

Should we just handle the sqlbody func case separately(by that i mean having different if conditions here for that) and not update the existing ones.
Just want to avoid any unknowns

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: I don't think we need to ensure that as it could be anywhere  not necessarily in starting of the line e.g.
```
CREATE FUNCTION public.asterisksdfsf(n integer) RETURNS SETOF text
    LANGUAGE sql IMMUTABLE STRICT PARALLEL SAFE BEGIN ATOMIC
 SELECT repeat('*'::text, g.g) AS repeat
    FROM generate_series(1, asterisksdfsf.n) g(g);
END;

```

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: > Here the condition is getting changed to match with either same or lowercase variant of it.

the same or lowercase variant is only of the `codeBlockDelimiter` variable which could be dollarQuoteRegex    = regexp.MustCompile(`(\$.*\$)`) match or the END string for sqlBody. 

But okay, I can do that in separate if condition for just END case

priyanshi-yb: done

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -921,14 +921,30 @@ sqlParsingLoop:
...
 			}
 		case CODE_BLOCK_STARTED:
-			if strings.Contains(currLine, codeBlockDelimiter) {
-				dollarQuoteFlag = 2 //denotes end of code/body part
-				if isEndOfSqlStmt(currLine) {
-					break sqlParsingLoop
+			switch codeBlockDelimiter {
+			case "END":
+				if strings.Contains(currLine, codeBlockDelimiter) ||
+					strings.Contains(currLine, strings.ToLower(codeBlockDelimiter)) {
```
sanyamsinghal, line 932: nit but can we use `strings.Equalfold()` here to compare both and dont care about casing(covering mixed case scenario also).

sanyamsinghal (approved):

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: `strings.Equalfold()` checks if two strings are equal without case sensitivity but we need to check whether string `Contains` so can't use it here.

PR #2203, opened 2025-01-17
Fix: Do not redact password for conn pool connection uri while connecting to DB 

priyanshi-yb:
Fixing the bug where import data always fails to connect to DB via the connection pool with a password authentication as the connection URLs are getting redacted while passing to the connection pool.
 https://github.com/yugabyte/yb-voyager/issues/2204
redaction still happens- 
```
2025-01-18 12:46:14.483348 INFO yugabytedb.go:252 Initialized connection pool with settings: (tgtdb.ConnectionParams) {
 NumConnections: (int) 2,
 NumMaxConnections: (int) 4,
 ConnUriList: ([]string) (len=1 cap=1) {
  (string) (len=81) "postgresql://ybvoyager:xxxxx@10.9.14.120:5433/pg_datatypes_offline?sslmode=prefer"
 },
 SessionInitScript: ([]string) (len=3 cap=4) {
  (string) (len=29) "SET client_encoding TO 'UTF8'",
  (string) (len=39) "SET session_replication_role TO replica",
  (string) (len=53) "SET default_transaction_isolation = 'repeatable read'"
 }
}
```
Deep copying the connection params for redacting while logging using the package - https://github.com/tiendc/go-deepcopy because of following reasons-
1. It is quite active as recent change is a month ago.
2. It claims to have better benchmarks for deep copying.

Its a fix, no user-facing changes  

automation tests are enough

### DIFF
--- a/.github/workflows/pg-9-migtests.yml
+++ b/.github/workflows/pg-9-migtests.yml
@@ -85,7 +85,7 @@ jobs:
           docker run -d --name yugabytedb \
             -p7000:7000 -p9000:9000 -p15433:15433 -p5433:5433 -p9042:9042 \
             yugabytedb/yugabyte:${{ matrix.version }} \
-            bin/yugabyted start --background=false --ui=false
+            bin/yugabyted start --tserver_flags="ysql_hba_conf_csv={host all yugabyte all trust,host all all all md5}" --background=false --ui=false 
           sleep 20
 
       - name: Test YugabyteDB connection
--- a/yb-voyager/src/tgtdb/yugabytedb.go
+++ b/yb-voyager/src/tgtdb/yugabytedb.go
@@ -246,7 +246,11 @@ func (yb *TargetYugabyteDB) InitConnPool() error {
 	yb.connPool = NewConnectionPool(params)
 	redactedParams := &ConnectionParams{}
 	//Whenever adding new fields to CONNECTION PARAMS check if that needs to be redacted while logging
-	deepcopy.Copy(redactedParams, params)
+	err := deepcopy.Copy(redactedParams, params)
+	if err!= nil {
+		log.Errorf("couldn't get the copy of connection params for logging: %v", err)
+		return nil
+	}
 	redactedParams.ConnUriList = utils.GetRedactedURLs(redactedParams.ConnUriList)
 	log.Info("Initialized connection pool with settings: ", spew.Sdump(redactedParams))
 	return nil
--- a/yb-voyager/src/tgtdb/yugabytedb.go
+++ b/yb-voyager/src/tgtdb/yugabytedb.go
@@ -243,8 +243,12 @@ func (yb *TargetYugabyteDB) InitConnPool() error {
 		SessionInitScript: getYBSessionInitScript(yb.tconf),
 	}
 	yb.connPool = NewConnectionPool(params)
-	redactedParams := params
-	redactedParams.ConnUriList = utils.GetRedactedURLs(redactedParams.ConnUriList)
+	redactedParams := ConnectionParams{
+		NumConnections:    yb.tconf.Parallelism,
+		NumMaxConnections: yb.tconf.MaxParallelism,
+		ConnUriList:       utils.GetRedactedURLs(targetUriList),
+		SessionInitScript: getYBSessionInitScript(yb.tconf),
+	}
 	log.Info("Initialized connection pool with settings: ", spew.Sdump(redactedParams))
 	return nil
 }
--- a/yb-voyager/src/tgtdb/yugabytedb.go
+++ b/yb-voyager/src/tgtdb/yugabytedb.go
@@ -243,7 +243,12 @@ func (yb *TargetYugabyteDB) InitConnPool() error {
 		SessionInitScript: getYBSessionInitScript(yb.tconf),
 	}
 	yb.connPool = NewConnectionPool(params)
-	redactedParams := *params
+	redactedParams := ConnectionParams{
+		NumConnections:    yb.tconf.Parallelism,
+		NumMaxConnections: yb.tconf.MaxParallelism,
+		ConnUriList:       utils.GetRedactedURLs(targetUriList),
+		SessionInitScript: getYBSessionInitScript(yb.tconf),
+	}
 	log.Info("Initialized connection pool with settings: ", spew.Sdump(redactedParams))
 	return nil
 }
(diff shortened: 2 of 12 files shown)


yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -243,8 +243,12 @@ func (yb *TargetYugabyteDB) InitConnPool() error {
 		SessionInitScript: getYBSessionInitScript(yb.tconf),
 	}
 	yb.connPool = NewConnectionPool(params)
-	redactedParams := params
-	redactedParams.ConnUriList = utils.GetRedactedURLs(redactedParams.ConnUriList)
+	redactedParams := ConnectionParams{
```
makalaaneesh, line 246: We're logging the full struct below, so this will mean we'll have to come back to keep updating this copy here whenever we add a new field. 

Let's go with `redactedParams := *params` ?

.github/workflows/pg-9-migtests.yml
```diff
@@ -85,7 +85,7 @@ jobs:
           docker run -d --name yugabytedb \
             -p7000:7000 -p9000:9000 -p15433:15433 -p5433:5433 -p9042:9042 \
             yugabytedb/yugabyte:${{ matrix.version }} \
-            bin/yugabyted start --background=false --ui=false
+            bin/yugabyted start --tserver_flags="ysql_hba_conf_csv={host all yugabyte all trust,host all all all md5}" --background=false --ui=false
```
makalaaneesh, line 88: Shall we just have `host all all all md5` ? Just want to avoid "trust" altogether 😅

priyanshi-yb: yeah I started with that only, but somehow yugabyted wasn't coming up with that conf as yugabyte is a default user maybe trust is required for that, I can take that up later after figuring out how to do that

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -243,8 +243,7 @@ func (yb *TargetYugabyteDB) InitConnPool() error {
 		SessionInitScript: getYBSessionInitScript(yb.tconf),
 	}
 	yb.connPool = NewConnectionPool(params)
-	redactedParams := params
-	redactedParams.ConnUriList = utils.GetRedactedURLs(redactedParams.ConnUriList)
+	redactedParams := *params
```
sanyamsinghal, line 246: We should not do this.
https://go.dev/play/p/NjUBR-pdyd7
This is typical example of shallow copy vs deep copy and can potentially lead to a bug in future again.

I think as a standard practice also, we should always do Deep copy unless there is a usecase for shallow or significant perf implications of deepcopy.


1. Either copy each field separately.
2. Make use of standard deepcopy libraries available (to address @makalaaneesh's concern)
https://github.com/mohae/deepcopy
https://github.com/barkimedes/go-deepcopy (this seems to be better maintained)

sanyamsinghal (changes requested):

yb-voyager/src/tgtdb/yugabytedb.go
makalaaneesh: Ah, good point! 
+1 to use deep copy by default. (unless there are pointers like mutexes/connections/etc. but then again if the struct has such fields, we wouldn't necessarily want to deep copy)

priyanshi-yb: Yeah, missed it thanks for bringing it up @sanyamsinghal .
I made this change but missed one more thing to do redaction of urls back for logging.
I am thinking to not blindly deep copy that to just for logging or for the case where we can also add new fields which don't need to be copied, we can add any new field to logged params as well if required. 
I can create the params with redaction for all current fields and log basically revert the commit

sanyamsinghal: Yeah you can copy each field manually also, just the overhead of maintaining the codepath for logging. But since the code is at one place so it will be fine.

btw do you see any issue with simply creating a deep copy and printing it? I believe this is a one time thing only?

priyanshi-yb: The issue with directly copying the params is that there will still be an overhead of redacting the URLs in ConnectionParams for logging, and this will still be present for any new fields we will add in the future to check if they need redaction. 
```
redactedParams := &ConnectionParams{}
deepcopy.Copy(redactedParams, params)
redactedParams.ConnUriList = utils.GetRedactedURLs(redactedParams.ConnUriList)
```
So, I think it is better to have the overhead of adding the new field to the logged parameters if required, as it is a sensitive part where we have connection info with passwords.

makalaaneesh: We're probably overthinking this :) Let's go with whatever solves the problem easily.
IMO, deepcopy + redacting URL should be okay because:

1. We definitely want to log all connection params (so the full struct). 
2. Passwords are most likely the only thing we'd not want to log for security reasons. Everything else is okay (this is not callhome, it's just logs on the client machine).

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -37,6 +37,7 @@ import (
 	_ "github.com/jackc/pgx/v5/stdlib"
 	"github.com/samber/lo"
 	log "github.com/sirupsen/logrus"
+	"github.com/tiendc/go-deepcopy"
```
makalaaneesh, line 40: Just as a general discussion, not suggesting you change it:
I'm guessing you picked this because it seems to be the fastest. In our case, though, we don't really care so much about speed (because we call it like once and log it). In such cases, it makes more sense to prioritize dependencies that are well maintained.

makalaaneesh (approved):
LGTM

yb-voyager/src/tgtdb/yugabytedb.go
priyanshi-yb: Not just fast, I also looked at activity on the repo and only this one seems quite active, rest had last commit years ago

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -243,7 +244,9 @@ func (yb *TargetYugabyteDB) InitConnPool() error {
 		SessionInitScript: getYBSessionInitScript(yb.tconf),
 	}
 	yb.connPool = NewConnectionPool(params)
-	redactedParams := params
+	redactedParams := &ConnectionParams{}
+	//Whenever adding new fields to CONNECTION PARAMS check if that needs to be redacted while logging
+	deepcopy.Copy(redactedParams, params)
```
sanyamsinghal, line 249: handle the error if its returning..

sanyamsinghal (approved):

PR #2234, opened 2025-01-22
Implement versioning for Assess Migration callhome payload

sanyamsinghal:
- it starts with 1.0
- with this version we are also sending MigrationComplexityExplanation and assessment issues struct instead of separate fields for each category
- For each assessment issue we are obfuscating the sensitive information where it is
   - either Description
   - ObjectName
   - Issue Suggestion coming in (which is eventually a part of description for assessment issue)
- https://yugabyte.atlassian.net/browse/DB-14988

Callhome payloads has changed.
So the dashboard queries needs to be updated for the payload version `1.0` (create a ticket for this?)

Manually tests on a local callhome server.
cc @shubham-yb for second round of testing before 

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -137,6 +137,10 @@ var descriptionsIncludingSensitiveInformationToCallhome = []string{
 	queryissue.INDEX_ON_COMPLEX_DATATYPE_ISSUE_DESCRIPTION + " " + queryissue.INDEX_ON_COMPLEX_DATATYPE_ISSUE_SUGGESTION,
 	queryissue.LARGE_OBJECT_DATATYPE_ISSUE_DESCRIPTION + " " + queryissue.LARGE_OBJECT_DATATYPE_ISSUE_SUGGESTION,
 	queryissue.MULTI_RANGE_DATATYPE_ISSUE_DESCRIPTION + " " + queryissue.MULTI_RANGE_DATATYPE_ISSUE_SUGGESTION,
+
+	// extras because suggestion contains sensitive information
+	queryissue.FOREIGN_TABLE_ISSUE_DESCRIPTION + " " + queryissue.FOREIGN_TABLE_ISSUE_SUGGESTION,
+	queryissue.ALTER_TABLE_DISABLE_RULE_ISSUE_DESCRIPTION + " " + queryissue.ALTER_TABLE_DISABLE_RULE_ISSUE_SUGGESTION,
 }
 
 func packAndSendAssessMigrationPayload(status string, errMsg string) {
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -193,17 +193,14 @@ func packAndSendAssessMigrationPayload(status string, errMsg string) {
 	var obfuscatedIssues []callhome.AssessmentIssueCallhome
 	for _, issue := range assessmentReport.Issues {
 		obfuscatedIssue := callhome.AssessmentIssueCallhome{
-			Category:              issue.Category,
-			CategoryDescription:   issue.CategoryDescription,
-			Type:                  issue.Type,
-			Name:                  issue.Name,
-			Description:           issue.Description,
-			Impact:                issue.Impact,
-			ObjectType:            issue.ObjectType,
-			ObjectName:            constants.OBFUSCATE_STRING,
-			SqlStatement:          constants.OBFUSCATE_STRING, // TODO(future): we can obfuscate sensitive info in SQL statement
-			DocsLink:              issue.DocsLink,
-			MinimumVersionFixedIn: issue.MinimumVersionFixedIn,
+			Category:            issue.Category,
+			CategoryDescription: issue.CategoryDescription,
+			Type:                issue.Type,
+			Name:                issue.Name,
+			Description:         issue.Description,
+			Impact:              issue.Impact,
+			ObjectType:          issue.ObjectType,
+			ObjectName:          constants.OBFUSCATE_STRING,
 		}
 
 		// allowing object name for unsupported extension issue type
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -199,32 +199,16 @@ func packAndSendAssessMigrationPayload(status string, errMsg string) {
 			CategoryDescription: issue.CategoryDescription,
 			Type:                issue.Type,
 			Name:                issue.Name,
-			Description:         issue.Description,
 			Impact:              issue.Impact,
 			ObjectType:          issue.ObjectType,
-			ObjectName:          constants.OBFUSCATE_STRING,
 		}
 
-		// allowing object name for unsupported extension issue type
-		if issue.Type == UNSUPPORTED_EXTENSION_ISSUE_TYPE {
-			// object name(i.e. extension name here) might be qualified with schema so just taking the last part
-			obfuscatedIssue.ObjectName = strings.Split(issue.ObjectName, ".")[len(strings.Split(issue.ObjectName, "."))-1]
-		}
+		// TODO: object name(for extensions) and other details like datatype/indextype(present in description) for callhome will be covered in next release
 
-		for _, sensitiveDescription := range descriptionsIncludingSensitiveInformationToCallhome {
-			if sensitiveDescription == UNSUPPORTED_PG_SYNTAX_ISSUE_REASON && strings.HasPrefix(obfuscatedIssue.Description, sensitiveDescription) {
-				obfuscatedIssue.Description = sensitiveDescription
-			} else {
-				match, err := utils.MatchesFormatString(sensitiveDescription, obfuscatedIssue.Description)
-				if match {
-					obfuscatedIssue.Description, err = utils.ObfuscateFormatDetails(sensitiveDescription, obfuscatedIssue.Description, constants.OBFUSCATE_STRING)
-				}
-				if err != nil {
-					log.Errorf("error while matching issue description with sensitive descriptions: %v", err)
-					obfuscatedIssue.Description = constants.OBFUSCATE_STRING
-				}
-			}
-		}
+		// if issue.Type == UNSUPPORTED_EXTENSION_ISSUE_TYPE {
+		// object name(i.e. extension name here) might be qualified with schema so just taking the last part
+		// 	obfuscatedIssue.ObjectName = strings.Split(issue.ObjectName, ".")[len(strings.Split(issue.ObjectName, "."))-1]
+		// }
 
 		// appending the issue after obfuscating sensitive information
 		obfuscatedIssues = append(obfuscatedIssues, obfuscatedIssue)
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -120,7 +120,24 @@ var assessMigrationCmd = &cobra.Command{
 	},
 }
 
-var descriptionsIncludingSensitiveInformationToCallhome = reasonsIncludingSensitiveInformationToCallhome
+// assessment issue descriptions to modify before sending to callhome as will have sensitive information
+// copy of reasonsIncludingSensitiveInformationToCallhome
+var descriptionsIncludingSensitiveInformationToCallhome = []string{
+	UNSUPPORTED_PG_SYNTAX_ISSUE_REASON,
+	queryissue.STORED_GENERATED_COLUMNS_ISSUE_DESCRIPTION + " " + queryissue.STORED_GENERATED_COLUMN_ISSUE_SUGGESTION,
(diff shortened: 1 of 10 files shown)


yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -120,23 +120,52 @@ var assessMigrationCmd = &cobra.Command{
...
 func packAndSendAssessMigrationPayload(status string, errMsg string) {
 	if !shouldSendCallhome() {
 		return
 	}
-	payload := createCallhomePayload()
 
+	payload := createCallhomePayload()
 	payload.MigrationPhase = ASSESS_MIGRATION_PHASE
+	payload.Status = status
+	if assessmentMetadataDirFlag == "" {
```
makalaaneesh, line 154: were we essentially not sending this before?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -120,23 +120,52 @@ var assessMigrationCmd = &cobra.Command{
...
 
 func packAndSendAssessMigrationPayload(status string, errMsg string) {
 	if !shouldSendCallhome() {
 		return
 	}
-	payload := createCallhomePayload()
 
+	payload := createCallhomePayload()
 	payload.MigrationPhase = ASSESS_MIGRATION_PHASE
+	payload.Status = status
```
makalaaneesh, line 153: Again, here, were we not sending this before?

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: should we also check `if source != nil` ? (just being more cautious in terms of avoiding nil pointer panics)

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -156,87 +185,68 @@ func packAndSendAssessMigrationPayload(status string, errMsg string) {
 		}),
 	}
 
-	unsupportedDatatypesList := lo.Map(assessmentReport.UnsupportedDataTypes, func(datatype utils.TableColumnsDataTypes, _ int) string {
-		return datatype.DataType
-	})
+	explanation, err := buildMigrationComplexityExplanation(source.DBType, assessmentReport, "")
```
makalaaneesh, line 188: Why are we building again? It should already be available in the assessment report, right?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -156,87 +185,68 @@ func packAndSendAssessMigrationPayload(status string, errMsg string) {
...
+	for _, issue := range assessmentReport.Issues {
+		obfuscatedIssue := callhome.AssessmentIssueCallhome{
+			Category:              issue.Category,
+			CategoryDescription:   issue.CategoryDescription,
+			Type:                  issue.Type,
+			Name:                  issue.Name,
+			Description:           issue.Description,
+			Impact:                issue.Impact,
+			ObjectType:            issue.ObjectType,
+			ObjectName:            constants.OBFUSCATE_STRING,
```
makalaaneesh, line 203: let's remove ObjectName from the struct? We will always obfuscate it, right? 
Even for sqlstatement, maybe in future we will send, but for now, let's just not send it?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -156,87 +185,68 @@ func packAndSendAssessMigrationPayload(status string, errMsg string) {
...
-					TotalOccurrences: 1,
-				})
-			}
-		} else {
-			var objects []string
-			if slices.Contains(featuresToSendObjectsToCallhome, feature.FeatureName) {
-				objects = lo.Map(feature.Objects, func(o ObjectInfo, _ int) string {
-					return o.ObjectName
-				})
+		for _, sensitiveDescription := range descriptionsIncludingSensitiveInformationToCallhome {
```
makalaaneesh, line 215: We had discussed that it does not make sense to send description, right? 🤔  
Do we see any value in sending "description"? It just contains a more verbose explanation of the issue type + sensitive object names. It's more for the user, not helpful when collecting data.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -156,87 +185,68 @@ func packAndSendAssessMigrationPayload(status string, errMsg string) {
...
+				obfuscatedIssue.Description = sensitiveDescription
+			} else {
+				match, err := utils.MatchesFormatString(sensitiveDescription, obfuscatedIssue.Description)
+				if match {
+					obfuscatedIssue.Description, err = utils.ObfuscateFormatDetails(sensitiveDescription, obfuscatedIssue.Description, constants.OBFUSCATE_STRING)
+				}
+				if err != nil {
+					log.Errorf("error while matching issue description with sensitive descriptions: %v", err)
+					obfuscatedIssue.Description = constants.OBFUSCATE_STRING
+				}
```
priyanshi-yb, line 226: I think we don't need to send the description at all in the assessment issues as the `issue.Name/issue.Type` should be sufficient but in the cases this is not sufficient like `index on complex datatypes, etc..` we should not worry about sending this currently as we were not sending this before as well and we should separate out issues to have `issue.Type/Name` differentiate that.
cc: @makalaaneesh

priyanshi-yb: we did this for analyze as we don't want to regress there with less information but I think here we can go without descriptions as well.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -156,87 +185,68 @@ func packAndSendAssessMigrationPayload(status string, errMsg string) {
...
+		obfuscatedIssue := callhome.AssessmentIssueCallhome{
+			Category:              issue.Category,
+			CategoryDescription:   issue.CategoryDescription,
+			Type:                  issue.Type,
+			Name:                  issue.Name,
+			Description:           issue.Description,
+			Impact:                issue.Impact,
+			ObjectType:            issue.ObjectType,
+			ObjectName:            constants.OBFUSCATE_STRING,
+			SqlStatement:          constants.OBFUSCATE_STRING, // TODO(future): we can obfuscate sensitive info in SQL statement
```
priyanshi-yb, line 204: I am not sure if that would be easily possible to obfuscate the SQL statement and if there is any need to do that.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -156,87 +185,68 @@ func packAndSendAssessMigrationPayload(status string, errMsg string) {
...
-	groupedByConstructType := lo.GroupBy(assessmentReport.UnsupportedQueryConstructs, func(q utils.UnsupportedQueryConstruct) string {
-		return q.ConstructTypeName
-	})
-	countByConstructType := lo.MapValues(groupedByConstructType, func(constructs []utils.UnsupportedQueryConstruct, _ string) int {
-		return len(constructs)
-	})
+		// allowing object name for unsupported extension issue type
+		if issue.Type == UNSUPPORTED_EXTENSION_ISSUE_TYPE {
+			// object name(i.e. extension name here) might be qualified with schema so just taking the last part
+			obfuscatedIssue.ObjectName = strings.Split(issue.ObjectName, ".")[len(strings.Split(issue.ObjectName, "."))-1]
```
priyanshi-yb, line 212: generally extension name is not qualified as the CREATE EXTENSION DDL has a separate option for mentioning the schema name and I believe we are not using the schema name. 
```
CREATE EXTENSION IF NOT EXISTS citext WITH SCHEMA public;
```

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: we were sending this info before, i just moved it a little above in this function, nothing else changed.

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: same for this, just did a bit of reordering.

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: the reason is, we will this twice... for json and html reports both, at the time of report generation.
whatever happens later will be stored in the struct's field.

So just to be on safer side i am building it again.

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: no, you remember we discussed that some info in this might be required.
Like column name or object name is not requrid
But object type / datatype like citext etc... can also be there which can be useful.. 
Although this might not be a good way to send that type of info to callhome.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -156,87 +185,68 @@ func packAndSendAssessMigrationPayload(status string, errMsg string) {
...
+			CategoryDescription:   issue.CategoryDescription,
+			Type:                  issue.Type,
+			Name:                  issue.Name,
+			Description:           issue.Description,
+			Impact:                issue.Impact,
+			ObjectType:            issue.ObjectType,
+			ObjectName:            constants.OBFUSCATE_STRING,
+			SqlStatement:          constants.OBFUSCATE_STRING, // TODO(future): we can obfuscate sensitive info in SQL statement
+			DocsLink:              issue.DocsLink,
+			MinimumVersionFixedIn: issue.MinimumVersionFixedIn,
```
priyanshi-yb, line 206: Maybe we can also skip sending these two as they are static information.

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: yeah for object name we will probably always, for sql statement we might start obfuscating and sending specific parts.

Just to be future proof i thought to have it there, and the queries we can ignore these fields for now??

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: only problem is we won't know have the info like index access method, datatype which is unsupported in the issue... anywhere...

The description after obfuscation of objectname can even act as a key for case like `index on complex datatypes 'citext'`

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: will remove ObjectName and SqlStatement field

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: will remove ObjectName and SqlStatement field

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: Agreed but based on what we had discussed earlier, the solution would be for us to define separate issues (index-not-supported-json, index-not-supported-citext, etc). 

Only in cases where that's not possible (for instance, extensions), we could somehow pass that information along using the TYPE itself, or a different field like ObjectName

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -156,87 +185,65 @@ func packAndSendAssessMigrationPayload(status string, errMsg string) {
...
+		}
 
-	groupedByConstructType := lo.GroupBy(assessmentReport.UnsupportedQueryConstructs, func(q utils.UnsupportedQueryConstruct) string {
-		return q.ConstructTypeName
-	})
-	countByConstructType := lo.MapValues(groupedByConstructType, func(constructs []utils.UnsupportedQueryConstruct, _ string) int {
-		return len(constructs)
-	})
+		// allowing object name for unsupported extension issue type
+		if issue.Type == UNSUPPORTED_EXTENSION_ISSUE_TYPE {
```
makalaaneesh, line 207: As part of https://yugabyte.atlassian.net/browse/DB-14476, we had decided to create separate  items for each extension type. Let's keep that behavior as is. Shall we modify the `Type` field to also include the extension name?

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: Ah, right. See, here's another reason why having different values for the same field (depending on report type) is not a good idea. `AssessmentReport.MigrationComplexityExplanation` should be properly defined in our struct (without including any html/json specific logic). 
- If we want any HTML specific logic, it should be in the template. 
- If we want any json-specific logic, it should be in a custom marshaller.
- If we want any callhome specific logic, it should be in the callhome layer (this function)

Pls add a comment here explaining why we're re-building. And let's clean this up soon

sanyamsinghal: yeah agree, only limitation which stopped me was that to the template we can pass only one struct.
In the original assessment report struct, i wanted to avoid a new child struct, just for the html report case.

makalaaneesh: Ah I see. It might not be such a bad idea to have a chid struct actually. But alternatively, you could just have all the fields in the struct directly, and then mark them as json:"-" if you don't want them in the json

makalaaneesh (approved):
LGTM!

PR #2241, opened 2025-01-24
Added warning to import.go for target-db-schema flag

atharva-kurhade-yb:
Added a warning to import.go file.

### DIFF
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -279,7 +279,7 @@ func validateTargetPortRange() {
 
 func validateTargetSchemaFlag() {
 	// do not check for source and source-replica imports
-	if !(importerRole == SOURCE_REPLICA_DB_IMPORTER_ROLE || importerRole == SOURCE_DB_IMPORTER_ROLE || importerRole == IMPORT_FILE_ROLE) {
+	if !slices.Contains([]string{SOURCE_REPLICA_DB_IMPORTER_ROLE, SOURCE_DB_IMPORTER_ROLE, IMPORT_FILE_ROLE}, importerRole) {
 		if tconf.Schema != "" && sourceDBType == "postgresql" {
 			utils.ErrExit("Error: --target-db-schema flag is not valid for export from 'postgresql' db type")
 		}
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -278,7 +278,9 @@ func validateTargetPortRange() {
 }
 
 func validateTargetSchemaFlag() {
-	// we want the check to run for import data to target and import schema commands
+	// we want to run this check only for import-data-to-target and import-schema commands.
+	// This is not applicable for import-data-to-source-replica (validateFFDBSchemaFlag)/import-data-to-source (no ability to pass schema).
+	// For import-data-file, we allow this flag and source is PG(dummy)
 	if !slices.Contains([]string{SOURCE_REPLICA_DB_IMPORTER_ROLE, SOURCE_DB_IMPORTER_ROLE, IMPORT_FILE_ROLE}, importerRole) {
 		if tconf.Schema != "" && sourceDBType == "postgresql" {
 			utils.ErrExit("Error: --target-db-schema flag is not valid for export from 'postgresql' db type")
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -279,14 +279,10 @@ func validateTargetPortRange() {
 
 func validateTargetSchemaFlag() {
 	// do not check for source and source-replica imports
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/import.go
```diff
@@ -285,6 +285,10 @@ func validateTargetSchemaFlag() {
 		}
```
makalaaneesh, line 285: @priyanshi-yb do we really need the `if` check on line 282? 
- For import-data-to-source-replica, the same limitation applies, right? 
- For import-data-to-source, there's no way for the user to specify target-db-schema in the first place. so, it should be okay? 

Can we do the tconf.Schema not empty check in all scenarios?

priyanshi-yb: > For import-data-to-source-replica, the same limitation applies, right?

yes user should not be allowed to pass schema names and should be same as source

> Can we do the tconf.Schema not empty check in all scenarios?

I think we can't do without importRole because this `validateTargetSchemaFlag()` is run in the commands - `import data` and `import data file`
and for the source-replica there is separate function `validateFFDBSchemaFlag` for this similar flag.

makalaaneesh: so we want this check to run in all scenarios except `import data file` and `import data to source replica`. Does that make sense?

priyanshi-yb: yeah and I think that is only import-data then.

makalaaneesh: and import-schema, for which this task was created :)

makalaaneesh: @atharva-kurhade-yb , let's do the following: 
- Remove the new check that you've added, we can re-use the existing check with a couple of modifications
- in the existing check (line 282), 
   - run the check only if `importerRole` is NOT one of (SOURCE_REPLICA_DB_IMPORTER_ROLE, SOURCE_DB_IMPORTER_ROLE, IMPORT_FILE_ROLE)

priyanshi-yb: > and import-schema, for which this task was created :)

oh yes, mybad I didn't realize that, I was just thinking about data commands 😅

makalaaneesh: It's definitely a bit weird to rely on importerRole for this because in case of import-schema it will be empty. But I think it might be okay for now.

atharva-kurhade-yb: So with this change will it throw an error with import schema instead of a warning?

makalaaneesh (approved):
LGTM!

yb-voyager/cmd/import.go
```diff
@@ -279,7 +279,7 @@ func validateTargetPortRange() {
 
 func validateTargetSchemaFlag() {
 	// do not check for source and source-replica imports
-	if importerRole == TARGET_DB_IMPORTER_ROLE {
+	if !(importerRole == SOURCE_REPLICA_DB_IMPORTER_ROLE || importerRole == SOURCE_DB_IMPORTER_ROLE || importerRole == IMPORT_FILE_ROLE) {
```
priyanshi-yb, line 282: nit: use `!slices.Contains([]string{SOURCE_REPLICA_DB_IMPORTER_ROLE, SOURCE_DB_IMPORTER_ROLE, IMPORT_FILE_ROLE}, importerRole)`

priyanshi-yb (approved):
LGTM with a nit

yb-voyager/cmd/import.go
sanyamsinghal, line 282: 1. Why have we removed the check on TARGET_DB_IMPORTER_ROLE here?
2. Can we please add a comment here, from the discussion/decision you guys made. 
cc @makalaaneesh @priyanshi-yb

sanyamsinghal (changes requested):

yb-voyager/cmd/import.go
atharva-kurhade-yb: I have added the comment specifying why the check was changed.

yb-voyager/cmd/import.go
```diff
@@ -278,7 +278,7 @@ func validateTargetPortRange() {
 }
 
 func validateTargetSchemaFlag() {
-	// do not check for source and source-replica imports
+	// we want the check to run for import data to target and import schema commands
```
makalaaneesh, line 281: @atharva-kurhade-yb let's add some more detail just so that we all don't forget this in future:  
`we want to run this check only for import-data-to-target and import-schema commands. This is not applicable for import-data-to-source-replica (validateFFDBSchemaFlag) /import-data-to-source (no ability to pass schema). For import-data-file,  we allow this flag and source is PG(dummy)`

sanyamsinghal (approved):
LGTM

PR #2249, opened 2025-01-27
Added code to prompt user when target-db-version flag not mentioned

atharva-kurhade-yb:
Changed the assessMigrationCommand.go file to prompt user when the target-db-version flag is not mentioned in the assessmigration/analyze-schema command.

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -1736,19 +1736,24 @@ func validateAssessmentMetadataDirFlag() {
 }
 
 func validateAndSetTargetDbVersionFlag() error {
+	isMentioned := true
 	if targetDbVersionStrFlag == "" {
-		targetDbVersion = ybversion.LatestStable
-		return nil
+		fmt.Printf("Specific YugabyteDB not mentioned. Defaulting to latest stable YugabyteDB version.\n")
+		isMentioned = false
 	}
+
 	var err error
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1736,19 +1736,24 @@ func validateAssessmentMetadataDirFlag() {
 }
 
 func validateAndSetTargetDbVersionFlag() error {
```
makalaaneesh, line 1738: Overall, there seem to be two cases where we prompt: 
1. User has not specified any value in target-db-version
2. User has specified some value but that target-db-version series in unsupported. 

I think the message that we would want to display to the user would be different in both these scenarios, so I think it's okay to have separate prompts for both of these. 

1. For the first case, let's just modify the `if targetDbVersionStrFlag == "" {` block. in the prompt, we can mention "No target-db-version has been specified. Do you want to continue with the latest stable YugabyteDB version xyz "
2. For the second case, we can keep the code as is. I don't believe we need any change.

makalaaneesh (changes requested):

yb-voyager/cmd/assessMigrationCommand.go
atharva-kurhade-yb: I have added the changes.

makalaaneesh (approved):
LGTM!

PR #2332, opened 2025-02-18
Checking mysql-community-devel package instead of mysql-devel package in the airgapped installation script when the OS is RHEL9

ShivanshGahlot:
Why is it needed?
The mysql-devel package is not available in RHEL9. Therefore, we need to install mysql-community-devel instead. Because of this reason, in case of RHEL9 airgapped installation we need to check whether the mysql-community-devel package is installed or not instead of the mysql-devel package.

No

Manually

### DIFF
--- a/installer_scripts/install-voyager-airgapped.sh
+++ b/installer_scripts/install-voyager-airgapped.sh
@@ -614,6 +614,18 @@ check_yum_package_version() {
 check_yum_dependencies() {
     for requirement in "${centos_yum_package_requirements[@]}"; do
         IFS='|' read -r package version_type required_version <<< "$requirement"
+       
+        # In case of rhel9 the mysql-devel package is not available. However, mysql-community-devel is available. So check for that instead.
+        # Check if OS is rhel9
+        version=$(source /etc/os-release; echo "$VERSION_ID")
+	    # Extract only the major version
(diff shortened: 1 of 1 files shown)


installer_scripts/install-voyager-airgapped.sh
```diff
@@ -614,6 +614,18 @@ check_yum_package_version() {
...
+        # In case of rhel9 the mysql-devel package is not available. However, mysql-community-devel is available. So check for that instead.
+        # Check if OS is rhel9
+        version=$(source /etc/os-release; echo "$VERSION_ID")
+	    # Extract only the major version
+	    majorVersion=$(echo $version | cut -d '.' -f 1)
+        if [ "$majorVersion" -eq 9 ]; then
+            if [[ "$package" == "mysql-devel" ]]; then
+                package="mysql-community-devel"
+            fi
+        fi
```
shubham-yb, line 627: Maybe we can simplify:

```
if [[ "$majorVersion" == "9" && "$package" == "mysql-devel" ]]; then
    package="mysql-community-devel"
fi
```

shubham-yb (approved):

installer_scripts/install-voyager-airgapped.sh
ShivanshGahlot: Done

PR #2348, opened 2025-02-20
[DB-15464][Sizing Calc] buffer to choose higher vCPU instance type

shaharuk-yb:
Pre-calculate the required number of cores and use it for selecting the appropriate instance type recommendation. 
Current sizing calculator decides final recommendation based on the number of nodes required in later stage( which is calculated using the required number of cores in the first place). Hence, the better way is to use required number of cores. 

example: Take into consideration following two sizing recommendation are suitable. 
- 8 vCPU setup needing 296 cores gives 37 nodes. 
- 16 vCPU setup needing 290 cores gives 19 nodes. 
But, currently, in pickBestRecommendation, while recalculating the cores, it uses ```nodes * vCPU```. Thus giving, 
8 * 37 ==> 296
16 * 19 ==> 304 
Even though the required cores were 290 but because of reverse calc, it says 304 cores required for 16 vCPU setup and hence current sizing calculator suggest 8vCPU setup as it favours lesser cores(even though the number of required cores originally were higher).

This change directly considers required cores instead on recalculating it. Giving better recommendation.  

NA

Manually and with CI jobs. 


sanyamsinghal (approved):
LGTM. Thanks for the clear PR description.

PR #2354, opened 2025-02-24
Restrict live migration to single table-list and added guardrails for that around table-list flags 

priyanshi-yb:
1. During resumption in the command `export data from source` we are now using the stored table list in MSR `TableListExportedFromSource` and `SourceExportedTableListWithLeafPartitions` based on the source type as for PostgreSQL source we allow filtering the leaf partitions as well so storing them in MSR.
2.  In the command `export data from target,` we use the stored table list in MSR `TableListExportedFromSource` to populate the leaf partitions from the target for all the partitioned tables. We then store this information in the `TargetExportedTableListWithLeafPartitions`  for further runs to avoid fetching new leaf partitions from the target.
3. Added guardrails around the `table-list` and `exclude-table-list`  flags to check for any discrepancy between these on resumption and the initial lists.
    1. If any new tables are passed via these flags, commands error out saying `Unknown tables found`.
    2. If any table-list changes are found, commands report them on the console output of the commands, like `Missing tables ...` / `Extra tables found ...`  than the initial list and command error out saying `Re-run with the table-list passed in the initial run`
4. For the case of the partition, if any new leaf tables are added during the migration on the partitioned tables that are getting migrated. On resumption, we will give a prompt to the user that some new leaf partitions are found and will not be considered for the migration if they are fine to continue or else restart the migration

Yes, as mentioned above there will be user-facing changes for some guardrails around table-list etc.. now user won't be able to change the table-list during the live migration

Behaviour in these scenarios 

- New table is added on database either source/target - 

 **With or without table-list flags:** export data from source/target commands will be unaware of newly added tables and won’t consider it either during the run or re-run.
If a newly added table is passed through the table-list/exclude-table-list - commands error out saying unknown table, valid tables names are […].

```sql
➜ yb-voyager export data --export-dir ... --export-type snapshot-and-changes  --table-list test_partitions_quences                           
...
Continuing streaming from where we left off...
Unknown table names [test_partitions_quences] in the include list
Valid table names are: [public.foo_bar public.sales_region public.test_seq1 public.try public.sequence_check1 public.london public.sydney public.boston public.test_partitions_sequences_b public.test_partitions_sequences_l public.test_partitions_sequences_s public.sequence_check2 public.se..]
```

- New leaf table is added on partitioned table in database either source/target - 

**With or without table-list flags**: 
         -- During the run, export data from source/target commands will be unaware of newly added leaf tables and won’t consider it. 
        -- On re-run, export data from source/target commands will report to the user that new leaf tables are detected for partitioned tables with below msg -

```sql
Detected new partition tables for the following partitioned tables. These will not be considered during migration:
Root table: public.test_partitions_sequences, new leaf partitions: public.test_partitions_sequences_h
Do you want to continue?? [Y/N]: Y
num tables to export: 1
table list for data export: [test_partitions_sequences (test_partitions_sequences_s)]
```

If a new leaf partition table is passed through the table-list/exclude-table-list flags - commands error out saying unknown table; valid tables names are […].

```sql
➜  yb-voyager export data --export-dir .... --export-type snapshot-and-changes  --table-list test_partitions_sequences_h
....
Continuing streaming from where we left off...
Unknown table names [test_partitions_sequences_h] in the include list
Valid table names are: [public.foo_bar public.sales_region public.test_seq1 public.try public.sequence_check1 public.london public.sydney public.boston public.test_partitions_sequences_b public.test_partitions_sequences_l public.test_partitions_sequences_s public.sequence_check2 publi.....]
```

- Table-list changes during resumption via flags

 In case some tables are removed from the initial list in resumption - commands will report that there is discrepancy found and error out as mentioned below

```sql
  TARGET_DB_PASSWORD=*** yb-voyager export data from target --export-dir ... --table-list public."accounts_list_partitioned",public."empty_partition_table2",public."orders_interval_partition" --exclude-table-list empty_partition_table2
migrationID: 822cd9e3-00bc-4c7f-8d8c-195cdd3e24b7
Note: Live migration is a TECH PREVIEW feature.
export of data for source type as 'yugabytedb'
Continuing streaming from where we left off...
Changing the table list during live-migration is not allowed.
Missing tables in the current run compared to the initial list: empty_partition_table2_p_west,empty_partition_table2_p_east
Table list passed in the initial run of migration - [public.accounts_list_partitioned_p_northwest .... public.orders_interval_partition_interval_partition_less_than_2018 public.orders_interval_partition]. 
Re-run the command with the table list passed in the initial run of migration.
```

In case some tables are added from the initial list in resumption - commands will report that there is discrepancy found and error out as mentioned below

```sql
 TARGET_DB_PASSWORD=*** yb-voyager export data from target --export-dir /home/centos/code/yb-voyager/migtests/tests/oracle/partitions/oracle_partitions_fallb_export-dir --table-list public."accounts_list_partitioned",public."empty_partition_table2",public."orders_interval_partition",empty_partition_table ...
migrationID: 822cd9e3-00bc-4c7f-8d8c-195cdd3e24b7
Note: Live migration is a TECH PREVIEW feature.
export of data for source type as 'yugabytedb'
Continuing streaming from where we left off...
Changing the table list during live-migration is not allowed.
Extra tables in the current run compared to the initial list: empty_partition_table2_p_extra,empty_partition_table_p_west,empty_partition_table_p_east,empty_partition_table
Table list passed in the initial run of migration - [public.accounts_list_partitioned_p_northwest .... public.orders_interval_partition]. 
Re-run the command with the table list passed in the initial run of migration.
```

Added unit tests with test container PG source for various cases.
  1. Validating the table-list in the first-run
  2. validating guardrails check prompts in subsequent runs
  3. validating new leaf additions
  4. validating unknown table case in table-list via error exit (any dummy table name or new table added on source)

| Sizing DB                                                 | o |


yb-voyager/cmd/common.go
```diff
@@ -1040,6 +1040,9 @@ func storeTableListInMSR(tableList []sqlname.NameTuple) error {
 	}))
 	err := metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
 		record.TableListExportedFromSource = minQuotedTableList
```
makalaaneesh, line 1042: nit: `minQuotedTableList` -> `minQuotedTableListWithRoots`

yb-voyager/cmd/exportData.go
```diff
@@ -798,6 +807,200 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
 	return partitionsToRootTableMap, finalTableList
 }
 
+func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
+	// store table list after filtering unsupported or unnecessary tables
+	msr, err := metaDB.GetMigrationStatusRecord()
+	if err != nil {
+		utils.ErrExit("error fetching migration status record: %v", err)
```
makalaaneesh, line 814: let's return error properly instead of utils.ErrExit?

yb-voyager/cmd/exportData.go
```diff
@@ -798,6 +807,200 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+	if source.DBType == POSTGRESQL && msr.SourceExportedTableListWithLeafPartitions != nil {
+		//list with partitioned root table and all leafs
+		storedTableList = msr.SourceExportedTableListWithLeafPartitions
+	}
+	for _, t := range storedTableList {
+		var tuple sqlname.NameTuple
+		var err error
+		switch source.DBType {
+		case POSTGRESQL:
+			defaultSchemaName, _ := getDefaultSourceSchemaName()
```
makalaaneesh, line 836: I'm a little concerned about createing nameTuples without referring to NameRegistry for all tables just to handle the leaf partition case. 

Is there a way we can distinguish between the two, and use name registry for the normal tables and manually construct for the leaf partitions? can we use the partitionsToRootTableMap maybe?

yb-voyager/cmd/exportData.go
```diff
@@ -798,6 +807,200 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+			if err != nil {
+				utils.ErrExit("lookup for table name failed: %s: %v", t, err)
+			}
+		}
+		fullTableList = append(fullTableList, tuple)
+	}
+	finalTableList = fullTableList
+
+	var partitionsToRootTableMap map[string]string
+	switch source.DBType {
```
makalaaneesh, line 851: Is this entire block only to support upgrade scenario? Let's discuss this. I don't think we should put extra effort into supporting upgrade scenario for this new feature. We should just ensure that upgrade doesn't fail.

yb-voyager/cmd/exportData.go
```diff
@@ -798,6 +807,200 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+	//guardrails around the table-list in case of re-run
+
+	//TODO: fix later this registered list also contains sequence names right now
+	//so in case some include/exclude table names are not proper we will display this registeredList as valid table names
+	registeredListObjNames, err := namereg.NameReg.GetRegisteredTableList()
+	if err != nil {
+		utils.ErrExit("error getting registered list in name registry: %v", err)
+	}
+
+	registeredList := lo.Map(registeredListObjNames, func(obj *sqlname.ObjectName, _ int) sqlname.NameTuple {
```
makalaaneesh, line 900: Why aren't we directly getting NameTuples from namereg.GetRegisteredTableList?

yb-voyager/cmd/exportData.go
```diff
@@ -798,6 +807,200 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+		if !lo.ContainsBy(registeredList, func(tbl sqlname.NameTuple) bool {
+			return tbl.AsQualifiedCatalogName() == leaf
+		}) {
+			//If this leaf table is not registered in name registry then it is a newly added leaf tables
+			newLeafTables[value] = append(newLeafTables[value], leaf)
+		}
+	}
+
+	if len(lo.Keys(newLeafTables)) > 0 {
+		utils.PrintAndLog("Detected new partition tables for the following partitioned tables. These will not be considered during migration:")
```
makalaaneesh, line 954: Should we prompt?

yb-voyager/cmd/exportData.go
priyanshi-yb: > I'm a little concerned about createing nameTuples without referring to NameRegistry for all tables just to handle the leaf partition case.

We are not creating nameTuples for all cases, its only the leaf tables. If you that `getNameTupleFromObjectName` function implementation, it first check whether this object has any parent if yes then only create rest use the name registry one.

yb-voyager/cmd/exportData.go
priyanshi-yb: As discussed, we need to handle nametuple look up in this code by checking if its a leaf of not

yb-voyager/cmd/exportData.go
priyanshi-yb: Not sure if a prompt is required, I think there is nothing we want users to do here right? so this might just FYI?

yb-voyager/cmd/exportData.go
```diff
@@ -788,14 +844,238 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+	return firstRunTableWithLeafsAndRoots, partitionsToRootTableMap, nil
+}
+
+func getInitialTableList() (map[string]string, []sqlname.NameTuple, error) {
+	// store table list after filtering unsupported or unnecessary tables
+	msr, err := metaDB.GetMigrationStatusRecord()
+	if err != nil {
+		utils.ErrExit("error fetching migration status record: %v", err)
+	}
+	storedTableListNotAvailable := func() bool {
```
makalaaneesh, line 921: shouldn't we handle this for export-data-from-target as well?

yb-voyager/cmd/exportData.go
```diff
@@ -735,50 +742,99 @@ func reportUnsupportedTables(finalTableList []sqlname.NameTuple) {
...
+	//apply table list flags filter on the nameTupleTableListFromDB
+	tableListInFirstRun, err = applyTableListFlagsOnFullListAndAddLeafPartitions(nameTupleTableListFromDB, source.TableList, source.ExcludeTableList)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error applying table list flags on full table list: %v", err)
+	}
+
+	isTableListModified := len(sqlname.SetDifferenceNameTuples(nameTupleTableListFromDB, tableListInFirstRun)) != 0
 	if exporterRole == SOURCE_DB_EXPORTER_ROLE {
 		metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
 			if isTableListModified {
```
makalaaneesh, line 840: unrelated to your changes, but do we need to set this differently for export-data-from-source/target?

yb-voyager/cmd/exportData.go
```diff
@@ -788,14 +844,238 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+		registeredList = append(registeredList, tuple)
+	}
+
+	rootTables := make([]sqlname.NameTuple, 0)
+	for _, v := range partitionsToRootTableMap {
+		tuple, err := namereg.NameReg.LookupTableName(v)
+		if err != nil {
+			utils.ErrExit("look up failed for the table name: %s: %v", v, err)
+		}
+		rootTables = append(rootTables, tuple)
```
makalaaneesh, line 965: this can have duplicates right? not a problem I'm guessing?

makalaaneesh (commented):
Went over it at a high level, looks good! 
Will give it a more detailed look soon along with the tests.

yb-voyager/cmd/exportData.go
priyanshi-yb: we can not go to DB for the export-data-from target as such, because we need to use the table list that is used from export (source one), which is available in msr. So, handling export data from target even first run in the below func`retrieveFirstRunListAndPartitionsRoot` which handles target case as well

yb-voyager/cmd/exportData.go
priyanshi-yb: Nope, this is mostly used for offline migration and display purposes only

yb-voyager/cmd/exportData.go
priyanshi-yb: Yeah, in the very first implementation, I had a map but changed it to a list later. I'm not sure why I missed filtering for unique items, but it shouldn't be a problem.

yb-voyager/cmd/exportData_test.go
```diff
@@ -0,0 +1,513 @@
...
+		getNameTuple("p1.London"),
+		getNameTuple("p1.Boston"),
+		getNameTuple("public.datatypes1"),
+	}
+	fetchTableListFromSourceAndAssertResult(t, expectedPartitionsToRootMap3, expectedTableList3)
+}
+
+/*
+ TODO: 
+ added test assertions for below cases
```
makalaaneesh, line 278: all great ideas for tests!

yb-voyager/cmd/exportData_test.go
```diff
@@ -0,0 +1,513 @@
...
+  1. validate guardrails check prompts
+  2. validate new leaf additions
+  3. Unknown table in table-list error exit
+ */
+func TestTableListInSubsequenceRunOfExportDataBasic(t *testing.T) {
```
makalaaneesh, line 283: nit: subsequent

yb-voyager/cmd/exportData_test.go
```diff
@@ -0,0 +1,513 @@
...
+	ddlChanges := []string{
+		`CREATE TABLE public.test_ddl_change(id int, val text);`,                                                                 // Adding normal table
+		`CREATE TABLE public.test_partitions_sequences_p PARTITION OF public.test_partitions_sequences FOR VALUES IN ('Paris');`, // Adding a new leaf on the test_partititons_sequences table (included in table-list)
+	}
+
+	testPostgresSource.ExecuteSqls(ddlChanges...)
+
+	// getInitialTableList for subsequent run with  start-clean false and basic without table-list flags so no guardrails
+	startClean = false
+	getInitialTableistAndAssertExpectedResult(t, expectedTableList, expectedPartitionsToRootMap)
```
makalaaneesh, line 362: here, we're asserting that you get back the same table-list, even though we have added some tables on the source, right? 
Pls add that to he comment.

makalaaneesh, line 362: you added a new partition leaf above. This should error out no?

yb-voyager/cmd/exportData_test.go
```diff
@@ -0,0 +1,513 @@
...
+	if err != nil {
+		t.Fatalf("error updating msr: %v", err)
+	}
+
+	//case1: getInitialTableList for subsequent run with  start-clean false and basic with same table-list flags so no guardrails
+	startClean = false
+
+	getInitialTableistAndAssertExpectedResult(t, expectedTableList, expectedPartitionsToRootMap)
+
+	//case2: getInitialTableList for subsequent run with  start-clean false and basic with no table-list flags so no guardrails
```
makalaaneesh, line 472: If first run has table-list, and subsequent run does not have table-list, we should fail no?

priyanshi-yb: We don't fail; we use the stored one directly. Not sure if we should fail or not. Maybe we can have a prompt in this case as well saying using the initial table list ...

yb-voyager/cmd/exportData_test.go
priyanshi-yb: yeah

yb-voyager/cmd/exportData.go
```diff
@@ -552,12 +559,12 @@ func addLeafPartitionsInTableList(tableList []sqlname.NameTuple, ifTableListSet
 		allLeafPartitions := GetAllLeafPartitions(table)
 		prevLengthOfList := len(modifiedTableList)
 		switch true {
-		case len(allLeafPartitions) == 0 && rootTable != table: //leaf partition
+		case len(allLeafPartitions) == 0 && !rootTable.Equals(table): //leaf partition
```
makalaaneesh, line 562: Does our `NameTuple.Equals` work if the pointers are different, but the underlying values are the same? 
Is that what prompted you to make this change?

yb-voyager/cmd/exportData.go
```diff
@@ -576,7 +583,12 @@ func addLeafPartitionsInTableList(tableList []sqlname.NameTuple, ifTableListSet
 func GetRootTableOfPartition(table sqlname.NameTuple) (sqlname.NameTuple, error) {
 	parentTable := source.DB().ParentTableOfPartition(table)
 	if parentTable == "" {
-		return table, nil
+		//now we know its a root table so return the tuple from the nameregistry
```
makalaaneesh, line 586: can you give more context in the comment on why you need to lookup from nameregistry again? 
For my clarification: this is because the original tuples do not have the target-side of names, right?

yb-voyager/cmd/exportData.go
```diff
@@ -735,50 +747,105 @@ func reportUnsupportedTables(finalTableList []sqlname.NameTuple) {
...
-		//TODO: avoid duplicate call to this function and optimize this function later
-		_, excludeTableList, err = addLeafPartitionsInTableList(excludeTableList, true)
+	return tuple, nil
+}
+
+func applyTableListFlagsOnFullListAndAddLeafPartitions(fullTableList []sqlname.NameTuple, tableListViaFlag string, excludeTableListViaFlag string) ([]sqlname.NameTuple, error) {
+	var err error
+	var includeTableList, excludeTableList []sqlname.NameTuple
+
+	applyFilterAndAddLeafTable := func(flagList string, flagName string) ([]sqlname.NameTuple, error) {
```
makalaaneesh, line 794: This structure of applying include/exclude is relatively easy to understand, thanks 👍

yb-voyager/cmd/exportData.go
```diff
@@ -735,50 +747,105 @@ func reportUnsupportedTables(finalTableList []sqlname.NameTuple) {
...
+	includeTableList = fullTableList
+	if tableListViaFlag != "" {
+		//Apply include table list filter if present
+		includeTableList, err = applyFilterAndAddLeafTable(tableListViaFlag, "include")
+		if err != nil {
+			return nil, err
+		}
 	} else {
-		tableList = fullTableList
+		//this is only for filtering the non-leaf and non-root tables - the mid level partitioned table from fullTableList
```
makalaaneesh, line 819: To remove non-leaf and non-root, you're still calling `addLeafPartitionsInTableList`? That sounds a bit weird. Does that function ignore all leafs, and then re-add them?

yb-voyager/cmd/exportData.go
```diff
@@ -788,14 +855,258 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+
+		//In case of YB we need to check the leaf condititon for the tables in table list first and then return the nametuple
+		//via namereg or handcrafted one
+		fetchNameTupleFromNameRegDirectly = false
+		// On subsequent run after the first we will use the stored table with leaf partitions
+		storedTableList = msr.TargetExportedTableListWithLeafPartitions
+		partitionsToRootTableMap = msr.TargetRenameTablesMap
+
+		//For the first run of export data from target we use the TableListExportedFromSource (which has only root tables)
+		if len(msr.TargetExportedTableListWithLeafPartitions) == 0 || msr.TargetRenameTablesMap == nil {
```
makalaaneesh, line 893: nit: create a variable for this since this is used below as well.

yb-voyager/cmd/exportData.go
```diff
@@ -788,14 +855,258 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+			utils.ErrExit("look up failed for the table name: %s: %v", v, err)
+		}
+		rootTables = append(rootTables, tuple)
+	}
+	rootTables = lo.UniqBy(rootTables, func(t sqlname.NameTuple) string {
+		return t.ForKey()
+	})
+
+	// Finding all the partitions of all root tables part of migration, and report if there any new partitions added
+	rootToNewLeafTablesMap, err := detectNewLeafPartitionsOnPartitionedTables(rootTables, registeredList)
```
makalaaneesh, line 972: should we move this block below to where we are actually checking `rootToNewLeafTablesMap` and doing guardrails?

yb-voyager/cmd/exportData.go
```diff
@@ -788,14 +855,258 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+		// so no need to do the guardrails check and all
+		return partitionsToRootTableMap, firstRunTableWithLeafsAndRoots, nil
+	}
+
+	firstRunTableWithLeafParititons, currentRunTableListWithLeafPartitions, err := applyTableListFlagsOnCurrentAndRemoveRootsFromBothLists(registeredList, source.TableList, source.ExcludeTableList, rootToNewLeafTablesMap, rootTables, firstRunTableWithLeafsAndRoots)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error applying table list flags for current list and remove roots: %v", err)
+	}
+
+	if len(lo.Keys(rootToNewLeafTablesMap)) > 0 {
```
makalaaneesh, line 988: shouldn't this check be done before the `source.TableList == "" && source.ExcludeTableList == ""` if block?

yb-voyager/cmd/exportData.go
```diff
@@ -788,14 +855,258 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+	}
+
+	//Reporting the guardrail msgs only on leaf tables to be consistent so filtering the root table from both the list
+	guardrailsAroundFirstRunAndCurrentRunTableList(firstRunTableWithLeafParititons, currentRunTableListWithLeafPartitions)
+
+	return partitionsToRootTableMap, firstRunTableWithLeafsAndRoots, nil
+
+}
+
+func applyTableListFlagsOnCurrentAndRemoveRootsFromBothLists(registeredList []sqlname.NameTuple, tableListViaFlag string, excludeTableListViaFlag string, rootToNewLeafTablesMap map[string][]string, rootTables []sqlname.NameTuple, firstRunTableWithLeafsAndRoots []sqlname.NameTuple) ([]sqlname.NameTuple, []sqlname.NameTuple, error) {
```
makalaaneesh, line 1004: nit: new lines for the args. Hard to read otherwise.

yb-voyager/src/namereg/namereg.go
```diff
@@ -159,11 +164,46 @@ func (reg *NameRegistry) registerSourceNames() (bool, error) {
...
 		}
 		m[schemaName] = append(m[schemaName], seqNames...)
+		m1[schemaName] = append(m1[schemaName], seqNames...)
 	}
 	reg.SourceDBTableNames = m
+	reg.SourceDBSequenceNames = m1
 	return true, nil
 }
 
+func (reg *NameRegistry) GetRegisteredTableList() ([]*sqlname.ObjectName, error) {
```
makalaaneesh, line 174: add a comment here on why we implemented this (partitions)

yb-voyager/cmd/exportData.go
priyanshi-yb: Yes, it does the `DeepEqual` internally on the two nametuples.

yb-voyager/cmd/exportData.go
priyanshi-yb: Yes, right. The original tuples are hand-crafted ones without target names, so for the root, we shouldn't use these but instead get a proper one from the name registry.

yb-voyager/cmd/exportData.go
priyanshi-yb: `addLeafPartitionsInTableList ` has a boolean parameter `addAllLeafPartitions` in case this is false, that function will consider that table-list to be and exhaustive one already with all the tables (root, middle partition, leaf partitions, so on..) and it will just do the filtering to remove the middle level partitions and just the keep the leaf and root tables in the list.

yb-voyager/cmd/exportData_test.go
```diff
@@ -0,0 +1,660 @@
...
+		getNameTuple("p1.boston"),
+		getNameTuple("public.foreign_test"),
+	}
+
+	expectedExtraTables2 := []sqlname.NameTuple{}
+
+	assertGuardrailsChecksForMissingAndExtraTablesInSubsequentRun(t, expectedMissingTables2, expectedExtraTables2, firstRunTableList, rootTables)
+	getInitialTableistAndAssertExpectedResult(t, firstRunTableList, firstRunPartitionsToRootMap)
+
+	//case5: getInitialTableList for subsequent run with  start-clean false and basic with table-list and exclude-table-list flags
```
makalaaneesh, line 625: is passing both include and exclude list allowed today? i would have thought we should disallow it? in what situations does this help?

yb-voyager/cmd/exportData_test.go
```diff
@@ -0,0 +1,660 @@
...
+	})
+	if err != nil {
+		t.Fatalf("error updating msr: %v", err)
+	}
+
+	testCasesWithDifferentTableListFlagValuesTest1(t, expectedTableList, expectedPartitionsToRootMap)
+
+}
+
+func testCasesWithDifferentTableListFlagValuesTest1(t *testing.T, firstRunTableList []sqlname.NameTuple, firstRunPartitionsToRootMap map[string]string) {
```
makalaaneesh, line 571: nice different set of test-cases!
If/when Shubham finds some other cases, pls do add it here.

yb-voyager/cmd/exportData_test.go
```diff
@@ -0,0 +1,660 @@
...
+		getNameTuple("public.datatypes1"),
+	}
+
+	expectedExtraTables4 := []sqlname.NameTuple{
+		getNameTuple("p1.london"),
+	}
+
+	assertGuardrailsChecksForMissingAndExtraTablesInSubsequentRun(t, expectedMissingTables4, expectedExtraTables4, firstRunTableList, rootTables)
+	getInitialTableistAndAssertExpectedResult(t, firstRunTableList, firstRunPartitionsToRootMap)
+}
```
makalaaneesh, line 660: test idea: ensure that 
first run: `foo*` (foo1, foo2)
second run : `foo1,foo2` does not result in any guardrails thrown.

yb-voyager/cmd/exportData_test.go
priyanshi-yb: yes we allow, it helps in scenario where:
tables on source - ab1,abc2,abc3,abc4,abc5, <few other tables>
table-list abc* (meaning all the abc prefixed tables)
exclude-table-list abc5 (excluding the abc5)

yb-voyager/cmd/exportData.go
priyanshi-yb: we use the rootToNewLeafTablesMap to calculate the currentList

yb-voyager/cmd/exportData.go
priyanshi-yb: moved it in the `detectNewLeafPartitionsOnPartitionedTable` function back.

yb-voyager/cmd/exportData.go
```diff
@@ -788,14 +855,258 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+	})
+
+	// Finding all the partitions of all root tables part of migration, and report if there any new partitions added
+	rootToNewLeafTablesMap, err := detectNewLeafPartitionsOnPartitionedTables(rootTables, registeredList)
+	if err != nil {
+		return nil, nil, fmt.Errorf("detecting new leaf tables on the partitioned tables: %v", err)
+	}
+
+	if source.TableList == "" && source.ExcludeTableList == "" {
+		//which mean no table-lists are passed in subsequent runs
```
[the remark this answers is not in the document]
priyanshi-yb: Hmm, okay let me remove this condition altogether and let the guardrails function do its of job of reporting the discrepancy.

yb-voyager/cmd/exportData_test.go
priyanshi-yb: yeah, I had it in mind for adding more tests around glob pattern support in the table list flags, will add it in a separate PR.

yb-voyager/cmd/exportData.go
```diff
@@ -788,14 +858,262 @@ func getInitialTableList() (map[string]string, []sqlname.NameTuple) {
...
+		if lo.ContainsBy(rootTables, func(r sqlname.NameTuple) bool {
+			return t.Equals(r)
+		}) {
+
+			//Remove root table
+			continue
+		}
+
+		//Filtering the new leaf tables if present in filteredListWithoutRootTable as we have reported above
+		if isNewLeafTable(t) {
```
makalaaneesh, line 1027: Why/how would the new leaf appear in currentRunTableListFilteredViaFlags?

priyanshi-yb: While applying include/exclude flags, we also add leaf partitions in case only the root is passed in flags.

yb-voyager/cmd/exportData.go
makalaaneesh: nit: write a separate specific function to remove the non-leaf/non-root tables. 
Add a TODO if defering.

yb-voyager/cmd/exportData_test.go
priyanshi-yb: Ticket for having a guardrail check during the ongoing live migration for new leaf tables https://github.com/yugabyte/yb-voyager/issues/2356

makalaaneesh (approved):
LGTM! pls update the PR description to reflect error vs prompt and if any other changes

PR #2371, opened 2025-03-03
Fix: nil pointer exception while sending the callhome payload for import data to source/source-replica

### LINKED ISSUE #2364: Fix "invalid memory address or nil pointer dereference" error during sending callhome on exit in import data to source
Jira Link: [DB-15568](https://yugabyte.atlassian.net/browse/DB-15568)

priyanshi-yb:
Fixing the nil pointer exception while sending the call home payload for exit/error case during initial phase of import data to source/source-replica. 
fixes #2364 

No

Manually

### DIFF
--- a/yb-voyager/cmd/importDataToSource.go
+++ b/yb-voyager/cmd/importDataToSource.go
@@ -98,7 +98,9 @@ func packAndSendImportDataToSourcePayload(status string, errorMsg string) {
 
 	sourceDBDetails := callhome.SourceDBDetails{
 		DBType:    tconf.TargetDBType,
-		DBVersion: targetDBDetails.DBVersion,
+	}
+	if targetDBDetails != nil {
+		sourceDBDetails.DBVersion = targetDBDetails.DBVersion
 	}
 	payload.SourceDBDetails = callhome.MarshalledJsonString(sourceDBDetails)
 
@@ -112,7 +114,7 @@ func packAndSendImportDataToSourcePayload(status string, errorMsg string) {
 
 	importDataPayload.Phase = importPhase
 
-	if importPhase != dbzm.MODE_SNAPSHOT {
+	if importPhase != dbzm.MODE_SNAPSHOT && statsReporter != nil {
 		importDataPayload.EventsImportRate = statsReporter.EventsImportRateLast3Min
 		importDataPayload.TotalImportedEvents = statsReporter.TotalEventsImported
 	}
--- a/yb-voyager/cmd/importDataToSourceReplica.go
+++ b/yb-voyager/cmd/importDataToSourceReplica.go
@@ -100,9 +100,11 @@ func packAndSendImportDataToSrcReplicaPayload(status string, errorMsg string) {
 	payload.MigrationType = LIVE_MIGRATION
 
 	sourceDBDetails := callhome.SourceDBDetails{
-		DBType:    tconf.TargetDBType,
-		DBVersion: targetDBDetails.DBVersion,
-		Role:      "replica",
+		DBType: tconf.TargetDBType,
+		Role:   "replica",
+	}
+	if targetDBDetails != nil {
+		sourceDBDetails.DBVersion = targetDBDetails.DBVersion
 	}
 	payload.SourceDBDetails = callhome.MarshalledJsonString(sourceDBDetails)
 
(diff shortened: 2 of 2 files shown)


yb-voyager/cmd/importDataToSource.go
```diff
@@ -112,7 +114,7 @@ func packAndSendImportDataToSourcePayload(status string, errorMsg string) {
 
 	importDataPayload.Phase = importPhase
 
-	if importPhase != dbzm.MODE_SNAPSHOT {
+	if importPhase != dbzm.MODE_SNAPSHOT && statsReporter != nil {
```
sanyamsinghal, line 117: what is the case when statsReporter will be null?

yb-voyager/cmd/importDataToSourceReplica.go
```diff
@@ -100,9 +100,11 @@ func packAndSendImportDataToSrcReplicaPayload(status string, errorMsg string) {
 	payload.MigrationType = LIVE_MIGRATION
 
 	sourceDBDetails := callhome.SourceDBDetails{
-		DBType:    tconf.TargetDBType,
-		DBVersion: targetDBDetails.DBVersion,
```
sanyamsinghal, line 104: same comment as above; use ternary.

yb-voyager/cmd/importDataToSource.go
priyanshi-yb: It can be nil until it is initialized. If there is any failure or exit before that, this function will try to send the payload, and we can get a nil pointer exception in such a case.

yb-voyager/cmd/importDataToSource.go
```diff
@@ -98,7 +98,9 @@ func packAndSendImportDataToSourcePayload(status string, errorMsg string) {
 
 	sourceDBDetails := callhome.SourceDBDetails{
 		DBType:    tconf.TargetDBType,
-		DBVersion: targetDBDetails.DBVersion,
+	}
+	if targetDBDetails != nil {
+		sourceDBDetails.DBVersion = targetDBDetails.DBVersion
```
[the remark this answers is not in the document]
priyanshi-yb: can't do this; it doesn't work if we do nil check in this function as it still tries to run the if clause. refer https://go.dev/play/p/sgeATTvMgMP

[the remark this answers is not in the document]
sanyamsinghal: Interesting. 
So this is because Go follows strict/eager evaluation of function argument

yb-voyager/cmd/importDataToSource.go
sanyamsinghal: https://www.honeybadger.io/blog/go-exception-handling/

makalaaneesh: @sanyamsinghal if I understood you correctly, you're suggesting that we have some sort of a try-catch for the exit handlers as a safety net so that they don't panic, is that correct? 

If so, atexit already sort of does that. Check this out:
```
func runHandler(handler func()) {
	defer func() {
		if err := recover(); err != nil {
			fmt.Fprintln(os.Stderr, "error: atexit handler error:", err)
		}
	}()

	handler()
}
```

priyanshi-yb: Yeah right, in this case I am fixing in this PR, it wasn't panicking with stack trace, it was just erroring with below msg
```
error: atexit handler error: runtime error: invalid memory address or nil pointer dereference
```

sanyamsinghal: @makalaaneesh @priyanshi-yb  what i was recommending is that we don't even error out if callhome payload sending failing.
Maybe just log that error, or max to max display on console.

Any failure in sending payload to callhome should not result in an error is what I think

priyanshi-yb: > Maybe just log that error, or max to max display on console.

exit handler also displays the msg on the console for any failure, and the issue we hit is for the exit case only.

sanyamsinghal: maybe i am missing something, lets discuss this overall tomorrow.

sanyamsinghal (approved):

PR #2389, opened 2025-03-07
Filtering the sequences based on table list and using two different maps in namereg for sequences and tables while lookup

priyanshi-yb:
1. Filtering sequences for source PG for live and offline based on table-list migrating in the workflow
2. Storing the Column to Sequence mapping used via debezium for the restoring sequences in MSR and retrieving it in the subsequent runs 
3. Changed the namereg lookup part to combine the two maps for looking up the table/sequence names from two maps SourceDBTableNames/YBTableNames and SourceDBSequenceNames/YBSequenceNames
4. Added the offline tests for various cases sequences not attached to tables, sequences with table-list running for both pg_dump/debezium cases in the pg/sequences test.  
5. Removed PG9 workflow as not supporting the source <PG11 officiallly.  

Yes, the Resume values of sequences on the tables that are not getting migrated or that are not attached to any tables are not restored during Live migration and Offline migration(source PG only). 

Added end-to-end-tests for offline 


yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -236,6 +234,31 @@ func generateOrGetStreamIDForYugabyteCDCClient(config *dbzm.Config) error {
...
+func fetchOrRetrieveColToSeqMap(msr *metadb.MigrationStatusRecord, tableList []sqlname.NameTuple) map[string]string {
+	var storedColToSeqMap map[string]string
+	//fetching the stored one in the MSR
+	switch exporterRole {
+	case SOURCE_DB_EXPORTER_ROLE:
+		storedColToSeqMap = msr.SourceColumnToSequenceMapping
+	case TARGET_DB_EXPORTER_FB_ROLE, TARGET_DB_EXPORTER_FF_ROLE:
+		storedColToSeqMap = msr.TargetColumnToSequenceMapping
+	}
+	if len(lo.Keys(storedColToSeqMap)) > 0 {
```
makalaaneesh, line 246: what if the source actually does not have any sequences? this will be empty and we will go to source? 
Can we probably do a NULL check instead of len(keys)?

yb-voyager/src/namereg/namereg.go
```diff
@@ -290,6 +288,19 @@ schema1.foobar, schema1."foobar", schema1.FooBar, schema1."FooBar", schema1.FOOB
 (fuzzy-case-match) schema1.fooBar, schema1."fooBar"
 */
 func (reg *NameRegistry) LookupTableName(tableNameArg string) (sqlname.NameTuple, error) {
+
+	sourceObjectNameMap := reg.SourceDBTableNames
+	targetObjectNameMap := reg.YBTableNames
+	if len(lo.Keys(reg.SourceDBSequenceNames)) > 0 {
+		for s, objs := range reg.SourceDBSequenceNames {
+			sourceObjectNameMap[s] = append(sourceObjectNameMap[s], objs...)
```
makalaaneesh, line 296: I think this will overwrite reg.SourceDBTableNames. a map is just a pointer I believe, so even if you copy, when you modify an entry in the map, it will update the underlying map structure.

https://go.dev/play/p/xN9dzHoBSAT

yb-voyager/src/namereg/namereg.go
```diff
@@ -290,6 +288,19 @@ schema1.foobar, schema1."foobar", schema1.FooBar, schema1."FooBar", schema1.FOOB
 (fuzzy-case-match) schema1.fooBar, schema1."fooBar"
 */
 func (reg *NameRegistry) LookupTableName(tableNameArg string) (sqlname.NameTuple, error) {
+
+	sourceObjectNameMap := reg.SourceDBTableNames
+	targetObjectNameMap := reg.YBTableNames
+	if len(lo.Keys(reg.SourceDBSequenceNames)) > 0 {
```
makalaaneesh, line 294: Is there any use-case for having to lookup both table names and sequence names together? should we have a separate function instead - `LookupSequenceName` ? Makes it more clear.

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -351,7 +351,7 @@ func (yb *YugabyteDB) getExportedColumnsListForTable(exportDir, tableName string
 }
 
 // GetAllSequences returns all the sequence names in the database for the given schema list
-func (yb *YugabyteDB) GetAllSequences() []string {
+func (yb *YugabyteDB) GetAllSequences(_ []sqlname.NameTuple) []string {
```
makalaaneesh, line 354: Shouldn't we also change the implementation for this to use FETCH_SEQUENCES_FOR_TABLE_LIST?

yb-voyager/src/srcdb/postgres.go
```diff
@@ -471,12 +471,36 @@ func GetAbsPathOfPGCommandAboveVersion(cmd string, sourceDBVersion string) (path
...
+   AND a.attnum = ad.adnum
+WHERE 
+  seq.relkind = 'S'
+  AND (tab_ns.nspname || '.' || tab.relname) IN (%s);`
+
 // GetAllSequences returns all the sequence names in the database for the given schema list
-func (pg *PostgreSQL) GetAllSequences() []string {
-	schemaList := pg.checkSchemasExists()
-	querySchemaList := "'" + strings.Join(schemaList, "','") + "'"
+func (pg *PostgreSQL) GetAllSequences(tableList []sqlname.NameTuple) []string {
```
makalaaneesh, line 497: I think good opportunity to  add unit tests using testcontainer for `GetAllSequences`. I asked copilot for different ways a sequence can be associated with a table; we should test all these scenarios :

1. Using SERIAL or BIGSERIAL Data Types
2. Using CREATE SEQUENCE and DEFAULT: (nextval)
3. Using ALTER SEQUENCE and OWNED BY (without having a default clause)
4. GENERATED ALWAYS AS IDENTITY

cc: @sanyamsinghal if you can also review this query.

yb-voyager/cmd/exportData.go
```diff
@@ -1170,10 +1170,12 @@ func finalizeTableColumnList(finalTableList []sqlname.NameTuple) ([]sqlname.Name
 			utils.ErrExit("Exiting at user's request. Use `--exclude-table-list` flag to continue without these tables")
 		} else {
 			var importingDatabase string
-			if importerRole == TARGET_DB_IMPORTER_ROLE {
+			if exporterRole == SOURCE_DB_EXPORTER_ROLE {
```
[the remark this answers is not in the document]
priyanshi-yb: yeah mostly UI msg fix

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -93,8 +93,11 @@ func prepareDebeziumConfig(partitionsToRootTableMap map[string]string, tableList
 		}
 		return true, nil
 	})
-
-	colToSeqMap := source.DB().GetColumnToSequenceMap(tableList)
+	msr, err := metaDB.GetMigrationStatusRecord()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get migration status record: %w", err)
+	}
+	colToSeqMap := fetchOrRetrieveColToSeqMap(msr, tableList)
```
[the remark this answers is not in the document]
priyanshi-yb: > fetching this again from the source in each run should be harmless

Yeah mostly it is harmless but thought if we can avoid the completely going to db in subsequent runs that will be better.

yb-voyager/src/srcdb/yugabytedb.go
priyanshi-yb: yeah we can, though this function is never used as such but for correctness will do

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: yeah will add

yb-voyager/src/namereg/namereg.go
priyanshi-yb: Hmm, there is no such use case for having the same function, just the difference is which map we are using to look up, maybe I can change it to a separate function LookupSequenceName let me see if its a lot of change

yb-voyager/src/srcdb/yugabytedb.go
makalaaneesh: Before this change, this would be used, right? 
After this change, this is not used because in export-data-from-target, we now rely on the sequence list from the source? 

Is my understanding correct?

makalaaneesh: Hmm, this is slightly concerning that we're not consistent in the offline and live part. 
For offline, we're using `GetAllSequences()` but for the debezium live part, we're using `GetColumnToSequenceMap`. 

Can we be consistent so that both the phases are using the same set of sequences?

yb-voyager/src/namereg/namereg.go
priyanshi-yb: Not making this change as I realised if I separate out the function it will be a breaking change because we are having the separate map for both and with this same function I handled by combining both the map so that should it will work even in upgrade.

migtests/scripts/run-test.sh
```diff
@@ -173,9 +173,10 @@ main() {
 	run_ysql ${TARGET_DB_NAME} "\dft" 
 
 	step "Run validations."
-	if [ -x "${TEST_DIR}/validate" ]
-	then
-		 "${TEST_DIR}/validate"
+	if [[ "${EXPORT_TABLE_LIST}" != "" && -x "${TEST_DIR}/validate-with-table-list" ]]; then
+		"${TEST_DIR}/validate-with-table-list"
+	elif [[ -x "${TEST_DIR}/validate" ]]; then
+		"${TEST_DIR}/validate"
```
sanyamsinghal, line 179: @priyanshi-yb @makalaaneesh 
Just thinking if special handling like this for table-list case might complicate the automation framework.
From what I understood we are handling multiple expected files differently here. like export data status and import data status. 

Instead can we just simply add a new test here - `pg/sequence-with-tablelist`?

yb-voyager/cmd/exportData.go
```diff
@@ -338,6 +338,10 @@ func exportData() bool {
 	fmt.Printf("num tables to export: %d\n", len(tableListToDisplay))
 	utils.PrintAndLog("table list for data export: %v", tableListToDisplay)
 
+	if source.DBType == POSTGRESQL {
+		utils.PrintAndLog("Only the sequences that are attached to the above tables will be restored during the migration.")
+	}
```
sanyamsinghal, line 343: How about also mentioning something like this - `tables being exported(refer above list) will be ...`

Reason being, lets say user didn't specify any table list, and then they don't necessarily need to go through the printed table list to ensure. They can infer that sequences attached to all the tables(being exported) will be restored.

yb-voyager/cmd/exportData.go
```diff
@@ -1223,14 +1229,23 @@ func exportDataOffline(ctx context.Context, cancel context.CancelFunc, finalTabl
 
 	if source.DBType == POSTGRESQL {
 		//need to export setval() calls to resume sequence value generation
-		sequenceList := source.DB().GetAllSequences()
-		for _, seq := range sequenceList {
+		colToSeqMap := source.DB().GetColumnToSequenceMap(finalTableList)
```
sanyamsinghal, line 1232: why not using `fetchOrRetrieveColToSeqMap` here?

yb-voyager/cmd/exportData.go
```diff
@@ -1223,14 +1229,23 @@ func exportDataOffline(ctx context.Context, cancel context.CancelFunc, finalTabl
...
 				utils.ErrExit("lookup for sequence failed: %s: err: %v", seq, err)
 			}
 			finalTableList = append(finalTableList, seqTuple)
 		}
+		err = metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
+			switch exporterRole {
+			case SOURCE_DB_EXPORTER_ROLE:
+				record.SourceColumnToSequenceMapping = colToSeqMap
+			}
+		})
```
sanyamsinghal, line 1245: same; would not using `fetchOrRetrieveColToSeqMap` also update the msr?

yb-voyager/src/metadb/migrationStatus.go
```diff
@@ -68,6 +68,9 @@ type MigrationStatusRecord struct {
 
 	ImportDataFileFlagFileTableMapping string `json:"ImportDataFileFlagFileTableMapping"` // Import data file command's file_table_mapping flag
 	ImportDataFileFlagDataDir          string `json:"ImportDataFileFlagDataDir"`          // Import data file command's data-dir flag
+
+	SourceColumnToSequenceMapping map[string]string `json:"SourceColumnToSequenceMapping"`
+	TargetColumnToSequenceMapping map[string]string `json:"TargetColumnToSequenceMapping"`
```
sanyamsinghal, line 73: I think we should start considering this tech debt now. MSR is getting bigger day by day.
We should organise the fields in it.

We can start with grouping the related fields
```
type CutoverFlags struct {
    RequestedToTarget        bool `json:"RequestedToTarget"`
    RequestedToSource        bool `json:"RequestedToSource"`
    RequestedToSourceReplica bool `json:"RequestedToSourceReplica"`
    DetectedByTargetImporter bool `json:"DetectedByTargetImporter"`
    // ... and so on
}

type ExportFlags struct {
    SchemaDone         bool `json:"SchemaDone"`
    DataDone           bool `json:"DataDone"`
    TargetFallForward  bool `json:"TargetFallForward"`
    TargetFallBack     bool `json:"TargetFallBack"`
    // ... etc.
}

type MigrationStatusRecord struct {
    MigrationUUID  string       `json:"MigrationUUID"`
    SourceDBConf   *srcdb.Source `json:"SourceDBConf"`
    TargetDBConf   *tgtdb.TargetConf `json:"TargetDBConf"`
    // ...
    Cutover        CutoverFlags  `json:"Cutover"`
    Export         ExportFlags   `json:"Export"`
    // ...
}
```

@priyanshi-yb Can you file a tech debt task for this?

yb-voyager/src/srcdb/yugbaytedb_test.go
```diff
@@ -59,6 +61,193 @@ func TestYugabyteGetAllTableNames(t *testing.T) {
 	testutils.AssertEqualSourceNameSlices(t, expectedTables, actualTables)
 }
 
+func TestYBGetColumnToSequenceMap(t *testing.T) {
```
sanyamsinghal, line 64: One high level comment(didn't go over the testcase thoroughly)
- here we are asserting twice(probably testing two different scenarios), problem is when this test fails, we(developers) might not clearly know what part of this test(or function) failed.

So I think its good to demarcate that either by breaking this into multiple functions.
You can name it like `TestYBGetColumnToSequenceMap_Case...`

Or if you want to retain single function. (chatgpt recommendation)
```
func TestPGGetColumnToSequenceMap(t *testing.T) {
	// Set up the test DB with required schemas, tables, sequences, etc.
	testPostgresSource.TestContainer.ExecuteSqls(
		`CREATE SCHEMA IF NOT EXISTS public;`,
		`CREATE SCHEMA IF NOT EXISTS custom_schema;`,
		//-- 1. SERIAL column (implicit)
		`CREATE TABLE public.serial_table (
			id SERIAL PRIMARY KEY,
			name TEXT
		);`,
		//-- 2. BIGSERIAL column
		`CREATE TABLE public.bigserial_table (
			id BIGSERIAL PRIMARY KEY,
			description TEXT
		);`,
		//-- 3. IDENTITY columns (Generated Always)
		`CREATE TABLE public.identity_always_table (
			id BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
			info TEXT
		);`,
		//-- 4. IDENTITY columns (Generated By Default)
		`CREATE TABLE public.identity_default_table (
			id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
			comments TEXT
		);`,
		//-- 5. DEFAULT nextval() with a manually created seq
		`CREATE SEQUENCE public.manual_seq START 1000;`,
		`CREATE TABLE public.default_nextval_table (
			id BIGINT NOT NULL DEFAULT nextval('public.manual_seq'),
			value TEXT
		);`,
		//-- 6. Same as above but cross-schema
		`CREATE SEQUENCE custom_schema.cross_schema_seq;`,
		`CREATE TABLE public.cross_schema_default_seq_table (
			id BIGINT NOT NULL DEFAULT nextval('custom_schema.cross_schema_seq'),
			payload TEXT
		);`,
		//-- 7. A table with a manually owned by sequence (DEFAULT)
		`CREATE SEQUENCE public.manual_linked_seq;`,
		`CREATE TABLE public.manual_linked_table (
			id BIGINT NOT NULL,
			meta TEXT
		);`,
		`ALTER TABLE public.manual_linked_table ALTER COLUMN id SET DEFAULT nextval('public.manual_linked_seq');`,
		`ALTER SEQUENCE public.manual_linked_seq OWNED BY public.manual_linked_table.id;`,
		//--8. Table with manually owned sequence on text column
		`CREATE SEQUENCE custom_schema.user_code_seq START WITH 1 INCREMENT BY 1;`,
		`CREATE OR REPLACE FUNCTION generate_user_code() RETURNS TEXT AS $$
		DECLARE
			new_code TEXT;
		BEGIN
			SELECT 'USR' || LPAD(nextval('custom_schema.user_code_seq')::TEXT, 4, '0') INTO new_code;
			RETURN new_code;
		END;
		$$ LANGUAGE plpgsql;`,
		`CREATE TABLE custom_schema.users (
			user_id INT PRIMARY KEY,
			user_code TEXT UNIQUE,
			name TEXT
		);`,
		`ALTER TABLE ONLY custom_schema.users ALTER COLUMN user_code SET DEFAULT generate_user_code();`,
		`ALTER SEQUENCE custom_schema.user_code_seq OWNED BY custom_schema.users.user_code;`,
		//--9. Normal sequences not attached to tables
		`CREATE SEQUENCE seq1_unattached;`,
		`CREATE SEQUENCE custom_schema.seq2_unattached;`,
		//--10. Manually linked sequence to multiple columns
		`CREATE SEQUENCE public.manual_linked_seq_another;`,
		`CREATE TABLE public.manual_linked_table_1 (
			id BIGINT NOT NULL,
			meta TEXT
		);`,
		`ALTER TABLE public.manual_linked_table_1 ALTER COLUMN id SET DEFAULT nextval('public.manual_linked_seq_another');`,
		`CREATE TABLE public.manual_linked_table_2 (
			id BIGINT NOT NULL,
			meta TEXT
		);`,
		`ALTER TABLE public.manual_linked_table_2 ALTER COLUMN id SET DEFAULT nextval('public.manual_linked_seq_another');`,
	)
	defer testPostgresSource.TestContainer.ExecuteSqls(
		`DROP TABLE public.serial_table;`,
		`DROP TABLE public.bigserial_table;`,
		`DROP TABLE public.identity_always_table;`,
		`DROP TABLE public.identity_default_table;`,
		`DROP TABLE public.default_nextval_table;`,
		`DROP TABLE public.cross_schema_default_seq_table;`,
		`DROP TABLE public.manual_linked_table;`,
		`DROP TABLE public.manual_linked_table_1;`,
		`DROP TABLE public.manual_linked_table_2;`,
		`DROP SEQUENCE IF EXISTS public.manual_linked_seq_another;`,
		`DROP TABLE custom_schema.users;`,
		`DROP SEQUENCE public.seq1_unattached;`,
		`DROP SEQUENCE IF EXISTS custom_schema.user_code_seq;`,
		`DROP FUNCTION generate_user_code;`,
		`DROP SEQUENCE manual_seq;`,
		`DROP SCHEMA custom_schema CASCADE;`,
	)

	// Set up common DB and schema information.
	sqlname.SourceDBType = "postgresql"
	testPostgresSource.Source.Schema = "public|custom_schema"
	_ = testPostgresSource.DB().Connect()
	defer testPostgresSource.DB().Disconnect()

	// helper function to run a subtest with a given table list and expected map.
	runTest := func(name string, tableList []sqlname.NameTuple, expected map[string]string) {
		t.Run(name, func(t *testing.T) {
			actual := testPostgresSource.DB().GetColumnToSequenceMap(tableList)
			assert.Equal(t, len(expected), len(actual), "Unexpected number of table entries")
			for key, expectedVal := range expected {
				actualVal, exists := actual[key]
				assert.True(t, exists, "Key %s not found", key)
				assert.Equal(t, expectedVal, actualVal, "Mismatch for key %s", key)
			}
		})
	}

	// Subtest with the full table list.
	fullTableList := []sqlname.NameTuple{
		testutils.CreateNameTuple("public.serial_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.bigserial_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.identity_always_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.identity_default_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.default_nextval_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.cross_schema_default_seq_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.manual_linked_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("custom_schema.users", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.manual_linked_table_1", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.manual_linked_table_2", "public", testPostgresSource.DBType),
	}
	expectedFullMap := map[string]string{
		"public.serial_table.id":                   `public."serial_table_id_seq"`,
		"public.bigserial_table.id":                `public."bigserial_table_id_seq"`,
		"public.identity_always_table.id":          `public."identity_always_table_id_seq"`,
		"public.identity_default_table.id":         `public."identity_default_table_id_seq"`,
		"public.default_nextval_table.id":          `public."manual_seq"`,
		"public.cross_schema_default_seq_table.id": `custom_schema."cross_schema_seq"`,
		"public.manual_linked_table.id":            `public."manual_linked_seq"`,
		"custom_schema.users.user_code":            `custom_schema."user_code_seq"`,
		"public.manual_linked_table_1.id":          `public."manual_linked_seq_another"`,
		"public.manual_linked_table_2.id":          `public."manual_linked_seq_another"`,
	}
	runTest("Full Table List", fullTableList, expectedFullMap)

	// Subtest with a subset of the tables.
	subsetTableList := []sqlname.NameTuple{
		testutils.CreateNameTuple("public.serial_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.bigserial_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.identity_always_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.cross_schema_default_seq_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.manual_linked_table", "public", testPostgresSource.DBType),
		testutils.CreateNameTuple("public.manual_linked_table_1", "public", testPostgresSource.DBType),
	}
	expectedSubsetMap := map[string]string{
		"public.serial_table.id":                   `public."serial_table_id_seq"`,
		"public.bigserial_table.id":                `public."bigserial_table_id_seq"`,
		"public.identity_always_table.id":          `public."identity_always_table_id_seq"`,
		"public.cross_schema_default_seq_table.id": `custom_schema."cross_schema_seq"`,
		"public.manual_linked_table.id":            `public."manual_linked_seq"`,
		"public.manual_linked_table_1.id":          `public."manual_linked_seq_another"`,
	}
	runTest("Subset of Tables", subsetTableList, expectedSubsetMap)
}
```

yb-voyager/src/srcdb/postgres.go
```diff
@@ -71,24 +71,61 @@ var PG_COMMAND_VERSION = map[string]string{
...
+seq.relname AS sequence_name
+FROM pg_class t
+JOIN pg_namespace tn ON tn.oid = t.relnamespace
+JOIN pg_attribute a ON a.attrelid = t.oid
+LEFT JOIN pg_attrdef ad ON ad.adrelid = t.oid AND ad.adnum = a.attnum
+LEFT JOIN pg_depend dep
+	ON (dep.objid = ad.oid ) 
+	OR (a.attidentity <> '' AND dep.refobjid = a.attrelid AND dep.refobjsubid = a.attnum)  -- identity sequence
+	OR (dep.refobjid = t.oid AND dep.refobjsubid = a.attnum) -- owned sequences
+LEFT JOIN pg_class seq ON seq.oid = dep.objid AND seq.relkind = 'S'
```
sanyamsinghal, line 96: Most of the part of the queries are same.
please point out  the differences in the queries with a `-- comment`

sanyamsinghal (changes requested):

yb-voyager/src/srcdb/yugbaytedb_test.go
priyanshi-yb: that can be handled with some print statements right?

yb-voyager/src/srcdb/yugbaytedb_test.go
```diff
@@ -59,6 +61,193 @@ func TestYugabyteGetAllTableNames(t *testing.T) {
...
+		);`,
+		`ALTER TABLE public.manual_linked_table_1 ALTER COLUMN id SET DEFAULT nextval('public.manual_linked_seq_another');`,
+
+		`CREATE TABLE public.manual_linked_table_2 (
+			id BIGINT NOT NULL,
+			meta TEXT
+		);`,
+		`ALTER TABLE public.manual_linked_table_2 ALTER COLUMN id SET DEFAULT nextval('public.manual_linked_seq_another');`,
+	)
+	defer testYugabyteDBSource.TestContainer.ExecuteSqls(
```
makalaaneesh, line 165: in general, i wonder if dropping the entire database will make it easier for clean up functions. (don't have to worry about DROPPING every object)

yb-voyager/test/utils/testutils.go
```diff
@@ -424,3 +424,10 @@ func CreateTempFile(dir string, fileContents string, fileFormat string) (string,
 
 	return file.Name(), nil
 }
+
+func CreateNameTuple(s string, defaultSchema string, dbType string) sqlname.NameTuple {
```
makalaaneesh, line 428: nit: Since you're only populating the sourceName, indicate that somehow in the function name.

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: added a comment before the query what is the difference

yb-voyager/src/srcdb/yugbaytedb_test.go
priyanshi-yb: yeah should be possible, will try to do this as whole separately for all integration tests we have

yb-voyager/src/srcdb/postgres_test.go
```diff
@@ -59,6 +61,194 @@ func TestPostgresGetAllTableNames(t *testing.T) {
...
+		`ALTER TABLE public.manual_linked_table_2 ALTER COLUMN id SET DEFAULT nextval('public.manual_linked_seq_another');`,
+	)
+	defer testPostgresSource.TestContainer.ExecuteSqls(
+		`DROP TABLE  public.serial_table;`,
+		`DROP TABLE  public.bigserial_table;`,
+		`DROP TABLE  public.identity_always_table;`,
+		`DROP TABLE  public.identity_default_table;`,
+		`DROP TABLE  public.default_nextval_table;`,
+		`DROP TABLE  public.cross_schema_default_seq_table;`,
+		`DROP TABLE  public.manual_linked_table;`,
```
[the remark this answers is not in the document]
priyanshi-yb: will try to do this as a whole separately for all integration tests we have, maybe with different database as the problem is we are working directly on postgres database in all tests.

yb-voyager/src/metadb/migrationStatus.go
priyanshi-yb: https://github.com/yugabyte/yb-voyager/issues/2409

migtests/scripts/run-test.sh
priyanshi-yb: Somehow, this seems easier to me to add the table-list test this way. We can take it up later when we plan to include live tests as well for these cases: the full list and subset of the table list.

yb-voyager/src/srcdb/yugbaytedb_test.go
sanyamsinghal: yeah thats a good idea.
Will have to handle the update of db name in container config and ability to use custom dbname for each test case.

yb-voyager/src/srcdb/yugbaytedb_test.go
sanyamsinghal: > that can be easily handled with some print statements right?

yeah that can also work. Using test name is cleaner approach.

migtests/scripts/run-test.sh
sanyamsinghal: yeah np, we can take this later.
To me this approach seems more useful when we plan to add table list related test/validation in a lot of tests. Since it increases the complexity of the framework.

Otherwise for a single test, we can separate this, even if there is some duplication or repeatition.

yb-voyager/src/srcdb/yugbaytedb_test.go
sanyamsinghal: > But also, running it locally makes it possible to determine where exactly it has failed (line numbers).

Its not just about this test case, I am more inclined toward best practice. 
Especially, when there will be thousands of tests running.
Lets stick to this practice if we agree.

sanyamsinghal (approved):
LGTM.
I didn't go through the end to end test added, assuming that is already done by @makalaaneesh

makalaaneesh (approved):
LGTM

makalaaneesh:
> LGTM. I didn't go through the end to end test added, assuming that is already done by @makalaaneesh

just went through the various cases we seem to be testing, looks good 👍

PR #2396, opened 2025-03-12
Using queryissues to report unsupported datatypes in assess-migration

ShivanshGahlot:
-Unsupported datatypes (including live and ff/fb) are now reported using the QueryIssues created for each unsupported datatype in a previous PR. 
-The reason to do this is that the QueryIssues for individual datatypes have more customized and extensive information about the issue rather than just a single standard issue for all the datatypes.
-Ticket: https://github.com/yugabyte/yb-voyager/issues/2240

Changes in the assessment report for all the reported unsupported datatypes:
image

Callhome data:
image

Locally

| Migration Assessment Report Json                          |  Yes * |

\* 
```
type TableColumnsDataTypes struct {
	SchemaName        string `json:"SchemaName"`
	TableName         string `json:"TableName"`
	ColumnName        string `json:"ColumnName"`
	DataType          string `json:"DataType"`
	IsArrayType bool   `json:"-"`
        IsEnumType bool `json:"-"
	IsUDTType         bool   `json:"-"`
}
```
Last three parameters were added but they are not being marshalled anywhere hence no breaking change


yb-voyager/src/query/queryissue/issues_ddl.go
```diff
@@ -443,7 +443,7 @@ func NewTopogeometryDatatypeIssue(objectType string, objectName string, sqlState
 var pointDatatypeIssue = issue.Issue{
 	Type:        POINT_DATATYPE,
```
sanyamsinghal, line 444: @ShivanshGahlot what is the reasoning behind keep the issue type as that particular datatype name?

I have not reviewed previous PRs so might be missing the context.

My understanding regarding this `Type` field in issue struct was to have control internally if we want to filter issues which are related like all unsupported datatypes, all unsupported live datatypes, or all unsupported extensions.

Using Type here same as the particular datatype might be of help I think.
That same thing can even be achieved using the `Name` field

cc @priyanshi-yb @makalaaneesh

yb-voyager/src/query/queryissue/detectors_ddl.go
```diff
@@ -493,7 +501,7 @@ func (f *ForeignTableIssueDetector) DetectIssues(obj queryparser.DDLObject) ([]Q
 	for _, col := range foreignTable.Columns {
 		isUnsupportedDatatype := utils.ContainsAnyStringFromSlice(srcdb.PostgresUnsupportedDataTypes, col.TypeName)
 		if isUnsupportedDatatype {
-			reportUnsupportedDatatypes(col, obj.GetObjectType(), foreignTable.GetObjectName(), &issues)
+			issues = append(issues, ReportUnsupportedDatatypes(col.TypeName, col.ColumnName, obj.GetObjectType(), foreignTable.GetObjectName()))
```
sanyamsinghal, line 504: didn't get this.
What is the difference here in before vs after.
Was passing reference to issues slice had any problem?

yb-voyager/src/query/queryissue/issues_ddl.go
sanyamsinghal: Conclusion: 
Lets use with `Type` field as some sort of `ID` for that issue.
Avoiding renaming it to `ID` right now due to callhome and ybd dependencies, and be consistent across.

@ShivanshGahlot lets add comments for in the code::
```
type Issue struct {
	// Type acts as ID for the issue; should be unique across all issues
	// for example: ...
	// This field is not to be changed frequently; if changed, be careful about impact on callhome and yugabyted
	Type                   string
	
	// readable name for the issue; used in UI, logs or any print statements 
	Name                   string
	Description            string
	Impact                 string
	Suggestion             string
	GH                     string
	DocsLink               string
	MinimumVersionsFixedIn map[string]*ybversion.YBVersion // key: series (2024.1, 2.21, etc)
}
```

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1390,19 +1395,36 @@ func fetchColumnsWithUnsupportedDataTypes() ([]utils.TableColumnsDataTypes, []ut
...
+				MinimumVersionsFixedIn: nil, // TODO
+			}
+		case POSTGRESQL:
+			// Datatypes can be of form public.geometry, so we need to extract the datatype from it
+			var datatype string
+			if strings.Contains(colInfo.DataType, ".") {
+				datatype = strings.Split(colInfo.DataType, ".")[1]
+			} else {
+				datatype = colInfo.DataType
+			}
```
sanyamsinghal, line 1422: nit: these operations are frequent; we can add a utils function for this:
```
datatype, ok := utils.SliceLastElement(strings.Split(colInfo.DataType, "."))
if !ok { // not expected in this case
   // return or just log and move ahead
}


utils.go
// SliceLastElement returns the last element of a slice (if non-empty) and a boolean indicating success.
func SliceLastElement[T any](slice []T) (T, bool) {
	if len(slice) == 0 {
		var zeroValue T // Default zero value of type T
		return zeroValue, false
	}
	return slice[len(slice)-1], true
}
```

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1390,19 +1395,36 @@ func fetchColumnsWithUnsupportedDataTypes() ([]utils.TableColumnsDataTypes, []ut
...
+			queryissue := queryissue.ReportUnsupportedDatatypes(datatype, colInfo.ColumnName, constants.COLUMN, qualifiedColName)
+			convertedAnalyzeIssue := convertIssueInstanceToAnalyzeIssue(queryissue, "", false, false)
+			issue = convertAnalyzeSchemaIssueToAssessmentIssue(convertedAnalyzeIssue, GetCategoryDescription(UNSUPPORTED_DATATYPES_CATEGORY), queryissue.MinimumVersionsFixedIn)
+		default:
```
sanyamsinghal, line 1426: add a comment about the conversion you are doing.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1532,19 +1557,44 @@ func addMigrationCaveatsToAssessmentReport(unsupportedDataTypesForLiveMigration
...
+				if strings.Contains(colInfo.DataType, ".") {
+					datatype = strings.Split(colInfo.DataType, ".")[1]
+				} else {
+					datatype = colInfo.DataType
+				}
+
+				var queryIssue queryissue.QueryIssue
+
+				if colInfo.IsArrayType && colInfo.IsEnumType {
+					queryIssue = queryissue.NewArrayOfEnumDatatypeIssue(
```
sanyamsinghal, line 1573: are these issue - array and enum types part of the list you are maintaining for all unsupported issues etc.

yb-voyager/src/query/queryissue/issues_ddl.go
```diff
@@ -523,7 +523,7 @@ func NewPathDatatypeIssue(objectType string, objectName string, sqlStatement str
 var polygonDatatypeIssue = issue.Issue{
 	Type:        POLYGON_DATATYPE,
```
sanyamsinghal, line 524: as discussed lets rename this and other ones
`POLYGON_DATATYPE` --> `UNSUPPORTED_DATATYPE_LIVE_MIGRATION_POLYGON`

yb-voyager/src/query/queryissue/issues_ddl.go
```diff
@@ -523,7 +523,7 @@ func NewPathDatatypeIssue(objectType string, objectName string, sqlStatement str
...
 	Name:        "Unsupported datatype for Live migration - Polygon",
```
sanyamsinghal, line 525: for Issue.Name lets maintain a variable in `constants.go` file of this package.
I think we are doing that for most of the issues, already. ?

sanyamsinghal (changes requested):

yb-voyager/src/query/queryissue/detectors_ddl.go
ShivanshGahlot: This was done so that I could re-use this function in assess migration. I wanted to return the queryissue instead of directly appending it to the list in the function. This returned queryissue could be then converted to Assessment Issue.

yb-voyager/src/query/queryissue/issues_ddl.go
ShivanshGahlot: Done

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1414,12 +1414,15 @@ func getAssessmentIssuesForUnsupportedDatatypes(unsupportedDatatypes []utils.Tab
...
-				datatype = colInfo.DataType
+			datatype, ok := utils.SliceLastElement(strings.Split(colInfo.DataType, "."))
+			if !ok {
+				log.Warnf("failed to get datatype from %s", colInfo.DataType)
+				continue
 			}
+
+			// We obtain the queryissue from the Report function. This queryissue is first converted to AnalyzeIssue and then to AssessmentIssue using pre existing function
+			// Coneverting queryissue directly to AssessmentIssue would have lead to the creation of a new function which would have required a lot of cases to be handled and led to code duplication
+			// This converted AssessmentIssue is then appended to the assessmentIssues slice
```
sanyamsinghal, line 1425: I am just wondering that for analyze schema issues we detect the unsupported datatypes based on scanning the DDL in schema files.

Here we are detecting based on system catalog containing each tables' column dataypes.

Two things I am not sure(not necessarily because of this PR):
1. How is it ensured that those issues are not reported twice.
2. Can we make use of same logic in both - assess and analyze, maybe just use the issues returned from analyze schema?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1560,15 +1566,17 @@ func addMigrationCaveatsToAssessmentReport(unsupportedDataTypesForLiveMigration
...
+				if !ok {
+					log.Warnf("failed to get datatype from %s", colInfo.DataType)
+					continue
 				}
 
 				var queryIssue queryissue.QueryIssue
 
+				// We obtain the queryissue from the Report function or directly as used below. This queryissue is first converted to AnalyzeIssue and then to AssessmentIssue using pre existing function
+				// Coneverting queryissue directly to AssessmentIssue would have lead to the creation of a new function which would have required a lot of cases to be handled and led to code duplication
+				// This converted AssessmentIssue is then appended to the assessmentIssues slice
```
sanyamsinghal, line 1579: above once is enough i guess

yb-voyager/src/query/queryissue/issues_ddl.go
sanyamsinghal, line 1: nit: in code, we can keep all issues of similar type like Unsupported Datatype together.
Then after that, issues of lets say PK_UK_ON_COMPLEX_DATATYPE

Maybe a comment to partition that also.
```
// ============================= Unsupported Datatypes Issues =================
```

Ignore if this is already followed.

sanyamsinghal (commented):
Looks good mostly, let me know once final changes are in and tests pass.

yb-voyager/src/query/queryissue/issues_ddl.go
ShivanshGahlot: Yeah they were already grouped together. Added these comments before Unsupported Datatypes, Index on on Unsupported Datatypes and PK or UK on Unsupported Datatypes issues.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -612,21 +610,18 @@ func checkPlPgSQLStmtsUsingParser(sqlInfoArr []sqlInfo, fpath string, objType st
...
-			issue := convertIssueInstanceToAnalyzeIssue(issueInstance, fpath, true)
+			issue := convertIssueInstanceToAnalyzeIssue(issueInstance, fpath, true, true)
 			schemaAnalysisReport.Issues = append(schemaAnalysisReport.Issues, issue)
 		}
 	}
 
 }
 
-func convertIssueInstanceToAnalyzeIssue(issueInstance queryissue.QueryIssue, fileName string, isPlPgSQLIssue bool) utils.AnalyzeSchemaIssue {
+func convertIssueInstanceToAnalyzeIssue(issueInstance queryissue.QueryIssue, fileName string, isPlPgSQLIssue bool, addToSummaryMap bool) utils.AnalyzeSchemaIssue {
```
sanyamsinghal, line 620: nit: few comments about this func args will help. 
Like when/why we need/don't-need the addToSummaryMap

Q. Can we not always add to summary map? I see its just store invalid count info in a set

yb-voyager/cmd/assessMigrationCommand.go
sanyamsinghal: Dual reporting won't happen because in assess we fetch each analyze issue specifically by passing the Issue Type info.

yb-voyager/src/issue/issue.go
```diff
@@ -21,8 +21,13 @@ import (
 )
 
 type Issue struct {
-	Type                   string // (advisory_locks, index_not_supported, etc)
-	Name                   string // for display
+	// Type acts as ID for the issue; should be unique across all issues
+	// for example: ...
```
sanyamsinghal, line 25: add example :)

yb-voyager/src/query/queryissue/constants.go
```diff
@@ -38,30 +38,38 @@ const (
...
+	UNSUPPORTED_DATATYPE_TOPOGEOMETRY    = "UNSUPPORTED_DATATYPE_TOPOGEOMETRY"
+	UNSUPPORTED_DATATYPE_RASTER          = "UNSUPPORTED_DATATYPE_RASTER"
+	UNSUPPORTED_DATATYPE_PG_LSN          = "UNSUPPORTED_DATATYPE_PG_LSN"
+	UNSUPPORTED_DATATYPE_TXID_SNAPSHOT   = "UNSUPPORTED_DATATYPE_TXID_SNAPSHOT"
+	UNSUPPORTED_DATATYPE_INT8_MULTIRANGE = "UNSUPPORTED_DATATYPE_INT8_MULTIRANGE"
+	UNSUPPORTED_DATATYPE_INT4_MULTIRANGE = "UNSUPPORTED_DATATYPE_INT4_MULTIRANGE"
+	UNSUPPORTED_DATATYPE_DATE_MULTIRANGE = "UNSUPPORTED_DATATYPE_DATE_MULTIRANGE"
+	UNSUPPORTED_DATATYPE_NUM_MULTIRANGE  = "UNSUPPORTED_DATATYPE_NUM_MULTIRANGE"
+	UNSUPPORTED_DATATYPE_TS_MULTIRANGE   = "UNSUPPORTED_DATATYPE_TS_MULTIRANGE"
+	UNSUPPORTED_DATATYPE_TSTZ_MULTIRANGE = "UNSUPPORTED_DATATYPE_TSTZ_MULTIRANGE"
```
sanyamsinghal, line 57: @ShivanshGahlot here also we should follow same format in issue type also.

Lets not separate out the datatype with `_` where actual datatype is continuous.

sanyamsinghal: Refer - https://www.postgresql.org/docs/current/rangetypes.html

yb-voyager/src/query/queryissue/constants.go
```diff
@@ -259,8 +260,37 @@ const (
...
+	UNSUPPORTED_DATATYPE_TOPOGEOMETRY_ISSUE_NAME    = "Unsupported datatype - topogeometry"
+	UNSUPPORTED_DATATYPE_RASTER_ISSUE_NAME          = "Unsupported datatype - raster"
+	UNSUPPORTED_DATATYPE_PG_LSN_ISSUE_NAME          = "Unsupported datatype - pg_lsn"
+	UNSUPPORTED_DATATYPE_TXID_SNAPSHOT_ISSUE_NAME   = "Unsupported datatype - txid_snapshot"
+	UNSUPPORTED_DATATYPE_INT8_MULTIRANGE_ISSUE_NAME = "Unsupported datatype - int8multirange"
+	UNSUPPORTED_DATATYPE_INT4_MULTIRANGE_ISSUE_NAME = "Unsupported datatype - int4multirange"
+	UNSUPPORTED_DATATYPE_DATE_MULTIRANGE_ISSUE_NAME = "Unsupported datatype - datemultirange"
+	UNSUPPORTED_DATATYPE_NUM_MULTIRANGE_ISSUE_NAME  = "Unsupported datatype - nummultirange"
+	UNSUPPORTED_DATATYPE_TS_MULTIRANGE_ISSUE_NAME   = "Unsupported datatype - tsmultirange"
+	UNSUPPORTED_DATATYPE_TSTZ_MULTIRANGE_ISSUE_NAME = "Unsupported datatype - tstzmultirange"
```
sanyamsinghal, line 278: lets remove underscore from the var name also, just to be consistent.

yb-voyager/src/query/queryissue/issues_ddl.go
```diff
@@ -632,6 +812,8 @@ func NewHstoreDatatypeIssue(objectType string, objectName string, sqlStatement s
 	return newQueryIssue(issue, objectType, objectName, sqlStatement, map[string]interface{}{})
 }
 
+// ============================= PK abd UK Constraints on Unsupported Datatypes Issues =================
```
sanyamsinghal, line 815: nit: `and`

sanyamsinghal (approved):
LGTM, thanks for this change, it will help to analyse callhome data better.

Please update the first screenshot in PR description.

yb-voyager/cmd/analyzeSchema.go
ShivanshGahlot: That summary map is used in the analyze schema part for TABLES, TRIGGERS etc to store summary related issues afair. We have no need of that in case of assessment report in datatypes. I have added a relevant comment there.

PR #2407, opened 2025-03-17
Added tests for failure of name-registry list for extra tables on eit…

priyanshi-yb:
…her source/target

### DIFF
--- a/yb-voyager/cmd/exportData_test.go
+++ b/yb-voyager/cmd/exportData_test.go
@@ -1081,3 +1081,270 @@ func TestTableListInFreshRunOfExportDataBasicYB(t *testing.T) {
 	}
 	assertGuardrailsChecksForMissingAndExtraTablesInSubsequentRun(t, missingTables3, nil, expectedNewTableListWithLessTables, rootTables)
 }
+
+func TestTableListInFreshRunOfExportDataForTablesExtraInSource(t *testing.T) {
+
+	testExportDir := setupPostgresDBSourceAndYugabyteDBTargetWithExportDependencies(t, pgSchemaSqls, pgSchemasTest1)
+
+	err := testPostgresSource.DB().Connect()
+	if err != nil {
+		utils.ErrExit("Failed to connect to postgres database: %w", err)
+	}
+	testPostgresSource.ExecuteSqls(`CREATE TABLE test_source_extra(id int, val text);`)
+	err = InitNameRegistry(testExportDir, SOURCE_DB_EXPORTER_ROLE, testPostgresSource.Source, testPostgresSource.DB(), nil, nil, false)
+	if err != nil {
+		t.Errorf("error initialising name reg for the source: %v", err)
+	}
+	defer testPostgresSource.DB().Disconnect()
+	defer testPostgresSource.ExecuteSqls(cleanUpSqls...)
+	defer testPostgresSource.ExecuteSqls(`DROP TABLE test_source_extra;`)
+
+	metaDB = initMetaDB(testExportDir)
+	//Running the command level functions
+	source = *testPostgresSource.Source
+	source.ExcludeTableList = "test_source_extra"
+
+	expectedPartitionsToRootMap := map[string]string{
+		"public.test_partitions_sequences_l": "public.test_partitions_sequences",
+		"public.test_partitions_sequences_s": "public.test_partitions_sequences",
+		"public.test_partitions_sequences_b": "public.test_partitions_sequences",
+		"p1.london":                          "public.sales_region",
+		"p1.sydney":                          "public.sales_region",
+		"p1.boston":                          "public.sales_region",
+	}
+	expectedTableList := []sqlname.NameTuple{
+		getNameTuple("public.test_partitions_sequences"),
+		getNameTuple("public.test_partitions_sequences_l"),
+		getNameTuple("public.test_partitions_sequences_s"),
+		getNameTuple("public.test_partitions_sequences_b"),
+		getNameTuple("public.sales_region"),
+		getNameTuple("p1.London"),
+		getNameTuple("p1.Sydney"),
+		getNameTuple("p1.Boston"),
+		getNameTuple("public.datatypes1"),
+		getNameTuple("public.foreign_test"),
+	}
+	assertTableListFilteringInTheFirstRun(t, expectedPartitionsToRootMap, expectedTableList)
+
+	//Create msr with required details for subsequent run
+
+	expectedTableListWithOnlyRootTable := []sqlname.NameTuple{
+		getNameTuple("public.test_partitions_sequences"),
+		getNameTuple("public.sales_region"),
+		getNameTuple("public.datatypes1"),
+		getNameTuple("public.foreign_test"),
+	}
+	err = metaDB.UpdateMigrationStatusRecord(func(msr *metadb.MigrationStatusRecord) {
+		msr.SourceDBConf = testPostgresSource.Source
+		msr.TargetDBConf = &testYugabyteDBTarget.Tconf
+		msr.TableListExportedFromSource = lo.Map(expectedTableListWithOnlyRootTable, func(t sqlname.NameTuple, _ int) string {
+			return t.ForOutput()
+		})
+		msr.SourceExportedTableListWithLeafPartitions = lo.Map(expectedTableList, func(t sqlname.NameTuple, _ int) string {
+			return t.ForOutput()
+		})
+		msr.SourceRenameTablesMap = expectedPartitionsToRootMap
+	})
+	if err != nil {
+		t.Fatalf("error updating msr: %v", err)
+	}
+
+	//Start import-data
+	err = testYugabyteDBTarget.Init()
+	if err != nil {
+		utils.ErrExit("Failed to connect to yugabyte database: %w", err)
+	}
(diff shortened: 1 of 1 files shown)


yb-voyager/src/namereg/namereg.go
```diff
@@ -375,6 +379,107 @@ func (reg *NameRegistry) LookupTableName(tableNameArg string) (sqlname.NameTuple
...
+
+func (reg *NameRegistry) LookupTableNameAndIgnoreOtherSideMappingIfNotFound(tableNameArg string) (sqlname.NameTuple, error) {
+	sourceName, targetName, err := reg.lookupSourceAndTargetTableNamesIgnoreIfEitherSideNotFound(tableNameArg)
+	if err != nil {
+		return sqlname.NameTuple{}, fmt.Errorf("error lookup source and target names for table [%v]: %v", tableNameArg, err)
+	}
+	ntup := NewNameTuple(reg.params.Role, sourceName, targetName)
+	return ntup, nil
+}
+func (reg *NameRegistry) lookupSourceAndTargetTableNamesIgnoreIfEitherSideNotFound(tableNameArg string) (*sqlname.ObjectName, *sqlname.ObjectName, error) {
```
makalaaneesh, line 398: Make sure `LookupTableName` and `LookupTableNameAndIgnoreOtherSideMappingIfNotFound` use the same underlying function with different arguments (whether or not to ignore)

yb-voyager/cmd/exportData.go
```diff
@@ -817,19 +818,20 @@ func applyTableListFlagsOnFullListAndAddLeafPartitions(fullTableList []sqlname.N
...
 		if err != nil {
-			return nil, fmt.Errorf("error keeping only leaf and root tables: %v", err)
+			return nil, nil, fmt.Errorf("error keeping only leaf and root tables: %v", err)
 		}
 	}
-	//return the final table list generated from the command flags  table-list (include list) / exclude-table-list in this run
-	return sqlname.SetDifferenceNameTuples(includeTableList, excludeTableList), nil
+	//return the include and exclude table list generated from the command flags  table-list (include list) / exclude-table-list in this run
+
+	return includeTableList, excludeTableList, nil
```
makalaaneesh, line 834: what's the motivation for this change?

yb-voyager/cmd/exportData_test.go
```diff
@@ -1081,3 +1082,270 @@ func TestTableListInFreshRunOfExportDataBasicYB(t *testing.T) {
...
+		})
+		msr.SourceRenameTablesMap = expectedPartitionsToRootMap
+	})
+	if err != nil {
+		t.Fatalf("error updating msr: %v", err)
+	}
+
+	//Start import-data
+	err = testYugabyteDBTarget.Init()
+	if err != nil {
```
makalaaneesh, line 1155: use testutils.FatalIfError

yb-voyager/cmd/exportData_test.go
```diff
@@ -1081,3 +1082,270 @@ func TestTableListInFreshRunOfExportDataBasicYB(t *testing.T) {
...
+		utils.ErrExit("Failed to connect to yugabyte database: %w", err)
+	}
+	err = InitNameRegistry(testExportDir, TARGET_DB_IMPORTER_ROLE, nil, nil, &testYugabyteDBTarget.Tconf, testYugabyteDBTarget.TargetDB, false)
+	if err != nil {
+		t.Errorf("error initialising name reg for the source: %v", err)
+	}
+	testYugabyteDBTarget.Finalize()
+	err = testYugabyteDBSource.DB().Connect()
```
makalaaneesh, line 1163: why do you need to initialize testYugabyteDBSource? 
Can't you run the clean up on testYugabyteDBTarget?

yb-voyager/src/namereg/namereg.go
priyanshi-yb: so the main underlying function `lookup(...)` is the same which returns an object name (for source/target with. respective call) and an error so ignoring the error this function `lookupSourceAndTargetTableNamesIgnoreIfEitherSideNotFound ` and returning a source and target name irrespective of error

yb-voyager/cmd/exportData.go
priyanshi-yb: So we have this common function for applying table-list flags and adding leaf partitions to it in the first run or subsequent run. Previously, we were using the `SetDifferenceNameTuples ` directly in this function to return a filtered list from include and exclude, but now, for subsequent runs, we are using a name registry list with a name tuple having all source-target names populated if found but only one of them in case not found.
The problem is now for subsequent runs we need to use `SetDifferenceNameTuplesWithKey` to get a filtered list for without worrying about the full NameTuple.
So made this change to accommodate that and the caller will call `SetDifferenceNameTuplesWithKey` / `SetDifferenceNameTuplesWithKey`respectively .

yb-voyager/cmd/exportData_test.go
priyanshi-yb: `testYugabyteDBSource ` is not just for cleanup, this test case is for `export-data-from-target` which uses source so have to initialize testYugabyteDBSource as well.

yb-voyager/src/namereg/namereg.go
priyanshi-yb: Oh I think I got you now, I have made the respective change without duplicating the logic. having a single function returning source and target names and ignore error of lookup in either of them if `ignoreIfOtherSideMapping` is true.

yb-voyager/cmd/exportData_test.go
makalaaneesh: it looks like this test function only deals with export-data-from-source flow. Just confirm once and remove this initialization if not required.

yb-voyager/cmd/exportData.go
```diff
@@ -1009,10 +1011,11 @@ func applyTableListFlagsOnCurrentAndRemoveRootsFromBothLists(
 	firstRunTableWithLeafsAndRoots []sqlname.NameTuple) ([]sqlname.NameTuple, []sqlname.NameTuple, error) {
 
 	//apply include/exclude flags and if a new table is passed (which is not present in name registry), then error out Unknown table
-	currentRunTableListFilteredViaFlags, err := applyTableListFlagsOnFullListAndAddLeafPartitions(registeredList, tableListViaFlag, excludeTableListViaFlag)
+	currentRunIncludeTableList, currentRunExlcudeTableList, err := applyTableListFlagsOnFullListAndAddLeafPartitions(registeredList, tableListViaFlag, excludeTableListViaFlag)
 	if err != nil {
 		return nil, nil, fmt.Errorf("error in apply table list filter on registered list for the flags in current run: %v", err)
 	}
+	currentRunTableListFilteredViaFlags := sqlname.SetDifferenceNameTuplesWithKey(currentRunIncludeTableList, currentRunExlcudeTableList)
```
makalaaneesh, line 1018: let's add a comment explaining why we're only comparing the nametuples by key

yb-voyager/src/namereg/namereg.go
```diff
@@ -345,8 +374,12 @@ func (reg *NameRegistry) LookupTableName(tableNameArg string) (sqlname.NameTuple
 				}
 			}
 			if err != nil {
-				// `err` can be: no default schema, no matching name, multiple matching names.
-				return sqlname.NameTuple{}, fmt.Errorf("lookup source table name [%s.%s]: %w", schemaName, tableName, err)
+				if ignoreIfOtherSideMappingNotFound {
```
makalaaneesh, line 377: how do we know for sure that the error is "not found"
Technically, it could be because of other scenarios as well, right? Ideally, we should be sure that the error we are ignoring is "not found" error. 

There seems to be an `ErrNameNotFound` error. We can check for that specifically.

yb-voyager/cmd/exportData_test.go
priyanshi-yb: oh yeah, must be a miss in copy-pasting the tests. Removed it.

yb-voyager/src/namereg/namereg.go
priyanshi-yb: Yeah, makes sense we should ignore the other errors. So just ignoring the `ErrNameNotFound` type of error only.

yb-voyager/src/namereg/namereg.go
```diff
@@ -374,7 +374,8 @@ func (reg *NameRegistry) lookupSourceAndTargetTableNames(tableNameArg string, ig
 				}
 			}
 			if err != nil {
-				if ignoreIfOtherSideMappingNotFound {
+				errNotFound := &ErrNameNotFound{}
+				if ignoreIfOtherSideMappingNotFound && errors.As(err, &errNotFound) {
```
makalaaneesh, line 378: `errNotFound` is already a pointer, do you need to pass `&errNotFound` again?

makalaaneesh (approved):
LGTM!

yb-voyager/src/namereg/namereg.go
priyanshi-yb: Yes, this is how its docs also [says](https://pkg.go.dev/errors#As:~:text=As%20panics%20if%20target%20is%20not%20a%20non%2Dnil%20pointer%20to%20either%20a%20type%20that%20implements%20error%2C%20or%20to%20any%20interface%20type.), this `errors.As` function panics if we pass doesn't pass it like this.

PR #2421, opened 2025-03-20
Update YB Latest Stable Version to 2024.2.2.0 and modified GHA to have placeholder for Latest Stable

sanyamsinghal:
- Update YB Latest Stable Version to `2024.2.2.0`

- Add 'LATEST_STABLE' version placeholder to GHA matrix and handle in final step
  - Introduce LATEST_STABLE as a placeholder in matrix.strategy
  - Override LATEST_STABLE with the actual stable YB release in Setup YugabyteDB step
  - Add a special test step (Test Latest Stable YB Version) that runs only when matrix.version == 'LATEST_STABLE'

NA

Existing tests

### DIFF
--- a/.github/workflows/issue-tests.yml
+++ b/.github/workflows/issue-tests.yml
@@ -32,7 +32,7 @@ jobs:
         # Note: Update val for LATEST_STABLE here along with 'src/ybversion/constants.go/LATEST_STABLE' variable
         if [ "${YB_VERSION}" == "LATEST_STABLE" ]; then
           ACTUAL_YB_VERSION="2024.2.2.0-b61"
-        esle
+        else
           ACTUAL_YB_VERSION="${YB_VERSION}"
         fi
 
--- a/.github/workflows/issue-tests.yml
+++ b/.github/workflows/issue-tests.yml
@@ -32,10 +32,10 @@ jobs:
         # Note: Update val for LATEST_STABLE here along with 'src/ybversion/constants.go/LATEST_STABLE' variable
         if [ "${YB_VERSION}" == "LATEST_STABLE" ]; then
           YB_VERSION="2024.2.2.0-b61"
-        fi
 
-        # write it to GitHub's environment to override previous value
-        echo "YB_VERSION=${YB_VERSION}" >> $GITHUB_ENV
+          # write it to GitHub's environment to override previous value
+          echo "YB_VERSION=${YB_VERSION}" >> $GITHUB_ENV
+        fi
 
         # using s3 release instead of docker image to allow testing against un-released versions
         wget https://s3.us-west-2.amazonaws.com/releases.yugabyte.com/${YB_VERSION}/yugabyte-${YB_VERSION}-centos-x86_64.tar.gz
--- a/.github/workflows/issue-tests.yml
+++ b/.github/workflows/issue-tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [2.23.1.0-b220, 2024.1.4.0-b108, 2024.2.1.0-b185, 2.20.9.0-b47, 2.25.0.0-b489]
+        version: [2.23.1.0-b220, 2024.1.4.0-b108, LATEST_STABLE, 2.20.9.0-b47, 2.25.0.0-b489]
     env:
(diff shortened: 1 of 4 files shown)


.github/workflows/issue-tests.yml
```diff
@@ -28,6 +28,14 @@ jobs:
 
     - name: Setup YugabyteDB
       run: |
+        # Note: Update val for LATEST_STABLE here along with 'src/ybversion/constants.go/LATEST_STABLE' variable
+        if [ "${YB_VERSION}" == "LATEST_STABLE" ]; then
```
makalaaneesh, line 32: Can you check if we can use env vars for this?  https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-environment-variables-for-a-single-workflow

Becomes slightly simpler because we don't have to set it explicitly here again.

.github/workflows/issue-tests.yml
```diff
@@ -50,7 +58,7 @@ jobs:
         go test -v ./... -tags 'issues_integration'
 
     - name: Test Latest Stable YB Version
-      if: ${{ matrix.version == '2024.2.0.0-b145' }}
+      if: ${{ matrix.version == 'LATEST_STABLE' }}
```
makalaaneesh, line 61: Also, we're doing all this just because we don't want this to be run in every matrix version, right? 
Can we just add this to another workflow? maybe misc-tests?

sanyamsinghal: As per my understanding, the matrix in this workflow is a list of voyager version, frequently updated with YB version, including latest ones which we want to test all the issues against.

So why create a new workflow? that will increase the work to maintain it and chances of missing to update.

.github/workflows/issue-tests.yml
sanyamsinghal: @anee I didn't how you are suggesting to use the env var.
Can you share a sample of it

sanyamsinghal: @makalaaneesh  I didn't how you are suggesting to use the env var.
Can you share a sample of it

.github/workflows/issue-tests.yml
priyanshi-yb: I think what Aneesh is saying it to shift this `Test Latest stable` one to another already existing workflow that doesn't have matrix.versions or runs as a single workflow so that we don't have to do this check.
We don't use this matrix version information for this particular test.

sanyamsinghal: Ok got your point.

sanyamsinghal: This new code is providing a placeholder to update the latest stable release and kind of enforcing the developers to update it when required so the issues-against thing also run for latest stable.

I see some value in that.
What do you guys think?

yb-voyager/src/ybversion/constants.go
```diff
@@ -66,5 +73,5 @@ func init() {
 		panic("could not create version 2.25.0.0")
 	}
```
priyanshi-yb, line 75: nit: Add the above note here as this is place we actually change the version -
```
// Note: Whenever LatestStable is updated, modify in issues-test.yml as well
```

priyanshi-yb (approved):
LGTM

.github/workflows/issue-tests.yml
```diff
@@ -29,6 +29,14 @@ jobs:
 
     - name: Setup YugabyteDB
       run: |
+        # Note: Update val for LATEST_STABLE here along with 'src/ybversion/constants.go/LATEST_STABLE' variable
+        if [ "${YB_VERSION}" == "LATEST_STABLE" ]; then
+          YB_VERSION="2024.2.2.0-b61"
+        fi
+
+        # write it to GitHub's environment to override previous value
```
makalaaneesh, line 37: don't we want to do this only in the if-block above?

makalaaneesh (approved):
LGTM with one comment

.github/workflows/issue-tests.yml
sanyamsinghal: yeah correct, although end result would be same but inside if block is correct place.

sanyamsinghal: done

PR #2433, opened 2025-03-25
Miscellaneous  UI/UX changes for the single table-list changes for live migration

priyanshi-yb:
Made the following changes for each of these in this ticket https://github.com/yugabyte/yb-voyager/issues/2398- 
1. Fixed the output with better UI -
Screenshot 2025-03-25 at 7 34 54 PM

2. For this, only Fixed only the table-list printed when the command is running to also have the root to leaf differentiation in output similar to export data from source-
Screenshot 2025-03-25 at 7 42 54 PM

 Not making any changes to the guardrails output as it will be a bit complicated to do now.

3. Fixed these msgs
             - Aborting, Start a fresh migration if required ….
             - Re-run the command with the table list passed in the initial migration run or start a fresh migration.

4. Made all the table-lists displayed to print qualified tables in all cases (see above outputs)

fixes https://github.com/yugabyte/yb-voyager/issues/2398 

A few misc UI/UX changes 

manually

### DIFF
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -303,7 +303,7 @@ func exportData() bool {
 	tableListTuplesToDisplay := lo.Map(finalTableList, func(table sqlname.NameTuple, _ int) sqlname.NameTuple {
 		renamedTable, isRenamed := renameTableIfRequired(table.ForOutput())
 		if isRenamed {
-			t := table.ForMinOutput()
+			t := table.ForOutput()
 			//Fine to lookup directly as this will root table in case of partitions
 			tuple, err := namereg.NameReg.LookupTableName(renamedTable)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -303,7 +303,7 @@ func exportData() bool {
 	tableListTuplesToDisplay := lo.Map(finalTableList, func(table sqlname.NameTuple, _ int) sqlname.NameTuple {
 		renamedTable, isRenamed := renameTableIfRequired(table.ForOutput())
 		if isRenamed {
-			t := table.ForMinOutput()
+			t := table.ForOutput()
```
makalaaneesh, line 306: what is the reason for changing it from `ForMinOutput` to `ForOutput`

makalaaneesh (approved):
LGTM!

yb-voyager/cmd/exportData.go
priyanshi-yb: this is regarding the 4th UI issue in the above ticket, where for the targe-db-schema non-public also we were displaying the only tables in MinOutput and qualified at some places with Output so made it consistent ForOutput in all scenarios

PR #2442, opened 2025-03-27
Resumption Test Enhancement: Capture the exit code in case of interruptions

shubham-yb:

### DIFF
--- a/migtests/scripts/resumption.py
+++ b/migtests/scripts/resumption.py
@@ -198,6 +198,8 @@ def run_command(command, allow_interruption=False, interrupt_after=None):
             print("\nCommand Errors:\n")
             for line in stderr.splitlines():
                 print(line)
+            # If there is any stderr output, treat it as an error and exit.
+            # In the interrupt-retry scenario, we do not expect stderr output. The command should be interrupted without errors.
             sys.exit(1)
 
         # If interrupted, check the exit code
--- a/migtests/scripts/resumption.py
+++ b/migtests/scripts/resumption.py
@@ -171,12 +171,15 @@ def run_command(command, allow_interruption=False, interrupt_after=None):
                     try:
                         process.terminate()
                         print("Terminate signal sent to process (PID: {}). Waiting for process to exit...".format(process.pid), flush=True)
-                        process.wait(timeout=10)
-                        print("Process (PID: {}) terminated gracefully.".format(process.pid), flush=True)
+
+                        process.wait(timeout=10)  # Wait for the process to exit
+                        print(f"Process (PID: {process.pid}) terminated gracefully with exit code: {process.returncode}", flush=True)
+
                     except subprocess.TimeoutExpired:
                         print("Process (PID: {}) did not terminate in time. Forcing termination...".format(process.pid), flush=True)
                         process.kill()
-                        print("Process (PID: {}) killed.".format(process.pid), flush=True)
+                        print(f"Process (PID: {process.pid}) killed with exit code: {process.returncode}", flush=True)
+
                     interrupted = True
                     break
             time.sleep(1)  # Avoid busy-waiting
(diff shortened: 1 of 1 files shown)


shubham-yb:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/5032/

migtests/scripts/resumption.py
```diff
@@ -191,16 +194,22 @@ def run_command(command, allow_interruption=False, interrupt_after=None):
...
         if stderr:
             print("\nCommand Errors:\n")
             for line in stderr.splitlines():
                 print(line)
+            sys.exit(1)
 
-        completed = process.returncode == 0 and not interrupted
+        # If interrupted, check the exit code
+        if interrupted:
+            if process.returncode not in {1, -9, 137}:  # -9 and 137 are SIGKILL variations
```
sanyamsinghal, line 205: there are atleast three signal we should test specifically(Not user if there are more or any custom/USR signal can do the same)
- SIGTERM, SIGKILL and SIGINT 

Lets handle these and update the comment.

sanyamsinghal: In case of these signal we should also capture the error message specifically of the import data command and print it here.

Is that already taken care of for every intermediate runs?

migtests/scripts/resumption.py
```diff
@@ -191,16 +194,22 @@ def run_command(command, allow_interruption=False, interrupt_after=None):
             print("\nCommand Output:\n")
             for line in stdout.splitlines():
                 print(line)
-
         if stderr:
```
sanyamsinghal, line 197: Are we printing stderr  for last execution only?

sanyamsinghal (changes requested):

migtests/scripts/resumption.py
shubham-yb: No it will print for intermediate runs as well if there is an stderr.

migtests/scripts/resumption.py
shubham-yb: The stdout or stderr of the commands will already be printed. Then if we encounter such a situation, we will exit as well. 
Yes it will happen for intermediate runs.

shubham-yb: Voyager handles SIGTERM internally and returns exit code 1 in that case. We aren't using SIGINT anywhere in the script so if we receive its exit code we should flag it.
Currently we are running SIGTERM and SIGKILL. If we should add SIGINT as well, we can take it up in the upcoming enhancements.

sanyamsinghal (approved):

migtests/scripts/resumption.py
```diff
@@ -171,12 +171,15 @@ def run_command(command, allow_interruption=False, interrupt_after=None):
                     try:
                         process.terminate()
                         print("Terminate signal sent to process (PID: {}). Waiting for process to exit...".format(process.pid), flush=True)
-                        process.wait(timeout=10)
-                        print("Process (PID: {}) terminated gracefully.".format(process.pid), flush=True)
+
+                        process.wait(timeout=10)  # Wait for the process to exit
```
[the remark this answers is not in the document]
shubham-yb: process.terminate() is not synchronous.
It sends a termination signal (SIGTERM), but does not wait for the process to actually exit.

migtests/scripts/resumption.py
```diff
@@ -191,16 +194,22 @@ def run_command(command, allow_interruption=False, interrupt_after=None):
...
+        if interrupted:
+            if process.returncode not in {1, -9, 137}:  # -9 and 137 are SIGKILL variations
+                print(f"Unexpected exit code after interruption: {process.returncode}", flush=True)
+                sys.exit(1)
 
+        completed = process.returncode == 0 and not interrupted
         return completed, stdout, stderr
 
+
 def run_and_resume_voyager(command):
```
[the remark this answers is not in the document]
shubham-yb: This is already handled — the `while process.poll() is None:` loop exits automatically when the command completes, so we don’t need explicit handling for that scenario. The process is checked on every iteration.

shubham-yb:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/5096/

PR #2451, opened 2025-04-01
Send information about control plane usage in the migration to callhome

priyanshi-yb:
Added information to phase_payload of every migration_phase about YugabyteD UI enable or not.
fixes - #2388 

no

manually
Added new field to all phase_payloads but its not a breaking change as its a json column

### DIFF
--- a/yb-voyager/src/callhome/diagnostics.go
+++ b/yb-voyager/src/callhome/diagnostics.go
@@ -112,6 +112,7 @@ type AssessMigrationPhasePayload struct {
 	IndexSizingStats               string                    `json:"index_sizing_stats"`
 	SourceConnectivity             bool                      `json:"source_connectivity"`
 	IopsInterval                   int64                     `json:"iops_interval"`
+	YugabyteDUIEnable              bool                      `json:"yugabyted_ui_enable"`
 }
 
 type AssessmentIssueCallhome struct {
@@ -145,8 +146,9 @@ type ObjectSizingStats struct {
 }
 
 type AssessMigrationBulkPhasePayload struct {
-	FleetConfigCount int    `json:"fleet_config_count"` // Not storing any source info just the count of db configs passed to bulk cmd
-	Error            string `json:"error"`
+	FleetConfigCount  int    `json:"fleet_config_count"` // Not storing any source info just the count of db configs passed to bulk cmd
+	Error             string `json:"error"`
+	YugabyteDUIEnable bool   `json:"yugabyted_ui_enable"`
 }
 
 type ExportSchemaPhasePayload struct {
@@ -155,6 +157,7 @@ type ExportSchemaPhasePayload struct {
 	UseOrafce              bool   `json:"use_orafce"`
 	CommentsOnObjects      bool   `json:"comments_on_objects"`
 	Error                  string `json:"error"`
+	YugabyteDUIEnable      bool   `json:"yugabyted_ui_enable"`
 }
 
 var ANALYZE_PHASE_PAYLOAD_VERSION = "1.0"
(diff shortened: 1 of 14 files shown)


yb-voyager/src/callhome/diagnostics.go
```diff
@@ -112,6 +112,7 @@ type AssessMigrationPhasePayload struct {
 	IndexSizingStats               string                    `json:"index_sizing_stats"`
 	SourceConnectivity             bool                      `json:"source_connectivity"`
 	IopsInterval                   int64                     `json:"iops_interval"`
+	YugabyteDUIEnabled             bool                      `json:"yugabyted_ui_enabled"`
```
makalaaneesh, line 115: Shall we just send "control_plane_type" instead of a boolean? 
That will cover for other control plane types as well in the future.

priyanshi-yb: I think sending the direct information helps; else we have to keep track of the values of control_plane_type in callhome queries, and in the future, if we change it we will have to update that.

sanyamsinghal: > Shall we just send "control_plane_type"

I think that might be more helpful and future-proof.
In anycase we will have to check for the flag's value

We can have this as fixed const var to incorporate multiple values - `yugabyted`, `yugabytedb_anywhere`, `yugabytedb_aeon`

If we are concerned about case - voyager sending to multiple control plane at one time then maybe use `[]string`

makalaaneesh: Yeah, I still feel like `control_plane_type` would be better. 

>we have to keep track of the values of control_plane_type in callhome queries, and in the future, if we change it we will have to update that. 

The way I look at it is: Send all the raw data to callhome, and write queries at that layer for interpretation (derived data). 
So, in this context, the control_plane_type is the raw data. "Whether yugabytedUI was used or not" is a derived data point, which can be defined at the callhome layer. 

If/when we introduce new control plane tyeps (yba,ybaeon), then there will be no change required in voyager, only in the callhome layer, which seems ideal, don't you think?

priyanshi-yb: Hmm, okay, I got your point. Let me make that change.

priyanshi-yb: Is it okay to assume a single control plane type for now?

sanyamsinghal: I am really not sure about this - how does it work in real world usecase.
Depends on if it is normal or frequent for a tool or service to be managed by multiple control planes.

cc @makalaaneesh

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -112,6 +112,7 @@ type AssessMigrationPhasePayload struct {
 	IndexSizingStats               string                    `json:"index_sizing_stats"`
 	SourceConnectivity             bool                      `json:"source_connectivity"`
 	IopsInterval                   int64                     `json:"iops_interval"`
+	ControlPlaneType               string                      `json:"control_plane_type"`
```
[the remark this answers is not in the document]
priyanshi-yb: > I believe the callhome queries check for >1.0

its >= 1.0.
I thought this is just an addition of a field to json so its not required but increased it to 1.1

yb-voyager/src/callhome/diagnostics.go
makalaaneesh: I think assumption of a single control plane is fine for now. Typically, that is how it is going to be set up. Don't see why one would need multiple control planes. 
I don't think our code allows multiple control planes either..

yb-voyager/src/callhome/diagnostics.go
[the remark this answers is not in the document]
makalaaneesh: True, it won't really matter so much (bigquery won't error out if the field is not present). But, just as a good practice...

makalaaneesh (approved):
LGTM

PR #2456, opened 2025-04-02
Monitoring the node status and replication running during import data for target cluster

priyanshi-yb:
-Added monitoring of the target cluster health on various parameters-
1. Node status—Ensured adaptation of the data load to only go to the live nodes and remove the connections from the pool for the down nodes for extra safety, the conn-pool already handles most of the adaptation and also alert the user with a message in import-data for a node going down so that user can take necessary actions.
2. Replication streams—Added monitoring for any replication stream on the target during the import process, asking the user to enable it after the data load is completed to keep the disk usage limited to the data load and the cluster stable. In this case, we abort the import data so that the user can take action and then re-run the import.

Func spec for reference https://docs.google.com/document/d/13P7xJ_bZ7d3RGcxeVszY3azAtSObqGNFXRaU0k6LYeE/edit?tab=t.0

-Added a way to report any extra information during the progress bar or the streaming live table on the console for anything important for the user.

yes, three new boolean flags in the import data for skipping any of the monitoring health checks. In case the node goes down user will msgs on the console to take necessary actions.
Sample msg for node status in case of progress bars and streaming phase - 

During progress-bar
Screenshot 2025-04-11 at 11 16 24 PM

During streaming events table

Screenshot 2025-04-12 at 7 27 45 PM

Manually.


debezium-server-voyager/yb-client-cdc-stream-wrapper/src/main/java/Main.java
```diff
@@ -51,8 +53,15 @@ public static void main(String[] args) throws Exception {
             masterAddressesList = masterAddressesList.replace("{", ""); //removing {}
             masterAddressesList = masterAddressesList.replace("}", "");
             System.out.println("Master Addresses: " + masterAddressesList);
+        } else if (parameters.getNumOfCDCStreams) {
+            ListCDCStreamsResponse cdcStreamsResponse = client.listCDCStreams(null, null, null);
```
makalaaneesh, line 57: what are the parameters to this function? (to which we are passing NULLs)?

yb-voyager/cmd/import.go
```diff
@@ -354,8 +354,17 @@ func registerFlagsForTarget(cmd *cobra.Command) {
 	BoolVar(cmd.Flags(), &tconf.EnableYBAdaptiveParallelism, "enable-adaptive-parallelism", true,
 		"Adapt parallelism based on the resource usage (CPU, memory) of the target YugabyteDB cluster.")
 	cmd.Flags().IntVar(&tconf.MaxParallelism, "adaptive-parallelism-max", 0,
-		"number of max parallel jobs to use while importing data when adaptive parallelism is enabled."+
+		"number of max parallel jobs to use while importing data when adaptive parallelism is enabled. "+
 			"By default, voyager will try if it can determine the total number of cores N and use N/2 as the max parallel jobs. ")
+	BoolVar(cmd.Flags(), &skipReplicationChecks, "skip-replication-checks", false,
```
makalaaneesh, line 359: In the help msg, we should indicate that voyager will abort if it detects replication on the target cluster.

makalaaneesh: How about something like `It is NOT recommended to have any form of replication (CDC/xcluster) running on the YB cluster during import-data. If detected, import-data is aborted. Use this flag to turn off the checks and continue importing data`

yb-voyager/cmd/import.go
```diff
@@ -354,8 +354,17 @@ func registerFlagsForTarget(cmd *cobra.Command) {
...
+		"Skips the monitoring of the replication enabled or not on the target YugabyteDB cluster. "+
+			"By default, voyager will keep monitoring of the replication enabled to keep the cluster stable.")
+	BoolVar(cmd.Flags(), &skipNodeHealthChecks, "skip-node-health-checks", false,
```
makalaaneesh, line 362: We are not really aborting in this case, right? Do we still need this flag?

yb-voyager/cmd/import.go
```diff
@@ -354,8 +354,17 @@ func registerFlagsForTarget(cmd *cobra.Command) {
...
+		"Skips the monitoring of the Node status checks on the target YugabyteDB cluster. "+
+			"By default, voyager will keep monitoring the node status to keep the cluster stable.")
+	BoolVar(cmd.Flags(), &skipDiskUsageHealthChecks, "skip-disk-usage-health-checks", false,
```
makalaaneesh, line 365: let's hide this for now, since there is no implementation for disk usage health check?

yb-voyager/cmd/importData.go
```diff
@@ -896,6 +902,45 @@ func createFileTaskImporter(task *ImportFileTask, state *ImportDataState, batchI
 	return taskImporter, nil
 }
 
+func startMonitoringTargetHealth() error {
```
makalaaneesh, line 905: `startMonitoringTargetYBHealth` to be specific.

yb-voyager/cmd/importData.go
```diff
@@ -896,6 +902,45 @@ func createFileTaskImporter(task *ImportFileTask, state *ImportDataState, batchI
...
+	}
+	if skipNodeHealthChecks && skipDiskUsageHealthChecks && skipReplicationChecks {
+		return nil
+	}
+	yb, ok := tdb.(*tgtdb.TargetYugabyteDB)
+	if !ok {
+		return fmt.Errorf("monitoring health is only supported if target DB is YugabyteDB")
+	}
+	go func() {
+		err := monitor.MonitorTargetHealth(yb, metaDB, skipDiskUsageHealthChecks, skipReplicationChecks, skipNodeHealthChecks, func(info string) {
```
makalaaneesh, line 917: `MonitorTargetYBHealth`

yb-voyager/cmd/importData.go
```diff
@@ -896,6 +902,45 @@ func createFileTaskImporter(task *ImportFileTask, state *ImportDataState, batchI
...
+}
+
+func displayMonitoringInformationOnTheConsole(info string) {
+	if info == "" {
+		return
+	}
+	if disablePb {
+		utils.PrintAndLog(info)
+	} else {
+		log.Warnf("monitoring: %v", info)
```
makalaaneesh, line 934: Is something like `ALERT:` better?

yb-voyager/cmd/importData.go
makalaaneesh, line 917: do we still need metadb, now that you are passing the callback function?

yb-voyager/src/dbzm/yb_cdc_client.go
```diff
@@ -159,6 +160,29 @@ func (ybc *YugabyteDBCDCClient) ListMastersNodes() (string, error) {
 	return masterAddresses, nil
 }
 
+func (ybc *YugabyteDBCDCClient) GetNumOfReplicationStreams() (int, error) {
+	tserverPort := "9100" //TODO: make it internally handled by yb-client
+	if os.Getenv("YB_TSERVER_PORT") != "" {
```
makalaaneesh, line 165: Use `GetEnvAsInt` ?

yb-voyager/src/dbzm/yb_cdc_client.go
```diff
@@ -159,6 +160,29 @@ func (ybc *YugabyteDBCDCClient) ListMastersNodes() (string, error) {
...
+	if ybc.sslRootCert != "" {
+		args += fmt.Sprintf(" -ssl_cert_file '%s'", ybc.sslRootCert)
+	}
+
+	stdout, err := ybc.runCommand(args)
+	if err != nil {
+		return 0, fmt.Errorf("running command with args: %s, error: %s", args, err)
+	}
+	//stdout - Streams: <num_of_streams>
+	numOfStreams, err := strconv.Atoi(strings.Trim(strings.Split(stdout, ": ")[1], " \n"))
```
makalaaneesh, line 179: we should check first that the output of split is a non-empty slice. Directly indexing the array `[1]` can lead to a panic

yb-voyager/src/tgtdb/conn_pool.go
```diff
@@ -280,6 +281,32 @@ func (pool *ConnectionPool) shuffledConnUriList() []string {
...
 
+func (pool *ConnectionPool) RemoveConnectionsForHosts(servers []string) error {
+	log.Infof("Checking for connections on host: %s", servers)
+	var conn *pgx.Conn
+	var gotIt bool
+	size := pool.size
+	for i := 0; i < size; i++ {
+		conn, gotIt = <-pool.conns
+		if !gotIt {
+			break
```
makalaaneesh, line 292: why break here?

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1483,3 +1505,23 @@ func IsCurrentUserSuperUser(tconf *TargetConf) (bool, error) {
 func (yb *TargetYugabyteDB) GetEnabledTriggersAndFks() (enabledTriggers []string, enabledFks []string, err error) {
 	return nil, nil, nil
 }
+
+func (yb *TargetYugabyteDB) NumOfLogicalReplicationSlots() (int64, error) {
+	query := "SELECT count(slot_name) from pg_replication_slots"
+	rows, err := yb.Query(query)
+	if err != nil {
+		return 0, fmt.Errorf("querying if user is superuser: %w", err)
```
makalaaneesh, line 1513: wrong error message here.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1483,3 +1505,23 @@ func IsCurrentUserSuperUser(tconf *TargetConf) (bool, error) {
 func (yb *TargetYugabyteDB) GetEnabledTriggersAndFks() (enabledTriggers []string, enabledFks []string, err error) {
 	return nil, nil, nil
 }
+
+func (yb *TargetYugabyteDB) NumOfLogicalReplicationSlots() (int64, error) {
+	query := "SELECT count(slot_name) from pg_replication_slots"
+	rows, err := yb.Query(query)
```
makalaaneesh, line 1511: I think you can use QueryRow?

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,205 @@
+/*
```
makalaaneesh, line 1: yet to review this.

makalaaneesh (commented):
First round of comments. 
Yet to review `monitor` pkg

yb-voyager/cmd/import.go
priyanshi-yb: Yeah not really sure. This is just for disabling the monitoring of the node status completely as well just in case if required.

yb-voyager/cmd/importData.go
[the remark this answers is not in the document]
priyanshi-yb: its coming from the monitor pkg

yb-voyager/cmd/importData.go
priyanshi-yb: yeah no, will remove it.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -222,7 +225,17 @@ func (yb *TargetYugabyteDB) PrepareForStreaming() {
 }
 
 func (yb *TargetYugabyteDB) InitConnPool() error {
-	tconfs := yb.getYBServers()
+	var tconfs []*TargetConf
+	loadBalancerUsed, confs, err := yb.GetYBServers()
+	if err != nil {
+		return fmt.Errorf("error fetching the yb servers: %v", err)
+	}
+	if loadBalancerUsed { // if load balancer is used no need to check direct connectivity
```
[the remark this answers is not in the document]
priyanshi-yb: hmm, I wanted to get this information out if load balancer is used or not, but yeah getting this logic out doesn't make sense now. will put it back in the function itself.

yb-voyager/src/tgtdb/conn_pool.go
priyanshi-yb: As per my understanding, I think this gotIt will be false in case the pool doesn't have connections or all conns are busy so in such case, it doesn't make sense to go further we keep on getting gotIT false only I think ?🤔

debezium-server-voyager/yb-client-cdc-stream-wrapper/src/main/java/Main.java
makalaaneesh: got it, thanks! 
Shouldn't we be only checking for cdc streams in that database to which we are ingesting? 
Also, can we have streams on CQL tables? That is also something we shouldn't be concerned about, right?

yb-voyager/cmd/import.go
makalaaneesh: Got it, but we don't seem to be doing much with the monitoring. So can't think of why we might want to expose this flag. If we're unsure, let's either remove it or hide it at least.

yb-voyager/src/tgtdb/yugabytedb.go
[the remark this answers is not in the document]
makalaaneesh: I think you had changed the logic of `GetCallhomeTargetDBInfo.NodeCount` to be based on the no. of servers, not tconfs (because that's not accurate when loadBalancerUsed. I remember seeing that in your previous commit. That is actually useful. 

Pls make the change accordingly. Let's just ensure that there is no duplication of logic. Create separate functions if necessary.

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,203 @@
...
+
+type TargetDBForMonitorHealth interface {
+	GetYBServers() (bool, []*tgtdb.TargetConf, error)
+	RemoveConnectionsForHosts([]string) error
+	NumOfLogicalReplicationSlots() (int64, error)
+	GetClusterMetrics() (map[string]tgtdb.NodeMetrics, error) // node_uuid:metric_name:metric_value
+}
+
+// bool variable is for indicating if a node is down or up
+var nodeStatuses map[string]bool
```
makalaaneesh, line 50: pls add a comment explaining what the key is.

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,203 @@
...
+		//Add the node in the main nodeStatuses initialised in the starting
+		//which weren't present the starting and during the run it came back.
+		if !slices.Contains(lo.Keys(nodeStatuses), conf.Host) {
+			nodeStatuses[conf.Host] = true
+		}
+	}
+	if len(downNodes) > 0 {
+		downNodeMsg := fmt.Sprintf("Following nodes are not healthy, please check them - [%v]", strings.Join(downNodes, ", "))
+		err := yb.RemoveConnectionsForHosts(downNodes)
+		if err != nil {
```
makalaaneesh, line 127: Shouldn't we be additionally updating `nodestatuses[node] = false` ? I guess it doesn't really make sense for us to be bombarding the user with a message "node is down" every single time we run the check. We should just do it once.

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,203 @@
...
+		return "", nil
+	}
+	_, currentNodes, err := yb.GetYBServers()
+	if err != nil {
+		return "", fmt.Errorf("error fetching servers informtion: %v", err)
+	}
+	downNodes := make([]string, 0)
+
+	for node, up := range nodeStatuses {
+		if up && !lo.ContainsBy(currentNodes, func(t *tgtdb.TargetConf) bool {
```
makalaaneesh, line 111: if nodeStatus was down, but now it has come back up (i.e. it is present in currentNodes), probably worth indicating to the user via a message on console that the node is back up?

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,203 @@
...
+			nodeStatuses[conf.Host] = true
+		}
+	}
+	if len(downNodes) > 0 {
+		downNodeMsg := fmt.Sprintf("Following nodes are not healthy, please check them - [%v]", strings.Join(downNodes, ", "))
+		err := yb.RemoveConnectionsForHosts(downNodes)
+		if err != nil {
+			return downNodeMsg, fmt.Errorf("error while removing the connections for the down nodes[%v]: %v", downNodes, err)
+		}
+		return downNodeMsg, nil
```
makalaaneesh, line 130: It might be worth passing the displayAlertFunc into these inner functions as well, instead of returning each msg, and then calling it in the main function. 

Reasons:
- You might want to send multiple alerts 
- Multiple functions (monitor*) might want to send alerts.

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,203 @@
...
+		return nil
+	}
+	numOfSlots, err := yb.NumOfLogicalReplicationSlots()
+	if err != nil {
+		return fmt.Errorf("error fetching logical replication slots for checking if replication enabled - %s", err)
+	}
+	ybServers := lo.Keys(nodeStatuses)
+	addresses := strings.Join(ybServers, ":7100,")
+	// no need to specify the empty params as we need to only get num of cdc streams
+	ybClient := dbzm.NewYugabyteDBCDCClient("", addresses, "", "", "", nil)
```
makalaaneesh, line 190: Doesn't make sense to initialize this every single time. (even though I suspect it might be lightweight). 

Can we initialize this once in the caller function, and pass it down to this function? 
Or use a global variable and check that it is inited. 

Alternatively, one option could be to create a struct for monitoring?
We seem to be saving node statuses. we need a handle to the ybclient to be able to make that call to get CDC streams..

yb-voyager/src/dbzm/yb_cdc_client.go
```diff
@@ -154,11 +156,39 @@ func (ybc *YugabyteDBCDCClient) ListMastersNodes() (string, error) {
...
+	}
+
+	stdout, err := ybc.runCommand(args)
+	if err != nil {
+		return 0, fmt.Errorf("running command with args: %s, error: %s", args, err)
+	}
+	//stdout - Streams: <num_of_streams>
+	numOfStreams := 0
+	stdoutSplits := strings.Split(stdout, ": ")
+	if len(stdoutSplits) > 1 {
```
makalaaneesh, line 183: wouldn't it be an error if len(stdoutSplits) is not greater than 1? 
Also, let's be specific. We expect no. of splits to be exactly 2, correct? Anything other than that, we should return an error.

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,203 @@
...
+			log.Errorf("error monitoring the disk and memory: %v", err)
+		}
+
+		err = monitorReplicationOnTarget(yb, skipReplicationChecks)
+		if err != nil {
+			log.Errorf("error monitoring the replication: %v", err)
+		}
+
+		if nodeAlert != "" {
+			alertMsg := color.RedString(fmt.Sprintf("Alert!\n%s\n", nodeAlert))
```
makalaaneesh, line 83: Let's have `ALERT: xyz` (ALERT in caps, no new line). 
I feel like that will be more concise, and similar to a "log" line

yb-voyager/src/tgtdb/conn_pool.go
makalaaneesh: I think you're right. 
Let's discuss this on call. I also want to go through the existing implementation of the conn pool to see if we might be missing anything.

yb-voyager/src/tgtdb/conn_pool.go
```diff
@@ -280,6 +281,32 @@ func (pool *ConnectionPool) shuffledConnUriList() []string {
 	return connUriList
 }
 
+func (pool *ConnectionPool) RemoveConnectionsForHosts(servers []string) error {
```
makalaaneesh, line 284: see if we can write unit tests for this part.

yb-voyager/src/tgtdb/conn_pool.go
```diff
@@ -280,6 +281,32 @@ func (pool *ConnectionPool) shuffledConnUriList() []string {
...
+	var conn *pgx.Conn
+	var gotIt bool
+	size := pool.size
+	for i := 0; i < size; i++ {
+		conn, gotIt = <-pool.conns
+		if !gotIt {
+			break
+		}
+
+		if conn == nil {
```
makalaaneesh, line 295: we should continue here, not break. 
Also return the connection back to the pool

yb-voyager/src/tgtdb/conn_pool.go
```diff
@@ -280,6 +281,32 @@ func (pool *ConnectionPool) shuffledConnUriList() []string {
...
+			break
+		}
+		if slices.Contains(servers, conn.Config().Host) {
+			log.Infof("Removing the connection for server: %s", conn.Config().Host)
```
makalaaneesh, line 299: `.. as it is down`

yb-voyager/src/tgtdb/conn_pool.go
makalaaneesh, line 284: document why this is needed clearly.

debezium-server-voyager/yb-client-cdc-stream-wrapper/src/main/java/Main.java
priyanshi-yb: as discussed, will create ticket for this and will do it separately

yb-voyager/cmd/importData.go
```diff
@@ -896,6 +902,48 @@ func createFileTaskImporter(task *ImportFileTask, state *ImportDataState, batchI
...
+		return nil
+	}
+	yb, ok := tdb.(*tgtdb.TargetYugabyteDB)
+	if !ok {
+		return fmt.Errorf("monitoring health is only supported if target DB is YugabyteDB")
+	}
+	go func() {
+		//for now not sending any other parameters as not required for monitor usage
+		ybClient := dbzm.NewYugabyteDBCDCClient(exportDir, "", "", "", "", nil)
+		monitorTDBHealth := monitor.NewMonitorTargetYBHealth(yb, skipDiskUsageHealthChecks, skipReplicationChecks, skipNodeHealthChecks, func(info string) {
```
makalaaneesh, line 919: nit, let the func callback be the last argument.. easier to read.

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,264 @@
...
+	SetYBServers(string)
+	SetSSLRootCert(string)
+	GetNumOfReplicationStreams() (int, error)
+}
+
+type MonitorTargetYBHealth struct {
+	// key is the target cluster node and value is bool variable for indicating if a node is down or up
+	nodesStatus map[string]bool
+	//ybClient instance to fetch streams
+	ybClient dbzm.YugabyteDBCDCClient
```
makalaaneesh, line 62: this should be `YugabyteDBClient` ?

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,264 @@
...
+
+func NewMonitorTargetYBHealth(yb TargetDBForMonitorHealth, skipDiskUsageHealthChecks utils.BoolStr, skipReplicationChecks utils.BoolStr, skipNodeHealthChecks utils.BoolStr, displayMsgFunc func(msg string), ybClient YugabyteDBClient, tconf tgtdb.TargetConf) MonitorTargetYBHealth {
+	return MonitorTargetYBHealth{
+		nodesStatus:          make(map[string]bool),
+		yb:                   yb,
+		skipNodeCheck:        bool(skipNodeHealthChecks),
+		skipDiskCheck:        bool(skipDiskUsageHealthChecks),
+		skipReplicationCheck: bool(skipReplicationChecks),
+		displayMsgFunc:       displayMsgFunc,
+		tconf:                tconf,
```
makalaaneesh, line 85: you don't seem to be setting `ybClient` here?

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,264 @@
...
+	for _, server := range servers {
+		m.nodesStatus[server.Host] = true
+	}
+
+	if loadBalancerEnabled {
+		m.skipNodeCheck = true
+	}
+
+	//skipping the disk usage monitoring for now, util the metrics function changes are made
+	m.skipDiskCheck = true
```
makalaaneesh, line 105: nit: instead of hardcoding it here, hardcode it while calling `NewMonitorTargetYBHealth`

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil
+	}
+	_, currentNodes, err := m.yb.GetYBServers()
+	if err != nil {
+		return fmt.Errorf("error fetching servers informtion: %v", err)
+	}
+	downNodes := make([]string, 0)
+	upNodes := make([]string, 0)
+
+	for node, up := range m.nodesStatus {
```
makalaaneesh, line 150: I'd recommend variable names that are slightly more helpful. How about something like:
`up` -> `previousUp`
`currentUp := lo.ContainsBy(currentNodes, ...`

Also, use either "up" or "live" consistently.

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	err = m.ybClient.Init()
+	if err != nil {
+		return fmt.Errorf("error intialising the yb client : %v", err)
+	}
+	ybServers := lo.Keys(m.nodesStatus)
+	addresses := strings.Join(ybServers, ":7100,")
+
+	m.ybClient.SetYBServers(addresses)
```
makalaaneesh, line 253: can this be done in the constructor?

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -0,0 +1,264 @@
...
+func (m *MonitorTargetYBHealth) monitorReplicationOnTarget() error {
+	if m.skipReplicationCheck {
+		return nil
+	}
+	numOfSlots, err := m.yb.NumOfLogicalReplicationSlots()
+	if err != nil {
+		return fmt.Errorf("error fetching logical replication slots for checking if replication enabled - %s", err)
+	}
+	if numOfSlots > 0 {
+		utils.ErrExit(color.RedString(REPLICATION_GUARDRAIL_ALERT_MSG))
```
makalaaneesh, line 243: We should mention in the error msg : "we found x no. of replication slots"
same for the replication streams case.

yb-voyager/src/tgtdb/conn_pool_test.go
```diff
@@ -204,3 +206,62 @@ func TestUpdateConnectionsRandom(t *testing.T) {
...
+		}
+	}
+
+	assert.Equal(t, size, len(pool.conns))
+	assert.Equal(t, size, len(pool.idleConns))
+
+	pool.RemoveConnectionsForHosts([]string{host})
+
+	assert.Equal(t, size, len(pool.conns))
+	assert.Equal(t, size, len(pool.idleConns))
```
makalaaneesh, line 264: let's also test that all the connections in conns, and idle are nil after the call in `261`? 
Also, (if easy to do), create a connection with a different host, port, and verify that `RemoveConnectionsForHosts` does not close that connection.

yb-voyager/src/monitor/monitor_target_health.go
priyanshi-yb: not in the constructor, moved it to the StartMonitoring() so that it is set only once.

yb-voyager/src/tgtdb/conn_pool_test.go
priyanshi-yb: > Also, (if easy to do), create a connection with a different host, port, and verify that RemoveConnectionsForHosts does not close that connection.

I think this might be easy to do as I need a multi-node cluster for this but I am not sure if that will be easily possible with testcontainers, it will require some work.
added a todo for now, maybe will create a ticket for enhancing the yugabyteTestContainer for this functionality then we can do this.

priyanshi-yb: > Also, (if easy to do), create a connection with a different host, port, and verify that RemoveConnectionsForHosts does not close that connection.

Tried It doesn't seem to be easy, it requires some changes in the test container to either make the host configurable (as host is always localhost) or configure it to start a multi-node YB cluster. So, adding a TODO for now.

yb-voyager/src/monitor/monitor_target_health.go
```diff
@@ -70,19 +68,17 @@ type MonitorTargetYBHealth struct {
 
 	//Display function for the printing any msg on the console
 	displayMsgFunc func(string)
-
-	tconf tgtdb.TargetConf
 }
 
-func NewMonitorTargetYBHealth(yb TargetDBForMonitorHealth, skipDiskUsageHealthChecks utils.BoolStr, skipReplicationChecks utils.BoolStr, skipNodeHealthChecks utils.BoolStr, displayMsgFunc func(msg string), ybClient YugabyteDBClient, tconf tgtdb.TargetConf) MonitorTargetYBHealth {
+func NewMonitorTargetYBHealth(yb TargetDBForMonitorHealth, skipDiskUsageHealthChecks utils.BoolStr, skipReplicationChecks utils.BoolStr, skipNodeHealthChecks utils.BoolStr, ybClient YugabyteDBClient, displayMsgFunc func(msg string)) MonitorTargetYBHealth {
```
makalaaneesh, line 73: we should avoid using Boolstr outside the command layer (specifically when accepting inputs for a flag). For downstream code, plain booleans make more sense.

yb-voyager/src/tgtdb/conn_pool_test.go
makalaaneesh: one idea is to create a new hostname that maps to the same IP, and create a connection to that new hostname. Idk if that will work for sure, because the underlying IP would be the same, but just something worth trying. Okay to take it up later 👍

makalaaneesh (approved):
LGTM!

PR #2483, opened 2025-04-11
[Resumption Test Enhancement] Added randomness of signals and framework for varying values of flags

shubham-yb:

Fixes: https://yugabyte.atlassian.net/browse/DB-15834 , https://yugabyte.atlassian.net/browse/DB-15837

Jenkins side PR: https://github.com/yugabyte/jenkins-helpers/pull/1031

- Added framework to vary the value of flags on each invocation of the command.
- Added test case to vary batch sizes and parallel jobs.
- Added handling for all three signals (SIGINT / SIGTERM / SIGKILL)
- Added README for the resumption script.


migtests/scripts/resumption.py
sanyamsinghal, line 1: Our resumption script seems to becoming quiet advanced with multiple configuration options.

Suggestion: A small readme for this directory would be help for everyone to understand the capabilities of this script.
Don't need to be very formal, just documenting all types of configuration possible and some examples of config file maybe.

shubham-yb:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline-test/1124/

migtests/scripts/resumption.py
```diff
@@ -155,30 +157,54 @@ def prepare_import_data_command(config):
...
 
+def inject_varying_flags(command):
+    global varying_flags
+
+    for flag, setting in varying_flags.items():
+        value_list = setting["value"]
+        if setting["type"] == "range":
+            value = random.randint(value_list[0], value_list[1])
+        elif setting["type"] == "choice":
+            value = random.choice(value_list)
```
sanyamsinghal, line 168: what does choice means?
whats the d/f b/w choice() and randint() here

migtests/scripts/resumption.sh
```diff
@@ -41,6 +45,18 @@ fi
 source ${SCRIPTS}/yugabytedb/env.sh
 source ${SCRIPTS}/functions.sh
 
+# Handle optional additional config file ($3)
+if [ -n "$3" ]; then
+    ADDITIONAL_CONFIG_FILE="${TESTS_DIR}/resumption/additional_configs/$3"
+    if [ ! -f "${ADDITIONAL_CONFIG_FILE}" ]; then
+        echo "Error: Additional config file $3 not found in resumption/additional_configs"
+        exit 1
+    fi
```
sanyamsinghal, line 54: Just to confirm, we are not restricting the config file to be placed here under here only - `migtests/tests/resumption/additional_configs`

Few questions:
1. Is this the only one test we have as of today?
2. For eg if we add one more resumption test tomorow, will that be a different directory, what will the structure look like?

migtests/scripts/resumption.sh
```diff
@@ -89,8 +105,18 @@ main() {
 	  ./generate_config.py
 	fi
 
-	step "Run import with resumptions"
+	# Backup original config
+	cp config.yaml config.yaml.original
 
+	# Append additional config to config.yaml if specified
```
sanyamsinghal, line 111: where is this config.yaml reside?
Is it being created at run by the script

migtests/scripts/resumption.sh
```diff
@@ -101,9 +127,17 @@ main() {
 
 	step "Verify import-data-status report"
 	verify_report ${expected_file} ${actual_file}
-	
+
 	step "Clean up"
 	rm -rf "${EXPORT_DIR}"
+
+	# Restore original config if it was modified
```
sanyamsinghal, line 134: some code comments about the flow/changes of config.yaml file would help.

migtests/scripts/resumption.py
shubham-yb: "Range" means a random integer will be picked from a start to end range using random.randint().
"Choice" means the value will be randomly picked from a predefined list of options in the config file using random.choice().

Sample config for clarity:
```
--parallel-jobs:
    type: range
    value: [1, 10]
  --batch-size:
    type: choice
    value: [1234, 2500, 10500, 20000]
```

migtests/scripts/resumption.sh
shubham-yb: `resumption/additional_configs` contain files which one can use to customise their runs. So the base tests will be under the resumption folder and within each test there will be a config file, or a means to generate it.
1. The files under additional_configs can be used with any of the base tests and thus giving flexibility. So think of it as one enhancement on top of the tests present.
2. If we want to add another test, it will have to be added in another directory like any test in the migtests folder.
Example test: migtests/tests/resumption/import-file/single-large-table

migtests/scripts/resumption.sh
sanyamsinghal: lets connect tomorrow once, need to understand the capabilities of the resumption script.

migtests/scripts/resumption.sh
shubham-yb: Each test will have its base config which will control the test. If you wish, you can enhance that test using any additional config file present in the `resumption/additional_configs` directory. Its like an add-on to the base config.

makalaaneesh:
@priyanshi-yb @sanyamsinghal , @shubham-yb  and I were discussing this as well. The current approach is not the simplest, but I can't immediately think of an alternative that is simpler. 
 
@sanyamsinghal , creating separate test folders is an option, but not a great one, because that means duplicating a lot of code, and the if/when the combinations of flags we want to test increase, that would lead to a lot of duplicated test folders. Furthermore, In this resumption case, for each test, the setup(how many tables, how much data), and validation (row counts) remains constant, no matter what flags we use. So it naturally points towards having a single test folder with all the details. 

If you guys have a better suggestion, let's definitely discuss! Otherwise, I'd recommend we not spend too much time overthinking this and move forward.

sanyamsinghal:
> creating separate test folders is an option, but not a great one, because that means duplicating a lot of code, and the if/when the combinations of flags we want to test increase, that would lead to a lot of duplicated test folders

@makalaaneesh @shubham-yb
Few thoughts on this:

1. **Tests should be independent:**
Right now, each normal/non-resumption test verifies a specific scenario, even if there’s some repetition (like SQL statements). That’s intentional and helps isolate issues of one from other.

2. **Don’t over-engineer test design:**
Principles like “extension over modification” are important for application code but not so much for tests. Tests aren’t usually changed frequently; they’re expanded as needed. 

3. **Share test code only where it makes sense:**
If there’s common logic (like data-generation scripts), isolate that clearly in a utility or helper function. That’s fine. But avoid tightly coupling tests together, because changes in shared logic can easily break multiple tests at once.

4. **Redundancy in test code isn’t necessarily bad:**
If one test covers multiple scenarios (e.g., batch size, parallelism), and another covers (batch size, parallelism + X parameter) that should be fine to replicate same setup for that. What we really need here are clearer utility scripts or helper functions to simplify common setup.


Just to clarify, are we assuming there will only ever be one resumption test? Tomorrow we might have multiple resumption tests, each focusing on different scenarios or parameter combinations. We should keep flexibility for that.


It might seem like over discussing, but its important, overtime our test suite might grow to huge, and some practices at this initial level are required to ensure maintenance

makalaaneesh:
@sanyamsinghal feels like there may be a disconnect, because points above are valid, but I don't necessarily think they apply to our specific case. 🤔 
Can you set up some time to discuss the problems and also the alternative approach that you are proposing?

makalaaneesh (approved):
High-level LGTM 👍 . Also wait for @sanyamsinghal for sign off

migtests/scripts/resumption.py
```diff
@@ -204,24 +230,39 @@ def run_command(command, allow_interruption=False, interrupt_after=None):
...
+        base_command (list): The base command to execute.
     """
     for attempt in range(1, max_restarts + 1):
         print(f"\n--- Attempt {attempt} of {max_restarts} ---")
 
+        # Clone base command
+        command = base_command.copy()
+
+        # Inject varying flags on each retry
+        command = inject_varying_flags(command)
```
sanyamsinghal, line 264: nit: `inject_varying_flags` -> `inject_varying_flags_values`

migtests/scripts/resumption.sh
```diff
@@ -2,14 +2,17 @@
...
+    echo "Usage: $0 TEST_NAME [env.sh] [config_file]"
+    exit 1
 fi
 
 set -x
 
+if [ $# -eq 2 ] && [[ "$2" == *.yaml ]]; then
+    set -- "$1" "" "$2"
+fi
+
```
sanyamsinghal, line 15: please add a comment.

sanyamsinghal (approved):
LGTM, good job.
Thanks for iterating on feedback and making it simpler for everyone.

PR #2484, opened 2025-04-11
Added code to integrate config file into cobra commands in config.go

ShivanshGahlot:

This PR introduces support for reading configuration values from a config file (`.yaml`) in the Voyager CLI. This allows users to define commonly used parameters in a file, simplifying repeated CLI usage.

- Added support for loading a config file (`--config` flag or defaults to `~/yb-voyager-config.yaml`)
- Config file can also be passed using the env variable: _YB_VOYAGER_CONFIG_FILE_. Precedence of these three ways is _CLI > ENV > Default_
- Introduced `validateConfigFile()` to ensure only supported keys and sections are allowed
- Implemented conflict detection across mutually exclusive sections (e.g., `export-data`, `export-data-from-source`)
- Added `bindCobraFlagsToViper()` to dynamically bind config values to Cobra command flags
- Implemented alias section support via `setToAliasPrefixIfSet()`
- CLI flags take precedence over config file values

Issue: https://yugabyte.atlassian.net/browse/DB-16087?atlOrigin=eyJpIjoiZjE4YTg1MmMxODFmNDY3Mjk0OTZkZWFmYWNmOTMzMjUiLCJwIjoiaiJ9

##### Notes

- Config structure follows a section-based hierarchy (e.g., `export-data.table-list`, `import-data.target-db-name`)
- Section/flag validation is enforced to prevent misuse
- No support for environment variable overrides in this version
- Global keys and per-command flags are validated using `allowedGlobalConfigKeys` and `allowedConfigSections` maps

##### Usage

You can pass a config file using the `--config` flag:

```sh
yb-voyager export data --config config.yaml
```

##### What is not implemented yet?
- Support for commands assess-migration-bulk
- Support for any kind of env variables
- Post snapshot phase of the import schema will have to be handled by the user by manually passing --post-snapshot-import and --refresh-mview (Not tested yet)
- Testing framework

Yes the user the will now be able to pass a config file using the new flag and not pass any other flags(which are allowed in the config file) for a command.

Manually tested. Will be adding integration tests in a future PR.

No

### DIFF
--- a/yb-voyager/cmd/config.go
+++ b/yb-voyager/cmd/config.go
@@ -179,7 +179,7 @@ func initConfig(cmd *cobra.Command) ([]ConfigFlagOverride, error) {
 
 	// If a config file is found, read it in.
 	if err := v.ReadInConfig(); err == nil {
-		fmt.Println("Using config file:", v.ConfigFileUsed())
+		fmt.Println("Using config file:", color.BlueString(v.ConfigFileUsed()))
 	} else {
 		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
 			return nil, err
--- a/yb-voyager/cmd/root.go
+++ b/yb-voyager/cmd/root.go
@@ -324,6 +324,8 @@ func validateExportDirFlag() {
 		}
 		exportDir = filepath.Clean(exportDir)
 	}
+
+	fmt.Println("Using export-dir: ", color.BlueString(exportDir))
 }
 
 func GetCommandID(c *cobra.Command) string {


yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,467 @@
...
+func initConfig(cmd *cobra.Command) error {
+	v := viper.New()
+
+	if cfgFile != "" {
+		// Use config file from the flag.
+		v.SetConfigFile(cfgFile)
+	} else {
+		// Find home directory.
+		home, err := os.UserHomeDir()
+		cobra.CheckErr(err)
```
makalaaneesh, line 229: what's this? what does this function do? why not just return the error?

yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,467 @@
...
+
+	This setup ensures CLI > ENV > Config precedence and allows environment variables
+	prefixed with YB_VOYAGER_ to override config values.
+*/
+func initConfig(cmd *cobra.Command) error {
+	v := viper.New()
+
+	if cfgFile != "" {
+		// Use config file from the flag.
+		v.SetConfigFile(cfgFile)
```
makalaaneesh, line 225: Can we also provide a way to pass in the config file via an env var? 
for example: 
export YB_VOYAGER_CONFIG_FILE = 'abc.yaml'
yb-voyager export schema

Basically if you set the env var, you don't have to explicitly pass it in every voyager command via a flag.

yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,467 @@
...
+	} else {
+		// Find home directory.
+		home, err := os.UserHomeDir()
+		cobra.CheckErr(err)
+
+		// Search config in home directory with name ".yb-voyager" (without extension).
+		v.AddConfigPath(home)
+		v.SetConfigType("yaml")
+		v.SetConfigName(".yb-voyager")
```
makalaaneesh, line 234: why the "." at the beginning? It is usually to make the file hidden by default. I don't think we want that. 

How about these options for the default: 
- /HOME/yb-voyager-config.yaml
- /HOME/yb-voyager/config.yaml

yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+
+	// If a config file is found, read it in.
+	if err := v.ReadInConfig(); err == nil {
+		fmt.Println("Using config file:", v.ConfigFileUsed())
+	} else {
+		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
```
makalaaneesh, line 241: Isn't there a way to check if viper found a config file, before calling `ReadInConfig`. If possible, try calling that first; would be cleaner, that's all..

yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,467 @@
...
+func validateConfigFile(v *viper.Viper) error {
+	var (
+		invalidGlobalKeys   []string
+		invalidSectionKeys  = make(map[string][]string)
+		invalidSections     []string
+		conflictingSections = [][]string{}
+		presentSections     = make(map[string]bool)
+	)
+
+	fmt.Println("Validating config file...")
```
makalaaneesh, line 286: not really required to print. we can log instead.

yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,467 @@
...
+	for _, key := range v.AllKeys() {
+		parts := strings.Split(key, ".")
+		if len(parts) == 1 {
+			// Check global level keys
+			if !allowedGlobalConfigKeys[key] {
+				invalidGlobalKeys = append(invalidGlobalKeys, key)
+			}
+		} else {
+			// Validate section-based keys
+			section, nestedKey := parts[0], parts[1]
```
makalaaneesh, line 296: Maybe not immediately pressing to incorporate, just a note for future: 

Today, we have only one level of nesting so this is okay. We store `allowedConfigSections` and validate the sections that are allowed, and the keys under that as well. 
But what if we have something like:
```
a: 
   b: 
      c: 
```

How do we validate `c` ? 

A more flexible way would be to just store all allowed viper keys in a single list: 
"a.b", "a.c", "a.d", "b.c", "b.c.d", etc. If you can't find the key, just throw an error saying unknown configuration parameter, and point the user to a sample config file.

yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,467 @@
...
+			fmt.Printf("%s %v\n", color.RedString("Invalid sections:"), invalidSections)
+		}
+		if len(conflictingSections) > 0 {
+			for _, conflict := range conflictingSections {
+				fmt.Printf("%s %v\n", color.RedString("Only one of the following sections can be used:"), conflict)
+			}
+		}
+
+		// Return a general error message
+		return fmt.Errorf("found invalid configurations in config file: %s", v.ConfigFileUsed())
```
makalaaneesh, line 342: we should point the user to a sample config file in documentation here. add a TODO.

yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,467 @@
...
+
+	It performs the following steps:
+	 1. Creates a new Viper instance to isolate config handling for the command.
+	 2. Loads the config file if explicitly provided via --config, or defaults to ~/.yb-voyager.yaml.
+	 3. Reads and validates the configuration file for allowed global keys, sections, and section keys.
+	 4. Binds Viper config values to Cobra flags, giving priority to command-line flags over config values.
+	 5. Returns an error if config file reading, validation, or flag binding fails.
+
+	This setup ensures CLI > ENV > Config precedence and allows environment variables
+	prefixed with YB_VOYAGER_ to override config values.
```
makalaaneesh, line 218: don't see the automaticEnv configuration here. are we planning to do that later? if so, remove the comment for now.

yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,467 @@
...
+	// Replace spaces with hyphens
+	configKeyPrefix := strings.ReplaceAll(subCmdPath, " ", "-")
+	configKeyPrefix = setToAliasPrefixIfSet(configKeyPrefix, v)
+
+	cmd.Flags().VisitAll(func(f *pflag.Flag) {
+		if bindErr != nil {
+			return // Stop further execution if an error is already encountered
+		}
+
+		if !f.Changed {
```
makalaaneesh, line 382: nit: use early return pattern. 
`If f.Changed: return`

yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,467 @@
...
+			// Check for <command_path>.<flagname>
+			if v.IsSet(configKeyPrefix + "." + f.Name) {
+				val := v.GetString(configKeyPrefix + "." + f.Name)
+				err := cmd.Flags().Set(f.Name, fmt.Sprintf("%v", val))
```
makalaaneesh, line 386: is the `Sprintf` necessary? `val` is already a string.

yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,467 @@
...
+				}
+			} else if strings.HasPrefix(f.Name, "target-") && v.IsSet("target."+strings.TrimPrefix(f.Name, "target-")) {
+				// Handle target db type flags
+				val := v.GetString("target." + strings.TrimPrefix(f.Name, "target-"))
+				err := cmd.Flags().Set(f.Name, fmt.Sprintf("%v", val))
+				if err != nil {
+					bindErr = err
+					return
+				}
+				// fmt.Printf("binding %s from viper to cmd flag: %s=%s\n", "target."+strings.TrimPrefix(f.Name, "target-"), f.Name, val)
```
makalaaneesh, line 424: add sufficient logs for every case. 
if we are ignoring a flag, because it was set in CLI
if we found it in global, if we found in cmd.key/ source.key/target-key.

yb-voyager/cmd/root.go
```diff
@@ -63,6 +63,13 @@ var rootCmd = &cobra.Command{
 Refer to docs (https://docs.yugabyte.com/preview/migrate/) for more details like setting up source/target, migration workflow etc.`,
 
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		// Initialize the config file
+		err := initConfig(cmd)
+		if err != nil {
+			// not using utils.ErrExit as logging is not initialized yet
+			fmt.Printf("ERROR: Failed to initialize config: %v\n", err)
+			atexit.Exit(0)
```
makalaaneesh, line 71: we shouldn't exit with 0. 0 is indicative of success. atexit.Exit(1) should be okay.

yb-voyager/cmd/root.go
```diff
@@ -283,37 +288,16 @@ func registerCommonGlobalFlags(cmd *cobra.Command) {
 
 	BoolVar(cmd.Flags(), &callhome.SendDiagnostics, "send-diagnostics", true,
 		"enable or disable the 'send-diagnostics' feature that sends analytics data to YugabyteDB.(default true)")
+
+	cmd.PersistentFlags().StringVarP(&cfgFile, "config-file", "c", "",
```
makalaaneesh, line 292: do you want to hide this now? once we have sufficient tests for each command, we can enable it one by one.

makalaaneesh (commented):
Thanks for the extensive code documentation, was very helpful!

yb-voyager/cmd/config.go
makalaaneesh: that makes sense!

yb-voyager/cmd/config.go
```diff
@@ -375,57 +384,55 @@ func bindCobraFlagsToViper(cmd *cobra.Command, v *viper.Viper) error {
...
+			}
+		} else if v.IsSet(f.Name) {
+			// Bind the global flag from viper to cmd
+			val := v.GetString(f.Name)
+			err := cmd.Flags().Set(f.Name, val)
+			if err != nil {
+				bindErr = err
+				return
+			}
+		} else if strings.HasPrefix(f.Name, "source-") && v.IsSet("source."+strings.TrimPrefix(f.Name, "source-")) {
```
makalaaneesh, line 408: let's use some constants like 
`SOURCE_DB_CONN_FLAG_PREFIX` , `SOURCE_DB_CONFIG_KEY_PREFIX` ,etc for `source-`, `source.`, `target-`, `target.`, `oracle-`, etc... 

Just helps to understand what the hyphen and dot are.

yb-voyager/cmd/config.go
makalaaneesh: Ah i see. 
Let's discuss this in standup to see if we can come up with something.

makalaaneesh: As discussed, pls collect all the config keys, and log them after logging is initialized.

yb-voyager/cmd/config.go
```diff
@@ -0,0 +1,474 @@
...
+)
+
+var allowedGlobalConfigKeys = map[string]bool{
+	"export-dir":            true,
+	"log-level":             true,
+	"send-diagnostics":      true,
+	"run-guardrails-checks": true,
+}
+
+var allowedSourceConfigKeys = map[string]bool{
```
makalaaneesh, line 22: nit: `mapset.NewThreadUnsafeSet` might be useful instead of creating a map

yb-voyager/cmd/config.go
ShivanshGahlot: Done

yb-voyager/cmd/config.go
ShivanshGahlot: Done

yb-voyager/cmd/config.go
ShivanshGahlot: Done

yb-voyager/cmd/config.go
```diff
@@ -397,6 +338,11 @@ func bindCobraFlagsToViper(cmd *cobra.Command, v *viper.Viper) error {
 				bindErr = err
 				return
 			}
+			overrides = append(overrides, ConfigFlagOverride{
+				FlagName:  f.Name,
+				ConfigKey: configKeyPrefix + "." + f.Name,
```
makalaaneesh, line 343: nit: use variable for configKey because it is also used above in `v.GetString` (for all cases)

makalaaneesh (approved):
LGTM with a nit! 
Pls also check why test cases seem to be failing.

PR #2490, opened 2025-04-11
[DB-16201] [Prerequisite] Non Txn Path: Fetching Primary Key Columns as part of ImportBatchArgsProto struct

sanyamsinghal:
https://yugabyte.atlassian.net/browse/DB-16201

- Need the list of Primary Key columns for adding in the `ON CONFLICT(target_list)` clause in the statement for example
```
INSERT ON table (col1, col2, col3...) VALUES (v1, v2, v3...)
    ON CONFLICT (pk1, pk2) DO NOTHING;
```

Added Unit tests

### DIFF
--- a/yb-voyager/cmd/importDataFileTaskImporter.go
+++ b/yb-voyager/cmd/importDataFileTaskImporter.go
@@ -241,11 +241,17 @@ func createImportDataTableMetrics(tableName string, countLiveRows int64, countTo
 }
 
 func getImportBatchArgsProto(tableNameTup sqlname.NameTuple, filePath string) *tgtdb.ImportBatchArgs {
-	columns, _ := TableToColumnNames.Get(tableNameTup)
-	columns, err := tdb.QuoteAttributeNames(tableNameTup, columns)
+	unquotedColumns, _ := TableToColumnNames.Get(tableNameTup)
+	columns, err := tdb.QuoteAttributeNames(tableNameTup, unquotedColumns)
 	if err != nil {
 		utils.ErrExit("if required quote column names: %s", err)
 	}
+
+	pkColumns, err := tdb.FilterPrimaryKeyColumns(tableNameTup, unquotedColumns)
+	if err != nil {
+		utils.ErrExit("getting primary key columns for table %s: %s", tableNameTup.ForMinOutput(), err)
+	}
+
 	// If `columns` is unset at this point, no attribute list is passed in the COPY command.
 	fileFormat := dataFileDescriptor.FileFormat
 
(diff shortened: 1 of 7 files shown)


yb-voyager/cmd/importDataFileTaskImporter.go
```diff
@@ -256,14 +262,15 @@ func getImportBatchArgsProto(tableNameTup sqlname.NameTuple, filePath string) *t
...
-		Columns:      columns,
-		FileFormat:   fileFormat,
-		Delimiter:    dataFileDescriptor.Delimiter,
-		HasHeader:    dataFileDescriptor.HasHeader && fileFormat == datafile.CSV,
-		QuoteChar:    dataFileDescriptor.QuoteChar,
-		EscapeChar:   dataFileDescriptor.EscapeChar,
-		NullString:   dataFileDescriptor.NullString,
+		TableNameTup:      tableNameTup,
+		Columns:           columns,
+		PrimaryKeyColumns: pkColumns,
```
makalaaneesh, line 267: Do we really need this as part of importBatchArgsProto? 
Can we abstract this inside yugabytedb.importBatch instead? 

Reason: We don't need this for any other target database, so adding this to importBatchArgs + `FilterPrimaryKeyColumns` to target db interface does not seem to be required?

yb-voyager/cmd/importDataFileTaskImporter.go
```diff
@@ -241,11 +241,17 @@ func createImportDataTableMetrics(tableName string, countLiveRows int64, countTo
...
 func getImportBatchArgsProto(tableNameTup sqlname.NameTuple, filePath string) *tgtdb.ImportBatchArgs {
-	columns, _ := TableToColumnNames.Get(tableNameTup)
-	columns, err := tdb.QuoteAttributeNames(tableNameTup, columns)
+	unquotedColumns, _ := TableToColumnNames.Get(tableNameTup)
+	columns, err := tdb.QuoteAttributeNames(tableNameTup, unquotedColumns)
 	if err != nil {
 		utils.ErrExit("if required quote column names: %s", err)
 	}
+
+	pkColumns, err := tdb.FilterPrimaryKeyColumns(tableNameTup, unquotedColumns)
```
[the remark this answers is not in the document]
sanyamsinghal: I think there are cases in export data where we don't export data for unsupported columns and skip those.
So just fetching the primary key columns from target might have extra columns also.
cc @priyanshi-yb

yb-voyager/cmd/importDataFileTaskImporter.go
sanyamsinghal: I thought it to place it parallel to columns field we already have.

I don't think strongly against using target_db_interface, given these are sort of standard and we already have done this in other cases, just for one target.

yb-voyager/cmd/importDataFileTaskImporter.go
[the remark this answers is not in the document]
priyanshi-yb: The unsupported columns are exported with NULL values, so I don't think it will be a problem if we pick the target columns.

yb-voyager/cmd/importDataFileTaskImporter.go
makalaaneesh: Wrt the method itself in isolation, there's no problem adding `FilterPrimaryKeyColumns` to target db interface. It fits. 
My concern is just that the target-db-interface is growing rather large in size, against the general "best practice" of keeping interfaces small, so I'm being a bit more mindful about whether we should add it to the interface or not. 

In this specific use-case, I find it more appropriate/clean for `targetYugabyteDB.importBatch` to have all the complexity of getting primary key columns, rather than expose a public method `FilterPrimaryKeyColumns` on `targetYugabyteDB` for which no other db type has an implementation. (Only one type implementing an interface method is also usually a sign that the interface method may not be useful.). Getting PK columns is needed in  a special case of retry of a non-txn COPY , ONLY applicable for YugabyteDB, so we should prefer to keep that complexity within targetYugabyteDB is what I think. 

Also, consider the point we discussed in standup: We should prefer getting the primary key columns from the target db directly rather than filtering the list of columns that we exported from the sourceDB.

makalaaneesh: Also, ON CONFLICT ON CONSTRAINT seems to be possible! https://stackoverflow.com/a/55040890
I think that would be both simpler and safer.

sanyamsinghal: yes, i will scrap this PR; lets go with fetching the constraint name and using that.

sanyamsinghal: But we need list of non-PK columns to write the update instruction in the statement
```
Sample Statement:
			INSERT INTO users (id, name, email) VALUES ($1, $2, $3)
				ON CONFLICT (id) DO UPDATE
				SET
					name = EXCLUDED.name,
					email = EXCLUDED.email
```
we can use constraint name for ON CONFLICT(%s) but we will still need the finer details

sanyamsinghal:
closing as its already covered

PR #2498, opened 2025-04-14
Implement COPY Fast Path in import data with recover/resume feature

sanyamsinghal:
- This PR implement non transactional import of batches via COPY command.
Taken care of Failure recovery to resume from same point remains as it is.

- For implementation details and cases handled, please refer Design Doc  - [link](https://docs.google.com/document/d/1Z2YfOLZMDLuYAxJKod9di7bHevFZHQ_SwMVtQ--nYDc/edit?usp=sharing)

- It will be best reviewed commit by commit

- Right not implementation doesn't restrict changing of action. For eg: from `ERROR` --> `UPDATE`/`IGNORE`. Tested changing it and worked. But still not sure when allow user to play/interchange between `UPDATE`/`IGNORE` is safe or not.

Two new boolean flags are added `--enable-false-path` and `-on-primary-key-conflict=ERROR|IGNORE|UPDATE`

TODO


yb-voyager/cmd/importDataState.go
```diff
@@ -796,7 +806,25 @@ type Batch struct {
...
 func (batch *Batch) Open() (*os.File, error) {
-	return os.Open(batch.FilePath)
+	file, err := os.Open(batch.FilePath)
+	if err != nil {
+		return nil, err
+	}
+	return file, err
+}
+
+func (batch *Batch) OpenDataFile() (datafile.DataFile, error) {
```
priyanshi-yb, line 816: this is for reading the batch file right? 
Rename function to OpenBatchAsDataFile()

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -501,13 +572,124 @@ func (yb *TargetYugabyteDB) importBatch(conn *pgx.Conn, batch Batch, args *Impor
...
+			utils.PrintAndLog("[debug] [table=%s | file=%s] empty line\n", batch.GetTableName().ForUserQuery(), filepath.Base(batch.GetFilePath()))
+			continue
+		}
+		utils.PrintAndLog("[debug] [table=%s | file=%s] line: %s\n", batch.GetTableName().ForUserQuery(), filepath.Base(batch.GetFilePath()), line)
+
+		fields := strings.Split(line, args.Delimiter)
+		stmtArgs := make([]interface{}, len(fields))
+		for i, val := range fields {
+			stmtArgs[i] = val
+		}
```
priyanshi-yb, line 652: Just realised one important thing:
Regarding the value converter, so for the CSV/Snapshot files, we do the value conversion without formatting the converted value, and for the streaming IUD events, we have a separate flag (formatIfRequired) for converting those in the respective format for Insert/Update/Delete stmts e.g.,String types are okay normal as is in CSV file but with inserts we need to have single quotes around them to make insert work.

So in recovery while using INSERT stmts for this we need the value conversion with. `formatIfRequired` flag.

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -125,8 +144,79 @@ func (args *ImportBatchArgs) GetPGCopyStatement() string {
...
+	2. Optimise the prepare statment to be cached and reused in connPool (same as for streaming events)
+*/
 
+/*
+GetInsertStatementBasedOn returns the INSERT prepared statement based on the requested conflict action.
+1. ERROR   -> raise error (no ON CONFLICT clause)
+2. IGNORE  -> ON CONFLICT DO NOTHING
+3. UPDATE  -> ON CONFLICT (pk…) DO UPDATE SET non‑pk = EXCLUDED.non‑pk, …
+*/
+func (args *ImportBatchArgs) GetInsertStatementBasedOn(conflictAction string) string {
```
priyanshi-yb, line 159: `GetYBInsertStatementBasedOn`

yb-voyager/cmd/import.go
```diff
@@ -238,7 +241,15 @@ func registerImportDataToTargetFlags(cmd *cobra.Command) {
...
 If you go ahead without truncating, then yb-voyager starts ingesting the data present in the data files with upsert mode.
 Note that for the cases where a table doesn't have a primary key, this may lead to insertion of duplicate data. To avoid this, exclude the table using the --exclude-file-list or truncate those tables manually before using the start-clean flag (default false)`)
-	BoolVar(cmd.Flags(), &truncateTables, "truncate-tables", false, "Truncate tables on target YugabyteDB before importing data. Only applicable along with --start-clean true (default false)")
+	BoolVar(cmd.Flags(), &truncateTables, "truncate-tables", false,
+		"Truncate tables on target YugabyteDB before importing data. Only applicable along with --start-clean true (default false)")
+	BoolVar(cmd.Flags(), &enableFastPath, "enable-fast-path", false,
+		"Enable fast path (non-transactional COPY) to improve import performance for colocated tables. Applies only to YugabyteDB target. Not applicable when importing into source/source-replica during fallforward or fallback. (default: false)")
```
priyanshi-yb, line 247: I think we can skip this `Not applicable when importing into source/source-replica during fallforward or fallback.` as this is in `registerImportDataToTargetFlags` so this flag is only for import data/import data file

yb-voyager/cmd/importDataFileTaskImporter.go
```diff
@@ -137,26 +222,47 @@ func (fti *FileTaskImporter) importBatch(batch *Batch) {
...
-	importBatchFunc := func() {
-		// There are `poolSize` number of competing go-routines trying to invoke COPY.
-		// But the `connPool` will allow only `parallelism` number of connections to be
-		// used at a time. Thus limiting the number of concurrent COPYs to `parallelism`.
-		fti.importBatch(batch)
-		if reportProgressInBytes {
-			fti.updateProgress(batch.ByteCount)
-		} else {
-			fti.updateProgress(batch.RecordCount)
+func (fti *FileTaskImporter) importBatchViaRecoverMode(batch *Batch, action string) {
```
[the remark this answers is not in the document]
sanyamsinghal: I have a different opinion on this.
The methods in target_db_interface are quiet low level; they take direct commands and do what is asked(query/execute/ingest on the target DB)

All the details of when to do what is not something we have embedded into our target DBs.

Also the other point we discussed in standup - Marking of batch as Inprogress/Done happens outside the target db classes. And at time of recovery we want to know what was the batch file state.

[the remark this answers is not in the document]
sanyamsinghal: But i do get your point, it would have been better to have abstract out the details of the path choosen(txn/non-txn) somehow.

Let me know if you have any suggestions how we can do that right now.
Otherwise it might be a significant effort of refactoring which can be deffered for now.

yb-voyager/cmd/assessMigrationCommand_test.go
```diff
@@ -60,16 +60,16 @@ func Test_AssessMigration(t *testing.T) {
 	}
 
 	// create table and initial data in it
-	postgresContainer.ExecuteSqls(`
-	CREATE TABLE test_data (
-		id SERIAL PRIMARY KEY,
-		value TEXT
-	);`,
-		`INSERT INTO test_data (value)
+	postgresContainer.ExecuteSqls(
```
makalaaneesh, line 63: curious, what prompted this change?

yb-voyager/cmd/import.go
```diff
@@ -238,7 +241,15 @@ func registerImportDataToTargetFlags(cmd *cobra.Command) {
...
+
+	// TODO: restrict changing of flag value after import data has started
+	// TODO: Detailed description of the flag
+	cmd.Flags().StringVar(&onPrimaryKeyConflictAction, "on-primary-key-conflict", "",
```
makalaaneesh, line 251: let's hide both these flags for now? We can enable them once we have done sufficient testing.

yb-voyager/cmd/import.go
```diff
@@ -429,3 +440,28 @@ func validateTruncateTablesFlag() {
 		utils.ErrExit("Error --truncate-tables true can only be specified along with --start-clean true")
 	}
 }
+
+func validateEnableFastPathFlag() {
+	if !enableFastPath && onPrimaryKeyConflictAction != "" {
+		utils.ErrExit("Error: To use --on-primary-key-conflict, you must also enable --enable-fast-path.")
```
makalaaneesh, line 446: hmm, this doesn't seem right. let's think about this. 
on-primary-key-conflict is a behavior decision. 
enable-fast-path is a performance optimization. 

I don't see why they have to be dependent on each other.

yb-voyager/cmd/importData.go
```diff
@@ -1265,6 +1270,7 @@ func prepareTableToColumns(tasks []*ImportFileTask) {
 			log.Infof("header row split using delimiter %q: %v\n", dataFileDescriptor.Delimiter, columns)
 			df.Close()
 		}
+		fmt.Printf("putting table name %q and columns %v in tableToColumns\n", task.TableNameTup.ForKey(), columns)
```
makalaaneesh, line 1273: convert to log?

sanyamsinghal: it was for debugging, will remove it.

yb-voyager/cmd/import.go
sanyamsinghal: We discussed this right.
We dont want to assume have behaviour for retries in case of fast path.
So user should voluntarily inform us what they are ok with.

makalaaneesh: Yeah, sorry I didn't communicate this properly. Internally for us, yes they are dependent. But from user's perspective, I'm not so sure if the combination of flags are simple enough to understand. 
Haven't thought about it completely, but maybe we only need "on-primary-key-conflict"?

yb-voyager/cmd/importDataFileTaskImporter.go
```diff
@@ -256,15 +373,28 @@ func getImportBatchArgsProto(tableNameTup sqlname.NameTuple, filePath string) *t
...
 }
+
+// returns true if non transaction path can be followed for this batch
+func IsNonTransactionalPath(batch *Batch) bool {
+	if batch == nil {
+		return false
+	}
+
+	nonTxnPath := bool(enableFastPath) &&
+		importerRole == TARGET_DB_IMPORTER_ROLE &&
```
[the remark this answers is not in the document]
sanyamsinghal: not required in changed.

yb-voyager/cmd/import.go
sanyamsinghal: hid the only flag `--on-primary-key-conflict` till QA is complete.

yb-voyager/cmd/import.go
```diff
@@ -404,6 +413,7 @@ func validateBatchSizeFlag(numLinesInASplit int64) {
 		defaultBatchSize = DEFAULT_BATCH_SIZE_YUGABYTEDB
 	}
 
+	// TODO: we might want to lift this restriction for non-transactional COPY (depends on testing of --batch-size flag)
```
makalaaneesh, line 416: good point. can you create a ticket for this, we can discuss this after this PR.

yb-voyager/cmd/importDataFileTaskImporter.go
```diff
@@ -99,22 +110,58 @@ func (fti *FileTaskImporter) ProduceAndSubmitNextBatchToWorkerPool() error {
...
 	importBatchArgs := *fti.importBatchArgsProto
 	importBatchArgs.FilePath = batch.FilePath
 	importBatchArgs.RowsPerTransaction = batch.OffsetEnd - batch.OffsetStart
 
-	var rowsAffected int64
 	sleepIntervalSec := 0
 	for attempt := 0; attempt < COPY_MAX_RETRY_COUNT; attempt++ {
 		tableSchema, _ := TableNameToSchema.Get(batch.TableNameTup)
-		rowsAffected, err = tdb.ImportBatch(batch, &importBatchArgs, exportDir, tableSchema)
+		isRecoveryCandidate := (recoveryBatch || attempt > 0)
```
makalaaneesh, line 163: nice. this is pretty clear!

yb-voyager/cmd/importDataFileTaskImporter.go
```diff
@@ -246,6 +271,17 @@ func getImportBatchArgsProto(tableNameTup sqlname.NameTuple, filePath string) *t
 	if err != nil {
 		utils.ErrExit("if required quote column names: %s", err)
 	}
+
+	// TODO: Implement caching since this is called for every batch
```
makalaaneesh, line 275: we might have to do this as part of this overall task before release. catalog queries on YB are quite slow, so I suspect perf degradation because of this.

yb-voyager/cmd/importDataState.go
```diff
@@ -796,7 +806,25 @@ type Batch struct {
 }
 
 func (batch *Batch) Open() (*os.File, error) {
-	return os.Open(batch.FilePath)
+	file, err := os.Open(batch.FilePath)
```
makalaaneesh, line 809: nit: the original `return os.Open(batch.FilePath)` suffices here, right?

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1036,6 +1211,9 @@ func getYBSessionInitScript(tconf *TargetConf) []string {
 	if checkSessionVariableSupport(tconf, SET_DEFAULT_ISOLATION_LEVEL_REPEATABLE_READ) {
 		sessionVars = append(sessionVars, SET_DEFAULT_ISOLATION_LEVEL_REPEATABLE_READ)
 	}
+	if checkSessionVariableSupport(tconf, SET_YB_FAST_PATH_FOR_COLOCATED_COPY) {
```
makalaaneesh, line 1214: shouldn't we only set this if `on-primary-key-conflict = IGNORE` (i.e. we're going with non-txn COPY)

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -545,18 +700,37 @@ func (yb *TargetYugabyteDB) GetListOfTableAttributes(nt sqlname.NameTuple) ([]st
...
 
 func (yb *TargetYugabyteDB) IsNonRetryableCopyError(err error) bool {
 	NonRetryCopyErrorsYB := NonRetryCopyErrors
-	NonRetryCopyErrorsYB = append(NonRetryCopyErrorsYB, "Sending too long RPC message")
+	NonRetryCopyErrorsYB = append(NonRetryCopyErrorsYB, RPC_MSG_LIMIT_ERROR)
 	return err != nil && utils.ContainsAnySubstringFromSlice(NonRetryCopyErrorsYB, err.Error())
 }
 
+// this function is meant to be used for insert errors during recovery mode
+func (yb *TargetYugabyteDB) IsNonRetryableInsertError(err error, onPrimaryKeyConflictAction string) bool {
```
makalaaneesh, line 721: Hey, i don't see this func being used. 
General question: 
- Today, if there is a duplicate PK error, do we retry or immediately fail. Based on the NonRetryCopyErrors, it seems like we retry? (idk if "violates unique constraints" is what we get when there is a duplicate PK)
- Do we plan/want to change that with this PR?

yb-voyager/cmd/importDataFileTaskImporter.go
```diff
@@ -88,6 +89,16 @@ func (fti *FileTaskImporter) AllBatchesSubmitted() bool {
 	return fti.batchProducer.Done()
 }
 
+func (fti *FileTaskImporter) TableHasPrimaryKey() bool {
+	return len(fti.importBatchArgsProto.PrimaryKeyColumns) > 0
+}
+
+func (fti *FileTaskImporter) shouldUseNonTransactionalPath() bool {
```
makalaaneesh, line 96: dont see usage of this func. remove if not needed

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -38,8 +40,9 @@ type TargetDB interface {
 	GetNonEmptyTables(tableNames []sqlname.NameTuple) []sqlname.NameTuple
 	TruncateTables(tableNames []sqlname.NameTuple) error
 	IsNonRetryableCopyError(err error) bool
-	ImportBatch(batch Batch, args *ImportBatchArgs, exportDir string, tableSchema map[string]map[string]string) (int64, error)
+	ImportBatch(batch Batch, args *ImportBatchArgs, exportDir string, tableSchema map[string]map[string]string, isRecoveryCandidate bool) (int64, error)
```
makalaaneesh, line 43: how about just calling it `recoveryMode` ?

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -442,31 +479,34 @@ func (yb *TargetYugabyteDB) TruncateTables(tables []sqlname.NameTuple) error {
...
 	copyFn := func(conn *pgx.Conn) (bool, error) {
-		rowsAffected, err = yb.importBatch(conn, batch, args)
+		if args.IsFastPath() {
+			if !isRecoveryCandidate {
+				rowsAffected, err = yb.importBatchFast(conn, batch, args)
+			} else {
+				rowsAffected, err = yb.importBatchFastRecover(conn, batch, args)
+			}
+		} else {
+			rowsAffected, err = yb.importBatch(conn, batch, args)
```
makalaaneesh, line 501: let's add a comment here explaining that in case of normal mode, we don't really have to care about recovery mode or not; since we run copy in a transaction.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -442,31 +479,34 @@ func (yb *TargetYugabyteDB) TruncateTables(tables []sqlname.NameTuple) error {
...
+3. Fast Path Recovery Mode - Importing a batch using COPY command without transaction but conflict handling.
+*/
+func (yb *TargetYugabyteDB) ImportBatch(batch Batch, args *ImportBatchArgs,
+	exportDir string, tableSchema map[string]map[string]string, isRecoveryCandidate bool) (int64, error) {
+
 	var rowsAffected int64
 	var err error
 	copyFn := func(conn *pgx.Conn) (bool, error) {
-		rowsAffected, err = yb.importBatch(conn, batch, args)
+		if args.IsFastPath() {
```
makalaaneesh, line 494: nit: `CanUseFastPath`

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -491,21 +531,49 @@ func (yb *TargetYugabyteDB) importBatch(conn *pgx.Conn, batch Batch, args *Impor
...
 		return rowsAffected, nil
 	}
 
-	// Import the split using COPY command.
+	// 4. Import the batch using COPY command.
 	var res pgconn.CommandTag
-	copyCommand := args.GetYBCopyStatement()
+	var copyCommand string
+
+	// TODO: maybe move this check into the args struct methods and just call GetYBCopyStatement()
```
makalaaneesh, line 569: Your intuition is right. You probably shouldn't be doing this check again, since you've already done an if-else check before. Also, fastPath or not is not relevant at this layer (i.e. this function where we're interested in just executing copy command)
How about: 
`copyBatchCore` has an arg `inTxn bool` . In case of fast path, you pass false, in case of normal path, you pass true. 
`copyBatchCore` chooses which type of statement it wants based on value of `inTxn`.

yb-voyager/src/constants/constants.go
```diff
@@ -51,3 +51,10 @@ const (
 const (
 	OBFUSCATE_STRING = "XXXXX"
 )
+
+// constants for --on-primary-key-conflict flag values
+const (
+	PRIMARY_KEY_CONFLICT_ACTION_ERROR  = "ERROR"
+	PRIMARY_KEY_CONFLICT_ACTION_IGNORE = "IGNORE"
+	PRIMARY_KEY_CONFLICT_ACTION_UPDATE = "UPDATE"
```
makalaaneesh, line 59: let's not support UPDATE for now. We will have to deprecate `--enable-upsert` and then also apply the session variables for upsert mode.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -514,13 +582,100 @@ func (yb *TargetYugabyteDB) importBatch(conn *pgx.Conn, batch Batch, args *Impor
...
+		}
+
+		// 5. Execute the INSERT statement one by one
+		res, err := conn.Exec(context.Background(), insertStmt, stmtArgs...)
+		if err != nil {
+			return 0, fmt.Errorf("insert into target table: %w", err)
+		}
+
+		if res.RowsAffected() > 0 {
+			rowsAffected++
```
makalaaneesh, line 650: lets add debug logs including the statment and the args in both cases (row affected, row ignored)

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -514,13 +582,100 @@ func (yb *TargetYugabyteDB) importBatch(conn *pgx.Conn, batch Batch, args *Impor
...
+		if readLinErr == io.EOF { // handles case 2
+			log.Infof("reached end of file %s", batch.GetFilePath())
+			break
+		}
+	}
+	log.Infof("total rows inserted for batch %q: %d", batch.GetFilePath(), rowsAffected)
+
+	// 6. Update the Metadata about the batch imported
+	/*
+		Thought: In case of onPrimaryKeyConflictAction=true total sum of rowsAffected for this batch in metadata can be less than total rows in batch(due to partial ingestion and retries)
```
makalaaneesh, line 663: @priyanshi-yb pls review the behavior for import-data status/ get migration report part

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -514,13 +582,100 @@ func (yb *TargetYugabyteDB) importBatch(conn *pgx.Conn, batch Batch, args *Impor
...
+	}
+	defer df.Close()
+
+	// 3. Read the batch file line by line(row) and build INSERT statment for it based on PK conflict action
+	var rowsIgnored int64 = 0
+	rowsAffected = 0
+
+	// [TODO: Optional] Use PreparedStatement Caching on coonPool for better performance
+	insertStmt := args.GetInsertPreparedStmtForBatchImport()
+	for {
```
makalaaneesh, line 617: can we create a separate function that returns a list of rows (row being a `[]interface`)? That function is probably best written with Batch struct as receiver.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -514,13 +582,100 @@ func (yb *TargetYugabyteDB) importBatch(conn *pgx.Conn, batch Batch, args *Impor
...
+		line, _, readLinErr := df.NextLine()
```
makalaaneesh, line 618: is next line appropriate here? Some questions (i don't know the answers to these): 
Are there cases where a \n in the text field causes problems with next line?

I see in fileBatchProducer that we use `line, currentBytesRead, readLineErr = p.dataFile.NextLine()` , so I'm guessing it shouldn't be a problem.
cc: @priyanshi-yb

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -514,13 +582,100 @@ func (yb *TargetYugabyteDB) importBatch(conn *pgx.Conn, batch Batch, args *Impor
...
+		}
+	}
+	log.Infof("total rows inserted for batch %q: %d", batch.GetFilePath(), rowsAffected)
+
+	// 6. Update the Metadata about the batch imported
+	/*
+		Thought: In case of onPrimaryKeyConflictAction=true total sum of rowsAffected for this batch in metadata can be less than total rows in batch(due to partial ingestion and retries)
+		But things like import-data-status cmd output shouldn't be affected by this (since it count imported row count based on .D files)
+	*/
+	err = yb.recordEntryInDB(conn, batch, int64(rowsAffected))
```
makalaaneesh, line 666: shouldn't we mark the entire batch as completed, therefore use `rowsAfftected+rowsIgnored`?

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -514,13 +582,100 @@ func (yb *TargetYugabyteDB) importBatch(conn *pgx.Conn, batch Batch, args *Impor
...
+			rowsAffected++
+		} else {
+			rowsIgnored++
+		}
+		if readLinErr == io.EOF { // handles case 2
+			log.Infof("reached end of file %s", batch.GetFilePath())
+			break
+		}
+	}
+	log.Infof("total rows inserted for batch %q: %d", batch.GetFilePath(), rowsAffected)
```
makalaaneesh, line 659: Can we do as assertion here - rowsAffected +rowsIgnored should be equal to batch.RecordCount. 
If not, we should error out.  

Maybe I'm being a bit paranoid here, what do you think?

yb-voyager/cmd/import.go
```diff
@@ -238,7 +240,14 @@ func registerImportDataToTargetFlags(cmd *cobra.Command) {
 If any table on YugabyteDB database is non-empty, it prompts whether you want to continue the import without truncating those tables; 
 If you go ahead without truncating, then yb-voyager starts ingesting the data present in the data files with upsert mode.
 Note that for the cases where a table doesn't have a primary key, this may lead to insertion of duplicate data. To avoid this, exclude the table using the --exclude-file-list or truncate those tables manually before using the start-clean flag (default false)`)
-	BoolVar(cmd.Flags(), &truncateTables, "truncate-tables", false, "Truncate tables on target YugabyteDB before importing data. Only applicable along with --start-clean true (default false)")
+	BoolVar(cmd.Flags(), &truncateTables, "truncate-tables", false,
+		"Truncate tables on target YugabyteDB before importing data. Only applicable along with --start-clean true (default false)")
+
+	// TODO: restrict changing of flag value after import data has started
+	// TODO: Detailed description of the flag
+	cmd.Flags().StringVar(&onPrimaryKeyConflictAction, "on-primary-key-conflict", "",
```
makalaaneesh, line 248: default should be ERROR?

yb-voyager/cmd/import.go
```diff
@@ -434,3 +444,18 @@ func validateTruncateTablesFlag() {
...
 	}
 }
+
+var onPrimaryKeyConflictActions = []string{
+	constants.PRIMARY_KEY_CONFLICT_ACTION_ERROR,
+	constants.PRIMARY_KEY_CONFLICT_ACTION_IGNORE,
+	constants.PRIMARY_KEY_CONFLICT_ACTION_UPDATE,
+}
+
+func validateOnPrimaryKeyConflictFlag() {
```
makalaaneesh, line 454: I think we should prevent user from changing this during resumption. 
What if user starts out with IGNORE, there is a batch that succeeded partially. Now, if the user switches to ERROR, it will start failing?

yb-voyager/cmd/import.go
makalaaneesh, line 248: Should we call this `on-primary-key-conflict-snapshot` ? 
For live-migration cdc phase, we sort of already do ON CONFLICT DO NOTHING when doing inserts

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: yeah i just relied on NextLine() since we use it at other places to get the row

yb-voyager/cmd/importDataState.go
sanyamsinghal: yes. redundant code.

yb-voyager/cmd/import.go
sanyamsinghal: its there - https://yugabyte.atlassian.net/browse/DB-15906

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: Good catch. Better to use it when required.
Although it should not cause any issue if transaction is used(no fast path in that case)

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: Thinking of this

```
func (args *ImportBatchArgs) GetYBCopyStatement() string {
	options := args.copyOptions()
	if !args.CanUseFastPath() { 
		// fast path on DB side not enabled if ROWS_PER_TRANSACTION is set in COPY
		options = append(options, fmt.Sprintf("ROWS_PER_TRANSACTION %v", args.RowsPerTransaction))
	}

	columns := ""
	if len(args.Columns) > 0 {
		columns = fmt.Sprintf("(%s)", strings.Join(args.Columns, ", "))
	}

	return fmt.Sprintf(`COPY %s %s FROM STDIN WITH (%s)`, args.TableNameTup.ForUserQuery(), columns, strings.Join(options, ", "))
}
```

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: Hmm, code from old implementation but this is important. 
Let me think it through once.

sanyamsinghal:
Resumption Tests - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/5329/

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: done

yb-voyager/cmd/import.go
sanyamsinghal: followup PR

sanyamsinghal: done

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: yes lets not add it here. We can add it on top layer where we call import batch irrespective of path.

yb-voyager/src/constants/constants.go
sanyamsinghal: followup PR

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: done. PTAL once.

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: done

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: We can't return all the rows by storing inmemory as multiple batches will be running parallel, and that can eatup all the memory(RAM)

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -125,8 +153,79 @@ func (args *ImportBatchArgs) GetPGCopyStatement() string {
...
+
+	/*
+		Sample Statement:
+			INSERT INTO users (id, name, email) VALUES ($1, $2, $3)
+				ON CONFLICT (id) DO UPDATE
+				SET
+					name = EXCLUDED.name,
+					email = EXCLUDED.email
+	*/
+	case constants.PRIMARY_KEY_CONFLICT_ACTION_UPDATE:
```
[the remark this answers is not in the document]
sanyamsinghal: @makalaaneesh, is it okay to do this also in a followup PR?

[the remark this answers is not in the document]
sanyamsinghal: done

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -491,21 +533,66 @@ func (yb *TargetYugabyteDB) importBatch(conn *pgx.Conn, batch Batch, args *Impor
...
+	rowsAffected, err := yb.copyBatchCore(conn, batch, args)
+
+	/*
+		Check if the error is violates unique constraint then use the fast path recovery to ingest this batch
+		Case: If the table already has the data(before import started) which is being imported
+
+		Lets say the importBatchFastRecover fails with some trasient DB error. In that case,
+		caller(fileTaskImporter.importBatch() function) takes care of retrying with importBatchFastRecover
+	*/
+	if err != nil && strings.Contains(err.Error(), VIOLATES_UNIQUE_CONSTRAINT_ERROR) {
```
makalaaneesh, line 554: The error that we are checking seems very generic. We should only retry if it's a PK violation, not any unique constraint error. Can we limit it only to PK violations?

yb-voyager/cmd/import.go
```diff
@@ -238,7 +240,17 @@ func registerImportDataToTargetFlags(cmd *cobra.Command) {
...
+	BoolVar(cmd.Flags(), &truncateTables, "truncate-tables", false,
+		"Truncate tables on target YugabyteDB before importing data. Only applicable along with --start-clean true (default false)")
+
+	// TODO: restrict changing of flag value after import data has started
+	// TODO: Detailed description of the flag
+	cmd.Flags().StringVar(&tconf.OnPrimaryKeyConflictAction, "on-primary-key-conflict", "ERROR",
+		`Action to take on primary key conflict during data import.
+Supported values:
+  ERROR(default): Imports data in this mode fails if any primary key conflict is encountered, assuming such conflicts are unexpected.
+  IGNORE		: Skips rows that already exist with the same primary key and enables fast-path import, which provides significantly better performance for large data loads.`)
```
makalaaneesh, line 252: - let's not use the word "significant".
- let's also qualify the exact case: "better performance for colocated tables"

yb-voyager/src/tgtdb/yugabytedb.go
makalaaneesh: let's address this in the follow-up PR

makalaaneesh (approved):
LGTM! 
Pls ensure all test pipelines pass before merging
Yet to review test cases, but don't want to block merge because of that. 
Pls make a list of open items for the follow-up PR. 
Pls have @priyanshi-yb also take a look once.

yb-voyager/cmd/import.go
sanyamsinghal: done

PR #2504, opened 2025-04-15
Updated the pg grant permission script to prompt the user for permission before transferring ownership of tables to replication group

ShivanshGahlot:
- Added a prompt to ask the user whether they want to proceed with the transfer of ownership of tables to the replication group
- This is done just before the transfer happens in the script
- The user can only provide 'yes' or 'no' as input. They cannot provide 'y' or 'n' since the psql \if does not support string comparision. 'yes' however is considered as 1 and 'no' as 0.

Issue: https://github.com/yugabyte/yb-voyager/issues/2485

Example:
```
This script will transfer ownership of all tables in the specified schemas to the specified replication group as required by debezium. User ybvoyager and the original owner of the tables will be added to the replication group.
Proceed with the transfer of ownership? (yes/no): no
Aborting the script.
```

In the pg permission grant script the user will now be asked whether they are okay with transfer of ownership of tables to replication group

Manually

### DIFF
--- a/guardrails-scripts/yb-voyager-pg-grant-migration-permissions.sql
+++ b/guardrails-scripts/yb-voyager-pg-grant-migration-permissions.sql
@@ -156,7 +156,8 @@ GRANT pg_read_all_stats to :voyager_user;
     -- Prompt the user to let them know that ownership of the tables will be transferred to the replication group and proceed only if they confirm
     \echo ''
     \echo 'This script will transfer ownership of all tables in the specified schemas to the specified replication group as required by debezium. User ybvoyager and the original owner of the tables will be added to the replication group.'
-    \prompt 'Proceed? (yes/no): ' proceed
+    -- Only 'yes' or 'no' are valid inputs. 'y' and 'n' are not valid inputs.
+    \prompt 'Proceed with the transfer of ownership? (yes/no): ' proceed
 
     \if :proceed
         \echo 'Proceeding with the transfer of ownership...'
--- a/guardrails-scripts/yb-voyager-pg-grant-migration-permissions.sql
+++ b/guardrails-scripts/yb-voyager-pg-grant-migration-permissions.sql
@@ -153,6 +153,18 @@ GRANT pg_read_all_stats to :voyager_user;
     \echo '--- Creating Replication Group ---'
     CREATE ROLE :replication_group;
 
+    -- Prompt the user to let them know that ownership of the tables will be transferred to the replication group and proceed only if they confirm
+    \echo ''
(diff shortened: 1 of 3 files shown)


guardrails-scripts/yb-voyager-pg-grant-migration-permissions.sql
```diff
@@ -153,6 +153,19 @@ GRANT pg_read_all_stats to :voyager_user;
     \echo '--- Creating Replication Group ---'
     CREATE ROLE :replication_group;
 
+    -- Prompt the user to let them know that ownership of the tables will be transferred to the replication group and proceed only if they confirm
+    \echo ''
+    \echo 'This script will transfer ownership of all tables in the specified schemas to the specified replication group. The migration user and the original owner of the tables will be added to the replication group.'
+    -- Only 'yes' or 'no' are valid inputs. 'y' and 'n' are not valid inputs.
```
makalaaneesh, line 159: nit: y/n might be simpler, so maybe choose that instead of yes/no

guardrails-scripts/yb-voyager-pg-grant-migration-permissions.sql
```diff
@@ -153,6 +153,19 @@ GRANT pg_read_all_stats to :voyager_user;
     \echo '--- Creating Replication Group ---'
     CREATE ROLE :replication_group;
 
+    -- Prompt the user to let them know that ownership of the tables will be transferred to the replication group and proceed only if they confirm
+    \echo ''
+    \echo 'This script will transfer ownership of all tables in the specified schemas to the specified replication group. The migration user and the original owner of the tables will be added to the replication group.'
```
makalaaneesh, line 158: nit: You don't need to mention 'this script'. How about something like :
`Transferring ownership of all tables in the specified schemas ...`

guardrails-scripts/yb-voyager-pg-grant-migration-permissions.sql
ShivanshGahlot: Hey so the psql script does not support string comparision in the \if statement. It only supports a boolean variables over there i.e. 0 /1 or true/false or yes/no - all of these are considered boolean values. So we can only use yes/no

guardrails-scripts/yb-voyager-pg-grant-migration-permissions.sql
ShivanshGahlot: Done

priyanshi-yb (approved):
LGTM!

PR #2492, opened 2025-04-14
Report timestamp or date indexes for the hotspots and range-sharding perf optimizations

priyanshi-yb:
### Describe the changes in this pull request
1. Adding a new category in assessment issues as `Performance Optimizations` for all the recommendation in the source schema/application required for better performance on YB.
2. Report the indexes on the Timestamp/date data types for the issues -
     - Hotspots with range-sharded indexes: In case these columns are used in range queries and the datatype is increasing in nature e.g. Timestamp values like now() etc. can cause the hotspots problem during reads/writes after migration on YugabyteDB. 
     - Use range-sharding for such indexes: Generally these columns are used in range-queries but by default currently YB creates Hash-sharded index so it is recommended to explicitly make the range-sharded index.
   fixes: #2480 
 3. Adding a generic Note in the report in case the indexes are present in the source that if there are columns of the indexes used in range queries, use range-sharding index.
 fixes: #2503 
 
### Describe if there are any user-facing changes

More issues for Performance optimizations are now reported for the certain cases.

YugabyteD UI
Screenshot 2025-04-15 at 7 25 36 PM

Assessment report-
Screenshot 2025-04-16 at 5 26 52 PM

### How was this pull request tested?

Added tests and updated existing complex schema tests having the indexes on timestamp/date columns.

### Does your PR have changes in callhome/yugabyted payloads? If so, is the payload version incremented?
NO

### Does your PR have changes that can cause upgrade issues? 
| Component        | Breaking changes? |
| :----------------------------------------------: | :-----------: |
| MetaDB                                                    | No |
| Name registry json                                        | No |
| Data File Descriptor Json                                 | No |
| Export Snapshot Status Json                               | No |
| Import Data State                                         | No |
| Export Status Json                                        | No |
| Data .sql files of tables                                 | No |
| Export and import data queue                              | No |
| Schema Dump                                               | No |
| AssessmentDB                                              | No |
| Sizing DB                                                 | No |
| Migration Assessment Report Json                          | No |
| Callhome Json                                             | No |
| YugabyteD Tables                                          | No |
| TargetDB Metadata Tables                                  | No |

PR #2516, opened 2025-04-17
Report the 'Use range-sharding' issue only in case the hotspot problem and ASC/DESC clauses is not present

priyanshi-yb:
On top of the previous PR #2492 - now only reporting the 'Use range-sharding indexes on timestamp/date columns' in case the Index doesn't have the Hotspot problem and the `ASC/DESC` clause is not present.

Yes, now for the Hotspot problem, only one issue will be reported and use range sharding is reporting only if ASC/DESC clause is not present in the INDEX

Integrated and unit tests


yb-voyager/src/query/queryissue/detectors_ddl.go
```diff
@@ -685,18 +705,29 @@ func (d *IndexIssueDetector) DetectIssues(obj queryparser.DDLObject) ([]QueryIss
...
 	default:
 		return issues, fmt.Errorf("unexpected type for the Hotspots on range indexes with timestamp/date types")
 	}
 	return issues, nil
 }
 
+func reportUseRangeShardingIndexes(typeName string, objType string, objName string) ([]QueryIssue, error) {
+	var issues []QueryIssue
+	switch typeName {
+	case "timestamp", "timestamptz":
```
makalaaneesh, line 721: lets use constants here.

yb-voyager/src/query/queryissue/detectors_ddl.go
```diff
@@ -643,11 +643,19 @@ func (d *IndexIssueDetector) DetectIssues(obj queryparser.DDLObject) ([]QueryIss
 						"",
 					))
 				} else if isHotspotType && idx == 0 {
+					//If first column is hotspot type then only report hotspot issue
 					hotspotIssues, err := reportHotspotsOnIndexes(param.ExprCastTypeName, obj.GetObjectType(), obj.GetObjectName())
 					if err != nil {
 						return nil, err
 					}
 					issues = append(issues, hotspotIssues...)
+				} else if isHotspotType && !slices.Contains(queryparser.RangeShardingClauses, param.SortByOrder) {
```
makalaaneesh, line 652: maybe for now it's okay (in the interest of time), but we seem to be combining the concerns of "which is a type which can lead to hotspot problem" and "which is a type for which we should nudge user to use range indexes". 

Currently, both of these are the same. But they might not be. 
pls add a todo.

makalaaneesh (approved):
LGTM, with a few comments.

yb-voyager/src/query/queryissue/detectors_ddl.go
priyanshi-yb: Yeah right, will add a TODO

PR #2531, opened 2025-04-22
Introduce Support for Setting Environment Variables from Config File

ShivanshGahlot:
### Overview

This PR introduces the `bindEnvVarsToViper` function, which sets up environment variables based on the values provided in the config file for a given Cobra command. This bridges the gap between Viper-based config parsing and legacy parts of the codebase that rely on `os.Getenv`.

### Behavior

- If an environment variable is **already set in the shell**, it is **respected and not overridden**.
- If it is **unset** and a corresponding value exists in the config file, the value is **exported to the process environment** using `os.Setenv`.
- If neither is set, the variable is left unset.

### Config Key Matching Logic

- Command-specific keys must be prefixed with the command path (e.g., `export-data.table-list`).
- Global config keys (e.g., `export-dir`) are allowed if they don't contain a `.`.
- Keys under nested sections like `source.`, `target.`, or `source-replica.` are also supported.

### Alias Command Support

Alias commands like `export data from source` are internally resolved to a known prefix (e.g., `export-data` or `export-data-from-source`). The config keys are then matched against this resolved prefix.

### Return Value

The function returns a list of all environment variables that were set via `os.Setenv` — useful for debugging or logging purposes.

### Notes

- These environment variables are **only available within the current Go process** and are **not visible to the parent shell** (i.e., cannot be accessed via `echo $VAR` after CLI exits).
- They are accessible within the application using `os.Getenv`.

Users can now specify the environment variables in the config file itself.

Manually

### DIFF
--- a/yb-voyager/cmd/config.go
+++ b/yb-voyager/cmd/config.go
@@ -26,6 +26,10 @@ const (
 
 var allowedGlobalConfigKeys = mapset.NewThreadUnsafeSet[string](
 	"export-dir", "log-level", "send-diagnostics", "run-guardrails-checks",
+	// environment variables keys
+	"control-plane-type", "yugabyted-db-conn-string", "java-home",
+	"local-call-home-service-host", "local-call-home-service-port",
+	"yb-tserver-port", "tns-admin",
 )
 
 var allowedSourceConfigKeys = mapset.NewThreadUnsafeSet[string](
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/config.go
```diff
@@ -182,23 +225,183 @@ func initConfig(cmd *cobra.Command) ([]ConfigFlagOverride, error) {
...
 	}
 
 	// Bind the config values to the Cobra command flags
 	overrides, err := bindCobraFlagsToViper(cmd, v)
 	if err != nil {
-		return nil, fmt.Errorf("failed to bind cobra flags to viper: %w", err)
+		return nil, nil, nil, fmt.Errorf("failed to bind cobra flags to viper: %w", err)
+	}
+
+	envVarsSet, envVarsAlreadySet, err := bindEnvVarsToViper(cmd, v)
```
priyanshi-yb, line 244: nit: envVarsSetViaConfig, envVarsAlreadyExported

priyanshi-yb (approved):
LGTM!

PR #2537, opened 2025-04-23
Created alias for command finalize-schema-post-data-import import schema --post-snapshot-data

ShivanshGahlot:
- Introduced a new alias command, `yb-voyager finalize-schema-post-data-import`, for the existing `yb-voyager import schema --post-snapshot-data`.

### Changes:
- **New Command**: 
  - Added a Go file `finalizeSchemaPostDataImport.go` that defines the new command.
  - Sets `flagPostSnapshotData = true` before invoking the `PreRun` and `Run` methods of the `import schema` command.
  - Hid the flags --post-snapshot-import and --refresh-mviews in the import schema command but did not remove them completely for backward compatibility
  
- **Test Framework Updates**:
  - Added a new function to facilitate running the `finalize_schema_post_data_import` command in tests.
  - Replaced all occurrences of `import_schema --post-snapshot-data --refresh-mviews` with `finalize-schema-post-data-import --refresh-mviews` in the test framework.

- **For Reference**:
```
yb-voyager finalize-schema-post-data-import --help
Finalize schema after data import is complete.

Usage:
  yb-voyager finalize-schema-post-data-import [flags]

Flags:
      --continue-on-error boolean       Ignore errors and continue with the import; accepted values (true, false, yes, no, 0, 1)
  -e, --export-dir string               export directory is the workspace used to keep the exported schema, data, state, and logs
  -h, --help                            help for finalize-schema-post-data-import
      --ignore-exist boolean            ignore errors if object already exists (default false); accepted values (true, false, yes, no, 0, 1)
  -l, --log-level string                log level for yb-voyager. Accepted values: (trace, debug, info, warn, error, fatal, panic) (default "info")
      --refresh-mviews boolean          Refreshes the materialised views on target during post snapshot import phase (default false); accepted values (true, false, yes, no, 0, 1)
      --run-guardrails-checks boolean   Run guardrails checks during import; accepted values (true, false, yes, no, 0, 1) (default true)
      --send-diagnostics boolean        enable or disable the 'send-diagnostics' feature that sends analytics data to YugabyteDB.(default true); accepted values (true, false, yes, no, 0, 1)
      --target-db-host string           host on which the YugabyteDB server is running (default "127.0.0.1")
      --target-db-name string           name of the database on the target YugabyteDB server on which import needs to be done
      --target-db-password string       password with which to connect to the target YugabyteDB server. Alternatively, you can also specify the password by setting the environment variable TARGET_DB_PASSWORD. If you don't provide a password via the CLI, yb-voyager will prompt you at runtime for a password. If the password contains special characters that are interpreted by the shell (for example, # and $), enclose the password in single quotes.
      --target-db-port int              port on which the YugabyteDB YSQL API is running (Default: 5433)
      --target-db-schema string         target schema name in YugabyteDB (Note: works only for source as Oracle and MySQL, in case of PostgreSQL you can ALTER schema name post import)
      --target-db-user string           username with which to connect to the target YugabyteDB server
      --target-ssl-cert string          Path of file containing target SSL Certificate
      --target-ssl-crl string           Path of file containing target SSL Root Certificate Revocation List (CRL)
      --target-ssl-key string           Path of file containing target SSL Key
      --target-ssl-mode string          specify the target SSL mode: (disable, allow, prefer, require, verify-ca, verify-full) (default "prefer")
      --target-ssl-root-cert string     Path of file containing target SSL Root Certificate
  -y, --yes                             assume answer as yes for all questions during migration (default false)
```

The users will be exposed to a new command as mentioned above.
Moreover, the `--help` flag for import schema will not show the flags  `--post-snapshot-import` and `--refresh-mviews`.

Manually and GH tests

### DIFF
--- a/migtests/scripts/live-migration-fallf-run-test.sh
+++ b/migtests/scripts/live-migration-fallf-run-test.sh
@@ -202,7 +202,7 @@ main() {
 	sleep 60
 
 	step "Import remaining schema (FK, index, and trigger) and Refreshing MViews if present."
-	post_data_import_finalize_schema --refresh-mviews true
+	finalize-schema-post-data-import --refresh-mviews true
 	
 	step "Run snapshot validations."
 	"${TEST_DIR}/validate" --live_migration 'true' --ff_enabled 'true' --fb_enabled 'false' || {
--- a/migtests/scripts/live-migration-fallf-run-test.sh
+++ b/migtests/scripts/live-migration-fallf-run-test.sh
@@ -202,7 +202,12 @@ main() {
 	sleep 60
 
 	step "Import remaining schema (FK, index, and trigger) and Refreshing MViews if present."
-	finalize_schema_post_data_import --refresh-mviews true
+	# Run either finalize_schema_post_data_import or import_schema based on a random condition until we support both the ways
+	if [ $((RANDOM % 2)) -eq 0 ]; then
+		finalize_schema_post_data_import --refresh-mviews true
+	else
+		import_schema --post-snapshot-import true --refresh-mviews true
+	fi
 	
 	step "Run snapshot validations."
 	"${TEST_DIR}/validate" --live_migration 'true' --ff_enabled 'true' --fb_enabled 'false' || {
--- a/migtests/scripts/live-migration-fallf-run-test.sh
+++ b/migtests/scripts/live-migration-fallf-run-test.sh
@@ -202,7 +202,7 @@ main() {
 	sleep 60
 
 	step "Import remaining schema (FK, index, and trigger) and Refreshing MViews if present."
-	finalize-schema-post-data-import --refresh-mviews true
+	finalize_schema_post_data_import --refresh-mviews true
 	
 	step "Run snapshot validations."
 	"${TEST_DIR}/validate" --live_migration 'true' --ff_enabled 'true' --fb_enabled 'false' || {
--- a/migtests/scripts/live-migration-fallf-run-test.sh
+++ b/migtests/scripts/live-migration-fallf-run-test.sh
@@ -202,7 +202,7 @@ main() {
 	sleep 60
 
 	step "Import remaining schema (FK, index, and trigger) and Refreshing MViews if present."
-	import_schema --post-snapshot-import true --refresh-mviews true
+	post_data_import_finalize_schema --refresh-mviews true
 	
 	step "Run snapshot validations."
 	"${TEST_DIR}/validate" --live_migration 'true' --ff_enabled 'true' --fb_enabled 'false' || {
--- a/migtests/scripts/live-migration-fallf-run-test.sh
+++ b/migtests/scripts/live-migration-fallf-run-test.sh
@@ -202,12 +202,7 @@ main() {
 	sleep 60
 
 	step "Import remaining schema (FK, index, and trigger) and Refreshing MViews if present."
-	# Run either finalize_schema_post_data_import or import_schema based on a random condition until we support both the ways
-	if [ $((RANDOM % 2)) -eq 0 ]; then
-		finalize_schema_post_data_import --refresh-mviews true
-	else
-		import_schema --post-snapshot-import true --refresh-mviews true
-	fi
+	finalize_schema_post_data_import --refresh-mviews true
 	
 	step "Run snapshot validations."
 	"${TEST_DIR}/validate" --live_migration 'true' --ff_enabled 'true' --fb_enabled 'false' || {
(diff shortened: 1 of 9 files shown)


migtests/scripts/live-migration-fallf-run-test.sh
```diff
@@ -202,7 +202,7 @@ main() {
 	sleep 60
 
 	step "Import remaining schema (FK, index, and trigger) and Refreshing MViews if present."
-	import_schema --post-snapshot-import true --refresh-mviews true
+	finalize_schema_post_data_import --refresh-mviews true
```
priyanshi-yb, line 205: Let's not change the commands in tests for all cases, we can keep the mix of both until we support both.

yb-voyager/cmd/finalizeSchemaPostDataImport.go
```diff
@@ -0,0 +1,41 @@
...
+
+func init() {
+	rootCmd.AddCommand(finalizeSchemaPostDataImportCmd)
+	registerCommonGlobalFlags(finalizeSchemaPostDataImportCmd)
+	registerCommonImportFlags(finalizeSchemaPostDataImportCmd)
+	registerTargetDBConnFlags(finalizeSchemaPostDataImportCmd)
+	BoolVar(finalizeSchemaPostDataImportCmd.Flags(), &tconf.IgnoreIfExists, "ignore-exist", false,
+		"ignore errors if object already exists (default false)")
+	BoolVar(finalizeSchemaPostDataImportCmd.Flags(), &flagRefreshMViews, "refresh-mviews", false,
+		"Refreshes the materialised views on target during post snapshot import phase (default false)")
```
priyanshi-yb, line 40: what about the `continue-on-error` flag?

yb-voyager/cmd/import.go
```diff
@@ -258,6 +258,11 @@ func registerImportSchemaFlags(cmd *cobra.Command) {
 		"Refreshes the materialised views on target during post snapshot import phase (default false)")
 	BoolVar(cmd.Flags(), &enableOrafce, "enable-orafce", true,
 		"enable Orafce extension on target(if source db type is Oracle)")
+
+	// --post-snapshot-import and --refresh-mviews flags will now be handled by the command post-data-import-finalize-schema
+	// Not removing these flags and just hiding them for backward compatibility.
+	cmd.Flags().MarkHidden("post-snapshot-import")
+	cmd.Flags().MarkHidden("refresh-mviews")
```
sanyamsinghal, line 265: Additionally: If someone still uses these, we can print a message saying these are deprecated and user should prefer using the new command.

migtests/scripts/live-migration-fallf-run-test.sh
ShivanshGahlot: Yeah I had mostly added the change everywhere to run the command once in every situation. Will do this for now:
```
if [ $((RANDOM % 2)) -eq 0 ]; then
	finalize_schema_post_data_import --refresh-mviews=true
else
	import_schema --post-snapshot-import=true --refresh-mviews=true
fi
```
This should randomly run either of the two.

yb-voyager/cmd/import.go
ShivanshGahlot: Okay I'll add something like this:
image

priyanshi-yb (approved):
LGTM!

migtests/scripts/live-migration-fallf-run-test.sh
makalaaneesh: It's probably not a big deal in this scenario but:
nit: It's generally NOT a good idea to introduce randomness into the test "scenarios". It becomes hard to reproduce if/when there is an error. Better to have separate tests that have the different scenarios.

yb-voyager/cmd/importSchema.go
```diff
@@ -59,6 +59,17 @@ var importSchemaCmd = &cobra.Command{
 	},
 
 	Run: func(cmd *cobra.Command, args []string) {
+		// If the flags --post-snapshot-import or --refresh-mviews are set and the command is import schema, then throw a warning that these flags are deprecated
+		// and instead use the finalize-schema-post-data-import command.
+		if (cmd.Flags().Changed("post-snapshot-import") || cmd.Flags().Changed("refresh-mviews")) && cmd.Name() == "schema" {
+			msg := "\nWarning: The flags --post-snapshot-import and --refresh-mviews are deprecated. " +
+				"Please use the finalize-schema-post-data-import command instead.\n" +
+				"For more details and examples, visit https://docs.yugabyte.com/preview/yugabyte-voyager/reference/schema-migration/finalize-schema-post-data-import/\n"
```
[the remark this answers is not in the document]
ShivanshGahlot: Removed it for now

migtests/scripts/live-migration-fallf-run-test.sh
ShivanshGahlot: Hmm anyways once we deprecate the command properly, or add this to the docs we can revert back to just using the new command. For now I have removed this random thing and using either command in two of the four scripts each.

PR #2549, opened 2025-04-28
Report Redundant indexes in assessment and analyze

priyanshi-yb:

Reporting the redundant index issues in the assessment and analyze schema.
The logic in the psql script is taken from here - https://github.com/yugabyte/yb-tools/blob/main/redundant_index/redundant_index_detector.sql
fixes https://github.com/yugabyte/yb-voyager/issues/2535

Yes new issues will be reported
Screenshot 2025-04-29 at 4 46 33 PM

Automation

There is an addition of a table in AssessmentDB `redundant_indexes`, but I think as we only use this assessment DB in assess-migration and the user can't resume this for the upgrade scenario, we should always be creating the table not present in Assessment DB, so it is okay.

| Schema Dump                                               |No |
| AssessmentDB                                              |  No|


yb-voyager/src/srcdb/data/gather-assessment-metadata/postgresql/redundant-indexes.psql
```diff
@@ -0,0 +1,293 @@
+
+CREATE TEMP TABLE ind_att AS
```
sanyamsinghal, line 2: can temp tables have same name as one of the existing table?

If not you might want to handle the name for tables created here.

priyanshi-yb: It is allowed to create temp tables with the same name as an existing table.
```
postgres=# CREATE TABLE temp_table(id int);
CREATE TABLE
postgres=# CREATE TEMP TABLE temp_table(id int);
CREATE TABLE
postgres=# DROP TABLE temp_table;
DROP TABLE
postgres=# \d
postgres=# \dt temp_table 
           List of relations
 Schema |    Name    | Type  |  Owner   
--------+------------+-------+----------
 public | temp_table | table | postgres
(1 row)

postgres=# \d temp_table 
             Table "public.temp_table"
 Column |  Type   | Collation | Nullable | Default 
--------+---------+-----------+----------+---------
 id     | integer |           |          | 

postgres=# 
```

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -868,6 +872,46 @@ func generateAssessmentReport() (err error) {
...
+			&redundantIndex.RedundantIndexDDL, &redundantIndex.ExistingIndexDDL)
+		if err != nil {
+			return fmt.Errorf("error scanning rows for redundant indexes: %w", err)
+		}
+
+		redundantIndexesInfo = append(redundantIndexesInfo, redundantIndex)
+	}
+	redundantIssues := parserIssueDetector.GetRedundantIndexIssues(redundantIndexesInfo)
+	for _, redundantIssue := range redundantIssues {
+		convertedAnalyzeIssue := convertIssueInstanceToAnalyzeIssue(redundantIssue, "", false, false)
```
makalaaneesh, line 908: nit: would have preferred directly converting redundantIssue to an AssessmentIssue, but ignore suggestion if this makes it easy.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -593,6 +593,72 @@ func checker(sqlInfoArr []sqlInfo, fpath string, objType string) {
 	}
 }
 
+func checkRedundantIndexes() {
+	msr, err := metaDB.GetMigrationStatusRecord()
+	if err != nil {
+		utils.ErrExit("error getting migration status record: %v", err)
+	}
+	if !msr.MigrationAssessmentDone { //TODO: add the case to handle if assessment is run in export schema
```
makalaaneesh, line 601: yes. Good point. 
create a function `isMigrationAssessmentDone(msr)` that can abstract out that logic (done directly/done via export-schema).

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -593,6 +593,72 @@ func checker(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+	}
+
+	if msr.SourceDBConf.DBType != POSTGRESQL {
+		return
+	}
+
+	//Only considering the case where assessment is run in this migration
+	//If not via assess-migration, there is going to be a change where internally in the export-schema we will run assessment
+	//so we should have this information most of the time.
+	assessmentReportPath := filepath.Join(exportDir, "assessment", "reports", fmt.Sprintf("%s.json", ASSESSMENT_FILE_NAME))
```
makalaaneesh, line 612: Just scanned the code, we seem to have too many instances of this logic is duplicated.
Can we have a function? 
We don't need to update all the existing usages, just going forward, it will help.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -593,6 +593,72 @@ func checker(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+	if !utils.FileOrFolderExists(assessmentReportPath) {
+		log.Infof("migration assessment report file doesn't exists at %q, skipping apply recommendations step...", assessmentReportPath)
```
makalaaneesh, line 614: fix error message.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -593,6 +593,72 @@ func checker(sqlInfoArr []sqlInfo, fpath string, objType string) {
 	}
 }
 
+func checkRedundantIndexes() {
+	msr, err := metaDB.GetMigrationStatusRecord()
+	if err != nil {
+		utils.ErrExit("error getting migration status record: %v", err)
```
makalaaneesh, line 599: let's not use utils.ErrExit here. let's return an error properly. We can throw a utils.ErrExit only in the cobra run command function.

Same for the entire function.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -593,6 +593,72 @@ func checker(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+	report, err := ParseJSONToAssessmentReport(assessmentReportPath)
+	if err != nil {
+		utils.ErrExit("error parsing json report file %q: %w", assessmentReportPath, err)
+	}
+	assessmentIssues := report.Issues
+	redundantIndexIssues := lo.Filter(assessmentIssues, func(i AssessmentIssue, _ int) bool {
+		return i.Category == PERFORMANCE_OPTIMIZATIONS_CATEGORY && i.Name == queryissue.REDUNDANT_INDEXES_ISSUE_NAME
+	})
+	for _, redundantIndexIssue := range redundantIndexIssues {
+		//this is the Unquoted object name
```
makalaaneesh, line 628: Here's an alternative that comes to mind:

Instead of reading from the report, read from the assessmentDB and create issues out of that. 
Basically, exactly what you're doing in `addAssessmentIssuesForRedundantIndexes` except you return analyzeIssues instead of assessmentIssues. 

Will that simplify things?

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -472,6 +473,55 @@ func (p *ParserIssueDetector) genericIssues(query string) ([]QueryIssue, error)
...
+	/*
+		e.g. INDEX idx1 on t(id); INDEX idx2 on t(id, id1); INDEX idx3 on t(id, id1,id2);
+		redundant index coming from the script can have
+		Redundant - idx1, Existing idx2
+		Redundant - idx2, Existing idx3
+		So in this case we need to report it like
+		Redundant - idx1, Existing idx3
+		Redundant - idx2, Existing idx3
+	*/
+	resolvedRedundantIndexInfo := func(currRedundantIndexInfo utils.RedundantIndexesInfo) utils.RedundantIndexesInfo {
```
makalaaneesh, line 491: I'm thinking  - might be better to phrase this as `getRootExistingIndex` ? idk, "resolved" isn't very clear immediately.

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -472,6 +473,55 @@ func (p *ParserIssueDetector) genericIssues(query string) ([]QueryIssue, error)
 	return result, nil
 }
 
+func (p *ParserIssueDetector) GetRedundantIndexIssues(redundantIndexes []utils.RedundantIndexesInfo) []QueryIssue {
+
+	redundantIndexToInfo := make(map[string]utils.RedundantIndexesInfo)
```
makalaaneesh, line 478: nit: Since we're already inside the `GetRedundantIndexIssues` function, you can skip the "redundant" prefix to variable names inside this func. 
They seem borderline long, and is slightly hard to read. 
(haven't thought about this too much though, you might need to distinguish between "redundant" and existing" in certain situations"?)

yb-voyager/cmd/analyzeSchema.go
makalaaneesh: also think about upgrade scenarios.

priyanshi-yb: I thought about this, but there is a case where assessment is run with a `metadata dir` where the user can use the script and get the information, and provide it. The assessment will generate the report, so for such we will need the `metadata-dir-path`, but that is not possible to get and will create problems in Docker even if we store that in MSR to use here. 
So, I thought in all cases we will have the report with and its in export-dir so shouldn't be a problem in any environment.

priyanshi-yb: > also think about upgrade scenarios.

In upgrade, if the  assessment is run withan  older version and analyze with a newer one, we will not have `redundantIndexes` issues filtered from assessment issues so we will not report them, which is okay I think?

yb-voyager/src/query/queryissue/parser_issue_detector.go
priyanshi-yb: > (haven't thought about this too much though, you might need to distinguish between "redundant" and existing" in certain situations"?)

yes it is required to differentiate between redundant and existing index, but let me see if I can clear it up any further.

yb-voyager/cmd/analyzeSchema.go
makalaaneesh: hmm, makes sense. Can you put this as a comment in this function as to why we chose this approach?

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: I think this is okay, we already had the handling for some fields like (category etc.. ) in the existing functions, so better to use them.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -593,6 +593,72 @@ func checker(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+		qualifiedTableName := ""
+		if len(splits) > 1 {
+			qualifiedTableName = splits[1]
+		}
+		tableObjectName := sqlname.NewObjectNameWithQualifiedName(POSTGRESQL, "", qualifiedTableName)
+		indexSQlObjectName := sqlname.NewObjectName(POSTGRESQL, "", tableObjectName.SchemaName, indexName)
+		qualifiedIndexName := fmt.Sprintf(`%s ON %s`, indexSQlObjectName.Unqualified.MinQuoted, tableObjectName.MinQualified.MinQuoted)
+		if slices.Contains(summaryMap["INDEX"].objSet, qualifiedIndexName) {
+			//If the index is present in exported schema then only report the issue
+			//In case of iterations where manually user has removed the index we shouldn't report the issue
```
makalaaneesh, line 642: good catch

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -593,6 +593,72 @@ func checker(sqlInfoArr []sqlInfo, fpath string, objType string) {
...
+		//this is to create a proper objectname with Quotes if required to do the check in the objSet of Index if its present or not.
+		splits := strings.Split(indexObjectName, " ON ")
+		indexName := splits[0]
+		qualifiedTableName := ""
+		if len(splits) > 1 {
+			qualifiedTableName = splits[1]
+		}
+		tableObjectName := sqlname.NewObjectNameWithQualifiedName(POSTGRESQL, "", qualifiedTableName)
+		indexSQlObjectName := sqlname.NewObjectName(POSTGRESQL, "", tableObjectName.SchemaName, indexName)
+		qualifiedIndexName := fmt.Sprintf(`%s ON %s`, indexSQlObjectName.Unqualified.MinQuoted, tableObjectName.MinQualified.MinQuoted)
```
makalaaneesh, line 639: Can we save it this way (i.e. qualified) while generating the assessment report itself? Rather than doing all this at this point? 

Let's discuss this once on call.

makalaaneesh (approved):
LGTM!

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: https://yugabyte.atlassian.net/browse/DB-16473

PR #2579, opened 2025-05-06
Added code to automatically invoke the 'import data to source' and 'export data from target' commands, taking into account the configuration file.

ShivanshGahlot:
- Enhanced the system to automatically invoke the **'import data to source'** and **'export data from target'** commands, utilizing the configuration file:

    - For command-specific flags, if a parameter is set in the config file, the flag is no longer passed along with the command invocation. Instead, the config file logic handles it when the command is executed.

- Implemented several refinements:
    - Removed the `max-retries` parameter from the config file, as it is marked as hidden in the CLI.
    - Updated the config file tests to reflect these changes.
    - Added the `transaction-ordering` parameter to the config file section for **'export data from target'**.

Example:
For this config file:
```
export-data-from-target:
  transaction-ordering: true
  disable-pb: true

import-data-to-source:
  parallel-jobs: 2
  disable-pb: true
```

The commands that are invoked are these:
`yb-voyager export data from target --export-dir /home/ubuntu/pg_export --send-diagnostics=false --log-level info --config-file /home/ubuntu/yb-voyager-config.yaml --table-list public."customers",public."orders",public."order_items",public."products" --target-ssl-mode disable`

`yb-voyager import data to source --export-dir /home/ubuntu/pg_export --send-diagnostics=false --log-level info --config-file /home/ubuntu/yb-voyager-config.yaml`

The params mentioned in the config file are not passed on as flags and are handled post the command starts.

None

Manually

None

### DIFF
--- a/yb-voyager/cmd/config.go
+++ b/yb-voyager/cmd/config.go
@@ -98,7 +98,7 @@ var allowedFinalizeSchemaPostDataImportConfigKeys = mapset.NewThreadUnsafeSet[st
 var allowedImportDataConfigKeys = mapset.NewThreadUnsafeSet[string](
 	"batch-size", "parallel-jobs", "enable-adaptive-parallelism", "adaptive-parallelism-max",
 	"skip-replication-checks",
-	"disable-pb", "exclude-table-list", "table-list",
+	"disable-pb", "max-retries", "exclude-table-list", "table-list",
 	"exclude-table-list-file-path", "table-list-file-path", "enable-upsert", "use-public-ip",
 	"target-endpoints", "truncate-tables",
 	// environment variables keys
@@ -109,14 +109,14 @@ var allowedImportDataConfigKeys = mapset.NewThreadUnsafeSet[string](
 )
 
 var allowedImportDataToSourceConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"parallel-jobs", "disable-pb",
+	"parallel-jobs", "disable-pb", "max-retries",
 	// environment variables keys
 	"num-event-channels", "event-channel-size", "max-events-per-batch",
 	"max-interval-between-batches", "max-batch-size-bytes",
 )
 
 var allowedImportDataToSourceReplicaConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"batch-size", "parallel-jobs", "truncate-tables", "disable-pb",
+	"batch-size", "parallel-jobs", "truncate-tables", "disable-pb", "max-retries",
 	// environment variables keys
 	"ybvoyager-max-colocated-batches-in-progress", "num-event-channels",
 	"event-channel-size", "max-events-per-batch", "max-interval-between-batches",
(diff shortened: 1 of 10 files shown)


yb-voyager/cmd/config.go
```diff
@@ -96,7 +96,7 @@ var allowedFinalizeSchemaPostDataImportConfigKeys = mapset.NewThreadUnsafeSet[st
 var allowedImportDataConfigKeys = mapset.NewThreadUnsafeSet[string](
 	"batch-size", "parallel-jobs", "enable-adaptive-parallelism", "adaptive-parallelism-max",
 	"skip-replication-checks",
-	"disable-pb", "max-retries", "exclude-table-list", "table-list",
+	"disable-pb", "exclude-table-list", "table-list",
```
makalaaneesh, line 99: why did we get rid of max-retries?

yb-voyager/cmd/exportData.go
```diff
@@ -1517,9 +1518,37 @@ func startFallBackSetupIfRequired() {
...
 		cmd = append(cmd, "--yes")
 	}
-	if disablePb {
+
+	passImportDataToSourceSpecificCLIFlags := map[string]bool{
+		"disable-pb": true,
+	}
+
+	// Check whether the command specifc flags have been set in the config file
+	if cfgFile != "" {
```
makalaaneesh, line 1527: It would be good to isolate all viper code only in the root command pre-run step. i.e. the first thing that we do is extract all the information that we want from viper and store it somewhere. Rest of the code only works upon those stored variables, and does not interact with viper directly. 

So, how about we capture all these flag values in `initConfig` and store them somewhere, which can then be used by this function subsequently?

ShivanshGahlot: Yep done that. You can have a look.

ShivanshGahlot (commented):
Along with the requested changes, I have also added config file flag to some more commands. It turns out it was missed as I was setting it in the function registerCommonGlobalFlags. But this function is not called in all the commands. Realised this while adding tests for the live migration commands.

yb-voyager/cmd/config.go
```diff
@@ -253,6 +256,9 @@ func initConfig(cmd *cobra.Command) ([]ConfigFlagOverride, []EnvVarSetViaConfig,
 		return nil, nil, nil, fmt.Errorf("failed to bind environment variables to viper: %w", err)
 	}
 
+	// Set the in-memory config file for later use
+	InMemoryConfigFile = v.AllSettings()
```
makalaaneesh, line 260: Minor aspect: 
The problem with saving the entire config file is that it opens up the possibility of someone using this data directly somewhere else. We don't want that because we want the flow to be viper -> cobra cli -> variables. (i.e. our main code always uses variables and not directly interact with cobra/viper). 

How about: 
- If the current command is export-data-from-source, only save the "import-data-to-source." keys in a specific variable `importDataToSourceConfigKeys` 
- if the current command  is import-data-to-target, only save the "export-data-from-target" keys in a specific variable `exportDataFromTargetConfigKeys`

ShivanshGahlot: Given the recent discussion added code to read the config file dynamically before starting the command itself. 
Since the user might change the config file while the import data to target is running and we need the latest config file before we run the export data from target.

yb-voyager/cmd/config.go
```diff
@@ -671,3 +672,102 @@ func setToAliasPrefixIfSet(configKeyPrefix string, v *viper.Viper) string {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read and validate config file: %w", err)
+	}
+	if v == nil {
+		return []string{}, nil
+	}
+
+	// Get the export-data-from-target keys that are set in the config file
+	exportDataFromTargetKeys := []string{}
+	keyPrefix := "export-data-from-target."
```
makalaaneesh, line 727: nit: use a const here; and for import-data-to-source

makalaaneesh (approved):
LGTM

PR #2581, opened 2025-05-07
Remove performance optimizations from schema analysis report and hash sharding issue reporting

priyanshi-yb:

- Removed reporting of Hash sharding on timestamp type column in the index other than the first column. #2566  
 **Reason -** Not required to report this issue as by default the Hash sharding is only applicable to the first column of the index, so the columns other than the first one are already range-sharded (ASC) by default. So, no action required by users.
- Removed the reporting of performance optimizations in the analyze-schema report. #2569 #2576
**Reason -** Schema Analysis report is an iterative process where the user can do manual changes to the exported schema, and the issues are not reported for those fixed in the schema. There are a couple of reasons that reporting performance optimisations is complicated and not required - 
  1. There can be various ways in which users can fix the schema for performance optimisations and understand all cases and handle all of them, so that the issue is not reported if the schema is fixed to handle that optimization is complicated.
  2. For most cases, reporting the performance optimization involves fetching some stats from the source database, and the analyze-schema is designed to only report issues that can be detected by understanding the exported schema DDLs. Users can modify the exported schema, which will eventually be used to create the schema on target databases, so involving the source DB can become tricky because we need to then somehow map both these information to report issues in analyze-schema.
- Updated Latest stable 2024.2.2.3 

1. Schema analysis won't report any performance optimizations
2. Hash sharding issue on timestamp type column in index other than first column won't be reported in assessment.

Existing tests.


yb-voyager/cmd/analyzeSchema.go
```diff
@@ -1213,10 +1149,13 @@ func analyzeSchema() {
 		utils.ErrExit("failed to get the migration status record: %s", err)
 	}
 	analyzeSchemaInternal(msr.SourceDBConf, true)
-	err = checkRedundantIndexes()
-	if err != nil {
-		utils.ErrExit("failed to get the redundant index issues from assessment report: %v", err)
-	}
+
+	//Not populate any perfomance optimizations in analyze-schema report when it is generated with  the command.
```
makalaaneesh, line 1153: This seems like a slightly round-about way of doing it. 
What i understand is: analyzeSchemaInternal can get you certain performance optimizations
1. in assess-migration, we want the performance optimizations. 
2. in analyze-schema, we don't want the performance optimizations. 

How about instead: analyzeSchemaInternal can have a flag `detectPerformanceOptimzations` and the caller can decide what they want.

makalaaneesh:
Can you pls add to the PR description WHY exactly are we removing the performance optimizations from analyze-schema and WHY we're not reporting the "hash sharding" issues

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: yeah makes sense

makalaaneesh: Ah good catch on the invalid counts! 
Seems like a good example of how taking shortcut approaches can sometimes lead to bugs 😄

makalaaneesh (approved):
LGTM

yb-voyager/cmd/analyzeSchema.go
priyanshi-yb: Yeah!

PR #2608, opened 2025-05-12
fix for missing left-padded 0s when migrating bit types in changes phase for postgres

makalaaneesh:
Fix for left-padded 0s getting skipped in postgres live migration. 

In postgres, there are 3 types of BIT type variables - BIT, BIT VARYING, BIT(n). debezium gives out bytes in all three cases. 
For fixed length types, debezium would also provide the length of the column, which would allow us to properly convert the bytes to a bit string of fixed length (that is insertable into YB). 
For BIT VARYING types, when converting the bytes to a bit string, there was no way for us to know how many 0s were there at the beginning of the bit string. Therefore, if the value had any left-padded 0s, they would get missed when inserting into YB. 

This PR eliminates that problem for PG by writing stringify logic in the debezium converter layer, so that debezium returns values for the BIT types in a string format. Therefore in voyager import layer, there is no need for any conversion. 

Note: For MySQL BIT type, there is potentially a similar problem (when using BETA_FAST_DATA_EXPORT), however handling that requires more thought and work. Therefore, we are keeping the existing conversion logic (bytes/debezium.Bits -> bit string) in the yugabyte converter suite.

Fixed existing tests.


debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/PostgresToYbValueConverter.java
```diff
@@ -28,17 +28,22 @@ public void configure(Properties props) {
     public void converterFor(RelationalColumn column,
             ConverterRegistration<SchemaBuilder> registration) {
 
+        LOGGER.info("Processing converter for column: {}, type: {}, JDBC type: {}", column.name(), column.typeName(), column.jdbcType());
```
priyanshi-yb, line 31: do we need these logs as info ones? I think we can keep these as debug.

migtests/tests/pg/datatypes/validate
```diff
@@ -119,15 +119,14 @@ def migration_completed_checks(tgt):
 		transform_func=str, expected_distinct_values=['1001100101','0001010101','1001000101'])
 	
 	print(f"distinct_bit_varying_values:")
-	# expected_values = ['0001010101','0001010','00101010101010101010101010001010100101010101010101000']
-	expected_values = ['1010101','1010','101010101010101010101010001010100101010101010101000']
+	expected_values = ['0001010101','0001010','00101010101010101010101010001010100101010101010101000']
+	# expected_values = ['1010101','1010','101010101010101010101010001010100101010101010101000']
```
priyanshi-yb, line 123: remove comment

priyanshi-yb (approved):
LGTM

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/PostgresToYbValueConverter.java
makalaaneesh: yeah, fair. I think these are useful for development, when we need to figure out the exact type of the column that we're dealing with..

PR #2622, opened 2025-05-15
[DB-16276] Missing metadata and inconsistent data fileName for non-public schema's data export

sanyamsinghal:

[DB-16276](https://yugabyte.atlassian.net/browse/DB-16276)

- it was not arising in normal COPY imports but for fast path with recovery where INSERT stmts are used it started coming up
- Fixed the code to use expected `NameTuple` for handling tables instead of manual strings.

NA

- Updated existing unit test to not workaround the schema list and rather use the single non public schema list if that's the case.

### DIFF
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -92,7 +92,7 @@ func updateFilePaths(source *srcdb.Source, exportDir string, tablesProgressMetad
 		for _, key := range sortedKeys {
 			tableName := tablesProgressMetadata[key].TableName
 			fullTableName := tableName.ForKey()
-			table := tableName.ForMinOutput()
+			table := tableName.ForDataFile()
 			if _, ok := requiredMap[fullTableName]; ok { // checking if toc/dump has data file for table
 				tablesProgressMetadata[key].InProgressFilePath = filepath.Join(exportDir, "data", requiredMap[fullTableName])
 				tablesProgressMetadata[key].FinalFilePath = filepath.Join(exportDir, "data", table+"_data.sql")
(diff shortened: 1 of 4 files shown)


sanyamsinghal:
All - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/5338/

yb-voyager/cmd/common.go
```diff
@@ -92,7 +92,7 @@ func updateFilePaths(source *srcdb.Source, exportDir string, tablesProgressMetad
 		for _, key := range sortedKeys {
 			tableName := tablesProgressMetadata[key].TableName
 			fullTableName := tableName.ForKey()
-			table := tableName.ForMinOutput()
```
priyanshi-yb, line 95: Instead of adding the schema name in the datafile for the single schema case, can't we fix the place where this name is being used to fetch the column name? 
I think this change can affect various other assumptions in code because we might be relying on this name to be unqualified for the single schema case.

sanyamsinghal: `ForMinOutput` function return the name qualified or not based on what is the default schema.
As per our implementation default schema for PG is:
1. if there are multiple and public is there then public
2. if single is there then that(irrespective of public/non-public)
3. Other no default schema

sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/5366/

yb-voyager/cmd/import_data_test.go
```diff
@@ -457,7 +457,7 @@ INSERT INTO test_schema.test_data (name) VALUES ('name_%d');`, i))
 
 	_, err := testutils.RunVoyagerCommand(postgresContainer, "export data", []string{
 		"--export-dir", exportDir,
-		"--source-db-schema", "public,test_schema", // adding public due to DB-16276
```
priyanshi-yb, line 460: Maybe we can keep one test with multi schema as well.

PR #2662, opened 2025-05-22
Implement fast path for import data file command

sanyamsinghal:
- Handling the column list for `INSERT ON CONFLICT` statement  in case of `import data file` command
- There are two cases:
  - Either no header is provided, in that COPY would assume all column and import but for INSERT we need to fetch the actual column list from target DB
  - If `--has-header=true` in that case, we can reuse that, just need to handle the first row(header row) in each batch file during fast path recovery mode

NA

Added go test (end to end) running import data file command with data already present - trigger fast path recovery

NA

### DIFF
--- a/yb-voyager/cmd/import_data_test.go
+++ b/yb-voyager/cmd/import_data_test.go
@@ -499,6 +499,108 @@ INSERT INTO test_schema.test_data (name) VALUES ('name_%d');`, i))
 	}
 }
 
+func TestImportData_FastPath_OnPrimaryKeyConflictAsIgnore_TableAlreadyHasData_MultiSchema(t *testing.T) {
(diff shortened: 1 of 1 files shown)


sanyamsinghal:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/5392/

sanyamsinghal:
Closing this in favour of https://github.com/yugabyte/yb-voyager/pull/2689

Right now we went with COPY command per row in a batch for fast path recovery mode.
That is not an optimal solution but serves the purpose with minimal implementation.

We can revisit temp table + INSERT approach later in future once TOAST temp table issue is fixed in YB
and we see any issues with the performance.

PR #2663, opened 2025-05-22
Miscellaneous changes for Import data fast path

sanyamsinghal:
- Restrict using `--enable-upsert` alongside `--on-primary-key-conflict = IGNORE`
- added Go test for all datatypes with `INSERT ON CONFLICT` statement
- Restrict changing of conflict from `ERROR` -> `IGNORE` or vice-versa once import has started.

### DIFF
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strings"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/slices"
 
@@ -456,7 +457,8 @@ var onPrimaryKeyConflictActions = []string{
 }
 
 func validateOnPrimaryKeyConflictFlag() {
-	conflictAction := strings.ToUpper(tconf.OnPrimaryKeyConflictAction)
+	log.Infof("passed value for --on-primary-key-conflict: %s", tconf.OnPrimaryKeyConflictAction)
+	tconf.OnPrimaryKeyConflictAction = strings.ToUpper(tconf.OnPrimaryKeyConflictAction)
 
 	// ensure that OnPrimaryKeyConflictAction is not changed in case of resumption
 	if !startClean {
@@ -465,20 +467,23 @@ func validateOnPrimaryKeyConflictFlag() {
 			utils.ErrExit("Error getting migration status record: %v", err)
 		}
 
-		if msr != nil && msr.OnPrimaryKeyConflictAction != tconf.OnPrimaryKeyConflictAction {
+		if msr != nil && msr.OnPrimaryKeyConflictAction != "" && msr.OnPrimaryKeyConflictAction != tconf.OnPrimaryKeyConflictAction {
 			utils.ErrExit("Error: --on-primary-key-conflict flag cannot be changed after the import has started."+
-				"Previous value was %s, current value is %s", msr.OnPrimaryKeyConflictAction, conflictAction)
+				"Previous value was %s, current value is %s", msr.OnPrimaryKeyConflictAction, tconf.OnPrimaryKeyConflictAction)
 		}
 	}
 
-	if conflictAction != "" {
-		if !slices.Contains(onPrimaryKeyConflictActions, conflictAction) {
+	if tconf.OnPrimaryKeyConflictAction != "" {
+		if !slices.Contains(onPrimaryKeyConflictActions, tconf.OnPrimaryKeyConflictAction) {
 			utils.ErrExit("Error: Invalid value for --on-primary-key-conflict. Allowed values are: [%s]", strings.Join(onPrimaryKeyConflictActions, ", "))
 		}
 	}
 
 	// restrict setting --enable-upsert as true if --on-primary-key-conflict is not set to ERROR
-	if tconf.EnableUpsert && conflictAction != constants.PRIMARY_KEY_CONFLICT_ACTION_ERROR {
+	if tconf.EnableUpsert && tconf.OnPrimaryKeyConflictAction != constants.PRIMARY_KEY_CONFLICT_ACTION_ERROR {
 		utils.ErrExit("Error: --enable-upsert=true can only be used with --on-primary-key-conflict=ERROR")
 	}
+
+	// once all validations passed we can save the action value in MSR
+	saveOnPrimaryKeyConflictActionInMSR()
 }
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -457,6 +457,20 @@ var onPrimaryKeyConflictActions = []string{
 
 func validateOnPrimaryKeyConflictFlag() {
 	conflictAction := strings.ToUpper(tconf.OnPrimaryKeyConflictAction)
+
+	// ensure that OnPrimaryKeyConflictAction is not changed in case of resumption
+	if !startClean {
+		msr, err := metaDB.GetMigrationStatusRecord()
+		if err != nil {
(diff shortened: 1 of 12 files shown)


yb-voyager/cmd/import.go
```diff
@@ -456,10 +457,33 @@ var onPrimaryKeyConflictActions = []string{
...
-	if conflictAction != "" {
-		if !slices.Contains(onPrimaryKeyConflictActions, conflictAction) {
+	log.Infof("passed value for --on-primary-key-conflict: %s", tconf.OnPrimaryKeyConflictAction)
+	tconf.OnPrimaryKeyConflictAction = strings.ToUpper(tconf.OnPrimaryKeyConflictAction)
+
+	// ensure that OnPrimaryKeyConflictAction is not changed in case of resumption
+	if !startClean {
+		msr, err := metaDB.GetMigrationStatusRecord()
+		if err != nil {
+			utils.ErrExit("Error getting migration status record: %v", err)
```
makalaaneesh, line 467: let's try to return proper errors everywhere. No `utils.ErrExit`. 
Only in the cobra command Run/pre-run functions, it should be okay to use `utils.ErrExit`

yb-voyager/cmd/import.go
```diff
@@ -456,10 +457,33 @@ var onPrimaryKeyConflictActions = []string{
...
+		}
+
+		if msr != nil && msr.OnPrimaryKeyConflictAction != "" && msr.OnPrimaryKeyConflictAction != tconf.OnPrimaryKeyConflictAction {
```
makalaaneesh, line 470: Under what conditions can msr be nil at this point? 
if we do NOT expect it, we should return an error

yb-voyager/cmd/import.go
```diff
@@ -456,10 +457,33 @@ var onPrimaryKeyConflictActions = []string{
...
+			utils.ErrExit("Error: --on-primary-key-conflict flag cannot be changed after the import has started."+
+				"Previous value was %s, current value is %s", msr.OnPrimaryKeyConflictAction, tconf.OnPrimaryKeyConflictAction)
+		}
+	}
+
+	if tconf.OnPrimaryKeyConflictAction != "" {
+		if !slices.Contains(onPrimaryKeyConflictActions, tconf.OnPrimaryKeyConflictAction) {
```
makalaaneesh, line 477: shouldn't we do this check before the above block (where we check if it is different compared to previous value)

yb-voyager/cmd/importData.go
```diff
@@ -1404,3 +1405,28 @@ func createInitialImportDataTableMetrics(tasks []*ImportFileTask) []*cp.UpdateIm
...
+}
+
+func cleanMSRForImportDataStartClean() {
+	if !startClean {
+		log.Infof("skipping cleaning migration status record for import data command start clean")
+	}
+
+	msr, err := metaDB.GetMigrationStatusRecord()
+	if err != nil {
+		utils.ErrExit("failed to get migration status record: %s", err)
```
makalaaneesh, line 1422: return error

yb-voyager/cmd/import_data_test.go
```diff
@@ -496,6 +379,97 @@ INSERT INTO test_schema.test_data (name) VALUES ('name_%d');`, i))
...
+		"--export-dir", exportDir,
+		"--disable-pb", "true",
+		"--on-primary-key-conflict", "IGNORE",
+		"--batch-size", "10",
+		"--start-clean", "true",
+		"--yes",
+	}
+	// second run: to test INSERT ON CONFLICT DO NOTHING statements with all datatypes
+	_, err = testutils.RunVoyagerCommand(yugabytedbContainer, "import data", importDataCmdArgs, nil, false)
+	if err != nil {
```
makalaaneesh, line 450: `nit: testutils.FatalIfError` for all these cases

yb-voyager/cmd/import_data_test.go
```diff
@@ -496,6 +379,97 @@ INSERT INTO test_schema.test_data (name) VALUES ('name_%d');`, i))
...
+		t.Fatalf("Error connecting to YugabyteDB: %v", err)
+	}
+
+	// Compare the full table data between Postgres and YugabyteDB.
+	// We assume the table "test_data" has a primary key "id" so we order by it.
+	if err := testutils.CompareTableData(ctx, pgConn, ybConn, "test_schema.all_types", "id"); err != nil {
+		t.Errorf("Table data mismatch between Postgres and YugabyteDB: %v", err)
+	} else {
+		t.Log("Success: Table data in Postgres and YugabyteDB match exactly.")
```
makalaaneesh, line 469: a success log is pretty much useless, since the test passing is an indication of success.

yb-voyager/cmd/import_data_test.go
```diff
@@ -496,6 +379,97 @@ INSERT INTO test_schema.test_data (name) VALUES ('name_%d');`, i))
...
+		"--yes",
+	}
+
+	// first run: import data command to load data from PG
+	_, err = testutils.RunVoyagerCommand(yugabytedbContainer, "import data", importDataCmdArgs, nil, false)
+	if err != nil {
+		t.Fatalf("Import command failed: %v", err)
+	}
+
+	// second run: test IGNORE on primary key conflict
```
makalaaneesh, line 439: The current test tests the scenario where the row is IGNORED on conflict. 

Idea to improve/add another test: 
Remove some rows randomly out of the 10000 rows. This will test the scenario where the `INSERT on CONFLICT` actually leads to an insert for a few rows (the ones you deleted).

sanyamsinghal: but random skipping is also a good idea

yb-voyager/cmd/import_data_test.go
sanyamsinghal: multiple parallel PRs have this. Will clean it for once, later.

yb-voyager/cmd/import.go
sanyamsinghal: done

yb-voyager/src/metadb/migrationStatus.go
```diff
@@ -53,6 +53,8 @@ type MigrationStatusRecord struct {
 	ExportDataSourceDebeziumStarted bool `json:"ExportDataSourceDebeziumStarted"`
 	ExportDataTargetDebeziumStarted bool `json:"ExportDataTargetDebeziumStarted"`
 
+	OnPrimaryKeyConflictAction string `json:"OnPrimaryKeyConflictAction"` // only used in import data or import data file commands
```
makalaaneesh, line 56: Should we have separate flags for `import-data-to-target` and `import-data-file` ? Could they conflict? I.e. is it possible to run import-data-file and import-data using the same export-dir? I think it's technically possible...

If we're going with separate MSR records, I'd recommend using this as an opportunity to use a separate struct/json, not MigrationStatusRecord. Perhaps `ImportDataToTargetState` / `ImportDataFileState` ?

sanyamsinghal: >  is it possible to run import-data-file and import-data using the same export-dir

Yeah it might be, but do we recommend doing that.

We can take this opportunity to restructure.
But to be same page, is this a case that mandates the restructuring?

sanyamsinghal: > if it's not broken (i.e. if technically you can run both import-data-file and import-data using the same export-dir)

I don't know this. Guideline from us is to use separate export-dirs for each migration/bulk-load.
That kind of workflow is not something we recommend.

If we want to allow that, then there are other considerations also - like they should not be allowed parallel (lock file implementation needs enhancement then)

makalaaneesh (approved):
LGTM

PR #2672, opened 2025-05-28
[DB-16642]: Fix behaviour with SSL mode allow in the migration workflow

sanyamsinghal:
Tried to cntralised the defining of SSL Modes

NA

### DIFF
--- a/yb-voyager/cmd/constants.go
+++ b/yb-voyager/cmd/constants.go
@@ -244,7 +244,7 @@ var validExportTypes = []string{SNAPSHOT_ONLY, CHANGES_ONLY, SNAPSHOT_AND_CHANGE
 
 var AllSSLModes = []string{constants.DISABLE, constants.PREFER, constants.ALLOW, constants.REQUIRE, constants.VERIFY_CA, constants.VERIFY_FULL}
 var ValidSSLModesForSourceDB = map[string][]string{
-	MYSQL:      utils.SetDifference(AllSSLModes, []string{constants.ALLOW}), // MySQL does not support ALLOW mode
+	MYSQL:      {constants.DISABLE, constants.PREFER, constants.REQUIRE, constants.VERIFY_CA, constants.VERIFY_FULL}, // MySQL does not support ALLOW mode
 	POSTGRESQL: AllSSLModes,
 	YUGABYTEDB: AllSSLModes,
 }
--- a/yb-voyager/cmd/constants.go
+++ b/yb-voyager/cmd/constants.go
@@ -242,10 +242,11 @@ const (
 var supportedSourceDBTypes = []string{ORACLE, MYSQL, POSTGRESQL, YUGABYTEDB}
 var validExportTypes = []string{SNAPSHOT_ONLY, CHANGES_ONLY, SNAPSHOT_AND_CHANGES}
 
-var validSSLModes = map[string][]string{
-	"mysql":      {"disable", "prefer", "require", "verify-ca", "verify-full"},
-	"postgresql": {"disable", "allow", "prefer", "require", "verify-ca", "verify-full"},
-	"yugabytedb": {"disable", "allow", "prefer", "require", "verify-ca", "verify-full"},
+var AllSSLModes = []string{constants.DISABLE, constants.PREFER, constants.ALLOW, constants.REQUIRE, constants.VERIFY_CA, constants.VERIFY_FULL}
+var ValidSSLModesForSourceDB = map[string][]string{
+	MYSQL:      utils.SetDifference(AllSSLModes, []string{constants.ALLOW}), // MySQL does not support ALLOW mode
+	POSTGRESQL: AllSSLModes,
+	YUGABYTEDB: AllSSLModes,
 }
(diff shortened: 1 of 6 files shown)


yb-voyager/cmd/export.go
```diff
@@ -296,10 +297,10 @@ func validatePortRange() {
 }
 
 func validateSSLMode() {
-	if source.DBType == ORACLE || slices.Contains(validSSLModes[source.DBType], source.SSLMode) {
+	if source.DBType == ORACLE || slices.Contains(ValidSSLModesForSourceDB[source.DBType], source.SSLMode) {
```
makalaaneesh, line 300: I'm confused by this existing condition. `If source = oracle OR sslmode is valid` ? for oracle, we don't check the valid ssl modes then?

Ah, I remember; this is because in oracle, we don't really support ssl modes, we rely on TNS ALIAS. Can you just add a comment here?

yb-voyager/cmd/exportDataFromTarget.go
```diff
@@ -92,7 +92,7 @@ Therefore, we allow users to override the ssl root cert in export-data-from-targ
 For logical-replication connecter, all SSL modes are supported.
 */
 func verifySSLFlags(cmd *cobra.Command, msr *metadb.MigrationStatusRecord) error {
-	allowedSSLModes := []string{constants.DISABLE, constants.PREFER, constants.ALLOW, constants.REQUIRE, constants.VERIFY_CA, constants.VERIFY_FULL}
+	allowedSSLModes := AllSSLModes
```
makalaaneesh, line 95: would it make more sense to use `ValidSSLModesForSourceDB[YUGABYTEDB]` ?

yb-voyager/cmd/import.go
```diff
@@ -344,10 +350,15 @@ func validateImportObjectsFlag(importObjectsString string, flagName string) {
...
+
 	if tconf.SSLMode == "" {
-		tconf.SSLMode = "prefer"
-	} else if tconf.SSLMode != "disable" && tconf.SSLMode != "prefer" && tconf.SSLMode != "require" && tconf.SSLMode != "verify-ca" && tconf.SSLMode != "verify-full" {
-		utils.ErrExit("Invalid sslmode %q. Required one of [disable, allow, prefer, require, verify-ca, verify-full]", tconf.SSLMode)
+		tconf.SSLMode = constants.PREFER
+		return
+	}
+
+	if !slices.Contains(supportedSSLModesOnTargetForImport, tconf.SSLMode) {
```
makalaaneesh, line 360: This seems to be run in all import cases (import-data-to-target, import-data-to-source-replica, import-data-to-source), so the supportedSSLmodes should be determined based on the command (since we're defining `supportedSSLModesOnTargetForImport` and `supportedSSLModesOnSourceOrSourceReplica`

yb-voyager/cmd/constants.go
```diff
@@ -242,10 +242,11 @@ const (
 var supportedSourceDBTypes = []string{ORACLE, MYSQL, POSTGRESQL, YUGABYTEDB}
 var validExportTypes = []string{SNAPSHOT_ONLY, CHANGES_ONLY, SNAPSHOT_AND_CHANGES}
 
-var validSSLModes = map[string][]string{
-	"mysql":      {"disable", "prefer", "require", "verify-ca", "verify-full"},
-	"postgresql": {"disable", "allow", "prefer", "require", "verify-ca", "verify-full"},
-	"yugabytedb": {"disable", "allow", "prefer", "require", "verify-ca", "verify-full"},
+var AllSSLModes = []string{constants.DISABLE, constants.PREFER, constants.ALLOW, constants.REQUIRE, constants.VERIFY_CA, constants.VERIFY_FULL}
+var ValidSSLModesForSourceDB = map[string][]string{
+	MYSQL:      utils.SetDifference(AllSSLModes, []string{constants.ALLOW}), // MySQL does not support ALLOW mode
```
makalaaneesh, line 247: nit: (or rather just a general discussion on code): 

While doing `allSSLModes - allow` is fine; I'd recommend explicitly listing out all the modes. Let's say we need to support a new ssl mode for pg/yb in the future; we will go add it to `AllSSLModes`, but the side effect of that is that it will get added to the MySQL list as well. That makes it slightly error-prone. 

It's also arguably more readable to just read all the modes in a list rather than reading some set logic :) 

To be clear, it's a minor point, but worth thinking about, just in the broader context of writing clean code. See Zen of Python - https://peps.python.org/pep-0020/. One of the core principles of golang as well is "simplicity".

```
Explicit is better than implicit.
Simple is better than complex.
```

yb-voyager/cmd/exportDataFromTarget.go
sanyamsinghal: hmm, probably. let me take a look.

yb-voyager/cmd/import.go
sanyamsinghal: updated. please check

makalaaneesh (approved):
LGTM

PR #2689, opened 2025-06-02
New Implementation for import data fast path recovery mode

sanyamsinghal:
- instead of using INSERT ON CONFLICT statements, use COPY statement per row
- ignoring all unique constraint violation error right now, followup to handle it properly

Add go test for invoking import data and import data file commands.

### DIFF
--- a/yb-voyager/src/tgtdb/target_db_interface.go
+++ b/yb-voyager/src/tgtdb/target_db_interface.go
@@ -134,23 +134,11 @@ func (args *ImportBatchArgs) GetYBTxnCopyStatement() string {
 
 // returns YB COPY statement for fast path
 // To trigger COPY fast path, no transaction and ROWS_PER_TRANSACTION should be used
+// To trigger fast path on server side there should be:
+// 1. No ROWS_PER_TRANSACTION set in COPY statement
+// 2. COPY stmt should not be wrapped in a transaction
 func (args *ImportBatchArgs) GetYBNonTxnCopyStatement() string {
 	options := args.copyOptions()
-	columns := ""
-	if len(args.Columns) > 0 {
-		columns = fmt.Sprintf("(%s)", strings.Join(args.Columns, ", "))
-	}
-
-	return fmt.Sprintf(`COPY %s %s FROM STDIN WITH (%s)`, args.TableNameTup.ForUserQuery(), columns, strings.Join(options, ", "))
-}
-
-func (args *ImportBatchArgs) GetYBCopyStatement() string {
-	options := args.copyOptions()
-	if !args.ShouldUseFastPath() {
-		// fast path on DB side not enabled if ROWS_PER_TRANSACTION is set in COPY
-		options = append(options, fmt.Sprintf("ROWS_PER_TRANSACTION %v", args.RowsPerTransaction))
-	}
-
 	columns := ""
 	if len(args.Columns) > 0 {
 		columns = fmt.Sprintf("(%s)", strings.Join(args.Columns, ", "))
@@ -167,6 +155,7 @@ func (args *ImportBatchArgs) GetPGCopyStatement() string {
 	}
 	return fmt.Sprintf(`COPY %s %s FROM STDIN WITH (%s)`, args.TableNameTup.ForUserQuery(), columns, strings.Join(options, ", "))
 }
+
 
 /*
 TODOs:
--- a/yb-voyager/src/tgtdb/target_db_interface.go
+++ b/yb-voyager/src/tgtdb/target_db_interface.go
@@ -134,7 +134,6 @@ func (args *ImportBatchArgs) GetYBTxnCopyStatement() string {
 
 // returns YB COPY statement for fast path
 // To trigger COPY fast path, no transaction and ROWS_PER_TRANSACTION should be used
-// To trigger fast path on server side there should be:
 // 1. No ROWS_PER_TRANSACTION set in COPY statement
 // 2. COPY stmt should not be wrapped in a transaction
 func (args *ImportBatchArgs) GetYBNonTxnCopyStatement() string {
@@ -155,7 +154,6 @@ func (args *ImportBatchArgs) GetPGCopyStatement() string {
 	}
 	return fmt.Sprintf(`COPY %s %s FROM STDIN WITH (%s)`, args.TableNameTup.ForUserQuery(), columns, strings.Join(options, ", "))
 }
-
 
 /*
 TODOs:
--- a/yb-voyager/src/tgtdb/yugabytedb.go
+++ b/yb-voyager/src/tgtdb/yugabytedb.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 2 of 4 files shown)


yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -550,6 +550,9 @@ func (yb *TargetYugabyteDB) importBatchFast(conn *pgx.Conn, batch Batch, args *I
 
 		Lets say the importBatchFastRecover fails with some trasient DB error. In that case,
 		caller(fileTaskImporter.importBatch() function) takes care of retrying with importBatchFastRecover
+
+		The violation error will be because of PK in this code path, not Non-PK table with unique constraint.
+		Even if table has PK + UK on different columns, violation will be on PK for sure, given that the data is existing on source database with same schema
```
makalaaneesh, line 555: I didn't understand this. How are we sure that this is only going to be PK error? If a table has both a PK and a UK on different columns, it could fail on the UK only.

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -134,6 +134,9 @@ func (args *ImportBatchArgs) GetYBTxnCopyStatement() string {
 
 // returns YB COPY statement for fast path
 // To trigger COPY fast path, no transaction and ROWS_PER_TRANSACTION should be used
+// To trigger fast path on server side there should be:
```
makalaaneesh, line 137: redundant comment?  line above already mentions this

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -656,23 +661,40 @@ func (yb *TargetYugabyteDB) importBatchFastRecover(conn *pgx.Conn, batch Batch,
...
 		}
 
-		// 5. Execute the INSERT statement one by one
-		res, err := conn.Exec(context.Background(), insertStmt, stmtArgs...)
+		// 5. Execute the COPY statement with the line read from the file
+		res, err := conn.PgConn().CopyFrom(context.Background(), singleLineReader, copyCommand)
 		if err != nil {
-			return 0, fmt.Errorf("insert into target table: %w", err)
+			// Ignore err if VIOLATES_UNIQUE_CONSTRAINT_ERROR only
+			if strings.Contains(err.Error(), VIOLATES_UNIQUE_CONSTRAINT_ERROR) {
```
makalaaneesh, line 675: here, the check should be for PK violations specifically, no? not any generic `VIOLATES_UNIQUE_CONSTRAINT_ERROR`

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -656,23 +661,40 @@ func (yb *TargetYugabyteDB) importBatchFastRecover(conn *pgx.Conn, batch Batch,
...
+				// logging lineNum might not be useful as batches are truncated later on
+				log.Debugf("ignoring error %q for line=%q in batch %q", err.Error(), line, batch.GetFilePath())
+				rowsIgnored++ // increment before continuing to next line
+				continue
+			}
+
+			var pgerr *pgconn.PgError
+			if errors.As(err, &pgerr) {
+				err = fmt.Errorf("%s, %s in %s", err.Error(), pgerr.Where, batch.GetFilePath())
+			}
```
makalaaneesh, line 685: what is the purpose of this block?

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -656,23 +661,40 @@ func (yb *TargetYugabyteDB) importBatchFastRecover(conn *pgx.Conn, batch Batch,
...
+			return rowsAffected + rowsIgnored, err
 		}
 
+		// rowsAffected will be either 0 or 1
 		if res.RowsAffected() > 0 {
 			rowsAffected++
 		} else {
+			// at this point it is not expected to have 0 rows affected
```
makalaaneesh, line 693: should we error out here?

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: If it is import data command then we are sure right?
Import data file i know anything is possible.

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: i think this just adds more context to the error message.
Copied it from the copyBatchCore() function

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: hmm, since it is unknown and unexpected case?
Can error policy take care of this?

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: Yeah as i mentioned in other comment.
In case of import data command i think this wont be an issue, in case of import data file this comparison can be wrong. Right? But yeah agree we need the complete check

yb-voyager/src/tgtdb/yugabytedb.go
makalaaneesh: Ideally, yes. But consider the cases: 
- the user already ingested some rows into the table before running voyager (for whatever reason). 
- there is no UK on the source, but during migration, user decided to add a UK on the target side. The UK is actually not appropriate. 

In both cases, with the current check (only for UK), we would silently fail, correct?

yb-voyager/src/tgtdb/yugabytedb.go
makalaaneesh: on second thoughts, it's probably okay to warn. We also do that in `logDiscrepancyInEventBatchIfAny` . It is definitely unexpected. COPY should not return 0 rows affected. 

Error policy will kick in anyways if there is an error(at the FileTaskImporter.importBatch layer). I didn't understand what you mean by "take care of this"?

yb-voyager/src/tgtdb/yugabytedb.go
makalaaneesh: yeah, i meant we would ignore that error.

sanyamsinghal: Handling unique constraint case in a subsequent PR, lets get this one merged.

makalaaneesh: Cool 👍 
Okay to do it in a separate PR obviously, ofc but just to make sure we're on the same page, this is essentially a must-have, not a good-to-have. 

> While in case of Unique Key or non-fast path it should be NonRetryable Error and exit.

did not completely understand this statement. getting a UK violation needs to be treated as a NonRetriableError irrespective of fast-path or normal-path. 


also, pls remove the comment in code before merging. Can be confusing later.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -656,23 +661,40 @@ func (yb *TargetYugabyteDB) importBatchFastRecover(conn *pgx.Conn, batch Batch,
...
+			}
+
+			var pgerr *pgconn.PgError
+			if errors.As(err, &pgerr) {
+				err = fmt.Errorf("%s, %s in %s", err.Error(), pgerr.Where, batch.GetFilePath())
+			}
+			return rowsAffected + rowsIgnored, err
 		}
 
+		// rowsAffected will be either 0 or 1
```
makalaaneesh, line 689: at this point, rowsAffected is expected to be 1, right? Pls fix this comment.

makalaaneesh (approved):
LGTM with minor comments

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: yes its a must have.

yb-voyager/src/tgtdb/yugabytedb.go
sanyamsinghal: done

PR #2714, opened 2025-06-06
Handling for constraint violation error from PK vs UK

sanyamsinghal:
- Unique Constraint violation check response from voyager is based on the on primary key conflict mode
  - `ERROR`: All errors whether due to Primary Key or Unique Key, should be non retryable and error out
  - `IGNORE`: If the error is due to Primary Key conflict only then ignore it since user has selected that mode. Otherwise for unique key case it will non retry and error.
  
Moreover the overall check for violation is much more strict now.
i.e. when ignoring we check the expected primary key constraint name in the error msg

Syntax: `violates unique constraint "%s"`
Here we check with the specific primary key constraint name that is expected for that batch
  

 Unit tests added 

### DIFF
--- a/yb-voyager/cmd/import_data_test.go
+++ b/yb-voyager/cmd/import_data_test.go
@@ -26,6 +26,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/tgtdb"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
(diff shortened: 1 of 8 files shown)


yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -43,6 +43,7 @@ type TargetDB interface {
 	ImportBatch(batch Batch, args *ImportBatchArgs, exportDir string, tableSchema map[string]map[string]string, isRecoveryCandidate bool) (int64, error)
 	QuoteAttributeNames(tableNameTup sqlname.NameTuple, columns []string) ([]string, error)
 	GetPrimaryKeyColumns(table sqlname.NameTuple) ([]string, error)
+	GetPrimaryKeyConstraintName(tableNameTup sqlname.NameTuple) (string, error)
```
makalaaneesh, line 46: consider implementing this inside YB ONLY and not expose it via the interface. Reasons being:
- We don't really need this for other DBs now. The interface has become too huge already. In fact, the outer layer (FileTaskImporter) doesn't really need to know what the PKColumns/PKConstraintName exactly are, right? We seem to be just getting this value and simply passing it back to tdb?
- It's only needed for the fast path, which is an internal detail of the yb.ImportBatch. 
- It seems a bit weird that `ImportBatchArgs` has `pkColumns` or `pkConstraintName`. We essentially call `tdb.ImportBatch(table, importBatchArgs)`, so it's a bit unexpected that the tdb wouldn't be able to get the pkColumns for the table. Make sense?

yb-voyager/src/tgtdb/yugabytedb_test.go
```diff
@@ -192,3 +192,55 @@ func TestYugabyteGetNonEmptyTables(t *testing.T) {
...
+			table:          sqlname.NameTuple{CurrentName: sqlname.NewObjectName(POSTGRESQL, "test_schema", "test_schema", "bar")},
+			expectedPKName: "bar_primary_key",
+		},
+		{
+			table:          sqlname.NameTuple{CurrentName: sqlname.NewObjectName(POSTGRESQL, "test_schema", "test_schema", "baz")},
+			expectedPKName: "",
+		},
+		{
+			table:          sqlname.NameTuple{CurrentName: sqlname.NewObjectName(POSTGRESQL, "test_schema", "test_schema", "CASE_sensitive")},
+			expectedPKName: "CASE_sensitive_pkey",
```
makalaaneesh, line 238: nice. Have we verified that this appears in this exact manner in the error message?

makalaaneesh (approved):
LGTM with a few questions/considerations!

yb-voyager/cmd/import_data_test.go
```diff
@@ -832,6 +832,190 @@ CREATE TABLE public.foo (
...
+		"--on-primary-key-conflict", "IGNORE",
+		"--data-dir", dataDir,
+		"--file-table-map", fmt.Sprintf("%s:test_schema.test_data", filepath.Base(dataFilePath)),
+		"--format", "CSV",
+		"--has-header", "true",
+		"--truncate-splits", "false",
+		"--yes",
+	}
+
+	// TODO: planning to enhance RunVoyagerCommand to return the actual error messages, currently it returns a exit status
```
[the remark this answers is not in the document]
sanyamsinghal: Sure, will consider writing FileTaskImporter for this, agree might be a better place.

Will enhance the assertion in a later PR

yb-voyager/cmd/import_data_test.go
```diff
@@ -832,6 +832,190 @@ CREATE TABLE public.foo (
...
+		"--on-primary-key-conflict", "IGNORE",
+		"--data-dir", dataDir,
+		"--file-table-map", fmt.Sprintf("%s:test_schema.test_data", filepath.Base(dataFilePath)),
+		"--format", "CSV",
+		"--has-header", "true",
+		"--yes",
+	}
+
+	_, err = testutils.RunVoyagerCommand(yugabytedbContainer, "import data file", importDataFileCmdArgs, nil, false)
+	if err != nil {
```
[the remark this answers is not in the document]
sanyamsinghal: yeah makes sense, will take care of using asserts for test errors moving forward.

yb-voyager/src/tgtdb/target_db_interface.go
sanyamsinghal: Yeah, the flow of information seems to be messed up a bit.

I have noted this down, would like to merge this merge as it has been pending from some time.

Is it ok if i take this in a separate PR @makalaaneesh ?

yb-voyager/src/tgtdb/yugabytedb_test.go
sanyamsinghal: yes, case senstivity is covered as part of `VIOLATES_UNIQUE_CONSTRAINT_ERROR_RETRYABLE_FAST_PATH` error detection logic

yb-voyager/src/tgtdb/target_db_interface.go
makalaaneesh: Yeah I had approved the PR already 👍

PR #2747, opened 2025-06-12
Fix: Only import data cmd should set on-primary-key-conflict flag in metaDB

sanyamsinghal:
- previously it was getting set by import schema as per default value and import data never able to set it to IGNORE

- Only setting and checking this flag in import data cmd workflow

### DIFF
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -97,10 +97,15 @@ func validateImportFlags(cmd *cobra.Command, importerRole string) error {
 	validateParallelismFlags()
 	validateTruncateTablesFlag()
 
-	err = validateOnPrimaryKeyConflictFlag()
+	return nil
+}
+
+func validateImportDataFlags() error {
+	err := validateOnPrimaryKeyConflictFlag()
 	if err != nil {
 		return fmt.Errorf("error validating --on-primary-key-conflict flag: %w", err)
 	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -101,6 +101,11 @@ var importDataCmd = &cobra.Command{
 		if err != nil {
 			utils.ErrExit("Error validating import flags: %s", err.Error())
 		}
+
+		err = validateImportDataFlags()
```
makalaaneesh, line 105: Couple of things: 
- This is also something we'd want to do for import-data-file. We;ll have to add this validation there as well, right? 
- ImportDataCmd.PreRun is also called by `import-data-to-source-replica` and `import-data-to-source` ? Can we check if it's going to be a similar clash with those commands as well?

sanyamsinghal: > This is also something we'd want to do for import-data-file

Yeah correct, i assumed reusing import data command might be already taking care of it.
But thats not the case.

On digging further, i noticed that `validateImportFlags()` which has some common import data/file checks is not even called for `import data file`


Validations are missing for import data file cmd

makalaaneesh: also some tests seem to be failing with a panic, pls check

yb-voyager/cmd/import.go
```diff
@@ -496,12 +498,17 @@ func validateOnPrimaryKeyConflictFlag() error {
 	}
 
 	// ensure that OnPrimaryKeyConflictAction is not changed in case of resumption
-	if !startClean {
+	/*
+		Most of our cmd validations(including this) run in PreRun phase
+		which can happen before the migration status record or the metaDB is created.
+		For example, in case of import data file command
+	*/
+	if !bool(startClean) && metaDBIsCreated(exportDir) {
```
makalaaneesh, line 506: Hmm, slightly confused. 
We call initMetaDB in the root command's PersistentPreRun. Isn't that run before the pre-run of the specific command?

PR #2750, opened 2025-06-12
Import Data check if there is already data in the table to avoid unique constraint conflicts

sanyamsinghal:
https://github.com/yugabyte/yb-voyager/issues/2541

image

Yes, attached screenshot for that.

Manual Testing.

NA

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -36,6 +36,7 @@ import (
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/adaptiveparallelism"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/callhome"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/config"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/constants"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/cp"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/datafile"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/datastore"
@@ -657,6 +658,8 @@ func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorP
 	log.Infof("pending tasks: %v", pendingTasks)
 	log.Infof("completed tasks: %v", completedTasks)
 
+	checkPKConflictModeOnFreshStart(importFileTasks, pendingTasks, completedTasks)
+
 	//TODO: BUG: we are applying table-list filter on importFileTasks, but here we are considering all tables as per
 	// export-data table-list. Should be fine because we are only disabling and re-enabling, but this is still not ideal.
 	sourceTableList := msr.TableListExportedFromSource
@@ -796,6 +799,45 @@ func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorP
 		displayImportedRowCountSnapshot(state, importFileTasks, errorHandler)
 	}
 	fmt.Printf("\nImport data complete.\n")
+}
+
+// For a fresh start but non empty tables in tableList && OnPrimaryKeyConflict is set to IGNORE -> notify user
+func checkPKConflictModeOnFreshStart(importFileTasks, pendingTasks, completedTasks []*ImportFileTask) {
+	if !isFreshStart(importFileTasks, pendingTasks, completedTasks) {
+		log.Infof("Not a fresh start, skipping primary key conflict mode check.")
+		return
+	}
+
+	// in case of ERROR mode, no need to check for non-empty tables
+	// but for IGNORE or UPDATE(in future), we need to prompt user
+	if tconf.OnPrimaryKeyConflictAction == constants.PRIMARY_KEY_CONFLICT_ACTION_ERROR {
+		return
+	}
+
+	pendingTablesList := importFileTasksToTableNameTuples(pendingTasks)
+	// TODO: skip if those are non PK tables?
+	tablesHavingData := tdb.GetNonEmptyTables(pendingTablesList)
+	if len(tablesHavingData) == 0 {
+		return
+	}
+
+	utils.PrintAndLog(
+		"Importing into tables that already hold data: %v\n\n"+
+			"Current conflict mode: --on-primary-key-conflict-action=IGNORE\n\n"+
+			"What happens now:\n"+
+			"  • Any incoming row whose PK matches an existing row will be dropped.\n"+
+			"  • No error will be thrown for those skipped rows.\n"+
+			"  • You'll end up with only the new non-conflicting rows added.\n\n"+
+			"Please confirm you understand and want to continue.\n", tablesHavingData,
+	)
+	if !utils.AskPrompt("Do you want to continue with the import?") {
+		utils.ErrExit("Aborting import.")
+	}
+}
+
+func isFreshStart(importFileTasks, pendingTasks, completedTasks []*ImportFileTask) bool {
+	// A fresh start is when all tasks are pending(non-zero) and no tasks are completed.
+	return len(pendingTasks) == len(importFileTasks) && len(completedTasks) == 0 && len(pendingTasks) > 0
 }
 
 func restoreGeneratedIdentityColumns(importTableList []sqlname.NameTuple) error {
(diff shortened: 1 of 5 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -798,6 +801,60 @@ func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorP
...
+			"  • Duplicate Primary Key rows are silently skipped.\n"+
+			"  • Import proceeds without error for skipped entries.\n"+
+			"  • Only unique rows will be imported.\n\n", sqlname.NameTupleListToStrings(tablesHavingData),
+	)
+	if !utils.AskPrompt("Please confirm whether to proceed") {
+		utils.ErrExit("Aborting import.")
+	}
+}
+
+func isFreshStart(importFileTasks, pendingTasks, completedTasks []*ImportFileTask) bool {
```
makalaaneesh, line 853: just confirming: will start-clean also lead to this function returning true?

yb-voyager/cmd/importData.go
```diff
@@ -798,6 +801,60 @@ func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorP
 	fmt.Printf("\nImport data complete.\n")
 }
 
+// For a fresh start but non empty tables in tableList && OnPrimaryKeyConflict is set to IGNORE -> notify user
+func checkPKConflictModeOnFreshStart(importFileTasks, pendingTasks, completedTasks []*ImportFileTask) {
```
makalaaneesh, line 805: nit: 
- either name it specifically: something like  `checkForNonEmptyTablesWhenPkConflictIgnore`
- or name it generically something like `runPKConflictModeGuardrails`

yb-voyager/cmd/importData.go
```diff
@@ -798,6 +801,60 @@ func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorP
...
+	if !isFreshStart(importFileTasks, pendingTasks, completedTasks) {
+		log.Infof("Not a fresh start, skipping primary key conflict mode check.")
+		return
+	}
+
+	// in case of ERROR mode, no need to check for non-empty tables
+	// but for IGNORE or UPDATE(in future), we need to prompt user
+	if tconf.OnPrimaryKeyConflictAction == constants.PRIMARY_KEY_CONFLICT_ACTION_ERROR {
```
makalaaneesh, line 813: nit: not going to affect anything, it seems more logical to do this check before the `isFreshStart` check.

yb-voyager/cmd/importData.go
```diff
@@ -798,6 +801,60 @@ func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorP
...
+	tablesHavingData := tdb.GetNonEmptyTables(pendingTablesList)
+	if len(tablesHavingData) == 0 {
+		return
+	}
+
+	allNonPKTables := true
+	for _, table := range tablesHavingData {
+		colList, err := tdb.GetPrimaryKeyColumns(table)
+		if err != nil {
+			utils.ErrExit("Failed to get primary key columns for table %s: %s", table.ForOutput(), err)
```
makalaaneesh, line 827: no utils.ErrExit pls, let's return proper errors

yb-voyager/cmd/importData.go
```diff
@@ -798,6 +801,60 @@ func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorP
...
+	}
+
+	utils.PrintAndLog(
+		"Target tables with pre-existing data: %v\n\n"+
+			"Selected mode: IGNORE on primary-key conflicts\n"+
+			"This mode offers improved import speed when your data contains no PK constraint violations.\n\n"+
+			"Under this mode:\n"+
+			"  • Duplicate Primary Key rows are silently skipped.\n"+
+			"  • Import proceeds without error for skipped entries.\n"+
+			"  • Only unique rows will be imported.\n\n", sqlname.NameTupleListToStrings(tablesHavingData),
```
makalaaneesh, line 846: I'm wondering: do we need to exclude the non-pk tables from this list as well? since it's basically not applicable?

yb-voyager/cmd/importData.go
makalaaneesh, line 805: shall we have a check at the beginning of this function to make sure it  only runs for certain importer-roles - import-data-to-target, import-data-file?

yb-voyager/cmd/importData.go
makalaaneesh, line 846: can you put a screenshot of this in the PR description?

yb-voyager/cmd/importDataFileCommand.go
```diff
@@ -225,9 +225,9 @@ func checkImportDataFileFlags(cmd *cobra.Command) {
 	validateTargetSchemaFlag()
 	validateParallelismFlags()
 
-	err := validateOnPrimaryKeyConflictFlag()
+	err := validateImportDataFlags()
```
makalaaneesh, line 228: I remember this change from another PR. Is there some merge issue?

yb-voyager/cmd/importData.go
sanyamsinghal: yes good catch

yb-voyager/cmd/importData.go
```diff
@@ -798,6 +804,71 @@ func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorP
...
+
+	// all nonEmptyTables have no primary key columns
+	if len(nonPKTables) == len(nonEmptyTables) {
+		log.Infof("All tables in %v have no primary key columns, skipping primary key conflict mode check.", sqlname.NameTupleListToStrings(nonEmptyTables))
+		return nil
+	}
+
+	pkTables := lo.Filter(nonEmptyTables, func(table sqlname.NameTuple, _ int) bool {
+		return !nonPKTables[table]
+	})
```
makalaaneesh, line 849: I was slightly confused by this block. 
How about we: 
- get all the nonEmptyTablesHavingPK. 
- if nonEmptyTablesHavingPK == 0, return (same as checking if all are non-pk)
- for all the nonEmptyTablesHavingPK, prompt.

We can avoid getting nonPk tables first and then flipping it to pkTables

yb-voyager/cmd/importData.go
```diff
@@ -798,6 +804,71 @@ func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorP
...
+	// but for IGNORE or UPDATE(in future), we need to prompt user
+	if tconf.OnPrimaryKeyConflictAction == constants.PRIMARY_KEY_CONFLICT_ACTION_ERROR {
+		return nil
+	}
+
+	if !isTargetDBImporter(importerRole) {
+		return nil
+	}
+
+	if !isFreshStart(importFileTasks, pendingTasks, completedTasks) {
```
makalaaneesh, line 819: (for discussion, not blocker)
I'm wondering if we really need this check. 
Can we blindly just check for all pendingTasks every time? 

the good: 
- this will help with resumption cases as well. I.e. if you stop in the middle, and then resume, and there are some tables for which we have not yet imported, this will check and prompt for those tables as well. 
the bad: 
- in import-data-file case, there can be multiple tasks for the same table, so we will unnecessarily prompt in those cases

makalaaneesh (approved):
LGTM with a couple of comments

yb-voyager/cmd/importData.go
```diff
@@ -1292,6 +1363,52 @@ func classifyTasks(state *ImportDataState, tasks []*ImportFileTask) (pendingTask
 	return append(inProgressTasks, notStartedTasks...), completedTasks, nil
 }
 
+func getNotStartedTasks(state *ImportDataState, tasks []*ImportFileTask) []*ImportFileTask {
```
makalaaneesh, line 1366: Can we instead club all this logic with `classifyTasks` since it's already doing very similar stuff?
Currently, it returns `pendingTasks, completedTasks`
What if we modify it to return `inProgressTasks,notStartedTasks,completedTasks` 

That should solve it right? We can avoid having separate functions. 
What do you think?

sanyamsinghal: i think its fine to have separate getter function. In classifyTasks we can reuse these.

makalaaneesh: I still think it's better to merge this logic in classifyTasks. 
classifyTasks does not have to re-use these methods, because it is iterating through all the tasks and then classifying them. 

If you are keeping these methods, pls fix them to return proper errors, instead of utils.ErrExits

PR #2753, opened 2025-06-12
Major Enhancements in Go Test Framework

sanyamsinghal:

- Implement `VoyagerCommandRunner` to capture output and manage voyager command execution in tests
  This can be further be enhanced as per growing needs:

```
VoyagerCommandRunner {
   Prepare() // just prepare the cmd args
   Run() // prepares if not prepared, and create/execute the cmd. 
   Wait()
   Stdout()
   Stderr() <- stores the full error message generated by the command to compare later in test with expectedErr
}
```

- Some additions to testcontainer interface: `GetVersion()`, `Query()` 

### DIFF
--- a/yb-voyager/test/utils/voyager_cmd_runner.go
+++ b/yb-voyager/test/utils/voyager_cmd_runner.go
@@ -64,7 +64,6 @@ func NewVoyagerCommandRunner(container testcontainers.TestContainer, cmdName str
 		CmdArgs:     cmdArgs,
 		doDuringCmd: doDuringCmd,
 		isAsync:     isAsync,
-		exitCode:    ExitCodeSuccess,
 	}
 	log.Debugf("Creating CommandRunner for command: %s with args: %s", cmdName, strings.Join(cmdArgs, " "))
 	return &cmdRunner
--- a/yb-voyager/test/utils/voyager_cmd_runner.go
+++ b/yb-voyager/test/utils/voyager_cmd_runner.go
@@ -116,90 +116,90 @@ func (this *VoyagerCommandRunner) Prepare() error {
 	return nil
 }
 
-func (this *VoyagerCommandRunner) newCmd() {
-	this.StdoutBuf = &bytes.Buffer{}
-	this.StderrBuf = &bytes.Buffer{}
+func (v *VoyagerCommandRunner) newCmd() {
+	v.StdoutBuf = &bytes.Buffer{}
+	v.StderrBuf = &bytes.Buffer{}
 
-	this.Cmd = exec.Command("yb-voyager", this.finalArgs...)
-	this.Cmd.Stdout = io.MultiWriter(os.Stdout, this.StdoutBuf)
-	this.Cmd.Stderr = io.MultiWriter(os.Stderr, this.StderrBuf)
+	v.Cmd = exec.Command("yb-voyager", v.finalArgs...)
+	v.Cmd.Stdout = io.MultiWriter(os.Stdout, v.StdoutBuf)
+	v.Cmd.Stderr = io.MultiWriter(os.Stderr, v.StderrBuf)
 	// disable callhome diagnostics during tests
-	this.Cmd.Env = append(os.Environ(), "YB_VOYAGER_SEND_DIAGNOSTICS=false")
+	v.Cmd.Env = append(os.Environ(), "YB_VOYAGER_SEND_DIAGNOSTICS=false")
 }
 
-func (this *VoyagerCommandRunner) Run() error {
-	if err := this.Prepare(); err != nil {
+func (v *VoyagerCommandRunner) Run() error {
+	if err := v.Prepare(); err != nil {
 		return fmt.Errorf("failed to prepare command: %w", err)
 	}
 
-	this.newCmd()
+	v.newCmd()
 
-	log.Debugf("running command: %s", this.Cmd.String())
-	err := this.Cmd.Start()
+	log.Debugf("running command: %s", v.Cmd.String())
+	err := v.Cmd.Start()
 	if err != nil {
 		return fmt.Errorf("failed to start command: %w", err)
 	}
 
-	if this.doDuringCmd != nil {
+	if v.doDuringCmd != nil {
 		// Small delay to let the command initiate.
 		time.Sleep(2 * time.Second)
-		this.doDuringCmd()
+		v.doDuringCmd()
 	}
 
-	if !this.isAsync {
-		return this.Wait()
+	if !v.isAsync {
+		return v.Wait()
 	}
 	return nil
 }
 
-func (this *VoyagerCommandRunner) Wait() error {
-	err := this.Cmd.Wait()
+func (v *VoyagerCommandRunner) Wait() error {
+	err := v.Cmd.Wait()
 	if err != nil {
 		if ee, ok := err.(*exec.ExitError); ok {
-			this.exitCode = ExitCode(ee.ExitCode())
+			v.exitCode = ExitCode(ee.ExitCode())
 		} else {
-			this.exitCode = ExitCodeFailure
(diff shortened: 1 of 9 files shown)


yb-voyager/test/utils/voyager_cmd_runner.go
```diff
@@ -0,0 +1,186 @@
...
+		CmdArgs:     cmdArgs,
+		doDuringCmd: doDuringCmd,
+		isAsync:     isAsync,
+		exitCode:    ExitCodeSuccess,
+	}
+	log.Debugf("Creating CommandRunner for command: %s with args: %s", cmdName, strings.Join(cmdArgs, " "))
+	return &cmdRunner
+}
+
+func (this *VoyagerCommandRunner) Prepare() error {
```
makalaaneesh, line 73: `this` is generally not golang idiomatic :) https://go.dev/wiki/CodeReviewComments#receiver-names

I'm curious as to what made you choose "this", since it's more a java thing 😁

yb-voyager/test/utils/voyager_cmd_runner.go
```diff
@@ -0,0 +1,186 @@
...
+		return "Failure"
+	default:
+		return fmt.Sprintf("Exitcode(%d)", int(ec))
+	}
+}
+
+// wrapper for exec.Cmd to capture stdout and stderr and to extend it as per our test framework needs
+type VoyagerCommandRunner struct {
+	// basic command information to be provided
+	container   testcontainers.TestContainer
```
priyanshi-yb, line 38: I am not sure if we should keep the container as part of this struct information as there could be multiple cases where this is not very useful -
1. In case we want to run an `export data` command runner with different source configs (like different dbs ), like create an instance of an export-data command runner and `Run` with different configs?
2. For the `get data-migration-report` type of command, which interacts with different containers.

yb-voyager/test/utils/voyager_cmd_runner.go
```diff
@@ -0,0 +1,186 @@
...
+	CmdArgs     []string
+	isAsync     bool
+	doDuringCmd func()
+
+	// as per test framework needs
+	Cmd       *exec.Cmd
+	StdoutBuf *bytes.Buffer
+	StderrBuf *bytes.Buffer
+	exitCode  ExitCode
+	lastErr   error // lastErr is used to capture the last error encountered during command execution
```
priyanshi-yb, line 49: what is the difference of `Stderr()` and `lastErr` ?

yb-voyager/test/utils/voyager_cmd_runner.go
```diff
@@ -0,0 +1,186 @@
...
+		log.Fatal("Container cannot be nil for source/target commands")
+	}
+
+	cmdRunner := VoyagerCommandRunner{
+		container:   container,
+		CmdName:     cmdName,
+		CmdArgs:     cmdArgs,
+		doDuringCmd: doDuringCmd,
+		isAsync:     isAsync,
+		exitCode:    ExitCodeSuccess,
```
priyanshi-yb, line 67: Why are we starting with the Success exit code?

yb-voyager/test/utils/voyager_cmd_runner.go
sanyamsinghal: > what made you choose "this"

One of two hard problems in CS 😅 - https://martinfowler.com/bliki/TwoHardThings.html

Was not able to comeup with a convincing var name at that moment

yb-voyager/test/utils/voyager_cmd_runner.go
sanyamsinghal: Rightnow i assumed all commands would need this.
But we can make container field of the class/struct optional in that case.
Wouldn't that solve the problem? 

>  create an instance of an export-data command runner and Run with different configs?
I am not sure if i got you but I think that is the specific usecase which will be cleanly covered by this class.
For every case you can have different runner objects with specific fields(like container)

yb-voyager/test/utils/voyager_cmd_runner.go
```diff
@@ -0,0 +1,186 @@
...
+	this.Cmd = exec.Command("yb-voyager", finalArgs...)
+	this.Cmd.Stdout = io.MultiWriter(os.Stdout, this.StdoutBuf)
+	this.Cmd.Stderr = io.MultiWriter(os.Stderr, this.StderrBuf)
+
+	// disable callhome diagnostics during tests
+	this.Cmd.Env = append(os.Environ(), "YB_VOYAGER_SEND_DIAGNOSTICS=false")
+	return nil
+}
+
+func (this *VoyagerCommandRunner) Run() error {
```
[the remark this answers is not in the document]
sanyamsinghal: > what motivated the two steps? (prepare, run)

I think significant amount of stuff is happening in each of the two methods, which IMO by default also makes sense to keep separate. Benefits - Single Responsibility, More flexible/extensible code.

we can write wrapper over these if two step feels inconvenient.

yb-voyager/test/utils/voyager_cmd_runner.go
sanyamsinghal: nvm it is already optional unless it is a source/target db command
```
if container == nil && (isSourceCmd(cmdName) || isTargetCmd(cmdName)) {
		log.Fatal("Container cannot be nil for source/target commands")
	}
```

yb-voyager/test/utils/voyager_cmd_runner.go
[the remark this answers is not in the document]
makalaaneesh: >If i answer in short, code quality -- wrt to just writing independent functions vs struct/classes.
I see the need of this part of code getting enhanced further if we continue to write Go integration test executing voyager command.


(just for discussion) Got it. As we were discussing some time back, consider trying functions first in general :). A class usually makes sense when there is state to maintain over time and multiple calls. in the most common case here - `run command get output`, there is no state really. I could just call a simple function. Code quality can still be maintained (without having to write a class) by splitting that function into `prepare command` and `run command` functions.

The async use-case is where I think you might actually need a class. You could write it in a way where it runs the command async, and stores the stdout, stderr, exitCode in class variables. (idk if you're doing that already?).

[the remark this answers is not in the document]
makalaaneesh: High-level, agreed 👍

priyanshi-yb (approved):
LGTM with couple of comments

yb-voyager/test/utils/voyager_cmd_runner.go
```diff
@@ -0,0 +1,210 @@
...
+	// split the command name on spaces so that Cobra framework
+	// can detect parent and child commands correctly.
+	// e.g. CmdName == "export data" → parts == ["export", "data"]
+	parts := strings.Fields(this.CmdName)
+
+	/*
+		append order is important here as the CmdArgs passed in CommmandRunner might want to override the default of ContainerConfig
+		For eg: append(this.CmdArgs, connectionArgs...) the default connection args with override the ones passed to CommandRunner
+	*/
+	this.finalArgs = append(parts, append(connectionArgs, this.CmdArgs...)...)
```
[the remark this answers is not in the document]
sanyamsinghal: yeah missed this

yb-voyager/test/utils/voyager_cmd_runner.go
```diff
@@ -0,0 +1,210 @@
...
+		return nil // already prepared
+	}
+
+	// Appending to CmdArgs based on the command type.
+	host, port, err := this.container.GetHostPort()
+	if err != nil {
+		return fmt.Errorf("failed to get host port for container: %v", err)
+	}
+
+	config := this.container.GetConfig()
```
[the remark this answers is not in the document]
sanyamsinghal: done

PR #2781, opened 2025-06-24
Lock file for the debezium exporter to error out if multiple debezium processes start

priyanshi-yb:

### DIFF
--- a/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
+++ b/debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
@@ -12,6 +12,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+
 import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.ConfigProvider;
 import org.slf4j.Logger;
@@ -47,17 +53,24 @@ public class YbExporterConsumer extends BaseChangeConsumer {
     boolean shutDown = false;
     Object flushingSnapshotFilesLock = new Object();
 
+    // Lock file logic
+    private File lockFile;
+
     public YbExporterConsumer(String dataDir) {
         this.dataDir = dataDir;
     }
 
     void connect() throws URISyntaxException {
         LOGGER.info("connect() called: dataDir = {}", dataDir);
+
         final Config config = ConfigProvider.getConfig();
 
         snapshotMode = config.getOptionalValue("debezium.source.snapshot.mode", String.class).orElse("");
         retrieveSourceType(config);
         exporterRole = config.getValue("debezium.sink.ybexporter.exporter.role", String.class);
+        
+        // Acquire lock file at startup
+        acquireLockFile();
 
         exportStatus = ExportStatus.getInstance(dataDir);
         exportStatus.setSourceType(sourceType);
@@ -80,6 +93,54 @@ public class YbExporterConsumer extends BaseChangeConsumer {
         flusherThread = new Thread(this::flush);
         flusherThread.setDaemon(true);
         flusherThread.start();
+
+        // Register shutdown hook to release lock on exit
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            releaseLockFile();
+        }));
+    }
+
+    /**
+     * Acquires a lock file in the dataDir. Fails if already locked.
+     */
+    private void acquireLockFile() {
+        lockFile = new File(dataDir, String.format(".%s.lock", exporterRole));
+        if (lockFile.exists()) {
+            String msg = String.format("Lock file %s already exists. Another process may be running for this dataDir.", lockFile.getAbsolutePath());
+            LOGGER.error(msg);
+            throw new IllegalStateException(msg);
+        }
+        try {
+            // Create the lock file atomically
+            boolean created = lockFile.createNewFile();
+            if (!created) {
+                String msg = String.format("Failed to create lock file %s. Another process may be running.", lockFile.getAbsolutePath());
+                LOGGER.error(msg);
+                throw new IllegalStateException(msg);
+            }
+            //write PID or timestamp for debugging
+            String pid = java.lang.management.ManagementFactory.getRuntimeMXBean().getName();
+            Files.writeString(lockFile.toPath(), "PID: " + pid + "\n", StandardOpenOption.WRITE);
+            LOGGER.info("Acquired lock file: {}", lockFile.getAbsolutePath());
+        } catch (IOException e) {
+            String msg = String.format("Error creating lock file %s: %s", lockFile.getAbsolutePath(), e.getMessage());
+            LOGGER.error(msg, e);
+            throw new IllegalStateException(msg, e);
+        }
+    }
+
+    /**
+     * Releases (deletes) the lock file if it exists.
+     */
+    private void releaseLockFile() {
+        if (lockFile != null && lockFile.exists()) {
+            try {
+                Files.delete(lockFile.toPath());
+                LOGGER.info("Released lock file: {}", lockFile.getAbsolutePath());
+            } catch (IOException e) {
+                LOGGER.warn("Failed to delete lock file {}: {}", lockFile.getAbsolutePath(), e.getMessage());
+            }
+        }
     }
 
     private ExportMode getExportModeToStartWith(String snapshotMode) {
@@ -171,6 +232,7 @@ public class YbExporterConsumer extends BaseChangeConsumer {
             LOGGER.info("{} processing complete. Exiting...", operation);
             shutDown = true; // to ensure that no event gets written after switch operation.
         }
+        releaseLockFile();
         System.exit(0);
     }
 
@@ -191,6 +253,7 @@ public class YbExporterConsumer extends BaseChangeConsumer {
             LOGGER.info("End migration processing complete. Exiting...");
             shutDown = true; // to ensure that no event gets written after switch operation.
         }
+        releaseLockFile();
         System.exit(0);
     }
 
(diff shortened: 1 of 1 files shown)


debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -80,6 +93,97 @@ void connect() throws URISyntaxException {
...
+                    String msg = String.format("Lock file %s already exists and process with PID %s is running. Another process may be running for this dataDir.", lockFile.getAbsolutePath(), pid);
+                    LOGGER.error(msg);
+                    throw new IllegalStateException(msg);
+                } else {
+                    // Process is not running, we can safely delete the lock file
+                    LOGGER.warn("Lock file {} exists but process with PID {} is not running. Deleting it.", lockFile.getAbsolutePath(), pid);
+                    releaseLockFile();
+                }
+            } catch (NumberFormatException e) {
+                // If PID is not a valid number, treat it as a stale lock
```
makalaaneesh, line 135: if we want to treat it as a stale lock, shouldn't we release it. We seem to be throwing an exception here?

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -80,6 +93,97 @@ void connect() throws URISyntaxException {
...
+            String pid = lines[0].trim();
+            // Check if the PID is a valid number
+            LOGGER.info("Lock file {} exists with PID: {}", lockFile.getAbsolutePath(), pid);
+            //Parse PID and check if the process is running
+            try {
+                long pidLong = Long.parseLong(pid);
+                // Check if the process with this PID is running
+                if (ProcessHandle.of(pidLong).isPresent()) {
+                    // Process is running, error out
+                    String msg = String.format("Lock file %s already exists and process with PID %s is running. Another process may be running for this dataDir.", lockFile.getAbsolutePath(), pid);
```
makalaaneesh, line 126: did you manually verify that this error is thrown?

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -80,6 +93,97 @@ void connect() throws URISyntaxException {
...
+            LOGGER.error(msg, e);
+            throw new IllegalStateException(msg, e);
+        }
+    }
+    /**
+     * Acquires a lock file in the dataDir with <exporter-role>.lck and stores the pid of the process 
+     * Fails if already locked.
+     */
+    private void acquireLockFile() {
+        lockFile = new File(dataDir, String.format(".%s.lck", exporterRole));
```
makalaaneesh, line 151: where do all the commands' lock files get created? Would be good to use the same directory as that.

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -80,6 +93,97 @@ void connect() throws URISyntaxException {
...
+            try {
+                long pidLong = Long.parseLong(pid);
+                // Check if the process with this PID is running
+                if (ProcessHandle.of(pidLong).isPresent()) {
+                    // Process is running, error out
+                    String msg = String.format("Lock file %s already exists and process with PID %s is running. Another process may be running for this dataDir.", lockFile.getAbsolutePath(), pid);
+                    LOGGER.error(msg);
+                    throw new IllegalStateException(msg);
+                } else {
+                    // Process is not running, we can safely delete the lock file
```
makalaaneesh, line 130: Hmm, I'm slightly skeptical of this. I would have just thrown an error, and the user can always manually delete the lock file. Do we delete it like this in the golang layer?

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -80,6 +93,97 @@ void connect() throws URISyntaxException {
         flusherThread = new Thread(this::flush);
         flusherThread.setDaemon(true);
         flusherThread.start();
+
+        // Register shutdown hook to release lock on exit
```
makalaaneesh, line 99: will the shutdown hook also run if there is some exception thrown? 
If so, let's move this block above immediately after line 71. That way,  even if there is some exception in the lines after that, it will gracefully release the lock.

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -80,6 +93,97 @@ void connect() throws URISyntaxException {
...
+            releaseLockFile();
+        }));
+    }
+
+    /**
+     * Reads the lock file and checks if the PID is still running.
+     * If the PID is not running, it deletes the lock file.
+     * If the PID is running, it throws an IllegalStateException.
+     */
+    private void readLockFileAndCheckPid() {
```
makalaaneesh, line 110: nit: `checkExistingLockFile`
you can move the isLockfileExists block inside this

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
makalaaneesh: OKay let's go ahead then 👍

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
makalaaneesh: Okay yeah makes sense. Just change the comment and the error msg to reflect that. 
(remove the "deleting lock file" since we're not actually deleting)

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -171,6 +275,7 @@ private void checkForSwitchOperationAndHandle(String operation) {
             LOGGER.info("{} processing complete. Exiting...", operation);
             shutDown = true; // to ensure that no event gets written after switch operation.
         }
+        releaseLockFile();
```
makalaaneesh, line 280: wouldn't system.Exit automatically handle it because of the exit hook?

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -191,6 +296,7 @@ private void checkForEndMigrationAndHandle() {
             LOGGER.info("End migration processing complete. Exiting...");
             shutDown = true; // to ensure that no event gets written after switch operation.
         }
+        releaseLockFile();
```
makalaaneesh, line 301: same question: wouldn't system.Exit automatically handle it because of the exit hook?

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
priyanshi-yb: yes it will be handled by the shutdown hook, but I thought initially that when we are calling `system.exit` we can release the lock file on our own and not rely on shutdown hooks as these can sometimes not run properly and so on but now when I think it is okay to depend on shutdown hook as that's the purpose of it and in go layer as well we rely on exit handler for all work to be done during exits/terminations and we don't call them where ever we call os.exit etc.. so removed it from the both places.

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
priyanshi-yb: yes

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -82,6 +102,92 @@ void connect() throws URISyntaxException {
...
+                    String msg = String.format("Lock file %s already exists and process with PID %s is running. Another process may be running for this dataDir.", lockFile.getAbsolutePath(), pid);
+                    LOGGER.error(msg);
+                    throw new IllegalStateException(msg);
+                } else {
+                    // Process is not running, we can safely delete the lock file
+                    LOGGER.warn("Lock file {} exists but process with PID {} is not running. Deleting it.", lockFile.getAbsolutePath(), pid);
+                    releaseLockFile();
+                }
+            } catch (NumberFormatException e) {
+                // If PID is not a valid number, treat it as a stale lock
```
makalaaneesh, line 141: pls fix this comment as well.

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -82,6 +102,92 @@ void connect() throws URISyntaxException {
         flusherThread.start();
     }
 
+    /**
+     * Reads the lock file and checks if the PID is still running.
+     * If the PID is not running, it deletes the lock file.
+     * If the PID is running, it throws an IllegalStateException.
+     */
+    private void checkExistingLockFile() {
+        lockFile = new File(exportDir, String.format(".debezium_%s.lck", exporterRole));
```
makalaaneesh, line 111: nit: 
i think this file should be created outside of this function. 
In acquire (lockFile.createNewFile();) , you sort of implicitly rely on this function being called. 

Better to create the File object outside in `connect()` is what I feel.

PR #2847, opened 2025-07-10
Improved error msgs in case of multiple debezium processes and replication slot being active

priyanshi-yb:
fixes: #2821 

Screenshot 2025-07-10 at 2 15 33 PM

### DIFF
--- a/yb-voyager/cmd/exportData.go
+++ b/yb-voyager/cmd/exportData.go
@@ -394,6 +394,8 @@ func exportData() bool {
 					// we have the PID of the process running debezium export
 					// so we can suggest the user to terminate it
 					errorMsg = fmt.Sprintf("%s. Terminate the process on the pid %s, and re-run the command.", errorMsg, pid)
+				} else {
+					log.Errorf("error getting debezium PID: %v", err)
 				}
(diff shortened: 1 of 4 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -388,7 +388,14 @@ func exportData() bool {
 				utils.ErrExit("error checking if replication slot is active: %v", err)
 			}
 			if isActive {
-				utils.ErrExit("Replication slot '%s' is active. Check and terminate if there is any internal voyager process running.", msr.PGPublicationName)
+				pid, err := dbzm.GetPIDOfDebeziumOnExportDir(exportDir, exporterRole)
```
makalaaneesh, line 391: slight restructure here: 

The core error here is that the replication is already active, which is something we want to report. I think it makes sense to  add the PID information on a best-effort basis (it is possible that the lock file is not there because dbzm is not running; it is possible we run into some other weird error with the lock file?, etc). 

So can we structure it like this: 
- form the base error message. 
- check for pid in lock file. If we get the pid, add the pid to the error message. 
- throw the error.

Does that make sense?

priyanshi-yb: yeah make sense! made the change

PR #2856, opened 2025-07-15
Fix: ANALYZE guardrail checks to consider the auto analyze on PG and gracefully error out if No is answered in prompt

### LINKED ISSUE #2820: Assess migration errors out in case user enters "NO" for ANALYSE prompt
Jira Link: [DB-17466](https://yugabyte.atlassian.net/browse/DB-17466)

priyanshi-yb:

fixes #2820 #2848 
If auto-analyze is enabled on the PG, then the `last_autoanalyze` column is updated with the timestamp of last analyze run automatically in the database. So, considering both the values `last_analyze` / `last_autoanalyze` to check if analyze is not run at all on the db

Erroring out gracefully in case the user answers no for the ANALYZE prompt.

In case auto analyze is enabled, the assessment won't prompt for ANALYZE, not run

manually


yb-voyager/src/srcdb/data/gather-assessment-metadata/postgresql/yb-voyager-pg-gather-assessment-metadata.sh
```diff
@@ -193,7 +193,7 @@ WHERE schemaname = ANY(ARRAY[string_to_array('$schema_list', '|')])
     SELECT 1 
     FROM pg_stat_all_tables pgst2 
     WHERE pgst2.schemaname = pgst1.schemaname 
-      AND pgst2.last_analyze IS NOT NULL
+      AND (pgst2.last_analyze IS NOT NULL OR pgst2.last_autoanalyze IS NOT NULL)
```
makalaaneesh, line 196: This seems to address the scenario where auto analyze was switched on; and there was no explicit analyze run. 

mihnea's doc has an additional item. Are we also planning to tackle this? 
"PG runs ANALYZE by default so I was surprised as this should basically never be a problem. In my example one table is very small (under 50 rows) so it was not auto-analyzed based on PG rules (pg auto-analyze default thresholds are 50 rows and 10% new changes). So I'd guess Voyager was complaining based on that.
I'd say that, if a use-case with hundreds of tables has 1-2 tiny tables that may not have stats, that is not so critical. I'd suggest we can just assume those tables are small and move on. We could also look at table sizes with pg_table_size or so to confirm."

yb-voyager/src/srcdb/data/gather-assessment-metadata/postgresql/yb-voyager-pg-gather-assessment-metadata.sh
```diff
@@ -209,7 +209,8 @@ WHERE schemaname = ANY(ARRAY[string_to_array('$schema_list', '|')])
             read -r user_input
             if [[ "$user_input" != "y" && "$user_input" != "Y" ]]; then
                 echo "You can run ANALYZE manually on the affected schemas before retrying."
```
makalaaneesh, line 211: Out of scope for this PR: but should we also report the tables involved, rather than just the schemas? 
is the concern that the output will have too many tables? 
cc: @shubham-yb; can't remember if you had brought it up or not.

makalaaneesh (approved):
LGTM with a couple of questions on further enhancements

yb-voyager/src/srcdb/data/gather-assessment-metadata/postgresql/yb-voyager-pg-gather-assessment-metadata.sh
priyanshi-yb: This is something we already covered. If you remember, we discussed this, and this is what we report. 
This is what we report

1. no tables analysed at all in schema - Yes report 
2. All tables analysed in schema - No
3. 1 table analysed in schema - No

The problem in his case was that the auto-analyze wasn't handled, and hence it was considering the first case where nothing was analyzed, and hence reporting it. But now this should also not be reported

PR #2878, opened 2025-07-16
Bug fix[DB-17592]: Import data with --on-primary-key-conflict ignore fails if import to source replica is started before it

sanyamsinghal:
- added checks for importer Role wherever required in import data path

### DIFF
--- a/yb-voyager/cmd/importData.go
+++ b/yb-voyager/cmd/importData.go
@@ -1634,6 +1634,10 @@ func createInitialImportDataTableMetrics(tasks []*ImportFileTask) []*cp.UpdateIm
 }
 
 func saveOnPrimaryKeyConflictActionInMSR() {
+	if !isPrimaryKeyConflictIgnoreModeValid() {
+		return
+	}
+
 	metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
 		record.OnPrimaryKeyConflictAction = tconf.OnPrimaryKeyConflictAction
 	})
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/importData.go
```diff
@@ -1643,6 +1647,8 @@ func cleanMSRForImportDataStartClean() error {
 	if !startClean {
 		log.Infof("skipping cleaning migration status record for import data command start clean")
 		return nil
+	} else if !isPrimaryKeyConflictIgnoreModeValid() {
```
makalaaneesh, line 1650: the function name is generic (`cleanMSRForImportDataStartClean` ), so it seems weird that we skip if `!isPrimaryKeyConflictIgnoreModeValid` . 

I'd recommend:
- either change the function name to make it specific(it doesn't do anything other than resetting OnPrimaryKeyConflictAction, so that makes sense)
- keep the function name as-is, but only run 1662-1664 if `isPrimaryKeyConflictIgnoreModeValid`

yb-voyager/cmd/importData.go
```diff
@@ -1660,6 +1666,10 @@ func cleanMSRForImportDataStartClean() error {
 	return nil
 }
 
+func isPrimaryKeyConflictIgnoreModeValid() bool {
```
makalaaneesh, line 1669: `isPrimaryKeyConflictValid` ? 
It's not specific to IGNORE, right? Even in case of ERROR, we'd like to to store and validate, right?

yb-voyager/cmd/importData.go
sanyamsinghal: I actually removed the check from `cleanMSRForImportDataStartClean`.
I don't think this required to be checked, we can erase irrespectively.

makalaaneesh (approved):

PR #2902, opened 2025-07-30
Shivansh/fk indexes

ShivanshGahlot:
Added logic to detect missing indexes on foreign key columns and implemented anonymization for sensitive information in callhome payloads.

**Detection Logic:**
- Detects foreign key constraints that lack corresponding indexes on the FK columns
- **Indexes that DO NOT trigger the issue:**
     - Exact match: FK (x,y,z) → Index (x,y,z) ✓
     - Permutation: FK (x,y,z) → Index (y,z,x) ✓
     - Prefix match: FK (x,y,z) → Index (x,y,z,other) ✓
     - Primary keys and unique constraints are considered as indexes
- **Indexes that DO trigger the issue:**
     - FK columns not at start: FK (x,y,z) → Index (other,x,y,z) ✗
     - Missing columns: FK (x,y,z) → Index (x,y) ✗
     - No index: FK (x,y,z) → No index ✗
 
**Anonymization:**
- Added `anonymizeIssueDetails()` function to anonymize `FK_COLUMN_NAMES` in callhome payloads
- Handles comma-separated FK column names by anonymizing each individually

New Assessment Issues: Users will now see "Missing index on foreign key columns" issues in assessment reports when FK constraints lack proper indexes.

Manually

No


yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -900,74 +889,40 @@ func (p *ParserIssueDetector) DetectMissingForeignKeyIndexes() []QueryIssue {
...
-		if len(current) == len(elements) {
-			// Make a copy of current
-			perm := make([]string, len(current))
-			copy(perm, current)
-			result = append(result, perm)
-			return
+// hasIndexCoverage checks if an index provides coverage for the given foreign key columns.
+// It checks if the FK columns form a subset of the index columns (in any order) and are at the beginning.
+func (p *ParserIssueDetector) hasIndexCoverage(index *queryparser.Index, fkColumns []string) bool {
+	// Extract non-expression columns from the index
```
makalaaneesh, line 909: what happens in case of expression columns? 
I am assuming we will not be considering that index. Is that in-line with what we want to do?

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -900,74 +889,40 @@ func (p *ParserIssueDetector) DetectMissingForeignKeyIndexes() []QueryIssue {
...
+	indexColumns := make([]string, 0)
+	for _, param := range index.Params {
+		if !param.IsExpression {
```
makalaaneesh, line 912: skipping if it is an expression can lead to a false-positive no? 
what if index is on (expr(col1), col2, col3) 
and fk = (col2, col3) 

You might decide that FK has index coverage, but that will not be true?

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -900,74 +889,40 @@ func (p *ParserIssueDetector) DetectMissingForeignKeyIndexes() []QueryIssue {
...
-		}
+	// Check if we have enough columns in the index
+	if len(indexColumns) < len(fkColumns) {
+		return false
 	}
 
-	used := make([]bool, len(elements))
-	backtrack(used, []string{})
-	return result
+	// Check if the FK columns are a subset of the first len(fkColumns) index columns
```
makalaaneesh, line 922: We're checking full set equality, not "subset", right? Can you pls fix the comment?

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -988,41 +943,49 @@ func (p *ParserIssueDetector) SetColumnStatistics(columnStats []utils.ColumnStat
...
+	// Add the index to the table's index list
+	if _, exists := p.tableIndexes[tableName]; !exists {
+		p.tableIndexes[tableName] = make([]*queryparser.Index, 0)
+	}
+	p.tableIndexes[tableName] = append(p.tableIndexes[tableName], index)
+}
+
+// addConstraintAsIndex creates a mock index object from a constraint and adds it to the table indexes map.
+// This is used for primary keys and unique constraints which are also indexes for missing foreign key index detection.
+func (p *ParserIssueDetector) addConstraintAsIndex(tableName string, columns []string, constraintType string) {
```
makalaaneesh, line 959: Nice. I think this is a good idea 👍  

@priyanshi-yb can you pls take a look at this function alone to review that we are populating the struct properly. ?

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -988,41 +943,49 @@ func (p *ParserIssueDetector) SetColumnStatistics(columnStats []utils.ColumnStat
...
+	// Create mock index parameters from columns
+	indexParams := make([]queryparser.IndexParam, len(columns))
+	for i, colName := range columns {
+		indexParams[i] = queryparser.IndexParam{
+			ColName:      colName,
+			IsExpression: false,
```
makalaaneesh, line 965: what if we have expression columns in a PK ? how would that be handled?

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -988,41 +943,49 @@ func (p *ParserIssueDetector) SetColumnStatistics(columnStats []utils.ColumnStat
...
 		}
 	}
 
-	// Generate all prefix combinations and store in coverage map
-	p.addIndexPrefixesToCoverageMap(tableName, columns)
-}
+	// Extract schema name from tableName (format: "schema.table")
+	schemaName := ""
+	tableNameOnly := tableName
```
makalaaneesh, line 971: Instead of figuring that out here, you can pass in the proper schema name, table name from the caller. At the caller function, you have access to `queryparser.Table` type, which has both the fields

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -125,20 +125,9 @@ type ParserIssueDetector struct {
...
-	//   "orders.user_id,product_id,order_date" = true
-	// - Foreign key on (user_id, product_id) is detected as covered
-	// - Primary key on (id) creates entry: "users.id" = true
-	// - Unique constraint on (email, status) creates entries:
-	//   "users.email" = true
-	//   "users.email,status" = true
-	indexPrefixCoverageMap map[string]bool
+	// Indexes stored per table
+	// Key is table name, value is slice of indexes on that table.
+	tableIndexes map[string][]*queryparser.Index
```
makalaaneesh, line 130: Just to elaborate, I was having a discussion with Priyanshi yesterday - 
It will be useful to store it like this (i.e. generic list of indexes) so that in future, if we want to extract any other information out of these indexes, we can re-use the same field. Alternatively, if we stored a very specific set of values here that is only useful for our current task (of checking fk index coverage), that wouldn't be very generically useful in the future.

yb-voyager/src/utils/utils.go
```diff
@@ -983,3 +983,22 @@ func GenerateAnonymisationSalt(n int) (string, error) {
 	// convert bytes to hex to make it printable ASCII string
 	return hex.EncodeToString(b), nil
 }
+
+// IsSetEqual checks if two string slices contain the same elements regardless of order.
+// It sorts both slices and compares them for equality.
+func IsSetEqual(a, b []string) bool {
+	if len(a) != len(b) {
```
makalaaneesh, line 990: you can just use `mapset.NewThreadUnsafeSet[string]` and then `a.Equal(b)` should work here

makalaaneesh (commented):
High level, looks good! 
Just a little unsure about expression indexes, let's double-check that.

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -826,12 +849,174 @@ func (p *ParserIssueDetector) IsUnloggedTablesIssueFiltered() bool {
...
+		schemaName = parts[0]
+		tableNameOnly = parts[1]
+	}
+
+	// Create a mock index object
+	mockIndex := &queryparser.Index{
+		SchemaName:   schemaName,
+		IndexName:    fmt.Sprintf("%s_%s", strings.ToLower(constraintType), strings.Join(columns, "_")),
+		TableName:    tableNameOnly,
+		AccessMethod: BTREE_ACCESS_METHOD, // Primary keys and unique constraints use btree by default
```
makalaaneesh, line 983: i don't think this is true. You can have other index types as well in a PK. Ideally, you should get the access method from the constraint type.

yb-voyager/src/query/queryissue/parser_issue_detector.go
ShivanshGahlot: Added a comment to say that PKs cant have expression columns

yb-voyager/src/query/queryissue/parser_issue_detector.go
ShivanshGahlot: Since it is mentioned in the docs that by default a btree is created, keeping this

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -1126,25 +1126,47 @@ func analyzeSchemaInternal(sourceDBConf *srcdb.Source, detectIssues bool, detect
...
 
+	// Run misc checks after all DDL has been processed
+	if detectIssues {
+		checkerMisc(detectPerfOptimizationIssues)
+
+		// Filter out issues that are already fixed in the target database version
+		// This should happen once after all issues have been collected
+		// Ideally all filtering of issues should happen in queryissue pkg layer,
+		// but until we move all issue detection logic to queryissue pkg, we will filter issues here as well.
+		schemaAnalysisReport.Issues = lo.Filter(schemaAnalysisReport.Issues, func(i utils.AnalyzeSchemaIssue, index int) bool {
```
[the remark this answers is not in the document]
makalaaneesh: Ah I see. dont remember the full context of why it's present here as well, but cool 👍

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -2152,3 +2158,46 @@ func validateSourceDBIOPSForAssessMigration() error {
...
+		return nil
+	}
+
+	anonymizedDetails := make(map[string]interface{})
+
+	for key, value := range details {
+		if key == queryissue.FK_COLUMN_NAMES {
+			if strValue, ok := value.(string); ok && strValue != "" {
+				// Split the string by comma to handle multiple FK columns
+				columnNames := strings.Split(strValue, ",")
```
makalaaneesh, line 2175: nit: we can just store it as a list instead of a string, no? Would be easier to process here.

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -798,12 +821,172 @@ func (p *ParserIssueDetector) IsUnloggedTablesIssueFiltered() bool {
 	return p.isUnloggedTablesIssueFiltered
 }
 
+// GetForeignKeyConstraints returns all foreign key constraints
+func (p *ParserIssueDetector) GetForeignKeyConstraints() []ForeignKeyConstraint {
+	return p.foreignKeyConstraints
+}
+
+// GetTableIndexes returns the table indexes map
```
makalaaneesh, line 829: Can't find where these getters are being used. let's remove them if they are not being used

ShivanshGahlot: Yep residue from older implemetation. Removed.

yb-voyager/cmd/analyzeSchema.go
```diff
@@ -1120,24 +1120,50 @@ func analyzeSchemaInternal(sourceDBConf *srcdb.Source, detectIssues bool, detect
...
-			// Ideally all filtering of issues should happen in queryissue pkg layer,
-			// but until we move all issue detection logic to queryissue pkg, we will filter issues here as well.
-			schemaAnalysisReport.Issues = lo.Filter(schemaAnalysisReport.Issues, func(i utils.AnalyzeSchemaIssue, index int) bool {
-				fixed, err := i.IsFixedIn(targetDbVersion)
-				if err != nil {
-					utils.ErrExit("error checking if analyze issue is supported: issue[%v]: %v", i, err)
-				}
-				return !fixed
-			})
+			if objType == "CONVERSION" {
```
makalaaneesh, line 1123: remove this code (as this was missed in merge)

PR #2996, opened 2025-08-26
Fix: send source password to assess-migration run internally in export schema if its not set via cli flag

priyanshi-yb:
1. Clear migration flags from the migration status record on cleanup.

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -375,9 +375,8 @@ func IsMigrationAssessmentDoneViaExportSchema() (bool, error) {
 
 func ClearMigrationAssessmentDone() error {
 	err := metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
-		if record.MigrationAssessmentDone {
-			record.MigrationAssessmentDone = false
-		}
+		record.MigrationAssessmentDone = false
+		record.MigrationAssessmentDoneViaExportSchema = false
 	})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/exportSchema.go
```diff
@@ -260,6 +265,12 @@ func runAssessMigrationCmdBeforExportSchemaIfRequired(exportSchemaCmd *cobra.Com
 		}
 	}
 
+	if !isSourcePasswordSetViaFlag {
```
makalaaneesh, line 268: why not always pass it? why do we need to check? 
In fact, can we set it in the ENV var just to be on the safe side?

priyanshi-yb: yeah, that we can do. I thought we can explicitly set for the but I like the idea of running it through ENV in all cases as its an internal command we are running.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -375,9 +375,8 @@ func IsMigrationAssessmentDoneViaExportSchema() (bool, error) {
 
 func ClearMigrationAssessmentDone() error {
 	err := metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
-		if record.MigrationAssessmentDone {
-			record.MigrationAssessmentDone = false
-		}
+		record.MigrationAssessmentDone = false
+		record.MigrationAssessmentDoneViaExportSchema = false
```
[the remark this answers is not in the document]
priyanshi-yb: Only state-wise, there were some weird issues. If we run the assessment and export in combinations\

PR #3148, opened 2025-10-23
Error Policy Miscellaneous Changes

makalaaneesh:
1. Changed on-primary-key-conflict option from ERROR -> ERROR-POLICY to make the expected behavior clear. 
Context: The expected behavior is that when user configures ERROR/ERROR-POLICY, the conflict error will be handled as per error-policy (abort or stash-and-continue). The current value (ERROR) can be misleading in case error-policy is stash-and-continue; therefore, we are changing it to ERROR-POLICY. 
2. In case we abort due to any error, we additionally print a msg to the on the console recommending them to use stash-and-continue if they want. 
image

Yes, changes were made to flag options. 

Manually. existing test cases suffice. this is just a UX change. Updated test cases for config files. 

No. 

No. 

### DIFF
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -536,7 +536,7 @@ func validateOnPrimaryKeyConflictFlag() error {
 		}
 	}
 
-	// restrict setting --enable-upsert as true if --on-primary-key-conflict is not set to ERROR
+	// --enable-upsert true and on-primary-key-conflict ignore is conflicting, therefore we only allow it if on-primary-key-conflict is set to ERROR-POLICY
 	if tconf.EnableUpsert && tconf.OnPrimaryKeyConflictAction != constants.PRIMARY_KEY_CONFLICT_ACTION_ERROR_POLICY {
 		return fmt.Errorf("--enable-upsert=true can only be used with --on-primary-key-conflict=ERROR-POLICY")
 	}


yb-voyager/cmd/import.go
```diff
@@ -537,8 +537,8 @@ func validateOnPrimaryKeyConflictFlag() error {
 	}
 
 	// restrict setting --enable-upsert as true if --on-primary-key-conflict is not set to ERROR
-	if tconf.EnableUpsert && tconf.OnPrimaryKeyConflictAction != constants.PRIMARY_KEY_CONFLICT_ACTION_ERROR {
-		return fmt.Errorf("--enable-upsert=true can only be used with --on-primary-key-conflict=ERROR")
+	if tconf.EnableUpsert && tconf.OnPrimaryKeyConflictAction != constants.PRIMARY_KEY_CONFLICT_ACTION_ERROR_POLICY {
+		return fmt.Errorf("--enable-upsert=true can only be used with --on-primary-key-conflict=ERROR-POLICY")
 	}
```
sanyamsinghal: can you help me remember the context here?
and also think with case of stash and continue

sanyamsinghal (approved):
lgtm with one question

yb-voyager/cmd/import.go
makalaaneesh: --enable-upsert true and on-primary-key-conflict ignore would be conflicting, therefore we only allow it if on-primary-key-conflict is set to ERROR-POLICY. added comment in code. 

enable-upsert will eventually be deprecated, so not really thinking much about stash-and-continue and enable-upsert. But either way, enable-upsert=true + error-policy=stash-and-continue just means that there will not be any conflict errors in the first place, so error-policy will kick in for all the rest of the errors. Don't see any confusion here. Lmk if you see any potential confusion with these flags.

PR #3151, opened 2025-10-24
Optimization in export data from source YugabyteDB after cutover

sanyamsinghal:
1. Optimize empty table fetching queries from source YugabyteDB
- instead of querying one by one for each table; query for all tables at once and just fetch/stream rows from server for the result

2. Optimize fetching of unsupported datatypes - user defined ones
- fetch for all table in one go by building query for complete table list
- add intergration test for the private filterUnsupportedUserDefinedDatatypes() method

3. Optimize GetColumnsWithSupportedTypes() to fetch all tables columns info in one go
- instead of N queries/roundtrips to the server

Corner case handled: Handling UDT with qualified names and rest of datatypes(native and extension defined as unqualified to match the unsupported datatypes list)

### DIFF
--- a/yb-voyager/src/srcdb/yugabytedb_test.go
+++ b/yb-voyager/src/srcdb/yugabytedb_test.go
@@ -599,7 +599,7 @@ func TestYugabyteGetColumnsWithSupportedTypes_AllScenarios(t *testing.T) {
 	tableList := []sqlname.NameTuple{
 		testutils.CreateNameTupleWithSourceName("custom_ext.ext_table", "custom_ext", constants.YUGABYTEDB),
 		testutils.CreateNameTupleWithSourceName("hr.employee_devices", "hr", constants.YUGABYTEDB),
-		testutils.CreateNameTupleWithSourceName("custom_ext.SearchTable", "custom_ext", constants.YUGABYTEDB),
+		testutils.CreateNameTupleWithSourceName("custom_ext.\"SearchTable\"", "custom_ext", constants.YUGABYTEDB),
 		testutils.CreateNameTupleWithSourceName("hr.projects", "hr", constants.YUGABYTEDB),
 	}
 


yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -611,74 +611,179 @@ func (yb *YugabyteDB) FilterUnsupportedTables(migrationUUID uuid.UUID, tableList
...
 }
 
-func (yb *YugabyteDB) getTableColumns(tableName sqlname.NameTuple) ([]string, []string, []string, error) {
-	var columns, dataTypes, dataTypesOwner []string
-	sname, tname := tableName.ForCatalogQuery()
-	query := fmt.Sprintf(GET_TABLE_COLUMNS_QUERY_TEMPLATE_PG_AND_YB, tname, sname)
+// tableColumnInfo holds column information for a table
+type tableColumnInfo struct {
+	Columns   []string
+	DataTypes []string
```
priyanshi-yb: instead you can have map column->datatype also right?

priyanshi-yb (commented):
There is one more optimization that can be done for `FilterUnsupportedTables`, where we are also doing a lot of stuff per table

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -689,16 +794,15 @@ func (yb *YugabyteDB) filterUnsupportedUserDefinedDatatypes(tableName sqlname.Na
 			log.Warnf("close rows for query %q: %v", query, closeErr)
 		}
 	}()
+
 	var userDefinedDataTypes []string
 	for rows.Next() {
-		var dataType, isUserDefined string
-		err = rows.Scan(&dataType, &isUserDefined)
+		var qualifiedTypeName string
```
priyanshi-yb: Since you have changed the datatype from unqualified to qualified, can you confirm if the below check of unsupported types is working as expected?

priyanshi-yb: Or alternatively you can return typename only since you are already filtering the types that are present in the schemas of the table list only.

yb-voyager/src/srcdb/yugabytedb.go
sanyamsinghal: @priyanshi-yb map doesn't ensure the order (its random) thats why i chose this as its safer also we have some logic relying on this.

Do you think it will not matter?
I don't have the exact reason but seems risky to me, as we pass this column list to debezium and pg uses this.

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -611,74 +611,179 @@ func (yb *YugabyteDB) FilterUnsupportedTables(migrationUUID uuid.UUID, tableList
...
+		FROM pg_attribute AS a
+		JOIN pg_type AS t ON t.oid = a.atttypid
+		JOIN pg_class AS c ON c.oid = a.attrelid
+		JOIN pg_namespace AS n ON n.oid = c.relnamespace
+		WHERE (n.nspname, c.relname) IN (%s)
+			AND a.attname NOT IN ('tableoid', 'cmax', 'xmax', 'cmin', 'xmin', 'ctid')
+			AND a.attnum > 0
+			AND NOT a.attisdropped
+		ORDER BY n.nspname, c.relname, a.attnum; -- attnum ensures columns appear in table definition order and keeps Columns[] and DataTypes[] arrays aligned deterministically
+	`, inClause)
```
sanyamsinghal: @priyanshi-yb this is the query, ordered by attribute number

yb-voyager/src/srcdb/yugabytedb.go
sanyamsinghal: https://github.com/yugabyte/yb-voyager/pull/3151#discussion_r2476982735

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -611,74 +611,179 @@ func (yb *YugabyteDB) FilterUnsupportedTables(migrationUUID uuid.UUID, tableList
...
+	// Build IN clause AND create reverse lookup map for table name to NameTuple
+	var tableTuples []string
+	// Key: "schema.table" -> Value: original NameTuple
+	tableLookup := make(map[string]sqlname.NameTuple)
+
+	for _, table := range tableList {
+		schema, name := table.ForCatalogQuery()
+		tableTuples = append(tableTuples, fmt.Sprintf("('%s', '%s')", schema, name))
+
+		lookupKey := fmt.Sprintf("%s.%s", schema, name)
```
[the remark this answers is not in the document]
priyanshi-yb: whereever you require a qualified catalog table, so you don't need a  logic of concatenating schema n table anywhere

yb-voyager/src/srcdb/yugabytedb.go
sanyamsinghal: > so even if there are two UDTs with same typename in two diffferent schemas that is okay because it doesn't, we just need whether the table type is a UDT, right?

I think you are mostly right but what if someone defines a UDT with name same as some native type.
Not sure if this is possible but same case like this For eg: custom_schema.int or custom_schema.varchar

It might end up detecting native cols as unsupported.

priyanshi-yb: > Not sure if this is possible but same case like this For eg: custom_schema.int or custom_schema.varchar
>It might end up detecting native cols as unsupported.

Even if that's a possible case - I don't think we were anyways not handling that earlier also so I guess since this is just an optimization PR we can keep it same and handle that separately (but we can file a bug if this is possible scenario and we have some false positives here) .

yb-voyager/src/srcdb/yugabytedb_test.go
```diff
@@ -0,0 +1,698 @@
...
+			search_vector TSVECTOR
+		);`,
+
+		// Case 4: All regular columns (baseline)
+		`CREATE TABLE hr.projects (
+			id INT PRIMARY KEY,
+			name VARCHAR,
+			budget NUMERIC
+		);`,
+	)
```
priyanshi-yb: Please add some case for case-sensitive table/column name

priyanshi-yb (approved):
LGTM

yb-voyager/src/srcdb/yugabytedb.go
sanyamsinghal: @priyanshi-yb Yes we were not handling that and i have already fixed that bug in this PR itself.
It's a very minor thing and no one reported so i will skip ticket for this and add in the commit message.

yb-voyager/src/srcdb/yugabytedb_test.go
priyanshi-yb: I understand that it might not be a problem since we are using the catalog names from nametuples everywhere, but I still feel that one case with the case sensitivity with unsupported datatype is important for verifying that query works as expected with these.

PR #3162, opened 2025-11-03
Optimize FilterUnsupportedTables method of srcdb Yugabyte using single query

sanyamsinghal:
Replace N+M queries for given table list with single SQL query using joins on pg_type and pg_namespace to identify tables with unsupported UDT array columns.

Refer to code comments for more details

TODO: validate the array of domain datatype support before Live Migration GA. cc @shubham-yb 

Added unit tests

NA

NA

---

| Component        


yb-voyager/cmd/exportData.go
```diff
@@ -1600,9 +1600,9 @@ func finalizeTableAndColumnList(finalTableList []sqlname.NameTuple) ([]sqlname.N
 		}
 	}
```
priyanshi-yb: I was thinking - We are querying the source for same information (table columns and udts etc.) again in these two functions FilterUnsupportedTables and GetColumnsWithUnsupportedTypes
so maybe something along the lines
1. Fetch all the table column information like table->columns ( datatype, arrayTypeOrNot )
2. Filter unsupported tables using the above information
3. Filter unsupported columns
what do you think?

yb-voyager/src/srcdb/yugabytedb_test.go
```diff
@@ -696,3 +696,300 @@ func TestYugabyteGetColumnsWithSupportedTypes_AllScenarios(t *testing.T) {
 		assert.Equal(t, 0, len(unsupportedCols.Keys()), "Expected empty unsupported map for offline migration")
 	})
 }
+
+// TestYugabyteFilterUnsupportedTables tests:
+// 1. Tables with arrays of composite types (UDTs) are always unsupported
+// 2. Tables with arrays of enums are unsupported only with YBGrpcConnector and supported with Logical connector
+// 3. Tables with regular arrays (int[], text[], etc.) are supported
+func TestYugabyteFilterUnsupportedTables_BasicScenarios(t *testing.T) {
```
priyanshi-yb: one case if a table having two columns one with udt array type and other with basic natie types. It should still be reported

yb-voyager/cmd/exportData.go
sanyamsinghal: Yeah we could do this as a next step(refactoring) and yes it will be a bit more optimized.

But for now i would like to stick with current implementation, it already tested and an incremental optimization over the older implementation.

yb-voyager/src/srcdb/yugabytedb_test.go
sanyamsinghal: UDT array and native type is covered in case-1 below

Do you mean native array datatype?
If so, i can just modify the case-1 here to have that too

priyanshi-yb: yes

PR #3188, opened 2025-11-17
Fix: handle tables having unique partial index with conflict detection cache

### LINKED ISSUE #3192: Handle partial unique index for live migration streaming  forward path
Jira Link: [DB-19206](https://yugabyte.atlassian.net/browse/DB-19206)

priyanshi-yb:

Fixes the case where a table has a unique constraint/index with a partial predicate e.g. `UNIQUE INDEX (col1) where col2 <> 'abc';`
Fix: Sending all the column values in the before fields of the cdc events to figure if the two events are dealing with same unique column value, and if yes, then treat them as conflicting events and serialise using the conflict Detection cache

fixes #3192 

yes mentioned above

Added test

---

There is a change in events from the exporter to have all before values of all the columns of the tables, so the partial unique index will not be supported with the upgrade if export is running with old and import is with new, which is okay since this was not working before itself so should be okay to start a fresh migration for that.

| Component        


yb-voyager/src/tgtdb/event.go
```diff
@@ -380,10 +380,26 @@ func getMapValuesForQuery(m map[string]*string) []interface{} {
 	return values
 }
 
-func (event *Event) IsUniqueKeyChanged(uniqueKeyCols []string) bool {
-	return event.Op == "u" &&
-		len(uniqueKeyCols) > 0 &&
-		lo.Some(lo.Keys(event.Fields), uniqueKeyCols)
```
makalaaneesh: we may have tested this before, but does our entire conflict detection work properly with case sensitive column names? 
Can we pls add a unit test (either in this or the subsequent PR) for this case.

yb-voyager/cmd/live_migration.go
```diff
@@ -295,7 +295,10 @@ func handleEvent(event *tgtdb.Event, evChans []chan *tgtdb.Event) error {
 			conflictDetectionCache.Put(event)
 		} else { // "i" or "u"
 			conflictDetectionCache.WaitUntilNoConflict(event)
-			if event.IsUniqueKeyChanged(uniqueKeyCols) {
+			if event.IsUniqueKeyPresent(uniqueKeyCols) { // This is just checking if any of the unique key columns are present in the before fields instead of fields since there can be cases where unique key
+				// column is not changed but the unique key is remove the index  because of partial predicate
+				//This condition is not neccessary now as such we can put all the update events in cache if the table has the unique key columns
+				//As this will essentially be true for all such cases as we are checking before fields instead of fields
```
makalaaneesh: let's just remove the IsUniqueKeyPresent because it makes it confusing. 
We seem to be adding to cache if:
1. it's an update
2. it has any unique key columns. 

check 2 is already done above (`if len(uniqueKeyCols) > 0 {`)
so we just need to add all updates as well to cache.

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -226,19 +342,66 @@ func (c *ConflictDetectionCache) eventsConfict(cachedEvent *tgtdb.Event, incomin
...
+			//If uk is not changed but the partial predicate is updated in cached and the same uk with before predicate is inserted in the incoming event then it is a partial conflict
+			//handles UPDATE-INSERT, DELETE-INSERT
 			log.Infof("conflict detected for table %s, column %s, between value of event1(vsn=%d, colVal=%s) and event2(vsn=%d, colVal=%s)",
-				cachedEvent.TableNameTup.ForKey(), column, cachedEvent.Vsn, *cachedEvent.BeforeFields[column], incomingEvent.Vsn, *incomingEvent.Fields[column])
+				cachedEvent.TableNameTup.ForKey(), column, cachedEvent.Vsn, cachedEventBefore, incomingEvent.Vsn, incomingEventAfter)
+			return true
+		case cachedEventBefore == incomingEventBefore:
+			//If two events are operating on same uk then it is a partial conflict
+			//handles UPDATE-UPDATE, DELETE-UPDATE
+			log.Infof("conflict detected for table %s, column %s, between value of event1(vsn=%d, colVal=%s) and event2(vsn=%d, colVal=%s)",
```
makalaaneesh: make sure the logs are different for both the cases so we're able to identify which case exactly it was.

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -226,19 +342,66 @@ func (c *ConflictDetectionCache) eventsConfict(cachedEvent *tgtdb.Event, incomin
...
+		if incomingEvent.Fields[column] != nil {
+			incomingEventAfter = *incomingEvent.Fields[column]
+		}
+
+		switch true {
+		case cachedEventBefore == incomingEventAfter:
+			//If uk column is changes then it is a pure conflict
+			//Handles all cases of UPDATE-UPDATE, UPDATE-INSERT, DELETE-INSERT, DELETE-UPDATE
+
+			//If uk is not changed but the partial predicate is updated in cached and the same uk with before predicate is inserted in the incoming event then it is a partial conflict
```
makalaaneesh: I think "partial conflict" terminology might be confusing :D It's more a potential conflict due to partial indexes.

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -226,19 +342,66 @@ func (c *ConflictDetectionCache) eventsConfict(cachedEvent *tgtdb.Event, incomin
...
+			log.Infof("conflict detected for table %s, column %s, between value of event1(vsn=%d, colVal=%s) and event2(vsn=%d, colVal=%s)",
+				cachedEvent.TableNameTup.ForKey(), column, cachedEvent.Vsn, "nil", incomingEvent.Vsn, "nil")
+			return true
+		}
+
+		if cachedEvent.BeforeFields[column] == nil && incomingEvent.BeforeFields[column] != nil {
+			//if c.before and i.before are nil then it can be a conflict
+			log.Infof("conflict detected for table %s, column %s, between value of event1(vsn=%d, colVal=%s) and event2(vsn=%d, colVal=%s)",
+				cachedEvent.TableNameTup.ForKey(), column, cachedEvent.Vsn, "nil", incomingEvent.Vsn, "nil")
+			return true
```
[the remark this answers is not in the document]
makalaaneesh: let's also document false positives for the two cases.

makalaaneesh (approved):
LGTM!

PR #3224, opened 2025-12-02
Parallel metadata collection for PostgreSQL assessment

ShivanshGahlot:
Refactored PostgreSQL assessment to collect metadata from all nodes (primary + replicas) **in parallel** with real-time progress display, reducing assessment time from sequential to parallel execution.

#### Key Features

**1. Concurrent Execution**

Launches one goroutine per node, all running simultaneously:

```
Sequential (Before):          Parallel (After):
Primary → Replica 1 → Replica 2    Primary   ┐
15 minutes total                   Replica 1 ├─ 5 minutes total
                                   Replica 2 ┘
```

Each goroutine runs the metadata collection script independently and reports progress via channels.

**2. Real-Time Progress Display**

In-place terminal updates showing current stage for each node:

```
Collecting metadata from 3 nodes in parallel...

  ⏳ Primary:                         Collecting table row counts...
  ⏳ Replica 1 (db-replica-1.aws...): Collecting column statistics...
  ⏳ Replica 2 (db-replica-2.aws...): Collecting table data types...
  
  ✓ Primary:                         Complete
  ✓ Replica 1 (db-replica-1.aws...): Complete
  ✓ Replica 2 (db-replica-2.aws...): Complete
```

**Features:**
- ANSI escape codes for in-place updates (no scroll spam)
- Auto-detects stages from script output (pass-through approach)
- Truncates long names to 80 chars to prevent line wrapping
- Adds "Replica N" prefix for clarity

**3. Graceful Error Handling**

Different strategies based on node criticality:

**Primary fails** → Abort immediately (critical data)  
**Replica fails** → Show warning, continue with available nodes  
**All succeed** → Report total node count

Example of partial failure:
```
  ✓ Primary:   Complete
  ⚠ Replica 1: Failed

WARNING: Metadata collection failed on 1 replica(s)
Continuing assessment with data from primary
```
**4. Thread-Safe Coordination**

**Channels:**
- `progressChan` (buffered): Workers → Display goroutine (stage updates)
- `resultChan` (buffered): Workers → Main (final success/failure)
- `displayDone` (unbuffered): Display → Main (completion signal)

**Synchronization:**
- `sync.WaitGroup`: Tracks collection goroutine completion
- `sync.Mutex`: Protects progress map from concurrent writes

**Deadlock Prevention:**
- 2 goroutines per node (stdout/stderr readers) prevent pipe buffer deadlocks
- Scripts can write to both streams; concurrent reading avoids blocking

**5. Code Organization**

Extracted ~580 lines of collection logic from `cmd/` to `src/migassessment/`:

**New file:** `src/migassessment/gather.go`
- `GatherAssessmentMetadataFromPG()` - Exported entry point for PostgreSQL
- `GatherAssessmentMetadataFromOracle()` - Exported entry point for Oracle
- All collection structs, progress tracking, and script execution logic

**Also added:** `Source.GetTNSAdmin()` method in `src/srcdb/source.go` to consolidate TNS_ADMIN logic (used by both assessment and export-data).

**6. Architecture**

```
Main Goroutine
 ├─► Display Goroutine (reads progressChan, updates UI)
 │
 ├─► Node 1 Goroutine ──┬─► stdout reader ─┐
 │                      └─► stderr reader ─┼─► progressChan
 │                                         │
 ├─► Node 2 Goroutine ──┬─► stdout reader ─┤
 │                      └─► stderr reader ─┼─► resultChan
 │                                         │
 └─► Cleanup Goroutine (closes channels)──┘
```

**7. Removed the use of application name**

Application name can be duplicated and is not a reliable way to name the replicas and use in the display for the users. Its better to use host:port instead as that should ideally be unique. 

**Flow:** Workers send updates → Display renders → Cleanup signals completion → Main processes results

#### Code Changes

**Modified:**
- `gatherAssessmentMetadataFromPG()` - orchestrates parallel collection

**New:**
- `runGatherAssessmentMetadataScriptBuffered()` - runs script with buffered I/O
- `detectStageFromOutput()` - parses script output for stages
- `progressTracker` - manages thread-safe display updates
- `collectionNode`, `collectionResult` - data structures

UI and performance changes explained above

Manually

No

No

---

| Component        

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -26,7 +26,6 @@ import (
 	"sort"
 	"strings"
 	"text/template"
-	"time"
 
 	goerrors "github.com/go-errors/errors"
 
@@ -575,8 +574,6 @@ func gatherAssessmentMetadata(validatedReplicas []srcdb.ReplicaEndpoint) error {
 	CreateMigrationProjectIfNotExists(source.DBType, exportDir)
 
 	utils.PrintAndLogf("\ngathering metadata and stats from '%s' source database...\n", source.DBType)
-
-	time.Sleep(10 * time.Second)
 
 	switch source.DBType {
 	case POSTGRESQL:


yb-voyager/src/migassessment/gather.go
```diff
@@ -0,0 +1,581 @@
...
+		return nil, fmt.Errorf("metadata collection failed on primary database (critical): %w", primaryErr)
+	}
+
+	// Print summary
+	if !primaryFailed && len(failedReplicasList) > 0 {
+		fmt.Println() // Blank line before warnings
+		color.Yellow("WARNING: Metadata collection failed on %d replica(s): %v", len(failedReplicasList), failedReplicasList)
+		utils.PrintAndLogfInfo("Continuing assessment with data from primary + %d successful replica(s)", successfulReplicaCount)
+		utils.PrintAndLogfWarning("Note: Sizing and metrics will reflect only the nodes that succeeded.")
+	}
```
makalaaneesh: why just warn and move on?
why not fail?

yb-voyager/src/migassessment/gather.go
```diff
@@ -0,0 +1,581 @@
...
+	return nil
+}
+
+// runGatherAssessmentMetadataScriptBuffered runs the metadata collection script with buffered output
+// and sends progress updates to the provided channel. This version is safe for parallel execution.
+func runGatherAssessmentMetadataScriptBuffered(
+	scriptPath string,
+	envVars []string,
+	nodeName string,
+	displayName string,
```
makalaaneesh: displayName is only the problem of displaying, not of gathering, so it seems to me that storing the display name should be the responsibility of the progressTracker alone, and only the nodeName (or the nodeId essentially) should be passed around everywhere. 


I don't have clarity on this, but let's discuss

yb-voyager/src/migassessment/gather.go
```diff
@@ -0,0 +1,581 @@
...
+			}
+		}(node)
+	}
+
+	// Goroutine to close channels after all collections complete
+	go func() {
+		wg.Wait()
+		close(progressChan)
+		close(resultChan)
+	}()
```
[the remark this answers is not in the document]
ShivanshGahlot: Yeah you are correct. This is more of a residue from an implementation I was trying earlier. The idea was to not block the main goroutine while it proceeds with the other operations. But doesn't seem like it is necessary anymore. Will remove it.

yb-voyager/src/migassessment/gather.go
ShivanshGahlot: Makes sense. Refactored the code around it. Only progressTracker now creates and saves displayName.

yb-voyager/src/migassessment/gather.go
```diff
@@ -0,0 +1,566 @@
...
+		return fmt.Errorf("error creating stderr pipe: %w", err)
+	}
+
+	err = cmd.Start()
+	if err != nil {
+		return fmt.Errorf("error starting gather assessment metadata script: %w", err)
+	}
+
+	// Report starting status
+	if progressChan != nil {
```
makalaaneesh: we seem to be checking `if progressChan != nil` multiple times in this function. 
Can we instead 
- check if once at the beginning. if it's nil, return error. 
- thereafter, we can safely just append onto that channel.

yb-voyager/src/migassessment/gather.go
makalaaneesh: I think it's simpler to just fail. Why complicate it? If it fails, the user gets to know what happend. If they want to exclude that replica, they can then just explicitly specify the ones to include via a config.

yb-voyager/src/migassessment/gather.go
```diff
@@ -0,0 +1,566 @@
...
+	nodes := []collectionNode{
+		{
+			nodeName:      "primary",
+			connectionUri: source.DB().GetConnectionUriWithoutPassword(),
+			isPrimary:     true,
+			replica:       nil, // nil for primary
+		},
+	}
+
+	for i := range validatedReplicas {
```
makalaaneesh: if there are no replicas, it might be better to follow the old path (where we list our all the steps one by one, because "primary"/"replica" terminology will be confusing to the user.

yb-voyager/src/migassessment/gather.go
```diff
@@ -462,11 +469,9 @@ func runGatherAssessmentMetadataScriptBuffered(
 	}
 
 	// Report starting status
-	if progressChan != nil {
-		progressChan <- NodeProgress{
-			NodeName: nodeName,
-			Stage:    "Starting collection...",
-		}
+	progressChan <- NodeProgress{
```
makalaaneesh: since you removed the nil checks everywhere, you should do one nil check at the beginning of the function and return err.

yb-voyager/src/migassessment/gather.go
```diff
@@ -171,22 +170,69 @@ func (pt *progressTracker) printSingleLine(progress NodeProgress) {
...
+// finalizeAndCheckResults processes the final results from all collection goroutines.
+// It logs each result and returns an error if any node failed.
+func (pt *progressTracker) finalizeAndCheckResults(resultChan <-chan collectionResult) error {
+	var failedNodes []string
+	for result := range resultChan {
+		displayName := pt.displayNames[result.nodeName]
+		if result.err == nil {
+			log.Infof("Successfully collected metadata from %s", displayName)
+		} else {
+			log.Errorf("Metadata collection failed on %s: %v", displayName, result.err)
```
makalaaneesh: so, right now, if metadata collection fails for some users, 
- it is just logged 
- on console, we simply print : 'metadata collection failed on one or more nodes" ?

This isn't ideal, because then the user has to take an extra step of looking through the logs to figure out what happened. 

Can we wrap the individual errors into the error we are returning, and print it properly? 

Let's treat this as an enhancement, don't want to block this PR further, as it's a big one.

PR #3225, opened 2025-12-02
Aggregate Query Stats from All Source Nodes in Compare Perf

ShivanshGahlot:

Modified `compare-performance` to aggregate query statistics from **all source nodes** (primary + replicas) instead of only the primary node.

Previously, the command filtered to `WHERE source_node = 'primary'`, discarding replica query data. This resulted in undercounting execution calls and times, providing an incomplete performance comparison.

1. **Removed primary-only filter** in `GetSourceQueryStats()` - now fetches from all nodes
2. **Created `MergeQueryStatsBasedOnQuery()`** function to aggregate same queries across nodes
   - Merges by query text (not queryid, which can differ across nodes)
   - Sums: calls, rows, total time
   - Recalculates: average time
   - Preserves: min/max times
3. **Added comprehensive tests** and documentation

**Example:**
**Before**: Query shows 1000 calls (primary only)  
**After**: Query shows 1800 calls (1000 primary + 500 replica1 + 300 replica2)

## Files Modified

- `src/types/query_stats.go` - New merge function
- `src/migassessment/assessmentDB.go` - Remove filter, call merge
- `src/types/query_stats_test.go` - Test coverage

Users will see query stats from replicas too

Manually

No

No

---

| Component        

### DIFF
--- a/yb-voyager/src/types/query_stats.go
+++ b/yb-voyager/src/types/query_stats.go
@@ -39,9 +39,8 @@ type QueryStats struct {
 // Mirrors the logic of pgss.MergePgStatStatementsBasedOnQuery but operates on QueryStats directly
 // to avoid conversion overhead between types and maintain a clean, direct code path.
 //
-// Merge key: Uses query text (not queryid) because:
-// - QueryIDs can differ for the same query across nodes due to user context, DB OIDs, or PG version differences
-// - Query text is the semantic truth - same text = same query, regardless of queryid
+// Merge key: Uses query text (not queryid).
+// TODO: revisit this in future and see if we can use QueryID instead as ideally it should work.
 func MergeQueryStatsBasedOnQuery(entries []*QueryStats) []*QueryStats {
 	queryMap := make(map[string]*QueryStats)
 


yb-voyager/src/types/query_stats.go
```diff
@@ -27,3 +33,64 @@ type QueryStats struct {
 	MinExecTime     float64 `json:"min_exec_time" db:"min_exec_time"`
 	MaxExecTime     float64 `json:"max_exec_time" db:"max_exec_time"`
 }
+
+// MergeQueryStatsBasedOnQuery merges query stats with the same query text.
+// Used to aggregate stats from multiple source nodes (primary + replicas).
+// Mirrors the logic of pgss.MergePgStatStatementsBasedOnQuery but operates on QueryStats directly
```
makalaaneesh: I don't fully understand why can't we reuse MergePgStatStatementsBasedOnQuery. Let's discuss this

yb-voyager/src/types/query_stats.go
```diff
@@ -27,3 +33,64 @@ type QueryStats struct {
...
+// to avoid conversion overhead between types and maintain a clean, direct code path.
+//
+// Merge key: Uses query text (not queryid) because:
+// - QueryIDs can differ for the same query across nodes due to user context, DB OIDs, or PG version differences
+// - Query text is the semantic truth - same text = same query, regardless of queryid
```
makalaaneesh: let's document that ideally queryID should work, but for now we're using queryText, and that we will revisit this later.

makalaaneesh (approved):
LGTM

PR #3228, opened 2025-12-04
Fix: handle conflict detection cache in case of unique index is only present on some partitions and not on root table

priyanshi-yb:
Fixed the scenario where a UNIQUE CONSTRAINT/INDEX is present on the leaf partitions and not on the root table by adding the unique of all the partitions to the root table.
1. Add the unique column of all partitions to the root table in the table-to-unique-columns config in metadata
2. Removed the entry of partitions from the table-to-unique columns map, since this config is read by importers and we only deal with root table on the importer side.
3. Added test cases

fix: #3226 

yes

added test

---
NO

| Component        

### DIFF
--- a/yb-voyager/src/srcdb/oracle.go
+++ b/yb-voyager/src/srcdb/oracle.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	goerrors "github.com/go-errors/errors"
 	"github.com/google/uuid"
 	"github.com/samber/lo"
 	log "github.com/sirupsen/logrus"
@@ -683,7 +684,7 @@ func (ora *Oracle) GetTableToUniqueKeyColumnsMap(tableList []sqlname.NameTuple) 
 		}
 		tableNameTuple, ok := tableStrToNameTupleMap[tableName]
 		if !ok {
-			return nil, fmt.Errorf("table %s not found in table list", tableName)
+			return nil, goerrors.Errorf("table %s not found in table list", tableName)
 		}
 		cols, ok := result.Get(tableNameTuple)
 		if !ok {
--- a/yb-voyager/src/srcdb/postgres.go
+++ b/yb-voyager/src/srcdb/postgres.go
@@ -28,9 +28,8 @@ import (
 	"strings"
 	"time"
 
-	goerrors "github.com/go-errors/errors"
-
 	"github.com/fatih/color"
+	goerrors "github.com/go-errors/errors"
 	"github.com/google/uuid"
 	"github.com/jackc/pglogrepl"
 	"github.com/jackc/pgx/v5/pgconn"
@@ -922,7 +921,7 @@ func (pg *PostgreSQL) GetTableToUniqueKeyColumnsMap(tableList []sqlname.NameTupl
 		tableName = fmt.Sprintf("%s.%s", schemaName, tableName)
 		tableNameTuple, ok := tableStrToNameTupleMap[tableName]
 		if !ok {
-			return nil, fmt.Errorf("table %s not found in table list", tableName)
+			return nil, goerrors.Errorf("table %s not found in table list", tableName)
 		}
 		cols, ok := result.Get(tableNameTuple)
 		if !ok {
--- a/yb-voyager/src/srcdb/yugabytedb.go
+++ b/yb-voyager/src/srcdb/yugabytedb.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	goerrors "github.com/go-errors/errors"
 	"github.com/google/uuid"
 	"github.com/jackc/pglogrepl"
 	"github.com/jackc/pgx/v5/pgconn"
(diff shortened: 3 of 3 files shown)


yb-voyager/cmd/exportData.go
```diff
@@ -1780,12 +1781,29 @@ func createUpdateExportedRowCountEventList(tableNames []string) []*cp.UpdateExpo
...
+		keyTbl := key.AsQualifiedCatalogName()
+		for _, leaf := range value {
+			keyLeafTbl := leaf.AsQualifiedCatalogName()
+			leafColumns, ok := res[keyLeafTbl]
+			if !ok {
+				continue
+			}
+			//Do not add leaf table key in the map since this config will be read by importer
+			delete(res, keyLeafTbl)
+			res[keyTbl] = append(res[keyTbl], leafColumns...)
```
makalaaneesh: what happens if all the leaf tables have the same set of unique columns. 
They will get added to the list multiple times, with duplicates?

yb-voyager/cmd/exportData.go
```diff
@@ -1780,12 +1781,29 @@ func createUpdateExportedRowCountEventList(tableNames []string) []*cp.UpdateExpo
...
 	res, err := source.DB().GetTableToUniqueKeyColumnsMap(tableList)
 	if err != nil {
 		utils.ErrExit("get table to unique key columns map: %w", err)
 	}
 
+	//Adding all the leaf partitions unique key columns to the root table unique key columns since in the importer all the events only have the root table name
+	leafPartitions.IterKV(func(key sqlname.NameTuple, value []sqlname.NameTuple) (bool, error) {
+		keyTbl := key.AsQualifiedCatalogName()
+		for _, leaf := range value {
+			keyLeafTbl := leaf.AsQualifiedCatalogName()
```
makalaaneesh: can we use terminology of rootTable/leafTable? 
so `rootTableKey` `leafTableKey`. or something like that?

yb-voyager/cmd/exportData.go
```diff
@@ -1780,12 +1781,29 @@ func createUpdateExportedRowCountEventList(tableNames []string) []*cp.UpdateExpo
...
-func saveTableToUniqueKeyColumnsMapInMetaDB(tableList []sqlname.NameTuple) {
+func saveTableToUniqueKeyColumnsMapInMetaDB(tableList []sqlname.NameTuple, leafPartitions *utils.StructMap[sqlname.NameTuple, []sqlname.NameTuple]) {
 	res, err := source.DB().GetTableToUniqueKeyColumnsMap(tableList)
 	if err != nil {
 		utils.ErrExit("get table to unique key columns map: %w", err)
 	}
 
+	//Adding all the leaf partitions unique key columns to the root table unique key columns since in the importer all the events only have the root table name
+	leafPartitions.IterKV(func(key sqlname.NameTuple, value []sqlname.NameTuple) (bool, error) {
+		keyTbl := key.AsQualifiedCatalogName()
```
makalaaneesh: this is a bit error-prone. 
`GetTableToUniqueKeyColumnsMap` is returning `map[string][]string` so the key structure is not fixed. IF we change the implementation of `GetTableToUniqueKeyColumnsMap` in the future, this will break. 

Ideally, it should be the case that: 
1. `GetTableToUniqueKeyColumnsMap` returns `StructMap[NameTuple][]string`
2. then i can directly just operate on a common key.

yb-voyager/cmd/live_migration_integration_test.go
```diff
@@ -1098,3 +1098,193 @@ END $$;`,
...
+	createTableWithUniqueIndexOnlyOnSomeLeafPartitionsSQL := `
+	CREATE TABLE test_schema.test_partitions (
+		id int,
+		name TEXT,
+		region TEXT,
+		branch TEXT,
+		PRIMARY KEY(id, region)
+	) PARTITION BY LIST (region);
+
+	CREATE TABLE test_schema.test_partitions_part1 PARTITION OF test_schema.test_partitions FOR VALUES IN ('London');
```
makalaaneesh: As we were discussing, shall we also add a test case where the leaf partitions are in a different schema?

yb-voyager/cmd/exportData.go
```diff
@@ -1780,16 +1781,54 @@ func createUpdateExportedRowCountEventList(tableNames []string) []*cp.UpdateExpo
...
+			utils.ErrExit("insert table to unique key columns map: %w", err)
+		}
+		return
+	}
+
+
+	//Adding all the leaf partitions unique key columns to the root table unique key columns since in the importer all the events only have the root table name
+	leafPartitions.IterKV(func(rootTable sqlname.NameTuple, value []sqlname.NameTuple) (bool, error) {
+		for _, leafTable := range value {
+			leafColumns, ok := res.Get(leafTable)
```
makalaaneesh: nit: `leafUniqueColumns`

yb-voyager/cmd/exportData.go
```diff
@@ -1780,16 +1781,54 @@ func createUpdateExportedRowCountEventList(tableNames []string) []*cp.UpdateExpo
...
+			if !ok {
+				continue
+			}
+			//Do not add leaf table key in the map since this config will be read by importer
+			res.Delete(leafTable)
+			rootColumns, ok := res.Get(rootTable)
```
makalaaneesh: nit: `rootUniqueColumns`

makalaaneesh (approved):
LGTM

PR #3231, opened 2025-12-08
Fix: expression based unique indexes if present only on partitions and not on root table

priyanshi-yb:

Fixed expression-based unique index query to fetch all the expression-based unique indexes of all the partitions. and return partitioned as having the expression index so that we can do partition by table strategy for that.
Added test cases.
fix: #3229 

yes

added test 

---

No

| Component        

### DIFF
--- a/yb-voyager/src/tgtdb/yugabytedb.go
+++ b/yb-voyager/src/tgtdb/yugabytedb.go
@@ -2223,6 +2223,9 @@ func (yb *TargetYugabyteDB) NumOfLogicalReplicationSlots() (int64, error) {
 	return numOfSlots, nil
 }
 
+// Function returns the table out of tableNames having the expression unique indexes
+// if returnPartitionRootTable is true, it will also check for the partitions of the partitioned table in tableNames and return the root table of the partition having expression unique indexes
+// else it will check for the tables in TableNames that have normal tables and root tables having expression unique indexes
 func (yb *TargetYugabyteDB) GetTablesHavingExpressionUniqueIndexes(tableNames []sqlname.NameTuple, returnPartitionRootTable bool) ([]sqlname.NameTuple, error) {
 	log.Infof("getting leaf table to root table map")
 	//returns a map of catalog leaf table name to catalog root table name


yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -2216,23 +2216,78 @@ func (yb *TargetYugabyteDB) GetTablesHavingExpressionUniqueIndexes(tableNames []
 	}
 
 	query := fmt.Sprintf(`
-SELECT 
-    n.nspname AS schema_name,
+WITH table_list(schema_name, table_name) AS (VALUES %s),
```
makalaaneesh: Question: 
In this PR, we have chosen to change `GetTablesHavingExpressionUniqueIndexes` to return only root tables, not leaf tables. 
In the other PR https://github.com/yugabyte/yb-voyager/pull/3228, we have NOT changed `GetTableToUniqueKeyColumnsMap` . We kept it consistent (i..e to return all the unique columns for all the interested tables as-is), and then remove the leaf partition tables, add the roots in the exportData layer. 

Why different approaches? 
I personally prefer approach2. The API of targetDB (in this case `GetTablesHavingExpressionUniqueIndexes` should properly return all the tables (root/leaf) , and any further changes should probably happen outside of this function in the caller. 


alternatively, if we want to do it inside this API, we should provide a flag so that the caller can decide what they want (i.e. whether to just return root tables, or the specific root/leaf tables)

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -2223,7 +2223,7 @@ func (yb *TargetYugabyteDB) NumOfLogicalReplicationSlots() (int64, error) {
 	return numOfSlots, nil
 }
 
-func (yb *TargetYugabyteDB) GetTablesHavingExpressionUniqueIndexes(tableNames []sqlname.NameTuple, considerPartitions bool) ([]sqlname.NameTuple, error) {
+func (yb *TargetYugabyteDB) GetTablesHavingExpressionUniqueIndexes(tableNames []sqlname.NameTuple, returnPartitionRootTable bool) ([]sqlname.NameTuple, error) {
```
makalaaneesh: pls add documentation as a comment for what `returnPartitionRootTable` does. It's quite complicated because it seems to depend on the input `tableNames` as well.

PR #3241, opened 2025-12-15
Skip multi range or custom range type columns in live migration

priyanshi-yb:

For the built-in MulitRange type or user-defined range type, the export from debezium is as BYTES data, but the value converter we have for BYTES is specifically for the bytea data and doesn't work for these datatypes.https://github.com/yugabyte/yb-voyager/issues/3212

For the XML datatype, it's an unsupported datatype of YB, as inserts don't work that way. We report it as unsupported completely for all workflows, and don't have handling for that in any of the workflows, and there is no way to skip this column in the snapshot part, as there is no way to skip any column via pg_dump. For that, added the `Unsupported XML feature` error msg in non-retryable errors to error out directly for that. https://github.com/yugabyte/yb-voyager/issues/3220

https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/6983/
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/6982

---

| Component        


yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Nit: Since we have added `OR t.typtype = 'r'` to the query, we should update this comment to mention that too.

PR #3247, opened 2025-12-16
Add SAVEPOINT usage detection in migration assessment

ShivanshGahlot:

Detects SAVEPOINT usage in PostgreSQL source databases during `assess-migration` using parser-based detection. When any savepoint-related statement (SAVEPOINT, ROLLBACK TO SAVEPOINT, or RELEASE SAVEPOINT) is detected across all queries, reports a single assessment issue (Impact Level 1) warning about YugabyteDB CDC's known limitation where DML operations rolled back via ROLLBACK TO SAVEPOINT are incorrectly emitted as CDC events, causing data inconsistencies in fall-forward/fall-back workflows.

**Implementation:**
- Added `SavepointDetector` that uses AST parsing to accurately detect savepoint operations
- Uses a callback pattern to update `ParserIssueDetector.isSavepointUsed` flag when detection occurs
- Issue is created in `assessMigrationCommand.go` after all queries are processed, ensuring only one issue is reported
- Added constants for savepoint transaction kinds in `queryparser/constants.go`
- Includes unit tests for both positive and negative detection cases

**Question:** **We are currently marking this as a DML query but it is in actual a Transaction Control Language(TCL) command. We should probably create an object for TCLs and use that. At other similar places too(two phase commit issue) we are using DML tag only. What should we do about this?**

Yes, new migration caveat will be reported
image

Manually

No

No

---

| Component        

### DIFF
--- a/yb-voyager/src/query/queryissue/issues_dml.go
+++ b/yb-voyager/src/query/queryissue/issues_dml.go
@@ -400,7 +400,7 @@ var savepointUsageIssue = issue.Issue{
 	Description: SAVEPOINT_USAGE_ISSUE_DESCRIPTION,
 	Suggestion:  SAVEPOINT_USAGE_ISSUE_SUGGESTION,
 	GH:          "",
-	DocsLink:    "https://docs.yugabyte.com/preview/yugabyte-voyager/known-issues/#savepoint-usage-in-transactions-cdc-limitation",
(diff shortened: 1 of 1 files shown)


makalaaneesh (approved):
high level LGTM. 
Pls have @priyanshi-yb review the details before merging

PR #3272, opened 2026-01-06
Fix: INTERVAL column case-sensitivity in fallback streaming

ShivanshGahlot:
#### Problem
INTERVAL columns were failing during fallback streaming (YugabyteDB → PostgreSQL/YugabyteDB) with the error:
```
Column INTERVAL_POSITIVE not found in table schema
```

**Root Cause**: The code was using `strings.ToUpper(column)` to look up columns in the source schema registry, which only worked for Oracle sources (uppercase identifiers) but failed for PostgreSQL/YugabyteDB sources (lowercase identifiers).

### Solution
- Implement 2-case fallback: try original case first, then uppercase
- Fixes PostgreSQL/YugabyteDB sources while maintaining Oracle support
- Handles quoted identifiers correctly

Users will be able to migration interval type in fall back situations.

Manually

No

No

---

| Component        

### DIFF
--- a/yb-voyager/cmd/live_migration_integration_test.go
+++ b/yb-voyager/cmd/live_migration_integration_test.go
@@ -2618,59 +2618,9 @@ VALUES (INTERVAL '7 years', INTERVAL '120 days', INTERVAL '15 hours', INTERVAL '
 
 	// Validate fallback streaming data consistency
 	// This ensures all INTERVAL columns (lowercase and mixed-case) were correctly replicated
+	// CompareTableData does a full SELECT * comparison of all rows and columns
 	err = lm.ValidateDataConsistency([]string{`test_schema."interval_test"`}, "id")
 	testutils.FatalIfError(t, err, "failed to validate fallback data consistency")
-
-	// Verify specific INTERVAL values were correctly replicated back to source
-	// This explicitly tests all INTERVAL columns (lowercase and mixed-case) during fallback
-	err = lm.WithSourceConn(func(source *sql.DB) error {
-		// Test 1: UPDATE on lowercase INTERVAL column (id=1)
-		var intervalYears string
-		err := source.QueryRow(`SELECT interval_years::text FROM test_schema.interval_test WHERE id = 1`).Scan(&intervalYears)
-		if err != nil {
-			return fmt.Errorf("failed to query interval_years: %w", err)
-		}
-		assert.Contains(t, intervalYears, "10 years", "interval_years (lowercase) should be updated via fallback")
-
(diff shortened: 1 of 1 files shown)


yb-voyager/cmd/live_migration_integration_test.go
```diff
@@ -2488,3 +2488,195 @@ END $$;
...
+VALUES
+	(INTERVAL '1 year', INTERVAL '30 days', INTERVAL '5 hours', INTERVAL '2 months', INTERVAL '1 day 3 hours', 'row1'),
+	(INTERVAL '2 years 6 months', INTERVAL '45 days', INTERVAL '10 hours', INTERVAL '3 months 15 days', INTERVAL '2 days', 'row2'),
+	(INTERVAL '5 years', INTERVAL '90 days', INTERVAL '24 hours', INTERVAL '1 year', INTERVAL '5 hours 30 minutes', 'row3');`,
+		},
+		SourceDeltaSQL: []string{
+			`INSERT INTO test_schema.interval_test (interval_years, interval_days, interval_hours, "IntervalMixed", "INTERVAL_UPPER", name)
+VALUES
+	(INTERVAL '3 years', INTERVAL '60 days', INTERVAL '8 hours', INTERVAL '6 months', INTERVAL '3 days', 'row4');`,
+		},
```
priyanshi-yb: Assuming you will have the variety of I/U/D in your datatypes PR for interval.

yb-voyager/cmd/live_migration_integration_test.go
```diff
@@ -2488,3 +2488,195 @@ END $$;
...
+	testutils.FatalIfError(t, err, "failed to wait for fallback streaming complete")
+
+	// Validate fallback streaming data consistency
+	// This ensures all INTERVAL columns (lowercase and mixed-case) were correctly replicated
+	err = lm.ValidateDataConsistency([]string{`test_schema."interval_test"`}, "id")
+	testutils.FatalIfError(t, err, "failed to validate fallback data consistency")
+
+	// Verify specific INTERVAL values were correctly replicated back to source
+	// This explicitly tests all INTERVAL columns (lowercase and mixed-case) during fallback
+	err = lm.WithSourceConn(func(source *sql.DB) error {
```
priyanshi-yb: What extra are we validating here ? as we already have ValidateDataConsistency for complete data. If not we can remove it also

priyanshi-yb (approved):
LGTM

yb-voyager/cmd/live_migration_integration_test.go
ShivanshGahlot: Yep removed

PR #3278, opened 2026-01-08
Sql Anonymizer: Return error in case of proc/func/trigger/view/matview

makalaaneesh:
There is no handling in sqlAnonymizer for proc/func/trigger/view/matview. Therefore, we are properly throwing an error now, so that any user of sqlAnonymizer will not accidentally send un-anonymized statements. 

No.

Unit tests

No
No
---

| Component        

### DIFF
--- a/yb-voyager/src/anon/sql_anonymizer.go
+++ b/yb-voyager/src/anon/sql_anonymizer.go
@@ -68,6 +68,8 @@ func checkUnsupportedStatementType(parseResult *pg_query.ParseResult) error {
 	}
 
 	// Check generic statements and special cases for unsupported object types
+	// TODO: Some of these statements (ALTER, DROP, RENAME, etc) do not contain the func/proc body in the DDL, so it should be easy to
+	// parse and anonymize, but at the moment, it is not supported.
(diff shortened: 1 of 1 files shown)


yb-voyager/src/anon/sql_anonymizer_pg_objects_test.go
```diff
@@ -1426,3 +957,314 @@ func TestPostgresDDLCornerCases(t *testing.T) {
...
+				for _, raw := range c.raw {
+					if strings.Contains(out, raw) {
+						t.Errorf("raw identifier %q leaked in %s", raw, out)
+					}
+				}
+				// Check that expected prefixes appear
+				for _, pref := range c.prefixes {
+					if !hasTok(out, pref) {
+						t.Errorf("expected prefix %q not found in %s", pref, out)
+					}
```
priyanshi-yb: nit: not required for this test as we don't need to validate anonymization if it fails, right?

makalaaneesh: I had just put it in there because initially i thought some of those will be supported, but in the end, none of those cases ended up being supported. 
Either way, will be useful for later.

PR #3287, opened 2026-01-19
Supporting data migration for the case sensitive PG source schema names

priyanshi-yb:
- changed source / targetConf structs from `Schema string` field to `Schemas  []sqlname.Identifier` and add a new field for old handling of string value from config param `source-db-schema/target-db-schema`
- Modified all usages of source.Schema/tconf.Schema to use list Schemas.
- 

---
metadb - changed SourceDBConf/TargetDbConf/etc..  fields so unmarshalling/marshalling will error out with upgrade scenario.

| Component        

### DIFF
--- a/yb-voyager/src/namereg/namereg.go
+++ b/yb-voyager/src/namereg/namereg.go
@@ -59,7 +59,6 @@ type NameRegistry struct {
 
 	// All schema and table names are in the format as stored in DB catalog.
 	SourceDBSchemaNames       []string
-	SourceDBSchemaIdentifiers []sqlname.Identifier `json:"-"`
 	DefaultSourceDBSchemaName string
 	// SourceDBTableNames has one entry for `DefaultSourceReplicaDBSchemaName` if `Mode` is `IMPORT_TO_SOURCE_REPLICA_MODE`.
 	SourceDBTableNames map[string][]string // nil for `import data file` mode.
@@ -133,6 +132,7 @@ func (reg *NameRegistry) registerNames() (bool, error) {
 		log.Infof("setting default source replica schema name in the name registry: %s", reg.DefaultSourceDBSchemaName)
 		targetSchema := ""
 		if len(reg.params.TargetDBSchema) > 0 {
+			//TODO: handle with oracle/mysql target db schema changes to error out if it doesn't have exactly one
 			targetSchema = reg.params.TargetDBSchema[0]
 		}
 		defaultSchema := lo.Ternary(reg.SourceDBType == constants.POSTGRESQL, reg.DefaultSourceDBSchemaName, targetSchema)
@@ -598,10 +598,10 @@ func SchemaNameMatcher(dbType string ,allSchemas []sqlname.Identifier, schemaCon
 	for _, schema := range allSchemas {
 		unquotedToIdentifierMap[schema.Unquoted] = schema
 	}
-	schemaNames := strings.Split(schemaConfig, ",")
+	configSchemaNames := strings.Split(schemaConfig, ",")
 	var schemaNotFound []string
 	var finalSchemaList []sqlname.Identifier
-	for _, schema := range schemaNames {
+	for _, schema := range configSchemaNames {
 		schemaName, err := matchName("schema", lo.Keys(unquotedToIdentifierMap), schema)
 		var errObj *ErrNameNotFound
 		if err != nil && errors.As(err, &errObj) {
--- a/yb-voyager/cmd/comparePerformanceCommand.go
+++ b/yb-voyager/cmd/comparePerformanceCommand.go
@@ -163,6 +163,7 @@ func validateComparePerfPrerequisites() {
 		utils.ErrExit("No query statistics found in assessment database. Please ensure pg_stat_statements extension was enabled during assess-migration and that workload was executed on the source database.")
 	}
 	
+	//TODO: fix later 
 	tconf.Schemas = sqlname.ParseIdentifiersFromString(dbType, tconf.SchemaConfig, ",")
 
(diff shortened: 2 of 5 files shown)


yb-voyager/cmd/export.go
```diff
@@ -276,21 +278,25 @@ func validateConflictsBetweenTableListFlags(tableList string, excludeTableList s
...
 		utils.ErrExit("Error --source-db-schema flag is not valid for 'MySQL' db type")
 	case ORACLE:
 		if len(schemaList) > 1 {
 			utils.ErrExit("Error single schema at a time is allowed to export from oracle. List of schemas provided: %s", schemaList)
 		}
+		source.Schemas = []sqlname.Identifier{sqlname.NewIdentifier(source.DBType, schemaList[0])}
 	case POSTGRESQL:
 		// In PG, its supported to export more than one schema
-		source.Schema = strings.Join(schemaList, "|") // clean and correct formatted for pg
+		source.Schemas = lo.Map(schemaList, func(s string, _ int) sqlname.Identifier {
```
makalaaneesh: in general, lots of lo.Map() spread everywhere. 
We should think about writing util functions or functions belonging to the right structs.

yb-voyager/src/namereg/namereg.go
```diff
@@ -481,6 +527,34 @@ func (reg *NameRegistry) lookupSourceAndTargetTableNames(tableNameArg string, ig
...
 
+func (reg *NameRegistry) LookupSchemaName(schemaName string) (sqlname.Identifier, error) {
+	var schemaNames []string
+	var dbType string
+	switch reg.params.Role {
+	case SOURCE_DB_EXPORTER_ROLE, SOURCE_DB_IMPORTER_ROLE, SOURCE_REPLICA_DB_IMPORTER_ROLE:
+		schemaNames = reg.SourceDBSchemaNames
+		dbType = reg.SourceDBType
+		if reg.params.Role == SOURCE_REPLICA_DB_IMPORTER_ROLE && dbType == constants.ORACLE {
+			//For oracle source replica , schema may or maynot be the sourceSchemaNames so taking the default source replica schema name
```
makalaaneesh: why not?

makalaaneesh:
@priyanshi-yb  on the lo.Map: 

See if this makes sense??

image
image
image

yb-voyager/cmd/common.go
```diff
@@ -286,7 +286,7 @@ func displayExportedRowCountSnapshot(snapshotViaDebezium bool) {
 	leafPartitions := getLeafPartitionsFromRootTable()
 	if !snapshotViaDebezium {
 		exportedRowCount := getExportedRowCountSnapshot(exportDir)
-		if source.Schema != "" {
+		if source.SchemaConfig != "" {
```
makalaaneesh: nit: although they're equivalent, we should make it a point to only use Schemas (identifiers) everywhere rather than SchemaConfig. SchemaCOnfig should only be used when parsing the config input into Schemas(identifiers)

yb-voyager/src/srcdb/oracle.go
```diff
@@ -196,13 +202,13 @@ func (ora *Oracle) GetAllTableNamesRaw(schemaName string) ([]string, error) {
 
 func (ora *Oracle) GetAllTableNames() []*sqlname.SourceName {
 	var tableNames []*sqlname.SourceName
-	tableNamesRaw, err := ora.GetAllTableNamesRaw(ora.source.Schema)
+	tableNamesRaw, err := ora.GetAllTableNamesRaw(ora.source.Schemas[0].Unquoted)
```
makalaaneesh: everywhere where we're using [0], are we sure that there will at least be a single schema in the list? Are we validating it?

yb-voyager/cmd/importSchema.go
```diff
@@ -93,14 +95,17 @@ var invalidTargetIndexesCache map[string]bool
 
 func importSchema() error {
 
-	tconf.Schema = strings.ToLower(tconf.Schema)
+	//TODO: see if namreg initialization is required here
+	tconf.Schemas = lo.Map(strings.Split(tconf.SchemaConfig, ","), func(s string, _ int) sqlname.Identifier {
```
makalaaneesh: looks like this will have to be done in multiple places (i.e. importdata, importschema). 
Can we write a single ParseSchemas method of sorts?

yb-voyager/cmd/exportData.go
```diff
@@ -292,23 +292,29 @@ func exportData() bool {
...
-	if !res {
-		utils.ErrExit("schema does not exist : %q", source.Schema)
+	clearMigrationStateIfRequired()
+
+	err = InitNameRegistry(exportDir, exporterRole, &source, source.DB(), nil, nil, false)
+	if err != nil {
+		utils.ErrExit("initialize name registry: %w", err)
+	}
+	validatedSchemas := []sqlname.Identifier{}
+	for _, schema := range source.Schemas {
```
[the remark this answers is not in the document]
priyanshi-yb: source.Schemas is populating in the very starting of the command in `validateSourceSchema` but its not with properly casing as user are allowed to pass in any casing so we are doing the lookups to fetch a proper list and then  using it further

yb-voyager/src/namereg/namereg.go
```diff
@@ -59,6 +59,7 @@ type NameRegistry struct {
 
 	// All schema and table names are in the format as stored in DB catalog.
 	SourceDBSchemaNames       []string
+	SourceDBSchemaIdentifiers []sqlname.Identifier `json:"-"`
```
makalaaneesh: let's also do this for target, source-replica

PR #3335, opened 2026-01-27
Enhance automation framework to work with case sensitive db and runni…

priyanshi-yb:
…ng pg/case-sensitive-reserved-words test with case sensitive db

---

| Component        


migtests/tests/pg/partitions/env.sh
shubham-yb: Should we add this to the case sensitive test case instead of partitions?

migtests/tests/pg/partitions/cleanup-db
shubham-yb: Is this from the previous change?

PR #3365, opened 2026-02-11
Changed the use-grpc-connector-flag to default false in the config files

ShivanshGahlot:
Changed the default values from true to false for the use-yb-grpc-connector flag in the config file templates.
Also modified the CLI help message for the same flag.

The default value in config file template and the help message

Manually

No

No

---

| Component        


makalaaneesh (approved):
LGTM

PR #3367, opened 2026-02-11
Use config file for the commands started after cutover to target

priyanshi-yb:

---

| Component        

### DIFF
--- a/yb-voyager/cmd/config.go
+++ b/yb-voyager/cmd/config.go
@@ -18,7 +18,7 @@ import (
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
 )
 
-var configurationDetails ConfigurationDetails
+var resolvedConfig ResolvedConfig
 
 const (
 	// Flag name prefixes (used in CLI flags)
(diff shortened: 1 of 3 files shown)


makalaaneesh (commented):
Let's discuss this logic once in stand-up along with Shivansh

yb-voyager/cmd/exportData.go
```diff
@@ -1618,6 +1576,40 @@ func startFallBackSetupIfRequired() {
...
+	cmd := []string{"yb-voyager", "import", "data", "to", "source"}
+	if utils.DoNotPrompt {
+		cmd = append(cmd, "--yes")
+	}
+
+	// If config file is provided, pass it to the command
+	if cfgFile != "" {
+		//If there are cli overrides for the command, pass them as cli overrides to the export data from target command
+		//Only disable-pb and log-level are the common flags of both the commands
+		for _, override := range cliOverrides {
```
ShivanshGahlot: So essentially the config file branch of the if else explicitly sets and passes global flags and the disable-pb flag.
And the no config branch sets log-level, export-dir, send-diagnostics and disable-pb.
Right?

ShivanshGahlot: Maybe we can add a comment explaining how this is working in both the branches?

ShivanshGahlot: Also log-level is sent via global flags in the config branch right?

yb-voyager/cmd/importData.go
```diff
@@ -389,48 +389,58 @@ func startExportDataFromTargetIfRequired() {
...
+		}
+	} else {
+		//else set some overrides for the command
+		cmd = append(cmd, "--log-level", config.LogLevel)
+		cmd = append(cmd, "--export-dir", exportDir)
+		if bool(disablePb) {
+			cmd = append(cmd, "--disable-pb=true")
+		}
+		cmd = append(cmd, fmt.Sprintf("--send-diagnostics=%t", callhome.SendDiagnostics))
 	}
```
ShivanshGahlot: This logic is more or less same as the `generateImportDataToSourceCommand` right? If so we can probably create a common function for easier management of the logic? Something like `appendGlobalFlagsToChildCmd` or `forwardGlobalFlagsToCmd`

yb-voyager/cmd/config.go
```diff
@@ -613,14 +621,14 @@ bindCobraFlagsToViper binds configuration values from a Viper instance to the fl
...
+	 6. Also returns a slice of Override structs, which represent the flags that were set from the config file. Should only be used if there are no errors.
 
 	This function allows users to configure flags through the config file or environment variables,
 	while still letting command-line input take precedence.
 */
-func bindCobraFlagsToViper(cmd *cobra.Command, v *viper.Viper) ([]ConfigFlagOverride, error) {
+func bindCobraFlagsToViper(cmd *cobra.Command, v *viper.Viper) error {
 	var bindErr error
-	var overrides []ConfigFlagOverride
+	var configOverrides, cliOverrides []CommandConfig
```
ShivanshGahlot: Nit: We should probably change the names here too to match with the global structures.

ShivanshGahlot (approved):
LGTM! Thanks!

PR #3381, opened 2026-02-18
Increase adaptive-parallelism-max default from totalCores/2 to totalCores and improve core detection fallback

ShivanshGahlot:

- Double `--adaptive-parallelism-max `default from **totalCores/2** to **totalCores** (Parallelism * 2 -> Parallelism * 4)
- Improve core detection for load-balanced setups: when `fetchCores` succeeds through a load balancer, it only reaches one node, so extrapolate to the full cluster using `nodeCount` from `yb_servers()`
- When core detection fails (YBM/managed clusters, permission issues), estimate **totalCores = nodeCount * YB_DEFAULT_CORES_PER_NODE (16)** using node count from `yb_servers()` instead of the **old len(tconfs) * 2** fallback which collapsed to 2 on **YBM** due to load balancer
- Extract `setDefaultParallelism` and split `fetchDefaultParallelJobs` into target-specific methods on `TargetYugabyteDB` and `TargetPostgreSQL` (PG behavior unchanged)
- Update `--parallel-jobs` and `--adaptive-parallelism-max` flag descriptions

- Default `--adaptive-parallelism-max` is now `totalCores` instead of `totalCores/2`. The adaptive algorithm still self-regulates via CPU/memory thresholds.
- On **YBM** (3-node, 4 vCPU default): defaults improve from start=2, cap=4 to start=12, cap=48
- For self-hosted behind a load balancer where core detection succeeds, the detected single-node cores are now correctly extrapolated to the full cluster
- Help text for `--parallel-jobs` and `--adaptive-parallelism-max` updated

Manually

No

No

---

| Component        

---

> [!NOTE]
> **Medium Risk**
> Changes default connection/parallelism sizing during import, which can impact cluster load and import stability/performance, especially on managed or load-balanced deployments.
> 
> **Overview**
> Adjusts import-time parallelism defaults for YugabyteDB targets: when adaptive parallelism is enabled and `--adaptive-parallelism-max` is unset, the cap now defaults to *total cores* (vs. half cores), and the default `--parallel-jobs` calculation now estimates cores as `nodeCount * 4` when core detection fails (using the full node list from `yb_servers()` even if a load balancer is used).
> 
> Refactors default parallelism selection into target-specific implementations (`TargetYugabyteDB.fetchDefaultParallelJobs` and `TargetPostgreSQL.fetchDefaultParallelJobs`), updates CLI help text to reflect the new behavior, and updates the expected callhome payload fixture to the new `adaptive_parallelism_max` value.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5a94aafe0136afaeb47c749af58cc57a31e5f627. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -463,7 +463,7 @@ func registerFlagsForTarget(cmd *cobra.Command) {
 	cmd.Flags().IntVar(&tconf.Parallelism, "parallel-jobs", 0,
 		"number of parallel jobs to use while importing data. By default, voyager will try to determine the total "+
 			"number of cores N across the cluster and use N/4 as parallel jobs. "+
-			"If core detection fails, it estimates N using the number of nodes * 4 vCPUs per node. "+
+			"If core detection fails, it estimates N using the number of nodes * 16 assumed vCPUs per node. "+
 			"Any value less than 1 reverts to the default calculation.")
 
 	cmd.Flags().Var(&tconf.AdaptiveParallelismMode, "adaptive-parallelism",


makalaaneesh:
@cursor review

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -243,7 +256,7 @@ func (pg *TargetPostgreSQL) InitConnPool() error {
 	log.Infof("targetUriList: %s", utils.GetRedactedURLs(targetUriList))
 
 	if pg.tconf.Parallelism == 0 {
-		pg.tconf.Parallelism = fetchDefaultParallelJobs(tconfs, PG_DEFAULT_PARALLELISM_FACTOR)
+		pg.tconf.Parallelism = pg.fetchDefaultParallelJobs(tconfs)
```
makalaaneesh: multiple tconfs does not make sense for pg. Can we simplify the logic?

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -305,14 +305,15 @@ func (yb *TargetYugabyteDB) InitConnPool() error {
 	}
 	log.Infof("targetUriList: %s", utils.GetRedactedURLs(targetUriList))
 
+	nodeCount := len(confs)
```
makalaaneesh: I'm not entirely sure this will work for ybm because of how the logic is written in `getTargetConfsAsPerLoadBalancerUsed`. Recommend testing this out

yb-voyager/cmd/import.go
```diff
@@ -461,9 +461,9 @@ func registerFlagsForTarget(cmd *cobra.Command) {
 	cmd.Flags().Int64Var(&batchSizeInNumRows, "batch-size", 0,
 		fmt.Sprintf("Size of batches in the number of rows generated for ingestion during import. default(%d)", DEFAULT_BATCH_SIZE_YUGABYTEDB))
 	cmd.Flags().IntVar(&tconf.Parallelism, "parallel-jobs", 0,
-		"number of parallel jobs to use while importing data. By default, voyager will try if it can determine the total "+
-			"number of cores N and use N/4 as parallel jobs. "+
-			"Otherwise, it fall back to using twice the number of nodes in the cluster. "+
+		"number of parallel jobs to use while importing data. By default, voyager will try to determine the total "+
+			"number of cores N across the cluster and use N/4 as parallel jobs. "+
+			"If core detection fails, it estimates N using the number of nodes * 4 vCPUs per node. "+
```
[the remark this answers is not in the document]
ShivanshGahlot: I checked the default config for YBAeon.

> The size of the cluster is based on the number of vCPUs. The default configuration for YugabyteDB Aeon clusters includes 4 vCPUs per node. Each vCPU comes with 50GB of storage. A node has a minimum of 2 vCPUs with 4GB of memory per vCPU. For the cluster to be [fault tolerant](https://docs.yugabyte.com/stable/yugabyte-cloud/cloud-basics/create-clusters-overview/#fault-tolerance), you need a minimum of 3 nodes.

This is why I chose 4.

By default 3 nodes and 4 cores per node are suggested. With this in mind calculations lead to this:
MaxParallelism: 12 (N)
Starting Paralllelism: 3 (N/4)

yb-voyager/src/tgtdb/postgres.go
ShivanshGahlot: Yep

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1456,20 +1460,17 @@ func fetchCores(tconfs []*TargetConf) (int, error) {
 	return totalCores, nil
 }
 
-func fetchDefaultParallelJobs(tconfs []*TargetConf, defaultParallelismFactor int) int {
+func (yb *TargetYugabyteDB) fetchDefaultParallelJobs(tconfs []*TargetConf, nodeCount int) int {
```
makalaaneesh: Can you add a comment here to explain what the difference between tconfs and nodeCount is. It's a bit weird that we're passing both tconfs and nodeCount because on the face of it, it seems like you could derive nodeCount from len(tconfs)

makalaaneesh: cursor: > When a load balancer is used, tconfs has only 1 entry (the LB URI). fetchCores(tconfs) connects to the LB and gets cores from whichever single node it routes to -- not the cluster total. So for a 3-node, 8-core cluster, the "success" path yields 8 / 4 = 2 parallel jobs instead of 24 / 4 = 6. The fallback path (with nodeCount * 4) actually gives a better estimate in this scenario.

I think this makes sense. If load-balancer is used, we would only fetch cores from one node, and return totalCores/4

ShivanshGahlot: Done

ShivanshGahlot: Added a case for Load Balancer as discussed.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1460,17 +1446,61 @@ func fetchCores(tconfs []*TargetConf) (int, error) {
...
+	coresFetched := err == nil
+
+	switch {
+	case coresFetched && !loadBalancerUsed:
+		// Case 1: No load balancer, use detected cores directly
+		clusterCores = detectedCores
+	case coresFetched && loadBalancerUsed:
+		// Case 2: fetchCores only reached one node via load balancer; extrapolate to the full cluster
+		clusterCores = detectedCores * nodeCount
+		log.Warnf("Load balancer detected: scaling single-node cores to cluster: %d cores/node * %d nodes = %d clusterCores",
```
makalaaneesh: It's not like a failure/fall-back scenario, so shouldn't be a warn imo.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1460,17 +1446,61 @@ func fetchCores(tconfs []*TargetConf) (int, error) {
...
+func (yb *TargetYugabyteDB) fetchDefaultParallelJobs(tconfs []*TargetConf, nodeCount int, loadBalancerUsed bool) int {
+	var clusterCores int
+	detectedCores, err := fetchCores(tconfs)
+	coresFetched := err == nil
+
+	switch {
+	case coresFetched && !loadBalancerUsed:
+		// Case 1: No load balancer, use detected cores directly
+		clusterCores = detectedCores
+	case coresFetched && loadBalancerUsed:
```
makalaaneesh: @priyanshi-yb just need a confirmation on this scenario. We seem to be making an assumption that if `loadBalancerUsed` we would have only fetched cores from only one node (i.e. only one node is accessible)

PR #3388, opened 2026-02-26
Data migration report support for multiple iterations to aggregate stats and provideoption for detailed report

priyanshi-yb:

Aggregated report 
Screenshot 2026-02-26 at 8 05 16 PM

Detailed report
Screenshot 2026-03-02 at 11 28 35 AM

---

| Component        

### DIFF
--- a/yb-voyager/cmd/getDataMigrationReportCommand.go
+++ b/yb-voyager/cmd/getDataMigrationReportCommand.go
@@ -42,7 +42,7 @@ var sourceReplicaDbPassword string
 var sourceDbPassword string
 var nameRegistryForSourceReplicaRole *namereg.NameRegistry
 
-var detailedReport utils.BoolStr
+var includeDetailedIterationsStats utils.BoolStr
 
 var getDataMigrationReportCmd = &cobra.Command{
 	Use:   "data-migration-report",
@@ -57,12 +57,12 @@ var getDataMigrationReportCmd = &cobra.Command{
 			utils.ErrExit("error while getting migration status: %w\n", err)
 		}
 		if migrationStatus.LatestIterationNumber == 0 {
-			if detailedReport {
+			if includeDetailedIterationsStats {
 				utils.ErrExit("Error: Detailed report is only applicable for multiple iterations of Live migration with fallback workflow")
 			}
 		}
 		if migrationStatus.FallForwardEnabled {
-			if detailedReport {
+			if includeDetailedIterationsStats {
 				utils.ErrExit("Error: Detailed report is only applicable for multiple iterations of Live migration with fallback workflow")
 			}
 		}
@@ -94,7 +94,7 @@ var getDataMigrationReportCmd = &cobra.Command{
 				utils.ErrExit("initializing name registry: %w", err)
 			}
 			color.Yellow("Generating data migration report for migration UUID: %s...\n", migrationStatus.MigrationUUID)
-			getDataMigrationReportCmdFn(migrationStatus, false)
+			getDataMigrationReportCmdFn(migrationStatus, false, true)
 		} else {
 			utils.ErrExit("Error: Data migration report is only applicable when export-type is 'snapshot-and-changes'(live migration)\nPlease run export data status/import data status commands.")
 		}
@@ -138,7 +138,7 @@ type rowDataForIteration struct {
 	ImportedDeletes int64  `json:"imported_deletes"`
 }
 
-func getDataMigrationReportCmdFn(msr *metadb.MigrationStatusRecord, donotPrint bool) {
+func getDataMigrationReportCmdFn(msr *metadb.MigrationStatusRecord, donotPrint bool, aggregateIterationsStats bool) {
 
 	var reportData []*rowData
 	fBEnabled = msr.FallbackEnabled
@@ -322,9 +322,11 @@ func getDataMigrationReportCmdFn(msr *metadb.MigrationStatusRecord, donotPrint b
 		}
 	}
 
-	reportData, err = aggregateDataWithIterationsIfRequired(reportData, msr)
-	if err != nil {
-		utils.ErrExit("error while aggregating data with iterations: %w", err)
+	if aggregateIterationsStats {
+		reportData, err = aggregateDataWithIterationsIfRequired(reportData, msr)
+		if err != nil {
+			utils.ErrExit("error while aggregating data with iterations: %w", err)
+		}
 	}
 
 	isIteration := msr.IterationNo > 0
@@ -360,7 +362,7 @@ func generateReportInJsonFormat(reportData []*rowData, msr *metadb.MigrationStat
 		}
 	}
 	if !isIteration && msr.LatestIterationNumber > 0 && !donotPrint {
-		if detailedReport {
+		if includeDetailedIterationsStats {
 			utils.PrintAndLogfPhase("\nDetailed data migration report for all iterations:")
 		} else {
 			utils.PrintAndLogfPhase("\nAggregated Data migration report for the overall migration:")
@@ -388,7 +390,7 @@ func printReport(forIteration bool, statsPerTable map[string][]*rowData, msr *me
 	printHeader(uitbl, forIteration)
 
 	if !forIteration && msr.LatestIterationNumber > 0 {
-		if detailedReport {
+		if includeDetailedIterationsStats {
 			utils.PrintAndLogfPhase("\nDetailed data migration report for all iterations:")
 		} else {
 			utils.PrintAndLogfPhase("\nAggregated Data migration report for the overall migration:")
@@ -417,7 +419,7 @@ func printReport(forIteration bool, statsPerTable map[string][]*rowData, msr *me
 				row.TableName = ""
 			}
 
-			if detailedReport && row.IterationNumber != lastIterationNumber {
+			if includeDetailedIterationsStats && row.IterationNumber != lastIterationNumber {
 				uitbl.AddRow()
 			}
 			addRowToTable(uitbl, row, forIteration)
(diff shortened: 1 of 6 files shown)


yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -295,48 +310,292 @@ func getDataMigrationReportCmdFn(msr *metadb.MigrationStatusRecord) {
...
+		fmt.Print(color.GreenString("Data migration report is written to %s\n", reportFilePath))
+
+	} else {
+		printReport(isIteration, reportData)
+	}
+
+}
+
+func printReport(forIteration bool, reportData []*rowData) {
+	msr, err := metaDB.GetMigrationStatusRecord()
```
makalaaneesh: can we pass the msr into this function instead of re-fetching?

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -448,6 +714,9 @@ func init() {
 	getDataMigrationReportCmd.Flags().StringVar(&reportOrStatusCmdOutputFormat, "output-format", "table",
 		"format in which report will be generated: (table, json) (default: table)")
 
+	BoolVar(getDataMigrationReportCmd.Flags(), &detailedReport, "all-iterations", false,
```
makalaaneesh: do we have a guardrail for erroring out if this is passed when: 
- there are no iterations
- fall-forward?

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -295,48 +310,292 @@ func getDataMigrationReportCmdFn(msr *metadb.MigrationStatusRecord) {
...
+	return reportData
 }
 
-func addRowInTheTable(uitbl *uitable.Table, row rowData, nameTup sqlname.NameTuple) {
-	if reportOrStatusCmdOutputFormat == "json" {
-		row.TableName = nameTup.ForKey()
-		row.FinalRowCount = getFinalRowCount(row)
-		reportData = append(reportData, &row)
-		return
+func getIterationDataMigrationReport(iterationExportDir string) ([]*rowData, error) {
```
makalaaneesh: Is there a way we can do this in-process? Just call getDataMigrationReportCmdFn (or some other reusable function that we can separate it out into)? 

The multiple processess thing is a bit error-prone and we'll need to have process shutdown handling, etc, etc. For example, if you ctrl-c this process, all those processes should get shut-down as well. (Maybe we already have this because of the exit handlers)

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -295,48 +310,292 @@ func getDataMigrationReportCmdFn(msr *metadb.MigrationStatusRecord) {
...
+	// Invoke the assess-migration command as a subprocess
+	cmd := exec.Command(voyagerExecutable, append([]string{"get", "data-migration-report"},
+		"--export-dir", iterationExportDir,
+		"--output-format", "json")...)
+	cmd.Stdout = &stdoutBuf
+	cmd.Stderr = &stderrBuf
+
+	cmd.Env = os.Environ()
+	cmd.Env = append(cmd.Env, "SOURCE_DB_PASSWORD="+sourceDbPassword)
+	cmd.Env = append(cmd.Env, "TARGET_DB_PASSWORD="+targetDBPassword)
```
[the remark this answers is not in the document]
makalaaneesh: pls test this manually with proper user-passwords (you can take Abhinay's help). If i remember correctly, even our test suite uses default passwords?

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -295,48 +318,318 @@ func getDataMigrationReportCmdFn(msr *metadb.MigrationStatusRecord) {
...
-		fmt.Print("\n")
+
+	if detailedReport {
+		//group the report data by table name and calcutate cumulative row count for iteration
+		reportData = groupByTableNameAndCalculateCumulativeRowCount(reportData)
 	}
+	return reportData, nil
+}
 
+func groupByTableNameAndCalculateCumulativeRowCount(reportData []*rowData) []*rowData {
```
makalaaneesh: some more issues by cursor (which i kindof agree): We can discuss which ones to pick up

1. Relies on row ordering within each map value, but that order is fragile

The accumulators assume that iteration-0 rows come before iteration-1 rows, which come before iteration-2 rows, etc. The input slice has this order because aggregateDataWithIterationsIfRequired first sets iteration-0 on the main data, then appends iteration-1 data, then iteration-2, etc. And append to the map preserves per-key insertion order.

So this happens to work — but it's an implicit contract. If anyone reorders reportData upstream, or if a table appears in iteration data but not main data, the accumulators start at 0 and produce wrong cumulative counts. There's no sort or validation enforcing the ordering.

2. Assumes exactly two DBTypes: "source" and "target"

The else branch on line 534/543 catches everything that isn't "source" — including "source-replica" if it were present. A "source-replica" row would be accumulated into finalRowCountTgt, which is semantically wrong. Currently safe because --all-iterations is blocked for fall-forward workflows, but it's a latent bug.

3. Map iteration order makes the output non-deterministic

Line 527 (for _, rows := range reportDataMap) iterates over a Go map. The order of tables in the output slice changes on every run. This affects:

The JSON output (rows appear in random table order)
Though not the table output (since printReport re-sorts by table name downstream)
If you care about stable JSON output (e.g., for diffing reports), this is a problem.

4. Grouping key is TableName, but rows within a table have mixed DBType

The grouping is only by TableName, not by TableName+DBType. So within each group, you have interleaved source/target rows. The accumulators handle this with the if row.DBType == "source" / else branch, which works, but it means the two accumulators are advanced at alternating steps. If a table ever had two consecutive "source" rows for the same iteration (a bug elsewhere), the cumulative count would silently double-count.

5. Iteration-0 row's FinalRowCount includes snapshot + changes

The seed value (row.FinalRowCount for iter=0) already includes snapshot rows + change events from the initial migration. Then for each iteration, only the net change (inserts - deletes) is added. This is correct — but Updates are intentionally excluded from the cumulative count (updates don't change row count). Worth confirming that's the intended semantics, since the column header says "CUMULATIVE FINAL_ROW_COUNT."

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -295,48 +318,318 @@ func getDataMigrationReportCmdFn(msr *metadb.MigrationStatusRecord) {
...
+	}
+	if iterationMsr.TargetDBConf != nil {
+		iterationMsr.TargetDBConf.Password = targetDBPassword
+	}
+	if iterationMsr.FallbackEnabled {
+		iterationMsr.SourceDBAsTargetConf.Password = sourceDbPassword
+	}
+	migrationUUID = uuid.MustParse(iterationMsr.MigrationUUID)
+
+	getDataMigrationReportCmdFn(iterationMsr, true)
```
makalaaneesh: cursor: getIterationDataMigrationReport calls getDataMigrationReportCmdFn(iterationMsr, true), which will itself call aggregateDataWithIterationsIfRequired. For iteration MSRs, LatestIterationNumber should be 0, so it returns early. However, if an iteration's MSR were ever misconfigured with LatestIterationNumber > 0, this would recurse, potentially corrupting global state or looping. There's no guard against this.

Recommendation: Add a guard or pass a parameter to prevent recursion (e.g., a skipAggregation bool parameter).

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -448,12 +748,15 @@ func init() {
...
 		"password with which to connect to the target Source-Replica DB server. Alternatively, you can also specify the password by setting the environment variable SOURCE_REPLICA_DB_PASSWORD. If you don't provide a password via the CLI, yb-voyager will prompt you at runtime for a password. If the password contains special characters that are interpreted by the shell (for example, # and $), enclose the password in single quotes.")
 
 	getDataMigrationReportCmd.Flags().StringVar(&sourceDbPassword, "source-db-password", "",
 		"password with which to connect to the target source DB server. Alternatively, you can also specify the password by setting the environment variable SOURCE_DB_PASSWORD. If you don't provide a password via the CLI, yb-voyager will prompt you at runtime for a password. If the password contains special characters that are interpreted by the shell (for example, # and $), enclose the password in single quotes")
 
-	getDataMigrationReportCmd.Flags().StringVar(&targetDbPassword, "target-db-password", "",
+	getDataMigrationReportCmd.Flags().StringVar(&targetDBPassword, "target-db-password", "",
```
makalaaneesh: why this change?

yb-voyager/cmd/getDataMigrationReportCommand.go
```diff
@@ -448,12 +748,15 @@ func init() {
 	getDataMigrationReportCmd.Flags().StringVar(&reportOrStatusCmdOutputFormat, "output-format", "table",
 		"format in which report will be generated: (table, json) (default: table)")
 
+	BoolVar(getDataMigrationReportCmd.Flags(), &detailedReport, "all-iterations", false,
```
makalaaneesh: nit: `--include-detailed-iterations-data` ? 
Also, let's name the variable similar to what the flag is (makes it easier to correlate)

priyanshi-yb: Keeping it `include-detailed-iterations-stats`

yb-voyager/cmd/getDataMigrationReportCommand.go
makalaaneesh: but this is in endMigration file, which seems more confusing. Maybe move that to some better location (something associated with target, like import.go maybe?)

PR #3407, opened 2026-03-09
Fix CP UI not able to show size and IOPS for the indexes and case-sensitive table names in assessment

praharsh-yb:
- Adds a new ParsedSchemaSummary field to both YB-Aeon and Yugabyted assessment payloads (upgraded to v1.1 and v1.8 respectively), providing structured Objects []ObjectPayload with individual ObjectName, ParentTableName and SchemaName fields instead of a single comma-separated ObjectNames string.
- Unquotes identifiers (e.g. "MyTable" becomes MyTable) so consumers don't need to handle SQL quoting. Index entries like idx_name ON schema.table are split into separate fields.
- The original SchemaSummary field is preserved unchanged for backward compatibility.

No

Manually
Yes, versions are upgraded 
No
---

| Component        

### DIFF
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -1590,24 +1590,18 @@ func parseObjectNamesToPayload(objectNames string, objectType string, dbType str
 		if slices.Contains([]string{"INDEX", "TRIGGER", "POLICY"}, objectType) {
 			parts := strings.SplitN(name, " ON ", 2)
 			if len(parts) == 2 {
-				tableObj := sqlname.NewObjectNameWithQualifiedName(dbType, "", strings.TrimSpace(parts[1]))
+				tableObj := sqlname.NewObjectNameFromMaybeQualifiedName(dbType, "", strings.TrimSpace(parts[1]))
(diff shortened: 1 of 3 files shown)


yb-voyager/cmd/common.go
```diff
@@ -1524,6 +1527,83 @@ type AssessmentIssueYBM struct {
...
+			parts := strings.SplitN(name, " ON ", 2)
+			if len(parts) == 2 {
+				obj.ObjectName = sqlname.NewIdentifier(dbType, strings.TrimSpace(parts[0])).Unquoted
+				tableObj := sqlname.NewObjectNameWithQualifiedName(dbType, "", strings.TrimSpace(parts[1]))
+				obj.ParentTableName = tableObj.Qualified.Unquoted
+			} else {
+				obj.ObjectName = sqlname.NewIdentifier(dbType, name).Unquoted
+			}
+		} else {
+			obj.ObjectName = sqlname.NewIdentifier(dbType, name).Unquoted
```
priyanshi-yb: An identifier is used for the basic identifier name in SQL, but an object name can also be qualified, so we need to use ObjectName as well.

PR #3421, opened 2026-03-11
Fix: Export side cutover should be resumable in case it fails at any point before cutover is marked processed

priyanshi-yb:
…

---

| Component        


yb-voyager/cmd/common.go
```diff
@@ -767,7 +767,7 @@ func getCutoverStatus() string {
 	}
 	if msr.FallForwardEnabled && a && b && c && msr.ExportFromTargetFallForwardStarted {
 		return COMPLETED
-	} else if msr.FallbackEnabled && a && b && c && msr.ExportFromTargetFallBackStarted {
+	} else if msr.FallbackEnabled && a && b && c && msr.ExportFromTargetFallBackStarted && msr.ImportDataToSourceStarted {
```
makalaaneesh: why also include the importDataToSourceStarted? The only thing we really need is for ExportDataFromTarget to have started, no? (to ensure that replication slot is created, and that we can resume... )

PR #3442, opened 2026-03-18
Fix xCluster replication detection on target database during import data

praharsh-yb:
Two bugs were preventing xCluster replication detection during import data:

**1. Port concatenation bug in monitor_target_health.go**

The master addresses passed to the yb-client wrapper were malformed. The old code used strings.Join(ybServers, ":7100,") which produced addresses like host1:7100,host2 -- the last server was missing its port. Fixed by appending :7100 to each server individually before joining with ,.

**2. xCluster streams not detected by yb-client-cdc-stream-wrapper (Main.java)**

The wrapper was calling listCDCStreams(null, namespaceID, IdTypePB.NAMESPACE_ID), which applies server-side filtering that excludes xCluster streams. The YB master's ListCDCStreams RPC (in xrepl_catalog_manager.cc) skips any stream that either lacks the id_type option or has it set to something other than NAMESPACEID when the request specifies IdTypePB.NAMESPACE_ID. xCluster streams don't carry this option (source_type=XCLUSTER, no id_type, no namespace_id), so they were unconditionally filtered out.

Fixed by:

- Fetching all CDC streams unfiltered via listCDCStreams(null, null, null)
- Post-filtering in Java to count only streams belonging to the target database: streams are counted if their namespaceId matches (CDCSDK streams) or if any of their tableIds belong to a table in the target namespace (xCluster streams)
- Also fixed a latent bug where string comparison used != (reference equality) instead of .isEmpty() (value equality)

No

Manually
No
No

---

| Component        

### DIFF
--- a/debezium-server-voyager/yb-client-cdc-stream-wrapper/src/main/java/Main.java
+++ b/debezium-server-voyager/yb-client-cdc-stream-wrapper/src/main/java/Main.java
@@ -84,13 +84,19 @@ public class Main {
                 String namespaceID = namespaceInfoResponse.getNamespaceId();
 
                 Set<String> namespaceTableIds = new HashSet<>();
-                ListTablesResponse tablesResp = client.getTablesList();
-                for (TableInfo tableInfo : tablesResp.getTableInfoList()) {
-                    String tableNamespaceId = tableInfo.getNamespace().getId().toStringUtf8();
-                    String tableId = tableInfo.getId().toStringUtf8();
-                    if (!tableNamespaceId.isEmpty() && !tableId.isEmpty() && namespaceID.equals(tableNamespaceId)) {
-                        namespaceTableIds.add(tableId);
+                try {
+                    ListTablesResponse tablesResp = client.getTablesList();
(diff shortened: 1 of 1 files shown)


debezium-server-voyager/yb-client-cdc-stream-wrapper/src/main/java/Main.java
```diff
@@ -74,14 +80,26 @@ public static void main(String[] args) throws Exception {
...
-            if (cdcStreamsResponse.hasError()) {
-              throw new RuntimeException("error getting the num of cdc streams");
+                }
+
+                int count = 0;
+                for (CDCStreamInfo stream : cdcStreamsResponse.getStreams()) {
+                    if (namespaceID.equals(stream.getNamespaceId())
+                            || stream.getTableIds().stream().anyMatch(namespaceTableIds::contains)) {
+                        count++;
+                    }
```
priyanshi-yb: this if also should be in a way 
```
if stream has namespace
        match namespace
else if stream has table IDs
        match table ids
```

PR #3446, opened 2026-03-23
Refactor archive changes iteration and add delete policy unit tests

AmeyaKasture:

This PR refactors the segment cleanup implementation so the delete and archive run loops delegate their work to single-iteration helpers, making the cleanup behavior easier to reason about and test. It introduces `DeleteProcessedSegments()` and `ArchiveProcessedSegments()` to encapsulate one cleanup pass, while preserving the existing stop-and-drain behavior in the ticker-driven loops. The PR also adds comprehensive unit coverage for segment cleanup delete-policy behavior across normal live migration, fall-forward pre-cutover, post-cutover, and fallback workflows. To support that coverage cleanly, the test setup helpers are kept local to the `segmentcleanup` package.

No user-facing changes.

Ran:
- `go test -tags unit ./src/segmentcleanup`
- `go vet -tags unit ./...`

New unit tests cover:
- `TestDeletePolicy_ProcessedSegmentsDeletedWithoutFF`: deletes only target-processed source segments when fall-forward is disabled.
- `TestDeletePolicy_FFPreCutoverRequiresBothImporters`: requires both target and source-replica processing before deletion in FF pre-cutover.
- `TestDeletePolicy_FFPostCutoverSourceSegmentPendingSR`: keeps old source-exported segments until source-replica processing completes after cutover.
- `TestDeletePolicy_FFPostCutoverTargetSegmentsSkippedBeforeSR`: keeps target-exported FF segments until source-replica processing completes.
- `TestDeletePolicy_FallbackWorkflowUsesImportCountOne`: verifies fallback still uses the non-FF import-count rule.
- `TestDeletePolicy_FSBelowThresholdNoDeletion`: verifies cleanup does not run below the configured FS threshold and only drains on stop.
- `TestIsValidPolicy`: validates accepted and rejected cleanup policies.
- `TestMarkSegmentDeletedAndArchived`: verifies DB flags are updated correctly.
- `TestDeleteProcessedSegmentsRemovesFile`: verifies processed segment files are removed and DB state is updated.
- `TestDeleteProcessedSegmentsMissingFile`: verifies missing files are handled gracefully without failing cleanup.

No.

No.

### DIFF
--- a/yb-voyager/src/segmentcleanup/cleanup.go
+++ b/yb-voyager/src/segmentcleanup/cleanup.go
@@ -140,7 +140,7 @@ func (sc *SegmentCleaner) runDeletePolicy() error {
 // atomically fetch both processed and pending segment lists, apply the cleanup
 // buffer, and delete each eligible segment. Logs counts for debugging.
 func (sc *SegmentCleaner) DeleteProcessedSegments() (segments []utils.Segment, pendingSegments []utils.Segment, deletedSegments []utils.Segment, err error) {
-	segments, pendingSegments, err := sc.metaDB.GetProcessedAndPendingSegments()
+	segments, pendingSegments, err = sc.metaDB.GetProcessedAndPendingSegments()
 	if err != nil {
 		return nil, nil, nil, goerrors.Errorf("get processed and pending segments: %v", err)
 	}
@@ -222,7 +222,7 @@ func (sc *SegmentCleaner) runArchivePolicy() error {
 // atomically fetch both processed and pending segment lists, copy each processed
 // segment to the archive directory, then delete it.
 func (sc *SegmentCleaner) ArchiveProcessedSegments() (segments []utils.Segment, pendingSegments []utils.Segment, archivedSegments []utils.Segment, err error) {
-	segments, pendingSegments, err := sc.metaDB.GetProcessedAndPendingSegments()
+	segments, pendingSegments, err = sc.metaDB.GetProcessedAndPendingSegments()
 	if err != nil {
 		return nil, nil, nil, goerrors.Errorf("get processed and pending segments: %v", err)
 	}


yb-voyager/src/segmentcleanup/cleanup_test.go
```diff
@@ -0,0 +1,821 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, 0, len(segsAfter), "No processed segments should remain after deletion")
+	t.Log("RESULT: PASS - Sum-based predicate importCount=1 works correctly for non-FF workflow")
+}
+
+// ============================================================
+// Test A2: MSR ArchivingEnabled tracking
+// Verifies: ArchivingEnabled set to true when cleanup starts
+// ============================================================
+func TestA2_MSRArchivingEnabled(t *testing.T) {
```
makalaaneesh: Don't quite understand what this test is testing? We're just setting a field in MSR and then checking if it's set properly?

yb-voyager/src/segmentcleanup/cleanup_test.go
```diff
@@ -0,0 +1,821 @@
...
+			s.ImportedBySource, s.Deleted)
+	}
+	t.Log("RESULT: PASS - Pre-cutover sum-based check with importCount=2 works correctly")
+}
+
+// ============================================================
+// Test B5: FF pre-cutover - SR lagging behind
+// Verifies: segment NOT deleted when SR hasn't processed it
+// ============================================================
+func TestB5_FFPreCutoverSRLagging(t *testing.T) {
```
makalaaneesh: Correct me if i'm wrong but this seems very similar to TestB4_FFPreCutoverBothImportersRequired? What extra are we testing here?

yb-voyager/src/segmentcleanup/cleanup_test.go
```diff
@@ -0,0 +1,821 @@
...
+	assert.Equal(t, 2, importCount)
+	assert.Equal(t, "", role)
+
+	segs, err := mdb.GetProcessedQueueSegments(importCount, role)
+	require.NoError(t, err)
+	t.Logf("Pre-cutover: GetProcessedQueueSegments(%d, %q) -> %d segments", importCount, role, len(segs))
+	for _, s := range segs {
+		t.Logf("  Eligible: seg=%d path=%s", s.Num, s.FilePath)
+	}
+	assert.Equal(t, 2, len(segs), "Seg 1 (source, sum=2) and seg 3 (target, sum=1) should be eligible")
```
makalaaneesh: don't think this is a potential real-scenario 
1. segment 2 will be produced before segment 3. 
2. segment 2 will be imported to source-replica before segment 3 (they go in order)
3. so it's impossible for segment 3 to be imported by source-replica but not segment 2.

yb-voyager/src/segmentcleanup/cleanup_test.go
```diff
@@ -0,0 +1,607 @@
...
+	deletedCount := 0
+	for _, s := range allSegs {
+		t.Logf("  seg=%d deleted=%d", s.SegmentNo, s.Deleted)
+		if s.Deleted == 1 {
+			deletedCount++
+		}
+	}
+	t.Logf("Total deleted: %d (during SignalStop draining)", deletedCount)
+	t.Log("NOTE: When SignalStop is called, isFSUtilizationExceeded returns true to allow draining")
+	t.Log("RESULT: PASS - Threshold gating works; segments only deleted during stop-drain")
```
makalaaneesh: we don't seem to be asserting anything here?

yb-voyager/src/segmentcleanup/cleanup_test.go
```diff
@@ -0,0 +1,340 @@
...
+func newArchiveCleaner(exportDir string, archiveDir string, mdb *metadb.MetaDB) *SegmentCleaner {
+	return NewSegmentCleaner(Config{
+		Policy:                 PolicyArchive,
+		ExportDir:              exportDir,
+		ArchiveDir:             archiveDir,
+		FSUtilizationThreshold: 0,
+	}, mdb)
+}
+
+// ============================================================
```
makalaaneesh: Can we order/name/group the files according to the workflow? (live/ live-fall-back/ live-fall-forward). There are multiple tests, and it's a bit confusing to figure out which ones correspond to which workflow, so i'm not able to tell whether we have good coverage or not. 

In fact, i think a single test for each workflow should be sufficient, no?

We should cover: 
- basic live flow: 
    - assert segments are being deleted. (- buffer of 3)
        - some segments should be processed, some should be pending. 
    - simulate cutover 
    - assert that all segments are deleted. 

(same for the rest of the flows).

yb-voyager/src/segmentcleanup/cleanup_test.go
```diff
@@ -0,0 +1,402 @@
...
+	})
+
+	paths := createSegmentFiles(t, exportDir, 2)
+	// Seg 1: fully processed by both target (1) and source-replica (1)
+	insertSegment(t, exportDir, SegmentRow{1, paths[0], "source_db_exporter", 1, 1, 0, 0, 0})
+	// Seg 2: only target processed, source-replica pending
+	insertSegment(t, exportDir, SegmentRow{2, paths[1], "source_db_exporter", 1, 0, 0, 0, 0})
+
+	cleaner := newDeleteCleaner(exportDir, mdb)
+	cleaner.SignalStop()
```
makalaaneesh: I think by default, for all tests, we should test our minimum-3 logic everywhere first.  
At the end of the test, we can additionally test our signalStop logic, and then verify that all the segments are deleted/archived.

Let's do this for all the tests

makalaaneesh (approved):
LGTM! I think there's some build issue, pls fix that before merging

PR #3447, opened 2026-03-23
Implement byte-offset seek resumption for cloud datastores (S3, GCS, Azure)

ShivanshGahlot:

A prior change (PR #3404) added byte-offset seeking for import data resumption on the local `DataStore`. This PR extends that to **Amazon S3** and **Google Cloud Storage** by implementing `OpenAt` with ranged reads:

- **S3**: `NewObjectReaderAt` in `s3utils` using `gocloud.dev/blob` range readers; `S3DataStore.OpenAt` matches the same symlink/URL resolution as `Open`.
- **GCS**: `NewObjectReaderAt` in `gcsutils` using the GCS client’s `NewRangeReader`; `GCSDataStore.OpenAt` mirrors `Open` for path handling.

**Azure Blob Storage** is **not** changed relative to `main`. `AzDataStore.OpenAt` remains a stub that returns `ErrOpenAtNotImplemented`, and the importer continues to use the `SkipLines` resumption path for data under `https://` Azure URLs. (Recent work on the branch that added Azure ranged reads and removed the `SkipLines` fallback was reverted so cloud resumption behavior for Azure stays as before, while S3 and GCS get fast byte-offset resume.)

No CLI, configuration, installation, or report changes. For **S3** and **GCS** data directories, resuming `import data` can seek to the last byte offset instead of re-reading and discarding prior lines. **Azure** behavior is unchanged from `main` (line-based skip on resume).

- `go build -tags unit` for affected packages (e.g. `./src/datastore`, `./src/utils/s3`, `./src/utils/gcs`, `./cmd`).
- Existing byte-offset resumption tests in `importDataSequentialFileBatchProducer_test.go` cover the `OpenAt` path with a local datastore; they do not require cloud credentials.
- **S3 / GCS**: `NewObjectReaderAt` uses provider SDK range APIs; validating against real buckets (resume after interrupt, confirm logs show byte-offset resume instead of `SkipLines`) is recommended.
- **Azure**: unchanged resumption path; no new ranged-read path to exercise for this PR.

No.

No.

---

| Component        

### DIFF
--- a/yb-voyager/cmd/importDataSequentialFileBatchProducer.go
+++ b/yb-voyager/cmd/importDataSequentialFileBatchProducer.go
@@ -16,7 +16,6 @@ limitations under the License.
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"sort"
@@ -27,7 +26,6 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/datafile"
-	"github.com/yugabyte/yb-voyager/yb-voyager/src/datastore"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/dbzm"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/importdata"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
@@ -41,16 +39,15 @@ type SequentialFileBatchProducer struct {
 
 	pendingBatches  []*Batch // pending batches after recovery
 	lastBatchNumber int64    // batch number of the last batch that was produced
-	lastOffset      int64    // file offset from where the last batch was produced, only used in recovery
 	fileFullySplit  bool     // if the file is fully split into batches
 	completed       bool     // if all batches have been produced
 
-	dataFile               datafile.DataFile
-	header                 string
-	headerByteCount        int64
-	numLinesTaken          int64 // number of lines read from the file
+	dataFile                  datafile.DataFile
+	header                    string
+	headerByteCount           int64
+	numLinesTaken             int64 // number of lines read from the file
 	lastBatchCumByteOffsetEnd int64 // cumulative byte offset end recovered from the last batch's state
-	cumByteOffsetEnd         int64 // running cumulative byte offset end tracking absolute file position
+	cumByteOffsetEnd          int64 // running cumulative byte offset end tracking absolute file position
 	// line that was read from file while producing the previous batch
 	// but not added to the batch because adding it would breach size/row based thresholds.
 	lineFromPreviousBatch string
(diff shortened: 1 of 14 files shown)


yb-voyager/cmd/importDataSequentialFileBatchProducer.go
```diff
@@ -360,26 +351,7 @@ func (p *SequentialFileBatchProducer) openDataFileAtByteOffset() error {
 	}
 	p.dataFile = dataFile
 
-	log.Infof("Resumed file %q at byte offset %d (skipped SkipLines)", p.task.FilePath, p.lastBatchCumByteOffsetEnd)
-	return nil
-}
-
-// openDataFileAndSkipLines resumes by opening the file from byte 0 and skipping
-// lines to reach the resumption point. Used when OpenAt is not supported.
-func (p *SequentialFileBatchProducer) openDataFileAndSkipLines() error {
```
makalaaneesh, line 369: should we keep the old code as a contingency (i.e. controlled by env var) for a few releases? (in case some user hits this issue, we can get them to revert to old slow behavior). 

I think this might be a good idea considering this is a big change.

makalaaneesh: @ShivanshGahlot as we had discussed, can you add in a config-variable(hidden) that allows us to pick between the old and new implementation?

ShivanshGahlot: I have made the implementation such that the code tries OpenAt, but for Azure and GCS it returns ErrOpenAtNotImpelemented and fallbacks to the older path. Kept the newer code in OpenAt for Azure and GCS commented for any future use. Also added a couple of tests to verify that the correct path is taken if OpenAt is not implemented.

makalaaneesh:
[AI] Automated review — no Critical/Warning findings. Implements byte-offset seek resumption for S3 (`OpenAt` + `s3.NewObjectReaderAt`) while GCS and Azure `OpenAt` intentionally return `ErrOpenAtNotImplemented` so the import path falls back to `SkipLines`. Reviewed for:

- **Import/compile correctness** — `s3Datastore.go` has all required imports (`fmt`, `os`, `strings`, `s3`).
- **Open/OpenAt parity** — `S3DataStore.OpenAt` mirrors `Open`'s `s3://`-prefix vs symlink resolution, and improves on it by returning the error instead of `utils.ErrExit` (aligns with the yb-voyager BUGBOT error-handling rule).
- **Resource handling** — `s3.NewObjectReaderAt` follows the existing `NewObjectReader` bucket pattern; no new leak relative to established convention.
- **Fallback contract** — covered by the new `datastore_test.go` and the `TestResumeFallsBackToSkipLinesWhenOpenAtUnsupported` end-to-end test.

Note (non-blocking, not posted as a finding): `gcs.NewObjectReaderAt` and `az.NewObjectReaderAt` are added but currently unused. This is a deliberate, clearly-commented decision to preserve them for later re-enablement, so it's flagged only as an FYI against the repo-root dead-code guideline rather than a defect.

---

---

yb-voyager/cmd/importDataSequentialFileBatchProducer_test.go
```diff
@@ -1676,3 +1678,88 @@ func assertNoProcessingErrorBatchFileExists(t *testing.T, lexportDir string, tas
...
+	assert.True(t, bp2.Done())
+	bp2.Close()
+
+	// The producer attempted the byte-offset seek path exactly once...
+	assert.Equal(t, 1, spy.openAtCalls,
+		"resumption should attempt OpenAt before falling back")
+	// ...and then, because OpenAt was unavailable, resumed via SkipLines, which
+	// opens the file from byte 0. (The byte-seek path would have used the reader
+	// returned by OpenAt and never called Open for the data.)
+	assert.Greater(t, spy.openCalls, 0,
```
makalaaneesh, line 1756: nice!

makalaaneesh (approved):
LGTM!

PR #3449, opened 2026-03-24
Add archive policy for segment cleanup

AmeyaKasture:
Add archive policy for segment cleanup 

Introduce a new "archive" cleanup policy that copies processed segments
to a user-specified --archive-dir before deleting the originals, without
gating on filesystem utilization. Replace the duplicate copySegmentFile
implementations in both segmentcleanup and archiveChangesCommand with
the existing utils.CopyFile.

### DIFF
--- a/yb-voyager/cmd/segmentCleanupCommand.go
+++ b/yb-voyager/cmd/segmentCleanupCommand.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/tebeka/atexit"
 
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/metadb"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/segmentcleanup"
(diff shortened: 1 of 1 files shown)


yb-voyager/src/segmentcleanup/cleanup.go
```diff
@@ -160,3 +165,61 @@ func (sc *SegmentCleaner) runRetainPolicy() error {
...
+			}
+		}
+	}
+	return nil
+}
+
+func (sc *SegmentCleaner) archiveAndDeleteSegment(seg utils.Segment) error {
+	archivePath := filepath.Join(sc.config.ArchiveDir, filepath.Base(seg.FilePath))
+
+	if utils.FileOrFolderExists(seg.FilePath) {
```
makalaaneesh: if seg.FilePath file does not exist, what then?

1. Firstly, when could this happen? Can you add comments? From what i can tell, if we crash after removing the file, but before updating metaDB, this could happen. Is that what you had in mind as well? 
2. If we are assuming that we've moved it before, we should validate that. I.e. check the archivePath if it exists. If it does exist, no problem continue and update the metadb. If it does NOT exist, then we have a problem, because it does not exist anywhere. In this case, it would be good to error out. I can't think of a case where this would happen, but it would be good to check and error out.

PR #3482, opened 2026-04-01
Add iteration test scenario with archive changes validation

Abhi-yb:
Adds a resumption test scenario for the iterative live-migration-with-fallback workflow. The orchestrator supports loop_start/loop_end actions to run variable iterations, stop_external_process to cleanly kill auto-transitioned Voyager processes, and apply_effective_export_dir to resolve per-iteration export directories. Archive changes validation covers both pre-cutover (file count + vsn continuity) and post-cutover-to-source (queue empty) states. Row count and row hash validations run at the end of each iteration.

Tested with iterations end-to-end covering forward/fallback phases, archive validation (archive and delete policies), and export resumptions.

---

> [!NOTE]
> **Medium Risk**
> Adds new looping test orchestration plus PID-based process termination and per-iteration export-dir resolution, which can affect how long-running `yb-voyager` processes are started/stopped and what data directories validations read. Main risk is accidentally stopping the wrong processes via `pgrep -f` matching or validating against an unexpected export-dir layout.
> 
> **Overview**
> Adds an **iterative live-migration-with-fallback resumption scenario** (`iteration-test`) that runs many forward/fallback cycles and performs row count/hash validations plus DB cleanup each iteration.
> 
> Extends the resumption orchestrator to support **scenario looping** (`loop_start`/`loop_end` with `num_iterations`), **per-iteration export directory resolution** (`apply_effective_export_dir`), and a new `stop_external_process` action that can terminate untracked `yb-voyager` processes by discovering PIDs via `pgrep`.
> 
> Introduces `yb-voyager archive changes` coverage: new `voyager_archive_changes_start`, command builder, and `validate_archive_changes` checks for both `archive` and `delete` policies (segment file counts and VSN continuity pre-cutover; queue-empty expectations post-cutover-to-source).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e01c70ac1e66c3e52d34ec6127563348f930ba92. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -0,0 +1,353 @@
...
+    timeout_sec: 30
+
+  - name: stop_import_resumptions
+    action: voyager_stop_resumptions
+    command: import_data
+    timeout_sec: 30
+
+  - name: wait_for_archiver
+    action: sleep
+    seconds: 180
```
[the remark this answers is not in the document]
Abhi-yb: Tried removing the sleep — archive validation started failing in long runs (8min, 17min iterations) because the archiver is async and hasn't finished processing segments when validation runs immediately. Keeping the sleep as a simple fix; can replace with a polling/retry loop later if needed.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -0,0 +1,353 @@
...
+  - name: loop_start
+    action: loop_start
+
+  - name: stop_forward_exporter
+    action: stop_external_process
+    process: export_data
+    graceful_timeout_sec: 60
+
+  - name: stop_forward_importer
+    action: stop_external_process
```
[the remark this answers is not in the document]
Abhi-yb: Done — renamed to stop_prev_iteration_exporter/stop_prev_iteration_importer (loop start) and stop_forward_exporter_for_fallback/stop_forward_importer_for_fallback (post-cutover).

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -0,0 +1,353 @@
...
+    action: row_hash_validations
+
+  - name: cleanup_source
+    action: run_sql
+    sql_path: ./cleanup.sql
+
+  - name: cleanup_target
+    action: run_sql
+    target: target
+    sql_path: ./cleanup.sql
```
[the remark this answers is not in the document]
Abhi-yb: Needed for the test scenario — the event generator reuses the same data patterns each iteration, so without cleanup we hit duplicate key errors during cutover. Real production iterations wouldn't need this, but the test does.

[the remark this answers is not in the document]
priyanshi-yb: I think ideally we should not do this as iteration workflow is not suppose to do cleanup in between, cc: @shubham-yb @yashtyagi-yb on the scenario and generator script problem, we should solve this somehow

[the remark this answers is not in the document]
Abhi-yb: Agree in principle — but for now cleanup is acting as a safety net against data collisions and non-idempotent validations, which can cause flaky failures across iterations.

It might be better to first address this at the generator/validation level, and then we can look at removing cleanup safely.

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -41,6 +43,46 @@ def __init__(self, cfg: Dict[str, Any], env: Dict[str, str], test_root: str | No
...
+def apply_effective_export_dir(ctx: Context) -> None:
+    """Point cfg['export_dir'] at the scenario parent for loop_iteration 0, else latest iteration export-dir."""
+    ctx.prev_archive_file_count = 0
+    base = ctx.export_dir_base or os.path.abspath(ctx.cfg.get("export_dir") or "")
+    if ctx.loop_iteration == 0:
+        ctx.cfg["export_dir"] = base
+        return
+
+    it_root = os.path.join(base, "live-data-migration-iterations")
+    best_n = -1
```
shubham-yb: Can you please name the variables to reflect their purpose?
example: it_root, best_n, cand etc

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -320,6 +410,36 @@ def stop_process(ctx: Context, name: str, graceful_timeout: int = 10) -> bool:
     return True
 
 
+def stop_external_process(ctx: Context, name: str, graceful_timeout: int = 10) -> bool:
```
shubham-yb: Why do we need this? What are we requiring to kill?

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -344,6 +464,18 @@ def _merge_flags(base: Dict[str, Any], extra: Dict[str, Any] | None) -> Dict[str
     return merged
 
 
+def _merge_flags_when_loop_iteration_gte(cfg: Dict[str, Any], op: str, loop_iteration: int) -> Dict[str, Any]:
```
shubham-yb: Can you give an example of what happens here?

migtests/scripts/live-migration-resumption/orchestrator.py
```diff
@@ -102,13 +104,56 @@ def import_to_source_replica_start_action(_stage, ctx: Any) -> None:
...
+    Optional stage key:
+      - policy: explicit policy ("delete" or "archive").
+                When omitted, a policy is chosen at random.
+    Skips if the archiver is already running (continuous archiver across iterations).
+    """
+    with ctx.process_lock:
+        existing = ctx.processes.get("archive_changes")
+        if existing and existing.poll() is None:
+            H.log("archive_changes: already running, skipping start")
+            return
```
shubham-yb, line 120: Do we need this extra handling?

migtests/scripts/live-migration-resumption/orchestrator.py
```diff
@@ -213,6 +258,23 @@ def sleep_action(stage: Dict[str, Any], ctx: Any) -> None:
...
+class _LoopEnd(Exception):
+    """Sentinel raised by loop_end to signal the runner to jump back."""
+    pass
+
+
+@action("loop_end")
+def loop_end_action(_stage, ctx: Any) -> None:
+    """Signal the runner to jump back to the matching loop_start."""
+    raise _LoopEnd()
+
```
shubham-yb, line 261: Can we simplify this?

migtests/tests/resumption/live-migration/common-sql/delete_marker.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM public.cutover_table;
```
shubham-yb: Why do we need to do this? How do the other tests handle multiple entries in the cutover table?

migtests/tests/resumption/live-migration/iteration-test/iteration_reset.sql
shubham-yb, line 1: These are DDL operations. Why are we doing this in the middle of the iterations?

migtests/tests/resumption/live-migration/iteration-test/temp_file_limit.sql
```diff
@@ -0,0 +1 @@
+ALTER DATABASE test_db SET temp_file_limit = 20971520; -- 20GB
```
shubham-yb, line 1: Why are we doing this?

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -0,0 +1,352 @@
...
+      faker_seed: 456
+      insert_rows: 6
+      update_rows: 4
+      delete_rows: 2
+      insert_max_retries: 5
+      update_max_retries: 1
+
+voyager:
+  export_data:
+    flags_when_loop_iteration_gte:
```
shubham-yb: What does this do? flags_when_loop_iteration_gte? Can we rename it?

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -0,0 +1,352 @@
...
+        export-type: changes-only
+  cutover_to_target:
+    flags:
+      prepare-for-fall-back: true
+  cutover_to_source:
+    flags:
+      restart-data-migration-source-target: true
+
+stages:
+  # Phase 1: forward live migration (no resumptions)
```
shubham-yb, line 99: Why don't we have any resumptions here?

migtests/tests/resumption/live-migration/iteration-test/iteration_reset.sql
shubham-yb, line 1: Also please rename the file. cleanup.sql is used across the repo to cleanup source and target DBs.

migtests/tests/resumption/live-migration/iteration-test/temp_file_limit.sql
shubham-yb, line 1: Please rename the file. init.sql is used across the repo to initialise the source and target DBs.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -0,0 +1,352 @@
...
+    action: run_sql
+    sql_path: ./init.sql
+    target: source
+    use_admin: true
+
+  - name: increase_temp_file_limit_target
+    action: run_sql
+    target: target
+    sql_path: ./init.sql
+    use_admin: true
```
shubham-yb, line 119: nit: Maintain the order of the flags.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -0,0 +1,352 @@
...
+
+  - name: stop_prev_iteration_exporter
+    action: stop_external_process
+    process: export_data
+    graceful_timeout_sec: 60
+
+  - name: stop_prev_iteration_importer
+    action: stop_external_process
+    process: import_data
+    graceful_timeout_sec: 60
```
shubham-yb: Is this something we need to do manually? Or do the iterations handle it? Asking for context. Also why can't we use "voyager_stop_command" here? 
Is it so that we control the resumptions gracefully? Do we test the transition flow elsewhere?
Need to understand the entire code around `stop_external_process` and if / why do we require it.
cc: @priyanshi-yb

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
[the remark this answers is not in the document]
shubham-yb: They are cleaning up the validation tools. Unsure why. Need to understand this better. Also what is the issue with the generator script?

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -0,0 +1,352 @@
...
+target:
+  type: yugabytedb
+  host: ${TARGET_DB_HOST}
+  port: ${TARGET_DB_PORT}
+  database: test_db
+  user: ${TARGET_DB_USER}
+  password: ${TARGET_DB_PASSWORD}
+  admin:
+    user: ${TARGET_DB_USER}
+    password: ${TARGET_DB_PASSWORD}
```
shubham-yb: admin:
    user: ${TARGET_DB_ADMIN_USER}
    password: ${TARGET_DB_ADMIN_PASSWORD}

migtests/scripts/live-migration-resumption/orchestrator.py
```diff
@@ -279,16 +341,46 @@ def main() -> None:
...
+                    raise RuntimeError(
+                        "loop_end: scenario has no loop_start stage"
+                    ) from None
+                iteration += 1
+                ctx.loop_iteration = iteration
+                H.apply_effective_export_dir(ctx)
+                if iteration >= num_iterations:
+                    idx += 1
+                else:
+                    idx = loop_start_idx
```
shubham-yb: `H.apply_effective_export_dir(ctx)` runs before the `iteration >= num_iterations` check, so on the final successful loop we retarget `cfg["export_dir"]` before artifact collection. That looks like it can snapshot the next iteration’s directory rather than the one that actually ran.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -0,0 +1,352 @@
...
+
+  - name: stop_forward_exporter_for_fallback
+    action: stop_external_process
+    process: export_data
+    graceful_timeout_sec: 60
+
+  - name: stop_forward_importer_for_fallback
+    action: stop_external_process
+    process: import_data
+    graceful_timeout_sec: 60
```
shubham-yb, line 238: Why aren't we using "voyager_stop_command" here? Refer the other tests.

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -483,13 +618,164 @@ def build_import_to_source_replica_cmd(cfg: Dict[str, Any]) -> list[str]:
...
+def _list_segment_files(directory: str) -> list[str]:
+    """Return sorted list of segment.N.ndjson filenames in *directory*."""
+    if not os.path.isdir(directory):
+        return []
+    return sorted(
+        (n for n in os.listdir(directory) if n.startswith("segment.") and n.endswith(".ndjson")),
+        key=lambda n: int(n.split(".")[1]),
+    )
+
+def _first_last_vsn(filepath: str) -> tuple[int | None, int | None]:
```
shubham-yb: nit: _return_first_last_vsn.
Similarly for other functions as well.

migtests/scripts/live-migration-resumption/helpers.py
Abhi-yb: This block lets the scenario flip on extra CLI flags for `yb-voyager` once the iteration counter crosses a threshold.

Concrete example from this PR:

```yaml
voyager:
  export_data:
    flags_when_loop_iteration_gte:
      "1":
        export-type: changes-only
```

- Iteration 0 → `yb-voyager export data` runs snapshot + CDC (default)
- Iteration ≥ 1 → `yb-voyager export data --export-type changes-only` runs CDC only

Iteration 0 already did the full snapshot, so from iteration 1 onwards we just stream changes from where we left off.

migtests/scripts/live-migration-resumption/helpers.py
Abhi-yb: Will rename to `_return_first_last_vsn` and apply the same `_return_*` prefix to neighboring helpers in this section.

migtests/scripts/live-migration-resumption/orchestrator.py
Abhi-yb: Yes — `start_archiver` is inside the loop body, so the action gets dispatched every iteration. Inside the handler we do `ctx.processes["archive_changes"] = spawn(...)`, which only overwrites the Python reference; it doesn't kill the previous OS process. So without the guard, every re-entry into the loop spawns a fresh `yb-voyager archive changes` process while the previous one is still running — leaving multiple archivers racing to read the same queue dir and write to the same archive-dir.

The guard makes the action a no-op when the archiver started earlier is still running, which is the intended "continuous archiver across iterations" behavior.

migtests/scripts/live-migration-resumption/orchestrator.py
Abhi-yb: The runner walks scenario stages one by one and dispatches each by action name. Action handlers just run and return — they don't have a way to tell the runner "jump back to a previous stage". So `loop_end` needs an out-of-band signal, and we use an exception (`_LoopEnd`) for that — the runner has a `try/except _LoopEnd:` block that handles the loop-back logic in one place.

One simpler-looking alternative would be to special-case `loop_end` in the dispatch loop directly (e.g. `if action == "loop_end": idx = loop_start_idx; continue`) and skip the exception. The trade-off is that it adds inline special-casing for two action names in the runner, breaking the otherwise-uniform "every stage = one action handler" pattern.

Open to switching to the inline approach if you prefer it — happy to refactor.

migtests/tests/resumption/live-migration/common-sql/delete_marker.sql
Abhi-yb: The marker pattern: `insert_marker.sql` inserts a sentinel row into `cutover_table` to signal "last event before cutover", and the orchestrator's `wait_for` condition (`marker_in_latest_queue_file`) checks that this row has propagated before declaring cutover complete.

Single-cutover tests (`fall-forward-resumption-test`, `fall-back-resumption-test`) are one-shot — insert once, wait, done. The iteration-test runs N forward/fallback cycles, so without `delete_marker.sql` between iterations the next cycle's `wait_for` would falsely succeed against the stale marker from the previous iteration.

`DELETE` (not `TRUNCATE`) so we keep the table + sequence intact.

migtests/tests/resumption/live-migration/iteration-test/iteration_reset.sql
Abhi-yb: This is test-harness state cleanup, not user-data DDL. The hash/segment validation framework installs the following on both source and target each iteration:
- `migration_validate_segments` table (records per-segment row counts)
- `compute_schema_segment_hashes` / `compute_table_segment_hashes` / `custom_hash_code_from_text` functions

Without the reset, stale rows in `migration_validate_segments` would skew the next iteration's segment math, and the framework's `CREATE FUNCTION` statements aren't idempotent so they'd conflict on re-install.

The `cutover_table` truncate is for a separate reason — during PR development we hit a duplicate-key error in `cutover_table` on the fallback marker insert across iterations (commit `aa37d09`). Truncating between iterations gives the next cycle's marker insert a clean slate. None of this DDL touches the actual application schema being migrated.

migtests/tests/resumption/live-migration/iteration-test/temp_file_limit.sql
Abhi-yb: The hash-validation step runs full-table aggregates (`compute_schema_segment_hashes` / `compute_table_segment_hashes`) that can spill significant temp files on large tables. The default `temp_file_limit` (~1 GB on YB) is too low for this scenario — the validation queries fail partway with "temp file size exceeds temp_file_limit". 20 GB is generous headroom for the data sizes iteration-test runs against. Applied on both source and target since both run validation queries.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
Abhi-yb: This block is a way to override CLI flags for a `yb-voyager` operation when the iteration counter crosses a threshold — e.g. the scenario uses it to switch `export-data` to `--export-type=changes-only` from iteration 1 onwards (since iteration 0 already did the snapshot).

Agree the current name is wordy. Proposing `flags_when_iteration_gte` — drops the redundant `loop_` prefix and keeps `gte` so the `>=` semantics stay explicit. Will apply the same to the helper `_merge_flags_when_iteration_gte`. Open to a different name if you'd prefer something else.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
Abhi-yb: The stages above this point are one-time setup — DB reset, schema export, schema import, cutover_table creation, source permissions — they run before the iteration loop starts. They're quick deterministic schema operations rather than data movement, so we didn't add resumptions for them; the resumption coverage lives in the data-migration stages inside the loop (`export_data`, `import_data`, `archive_changes`, fallback exporter+importer).

Happy to add resumption coverage to the setup stages too if you think it'd add value — let me know.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
Abhi-yb: `voyager_stop_command` only stops processes tracked in `ctx.processes` (i.e. ones the orchestrator itself started via `voyager_export_start` etc.). `stop_external_process` uses `pgrep -f` to find any matching `yb-voyager` process by semantic name, regardless of who started it.

Why we need this in iteration-test: when `cutover_to_target` completes, Voyager itself spawns successor processes for the fallback phase (e.g. import-to-source) via its auto-transition logic. Those are not tracked in `ctx.processes`, so `voyager_stop_command` can't reach them.

- At loop start (line 142–150): we stop any leftover untracked exporter/importer from the previous iteration before starting fresh ones.
- At line 234–242 (`*_for_fallback`): same reason — Voyager hands off to the fallback flow internally and the orchestrator loses tracking.

Other resumption tests are single-cycle, so they only deal with processes the orchestrator started directly — `voyager_stop_command` is sufficient there. Iteration-test needs both.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
Abhi-yb: Same reason as my reply on line 150 — `voyager_stop_command` only handles orchestrator-tracked processes. After `cutover_to_target`, Voyager auto-spawns the fallback exporter/importer internally; those PIDs aren't in `ctx.processes`, so we need `stop_external_process` (pgrep-based) to terminate them cleanly before the explicit fallback start.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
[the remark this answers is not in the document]
Abhi-yb: Two parts:

(1) Cleanup of validation tools — these aren't user-schema DDLs. They're the test's hash/segment validation harness (`migration_validate_segments` table + `compute_*` functions) which gets re-installed at the start of each iteration's validation phase. The CREATE statements aren't idempotent, so we drop them between iterations. The `cutover_table` truncate here is for the marker pattern, not user data.

(2) Issue with the generator script — the generator replays the same sequence of operations across iterations, which can lead to collisions on user-data tables. This isn't something the resumption-test scenario itself can solve — it's a generator-side concern.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
makalaaneesh: I am not sure if this is needed. afaik, voyager internally takes care of this change. cc: @priyanshi-yb

migtests/scripts/live-migration-resumption/helpers.py
priyanshi-yb: now since we don't have any flag that we want to change, do we need this merge?

priyanshi-yb (commented):
high level LGTM!

migtests/tests/resumption/live-migration/iteration-test/temp_file_limit.sql
makalaaneesh: in such cases, pls add detailed comments that explain the change

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -165,11 +220,8 @@ def poll_until(timeout_sec: int, interval_sec: int, fn: Callable[[], bool]) -> b
             return False
         time.sleep(interval_sec)
 
-
-
 def exporter_streaming(export_dir: str) -> bool:
-    """Heuristic: streaming considered started when first queue segment has at least one event.
-    """
+    """Heuristic: streaming started when some queue segment.0.ndjson has data (parent or iteration export-dir)."""
```
shubham-yb, line 224: nit: remove

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -417,14 +486,17 @@ def initiate_cutover(cfg: Dict[str, Any], env: Dict[str, str], direction: str) -
...
+    cfg = ctx.cfg
     voyager_flags = _get_voyager_flags(cfg, "export_data")
     base = _base_common_flags(cfg)
     base.update(_source_conn_flags(cfg))
-    # Live migration default
-    base["export-type"] = "snapshot-and-changes"
-    # data command defaults
     base["disable-pb"] = True
+    if ctx.loop_iteration >= 1:
+        base["export-type"] = "changes-only"
```
shubham-yb: This is not required.

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -417,14 +486,17 @@ def initiate_cutover(cfg: Dict[str, Any], env: Dict[str, str], direction: str) -
     run_checked(cmd, env, description=f"cutover_to_{direction}")
 
 
-def build_export_data_cmd(cfg: Dict[str, Any]) -> list[str]:
+def build_export_data_cmd(ctx: Context) -> list[str]:
```
shubham-yb: Why does the build command require context?

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -803,13 +1029,19 @@ def snapshot_msr_and_stats(export_dir: str, artifacts_dir: str) -> None:
     metainfo_dir = os.path.join(export_dir, "metainfo")
     dest_dir = os.path.join(artifacts_dir, "metainfo")
 
+    if not os.path.isdir(metainfo_dir):
+        log(f"snapshot_msr_and_stats: skipped, no metainfo dir at {metainfo_dir}")
+        return
```
shubham-yb: Not required.

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -803,13 +1029,19 @@ def snapshot_msr_and_stats(export_dir: str, artifacts_dir: str) -> None:
...
     shutil.copytree(metainfo_dir, dest_dir, dirs_exist_ok=True)
 
 
 def copy_logs_directory(export_dir: str, artifacts_dir: str) -> None:
     logs_dir = os.path.join(export_dir, "logs")
     dest_dir = os.path.join(artifacts_dir, "logs")
 
+    if not os.path.isdir(logs_dir):
+        log(f"copy_logs_directory: skipped, no logs dir at {logs_dir}")
+        return
```
shubham-yb: Not required.

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -1174,6 +1406,10 @@ def run_segment_hash_validations(
     Raises:
         RuntimeError if any segment shows a mismatch or is missing on one side.
     """
+    # 0) Cleanup stale validation rows from prior iterations (self-cleaning per Abhinay's guidance)
```
shubham-yb: Cleanup

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -1206,6 +1442,9 @@ def run_segment_hash_validations(
         raise RuntimeError(f"segment hash validation failed for segments: {preview}")
 
 
+_VALIDATION_INFRA_TABLES = frozenset({"migration_validate_segments", "cutover_table"})
+
+
```
shubham-yb, line 1432: Extra handling. Not required.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -0,0 +1,334 @@
...
+    action: run_sql
+    sql_path: ./temp_file_limit.sql
+    target: source
+    use_admin: true
+
+  - name: increase_temp_file_limit_target
+    action: run_sql
+    sql_path: ./temp_file_limit.sql
+    target: target
+    use_admin: true
```
shubham-yb, line 119: Run after commenting. Need to understand the requirement for this.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -0,0 +1,334 @@
...
+      insert_max_retries: 5
+      update_max_retries: 1
+
+voyager:
+  cutover_to_target:
+    flags:
+      prepare-for-fall-back: true
+  cutover_to_source:
+    flags:
+      restart-data-migration-source-target: true
```
priyanshi-yb, line 96: One issue: in the last iteration, when we are issuing cutover to source, we shouldn't use this flag as true; it should be false so that everything finishes cleanly.

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -500,7 +500,6 @@ def build_import_data_cmd(cfg: Dict[str, Any]) -> list[str]:
     voyager_flags = _get_voyager_flags(cfg, "import_data")
     base = _base_common_flags(cfg)
     base.update(_target_conn_flags(cfg))
-    base["max-retries-streaming"] = 1
```
shubham-yb, line 503: Why did we remove this?

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
priyanshi-yb: @Abhi-yb we can take it up in a separate PR.

shubham-yb (approved):
LGTM

PR #3496, opened 2026-04-14
Fix pg/datatypes_callhome test: update expected target_only_queries c…

Abhi-yb:
…ount

PR #3484 moved handleIdentityColumns() inside the changeStreamingIsEnabled block and removed restoreGeneratedIdentityColumns() from postSnapshotImportProcessing(). For offline migrations (like the callhome test), the GetIdentityColumnNamesForTables query no longer runs on the target YB. This is expected - for PG offline migrations, identity columns don't need to be handled before snapshot import. Confirmed by Aneesh.

Verified by running the callhome test locally on both v2026.3.3 (release) and latest main, comparing pg_stat_statements from the target YB. Exactly 1 query was missing on main, matching the count drop from 225 to 224.

Made-with: Cursor

---

| Component        

### DIFF
--- a/migtests/tests/pg/datatypes/expected_callhome_payloads/compare_performance_callhome.json
+++ b/migtests/tests/pg/datatypes/expected_callhome_payloads/compare_performance_callhome.json
@@ -10,8 +10,8 @@
     "payload_version": 1.0,
     "query_metrics": null,
     "source_only_queries": 142,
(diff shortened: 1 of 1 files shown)


makalaaneesh (approved):
LGTM. pls run the jenkins pipelines to verify (if not already done) before merging

PR #3497, opened 2026-04-14
Overhaul assess-migration command UX with structured preflight checks, progress spinner, and summary

AmeyaKasture:
## Describe the changes in this pull request

This PR redesigns the terminal UX for the `assess-migration` command to make it more structured, informative, and visually polished. The previous experience printed verbose log-style messages interleaved with prompts, making it hard to follow progress. The new UX is organized into three distinct phases:

1. **Preflight Checks** — All guardrail validations (password, connectivity, version compatibility, dependencies, schema resolution, replica discovery, permissions) are grouped at the top with green checkmarks, yellow skips, or red failure indicators, giving a clear pass/fail summary before any heavy work begins.

2. **Assessment Pipeline** — The core assessment work (metadata gathering, usage analysis, sizing, report generation) is presented as four discrete stages with a live spinner and step counter. The "Gathering metadata" stage tracks three key sub-steps (column statistics, redundant indexes, table index usage stats) with a `(X/3)` counter and inner sub-stage label, while all other verbose script output is redirected to the log file only.

3. **Summary** — A clean block showing issue count and clickable report paths (JSON/HTML).

A new file `assessMigrationUX.go` centralizes all UX components (banner, separator, preflight renderers, `AssessmentProgressTracker` with spinner/step counter, summary/failure printers). The metadata gathering script's stdout is now captured and parsed via `runGatherAssessmentMetadataScriptWithProgress`, which sends detected sub-step names to a callback instead of printing them directly. All `utils.PrintAndLogf` calls within the pipeline have been replaced with `log.Infof` to keep terminal output minimal.

## Describe if there are any user-facing changes

Yes — the terminal output of `yb-voyager assess-migration` is completely redesigned:
- A styled banner displays the target YugabyteDB version at startup.
- Preflight checks are listed with checkmark/skip/fail icons before the assessment begins.
- A spinner with stage names and step counters replaces verbose log output during assessment.
- A summary block with issue count and report file paths is shown at the end.
- No changes to command-line flags, configuration, installation, or report contents.

## How was this pull request tested?

- Existing unit tests pass: `go test -tags unit ./cmd/...` and `go test -tags unit ./src/migassessment/...`.
- Manual testing with `yb-voyager assess-migration` against a live PostgreSQL source, verifying:
  - All preflight checks render correctly (pass, skip, fail scenarios).
  - The spinner animates smoothly with the `(X/3)` counter ticking for the three tracked metadata sub-stages.
  - IOPS validation prompt appears cleanly between stages without spinner interference.
  - The `--assessment-metadata-dir` offline path skips source connectivity checks with appropriate skip indicators.
  - The `--yes` flag suppresses prompts as before.
- No new unit tests were added; the changes are purely cosmetic/UX and the underlying assessment logic is unchanged.

## Does your PR have changes in callhome/yugabyted payloads? If so, is the payload version incremented?

No.

## Does your PR have changes to on-disk structures that can cause upgrade issues?

No.

Screenshot 2026-04-21 at 4 09 47 PM


yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -221,141 +222,220 @@ func assessMigration() (err error) {
...
+	}
+	bannerRows = append(bannerRows, ux.BannerRow{Key: "Export directory", Value: exportDir})
+	if cfgFile != "" {
+		bannerRows = append(bannerRows, ux.BannerRow{Key: "Config file", Value: cfgFile})
+	}
+	if assessmentMetadataDirFlag != "" {
+		bannerRows = append(bannerRows, ux.BannerRow{Key: "Metadata directory", Value: assessmentMetadataDirFlag})
+	}
+	ux.PrintBanner("YugabyteDB Voyager — Migration Assessment", bannerRows)
+	log.Infof("assessing for migration to target YugabyteDB version %s", targetDbVersion)
```
makalaaneesh, line 258: It makes sense for this to be in some common pkg that all commands can re-use. 
Every command will have to print a banner and it can provide all those inputs to a function that then goes ahead and prints it. 

Maybe ux package itself?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -221,141 +222,220 @@ func assessMigration() (err error) {
...
+
+		// Connect
+		err = source.DB().Connect()
 		if err != nil {
+			ux.PrintPreflightFail("Connect to source database")
 			return fmt.Errorf("failed to connect source db for assessing migration: %w", err)
 		}
-		// We will require source db connection for the below checks
-		// Check if required binaries are installed.
+		ux.PrintPreflightCheck("Connected to source database")
```
makalaaneesh, line 281: One idea: 

How about we introduce a helper function `ux.RunPreflightCheck(checkName string, check func() err)`
this function just executes the check func, if it returns an error, it calls `PrintPreflightFail` and if it succeeds it calls `PrintPreflightCheckComplete`

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -221,141 +222,220 @@ func assessMigration() (err error) {
...
+		bannerRows = append(bannerRows, ux.BannerRow{Key: "Metadata directory", Value: assessmentMetadataDirFlag})
+	}
+	ux.PrintBanner("YugabyteDB Voyager — Migration Assessment", bannerRows)
+	log.Infof("assessing for migration to target YugabyteDB version %s", targetDbVersion)
+	ux.PrintSectionHeader("Preflight Checks")
+
 	var validatedReplicaEndpoints []srcdb.ReplicaEndpoint
 
-	if assessmentMetadataDirFlag == "" { // only in case of source connectivity
+	hasSourceConnectivity := assessmentMetadataDirFlag == ""
```
makalaaneesh, line 263: small but nice readability improvement 👍

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -221,141 +222,220 @@ func assessMigration() (err error) {
...
+		tracker.StartStage("Gathering metadata", gatherStepCount)
+	}
+	initAssessmentDB()
+	err = gatherAssessmentMetadata(validatedReplicaEndpoints, func(detail string) {
+		if detail != "Complete" {
+			tracker.IncrementStep(detail)
+		}
+	}, gatherStepCount)
 	if err != nil {
+		tracker.Finish()
```
makalaaneesh, line 374: whenever there is an error, why do we call tracker.Finish()? Shouldn't we fail the stage/sub-step?

yb-voyager/src/migassessment/gather.go
```diff
@@ -195,12 +212,16 @@ func (pt *progressTracker) finalizeAndCheckResults(resultChan <-chan collectionR
 // GatherAssessmentMetadataFromPG collects metadata from PostgreSQL primary and (optionally) replicas.
 // If no replicas are provided, uses a simpler sequential path.
 // If replicas are provided, collection is performed in parallel for better performance.
+// The optional progressFn callback receives sub-step descriptions as they happen;
+// when provided, direct stdout output from the script is suppressed.
 func GatherAssessmentMetadataFromPG(
```
makalaaneesh, line 217: this reminds me: we should just test oracle once. We don't necessarily have to improve the UI for that (if that's too much work), but we should at least test it to make sure it's not broken.

yb-voyager/src/migassessment/gather.go
```diff
@@ -437,6 +473,69 @@ func runGatherAssessmentMetadataScript(scriptPath string, envVars []string, work
...
+
+	err = cmd.Start()
+	if err != nil {
+		return fmt.Errorf("error starting gather assessment metadata script: %w", err)
+	}
+
+	go func() {
+		scanner := bufio.NewScanner(stderr)
+		for scanner.Scan() {
+			log.Errorf("[stderr of script]: %s", scanner.Text())
```
makalaaneesh, line 511: cursor: This is the riskier divergence: in the old flow, stderr from psql / pg_dump ("password authentication failed", ERROR: permission denied for schema X, etc.) was both logged and shown to the user. In the new flow, that information only lands in logs/yb-voyager-assess-migration.log. If the script errors out, the user just sees a generic wrapped error like "error waiting for gather assessment metadata script to complete: exit status 1" in the terminal — they have to open the log to find the real cause. Worth being aware of; this is a UX regression on the error path that's the flip-side of the "clean terminal" win on the success path.

yb-voyager/src/migassessment/gather.go
```diff
@@ -437,6 +473,69 @@ func runGatherAssessmentMetadataScript(scriptPath string, envVars []string, work
 	return nil
 }
 
+// runGatherAssessmentMetadataScriptWithProgress runs the script, captures stdout, detects sub-steps
+// and sends them to progressFn. If progressFn is nil, falls back to printing stdout directly.
+func runGatherAssessmentMetadataScriptWithProgress(
```
makalaaneesh, line 478: do we really need a separate function? Can we merge this with the existing function? If the `progressFn` is nil, we can just fallback to the previous behavior (of printing to stdout). Do you see any challenges in this?

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: same for stages: for example: 
```
err := tracker.RunStage("Assessing migration", 0, runAssessment); 
if  err != nil {
    return fmt.Errorf("failed to assess migration: %w", err)
}
```

The RunStage takes care of running and interpreting error/complete and updating the tracker.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -221,141 +222,220 @@ func assessMigration() (err error) {
...
+	if hasReplicas {
+		// Parallel replica path uses its own internal progress display,
+		// so set up the stage without a spinner to avoid a duplicate line.
+		tracker.PrepareStage("Gathering metadata", gatherStepCount)
+	} else {
+		tracker.StartStage("Gathering metadata", gatherStepCount)
+	}
+	initAssessmentDB()
+	err = gatherAssessmentMetadata(validatedReplicaEndpoints, func(detail string) {
+		if detail != "Complete" {
```
makalaaneesh, line 369: How about we pass an interface of a progressTracker instead of a callback function? 
Let `gatherAssessmentMetadata` call `StartStage` and figure out all the steps and what not. 

Furthermore, pass along a `ProgressTracker` interface even to `migassessment.GatherAssessmentMetadataFromPG`

yb-voyager/src/migassessment/gather.go
```diff
@@ -65,16 +65,20 @@ type progressTracker struct {
...
 	statuses     map[string]*NodeProgress
+	stepCounts   map[string]int // per-node completed step count
+	stepsTotal   int            // total steps per node (0 = unknown)
 	mutex        sync.Mutex
 	initialized  bool // Whether initial lines have been printed
 	maxNameLen   int  // Maximum length of display names for alignment
 }
 
-func newProgressTracker(nodes []collectionNode) *progressTracker {
+func newProgressTracker(nodes []collectionNode, stepsTotal int) *progressTracker {
```
makalaaneesh, line 75: we should discuss and reconcile what this progress tracker means in the context of the new progress tracker. 

The responsibilities of the two progress trackers are not really clear to me.

AmeyaKasture: The new progress tracker is invoked when for the read replicas as we have  to keep count for each cluster and check the name of each cluster for sizing.
So new progress tracker acts as an initialisation function

yb-voyager/src/migassessment/gather.go
AmeyaKasture: That would work too

yb-voyager/src/ux/progress.go
```diff
@@ -0,0 +1,235 @@
...
+	t.stopSpinnerLocked()
+
+	if t.subStage != "" {
+		counter := t.stepCounter()
+		CheckColor.Printf("    ✓ %s %s\n", counter, t.subStage)
+		t.linesPrinted++
+	}
+
+	if t.linesPrinted > 0 {
+		fmt.Printf("\033[%dA\033[J", t.linesPrinted)
```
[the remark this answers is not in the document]
makalaaneesh: @AmeyaKasture this is what we were discussing. (Not a blocker, we can take it up in a subsequent PR)

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -221,141 +222,220 @@ func assessMigration() (err error) {
...
 		replicaDiscoveryInfo, err := migassessment.HandleReplicaDiscoveryAndValidation(&source, sourceReadReplicaEndpoints, primaryOnly)
 		if err != nil {
+			ux.PrintPreflightFail("Replica discovery and validation")
 			return fmt.Errorf("failed to handle replica discovery and validation: %w", err)
 		}
 		validatedReplicaEndpoints = replicaDiscoveryInfo.ValidatedReplicas
-
-		// Store for callhome (including error scenarios)
 		replicaDiscoveryInfoForCallhome = &replicaDiscoveryInfo
+		ux.PrintPreflightCheck("Replica discovery and validation")
```
makalaaneesh, line 330: **Verbose output from `HandleReplicaDiscoveryAndValidation` undermines the clean Preflight Checks block.**

`HandleReplicaDiscoveryAndValidation` in `yb-voyager/src/migassessment/replicas.go` calls `utils.PrintAndLogfInfo` / `...Success` / `...Warning` ~20 times (lines 53, 58, 79, 114, 145, 169, 176, 185–200, 213–241, 290–293, 316–343, 371, 380, 415) and then this site prints a single `✓ Replica discovery and validation` line on top of all that.

Result: users see the verbose legacy output **and** the new summary line, so the Preflight Checks section isn't actually a clean checklist. Same pattern applies to `CheckAssessmentPermissionsOnAllNodes` a few lines below (see `yb-voyager/src/migassessment/permissions.go:76, 100, 102, 190–266`).

Suggestion — either:
- thread a `quiet bool` / progress callback through these helpers so their internal prints go log-only when a preflight summary is in use; or
- drop the summary `PrintPreflightCheck(...)` here and let the helpers' own output stand.

Leaving both is the worst of both worlds.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -221,141 +222,220 @@ func assessMigration() (err error) {
...
 
-		// Check permissions on all nodes (primary + replicas) after validation
+		// Permissions
 		if source.RunGuardrailsChecks {
-			// Check schema usage permissions first (no-op for non-PostgreSQL databases)
 			checkIfSchemasHaveUsagePermissions()
```
makalaaneesh: **`checkIfSchemasHaveUsagePermissions()` bypasses the new preflight summary.**

`checkIfSchemasHaveUsagePermissions` (in `yb-voyager/cmd/exportData.go:859-875`) calls `utils.ErrExit` on failure. Under the new preflight layout a permission failure here will print a raw ErrExit message instead of `✗ Permissions verified on all nodes`, and the subsequent `CheckAssessmentPermissionsOnAllNodes` summary line never prints because the process is already gone.

Per `yb-voyager/cmd/.cursor/BUGBOT.md`: *"Command-level functions should return errors, not call `utils.ErrExit` directly."*

Refactor `checkIfSchemasHaveUsagePermissions` to return an error so the caller can emit the correct `PrintPreflightFail` and return cleanly. This is pre-existing code, but the new preflight contract makes the latent issue user-visible.

yb-voyager/src/migassessment/gather.go
```diff
@@ -385,6 +403,24 @@ func findGatherMetadataScriptPath(dbType string) (string, error) {
...
+		log.Warnf("could not locate PG gather script to count steps: %v", err)
+		return 0
+	}
+	scriptDir := filepath.Dir(scriptPath)
+	matches, err := filepath.Glob(filepath.Join(scriptDir, "*.psql"))
+	if err != nil {
+		log.Warnf("could not glob .psql files in %s: %v", scriptDir, err)
+		return 0
+	}
+	return len(matches) + 1 // +1 for schema collection (pg_dump)
```
makalaaneesh, line 428: **Step count overshoots on replica nodes.**

`CountPGGatherSteps()` returns `len(*.psql) + 1` — i.e. 12 with the current 11 `.psql` files, +1 for the schema dump. But the gather shell script `yb-voyager-pg-gather-assessment-metadata.sh` only emits `"Collecting schema information..."` on the primary node:

```bash
if [ "$source_node_name" == "primary" ]; then
    ...
    print_and_log "INFO" "Collecting schema information..."
else
    log "INFO" "Skipping schema collection for replica '$source_node_name'"
fi
```

So replica nodes only emit 11 "Collecting" lines. With `stepsTotal = 12` applied to every node, replica counters peak at `(11/12)` and then jump to `(12/12) Complete`. Small but user-visible inconsistency.

Options:
- pass the primary/replica bit to the tracker so replicas use `stepsTotal - 1`;
- count `"Collecting"` lines actually emitted per node (defer the denominator until first completion); or
- treat schema collection as a primary-only extra stage outside the per-node counter.

makalaaneesh (commented):
[cursor] Review from branch-review pass — 6 `Should fix` items. Most are about the new Preflight / Progress UX leaking its details into older helpers, plus a latent concurrency nit in `ProgressTracker`. See inline comments.

yb-voyager/src/ux/progress.go
```diff
@@ -0,0 +1,235 @@
...
+	t.stopSpinnerLocked()
+}
+
+// stopSpinnerLocked signals the spinner goroutine to stop and waits for it.
+// The mutex is temporarily released to avoid deadlock since the spinner
+// goroutine also acquires it on each frame.
+func (t *ProgressTracker) stopSpinnerLocked() {
+	if t.spinnerStop != nil {
+		close(t.spinnerStop)
+		t.mu.Unlock()
```
makalaaneesh, line 213: **Latent double-close race in `stopSpinnerLocked`.**

This method temporarily releases `t.mu` while waiting on `spinnerDone`, then re-acquires it to zero out `spinnerStop`. If any other goroutine grabs `t.mu` during that release window, it will see a still-non-nil `spinnerStop`, call `close(t.spinnerStop)` on an already-closed channel, and panic.

Today the tracker is driven by a single external goroutine (the scanner loop in `gather.go` / the main `assessMigration` goroutine), so this doesn't trigger in practice. But the mutex nominally protects exactly this invariant, and the existing doc comment (*"The mutex is temporarily released to avoid deadlock since the spinner goroutine also acquires it on each frame."*) doesn't spell out the single-producer contract.

Fixes, in order of preference:
1. Zero `t.spinnerStop` (and the other fields) **before** the `Unlock`, then wait on `spinnerDone` without the lock. The nil-set doesn't need the spinner to have exited; subsequent callers will just treat the spinner as stopped.
2. Keep `t.mu` held for the full teardown and have `runSpinner` use a separate lock / condition so it doesn't contend with public API calls.
3. Document the single-producer contract explicitly so future readers don't regress this.

(1) is the minimal change.

yb-voyager/cmd/root.go
```diff
@@ -58,6 +58,7 @@ var (
 	currentCommand                     string
 	callHomeErrorOrCompletePayloadSent bool
 	controlPlaneConfig                 map[string]string // Holds control plane configuration from config file
+	suppressInfoMessages               bool              // set by commands that render their own banner (e.g. assess-migration)
```
makalaaneesh, line 61: **New global `suppressInfoMessages` violates the package-level layering rules.**

`yb-voyager/.cursor/BUGBOT.md`: *"Avoid introducing new global variables. Prefer passing dependencies explicitly through function arguments or struct fields."*

Two concrete concerns:
1. It's a package-level bool mutated in `PersistentPreRun` and read from three unrelated helper sites (`initConfig`, `validateExportDirFlag`, `retrieveMigrationUUID`).
2. It's keyed on the brittle literal `cmd.Name() == "assess-migration"` (line 83). If the command is ever renamed, the banner **and** the old info lines will silently both appear again.

Alternatives:
- Move the three info-prints into the commands that actually want them and out of the shared helpers (each command decides whether to print its own "Using config file" line).
- Attach a per-command capability via cobra annotations (e.g. `cmd.Annotations["ownsBanner"] = "true"`) and have the helpers read the *current* command instead of a mutable global.

yb-voyager/src/ux/ux.go
```diff
@@ -0,0 +1,128 @@
...
+//
+//	┌──────────────────────────────────────────────────────────┐
+//	│  YugabyteDB Voyager — Migration Assessment              │
+//	├──────────────────────────────────────────────────────────┤
+//	│  Voyager version   main                                 │
+//	│  Target DB         YugabyteDB 2025.2.2.1                │
+//	│  Source             postgresql (localhost:5432/mydb)     │
+//	│  Export directory   /path/to/export                     │
+//	└──────────────────────────────────────────────────────────┘
+func PrintBanner(title string, rows []BannerRow) {
```
makalaaneesh, line 86: **No unit tests for the new `ux` package.**

`yb-voyager/.cursor/BUGBOT.md`: *"Use the `unit` build tag for unit tests ... Prefer table-driven tests with `t.Run(name, ...)` for multiple scenarios."*

The package has non-trivial formatting / alignment logic that's easy to regress and perfect for table-driven tests:
- `PrintBanner` — width math (`keyWidth`, `titleLen`, `rowLen`, `inner`), including edge cases like empty `rows`, a title longer than any row, and multi-byte characters in values.
- `spinnerLabel` / `stepCounter` / `stepCounterFor` / `completedLabel` — branch coverage for `stepsTotal == 0`, `stepsDone == 0`, and the `(i/N)` vs `(N steps)` vs bare-stage-name branches.
- `CompleteStage` — behavior when `linesPrinted == 0` vs `> 0`, and the `stepsDone = stepsTotal` override.

Add `yb-voyager/src/ux/progress_test.go` and `ux_test.go` under `//go:build unit`.

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: We'll probably push this out to a subsequent PR.

yb-voyager/src/ux/ux.go
makalaaneesh: This may or may not be useful. you can take a call.

yb-voyager/src/ux/progress.go
makalaaneesh: It's a bit confusing to see a mutex being unlocked and locked again mid-way. Sort of error-prone. 
Golang actually has an interesting approach to dealing with shared memory - https://go.dev/blog/codelab-share . 
It doesn't mean mutexes are wrong always, but worth considering the channels approach (blog above) here IMO. 

Also, currently, the ProgressTracker only supports a single user, right? Do we want to make it concurrency-safe? (for the replicas case?)

yb-voyager/src/migassessment/gather.go
makalaaneesh: for now, we can leave it as is. In future, if we have more use-cases for multiple parallel stages, we can bring that logic into the ux.ProgressTracker.

yb-voyager/cmd/root.go
makalaaneesh: take a call here. either approach is fine

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: we can pick this up in a subsequent PR

yb-voyager/src/ux/ux.go
```diff
@@ -0,0 +1,128 @@
...
+package ux
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/fatih/color"
+)
+
+var (
```
makalaaneesh, line 28: @priyanshi-yb I remember you also created a common file somewhere to store all the styles. We should bring them under the ux package. Can you pls point @AmeyaKasture to that file?
(again not in this PR, we can do it subsequently)

priyanshi-yb: yeah, its in `src/utils/logging.go` file

yb-voyager/cmd/assessMigrationCommand.go
AmeyaKasture: Banner is present in the UX package only 
In assess migration command we are just injecting it with key value pairs
For Every command we just need to give the key value pairs and it will print banner

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -221,141 +222,220 @@ func assessMigration() (err error) {
...
+	// ── Phase 2: Assessment pipeline ────────────────────────────────────
 	startEvent := createMigrationAssessmentStartedEvent()
 	controlPlane.MigrationAssessmentStarted(startEvent)
 
-	initAssessmentDB() // Note: migassessment.AssessmentDir needs to be set beforehand
+	tracker := ux.NewProgressTracker("Preparing for migration assessment")
 
-	err = gatherAssessmentMetadata(validatedReplicaEndpoints)
+	// Stage 1: Gather metadata + export schema + load into DB
+	gatherStepCount := migassessment.CountPGGatherSteps()
```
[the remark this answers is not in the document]
AmeyaKasture: Prioritising Postgres Changes for this PR

yb-voyager/src/ux/progress.go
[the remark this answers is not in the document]
AmeyaKasture: Yes, this can be taken in subsequent PR

yb-voyager/src/ux/progress.go
```diff
@@ -0,0 +1,235 @@
...
+		select {
+		case <-t.spinnerStop:
+			fmt.Printf("\r\033[K")
+			return
+		default:
+		}
+
+		frame := spinnerFrames[i%len(spinnerFrames)]
+		fmt.Printf("\r\033[K")
+		StageColor.Printf("%s%s %s", indent, frame, label)
```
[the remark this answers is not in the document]
AmeyaKasture: Added terminal width utility in ux.go

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -221,141 +222,220 @@ func assessMigration() (err error) {
...
+			ux.PrintPreflightCheck(fmt.Sprintf("Source DB version compatible (%s)", source.DB().GetVersion()))
 
-			// Check if required binaries are installed.
 			binaryCheckIssues, err := checkDependenciesForExport()
 			if err != nil {
+				ux.PrintPreflightFail("Required dependencies present")
 				return fmt.Errorf("failed to check dependencies for assess migration: %w", err)
 			} else if len(binaryCheckIssues) > 0 {
+				ux.PrintPreflightFail("Required dependencies present")
 				return goerrors.Errorf("\n%s\n%s", color.RedString("\nMissing dependencies for assess migration:"), strings.Join(binaryCheckIssues, "\n"))
```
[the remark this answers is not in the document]
AmeyaKasture: done

yb-voyager/cmd/assessMigrationCommand.go
AmeyaKasture: Done

yb-voyager/cmd/root.go
AmeyaKasture: Its a temporary field and as the UX migration for each field is done, it will finally be removed completely 
Would also help in migration of newer commands as we just add them to suppress info

yb-voyager/src/ux/ux.go
AmeyaKasture: Will move into the refractoring commit

yb-voyager/src/ux/ux.go
AmeyaKasture: Can add unit tests in subsequent PR

yb-voyager/src/ux/progress.go
AmeyaKasture: Since the process.go is a standalone process and has only two go routines to interact with, maintaining with mutex are a better idea
Cost of creating the whole channel interface exceeds its need and goes towards realm of overengineering
If in future we need to spawn multiple scanners simultaneously we can think of channel based approach

This works for replicas too as for replicas only gathering metadata happens parallely which is taken care by an internal progress tracker 

Thus i think the current progress tracker works correctly

yb-voyager/src/migassessment/gather.go
AmeyaKasture: We havent made any logical changes to oracle so wont be an issue , but will check and test it out on oracle db

yb-voyager/cmd/assessMigrationCommand.go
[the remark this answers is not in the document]
makalaaneesh: Let's verify this specifically when verifying oracle output manually.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -221,141 +222,219 @@ func assessMigration() (err error) {
...
+	// Stage 1: Gather metadata + export schema + load into DB
+	gatherStepCount := migassessment.CountPGGatherSteps()
+	hasReplicas := len(validatedReplicaEndpoints) > 0
+	if hasReplicas {
+		// Parallel replica path uses its own internal progress display,
+		// so set up the stage without a spinner to avoid a duplicate line.
+		tracker.PrepareStage("Gathering metadata", gatherStepCount)
+	} else {
+		tracker.StartStage("Gathering metadata", gatherStepCount)
+	}
```
makalaaneesh, line 369: (can be done in subsequent PR) we can move this bit of logic inside `gatherAssessmentMetadata` function as well. I don't see a reason this needs to be done outside, is there?

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: Got it, but i think another helper function (whether it is in cmd or ux) will be helpful wherein i can just call 
`printBanner(uuid, exportDir, source, target, ...)`. 
Every command will eventually call this method, which will use the ux pkg to print the banner. 

Makes sense? 
(Again, not blocker, we can do this in a subsequent PR).

yb-voyager/src/ux/progress.go
makalaaneesh: I still think mutex temporary unlock-lock is a bit weird, but i can't think of a better alternative right now. I'll think more and get back later. 

If we're saying that the expectation is that there will be only a single user of ProgressTracker (i.e. it is not concurrency-safe), then pls add a comment above `ProgressTracker` struct to signify that. 

For now, we can go ahead. We can revisit later if required.

makalaaneesh (approved):
LGTM
1. As discussed, pls verify oracle output manually before merging. 
2. Pls make a list of the follow-ups from this PR so that we don't miss anything. (we've discussed a bunch of improvements/refactorings here and there)

PR #3502, opened 2026-04-15
Send db oid in the source db details to callhome in assess/export payloads

priyanshi-yb:

---

| Component        


yb-voyager/src/srcdb/source.go
```diff
@@ -126,6 +128,24 @@ func (s *Source) FetchDBSystemIdentifier() {
 	}
 }
 
+// FetchDBID fetches and stores a numeric database identifier for call-home when the source exposes one; otherwise leaves 0.
```
makalaaneesh: Why not define this as in interface method? with different implementations as per the db type.

priyanshi-yb: Yeah, I was also wondering why not do it that way, but we are also doing something similar for fetching `source_db_system_identifier`. So I had added it here itself, but now when I think System identifier is probably a PG  thing, so I renamed it to `FetchPGDBSystemIdentifier` and DBID can be an interface function implemented by every db, as it's mostly available for every db.

makalaaneesh (approved):
LGTM

PR #3503, opened 2026-04-16
Fix test cleanup failures by terminating stale sessions before DROP DATABASE

Abhi-yb:

Test scripts were intermittently failing during cleanup because `DROP DATABASE IF EXISTS` would error out when active sessions (from a prior migration run) were still connected to the target database. This PR introduces a shared helper function `ysql_terminate_and_drop_database` in `functions.sh` that first calls `pg_terminate_backend` to kill all active connections to the target database, waits briefly for them to close, and then drops the database. All nine test scripts that perform target database cleanup are updated to use this helper instead of calling `DROP DATABASE` directly. This also fixes a couple of pre-existing quoting bugs where the semicolon was incorrectly placed inside the database name quotes (e.g., `"${TARGET_DB_NAME};"` instead of `"${TARGET_DB_NAME}"`).

No user-facing changes. This only affects internal CI/test scripts.

The affected CI pipelines were triggered and all pass without cleanup failures.

No.

No.

### DIFF
--- a/migtests/scripts/callhome/run-callhome-test.sh
+++ b/migtests/scripts/callhome/run-callhome-test.sh
@@ -143,7 +143,7 @@ main() {
     fi
 
     step "Create target database."
-    run_ysql yugabyte "DROP DATABASE IF EXISTS \"${TARGET_DB_NAME};\""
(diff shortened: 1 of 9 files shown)


ShivanshGahlot (approved):
LGTM

PR #3508, opened 2026-04-16
Support Partitioned table without PK but all its child partitions have PK by importing event via partition table based on use-partition-root config

### LINKED ISSUE #3507: [Voyager] Support Live migration for partitioned tables where root has no PK but child partitions do
Jira Link: [DB-21011](https://yugabyte.atlassian.net/browse/DB-21011)
### Description

Live migration does not work for partitioned tables where the root table has no primary key, but child partitions do — particularly when the partition key differs from the PK.

This is a valid and common PostgreSQL schema pattern, but Voyager rejects it at export time and, even if the no-pk guardrail is bypassed, fails at import time with `ERROR: there is no unique or exclusion constraint matching the ON CONFLICT specification (SQLSTATE 42P10`. This is because voyager inserts with `ON CONFLICT` (at-least once semantics) and the root table has no PK. 

## Reproducer: 
```
CREATE TABLE orders (
    id       SERIAL,
    region   TEXT NOT NULL,
    amount   NUMERIC
) PARTITION BY LIST (region);

-- Child partitions WITH a primary key (id is unique within each partition)
CREATE TABLE orders_us PARTITION OF orders FOR VALUES IN ('US');
ALTER TABLE orders_us ADD PRIMARY KEY (id);

CREATE TABLE orders_eu PARTITION OF orders FOR VALUES IN ('EU');
ALTER TABLE orders_eu ADD PRIMARY KEY (id);

CREATE TABLE orders_apac PARTITION OF orders FOR VALUES IN ('APAC');
ALTER TABLE orders_apac ADD PRIMARY KEY (id);

yb-voyager export data from source \
  --source-db-type postgresql \
  --source-db-host localhost \
  --source-db-port 5432 \
  --source-db-user postgres \
  --source-db-password postgres \
  --source-db-name mydb \
  --source-db-schema public \
  --export-dir /tmp/export \
  --export-type snapshot-and-changes \
  --table-list 'orders'

Table names without a Primary key: [public.orders]
Currently voyager does not support live-migration for tables without a primary key.
You can exclude these tables using the --exclude-table-list argument.


If the guardrail is removed
Even if the PK check in reportUnsupportedTablesForLiveMigration is bypassed, import fails at the target. Here is why:

Export side: Debezium captures WAL events for the leaf partitions (orders_us, orders_eu, orders_apac).
Rename: Voyager's Debezium sink renames all leaf partition events to the root table (orders) via the debezium.sink.ybexporter.tables.rename config.
Import side: The importer generates:
INSERT INTO public.orders (id, region, amount) VALUES ($1, $2, $3)
  ON CONFLICT (id) DO NOTHING
Failure: The root table orders has no unique constraint on (id), so PostgreSQL/YugabyteDB rejects the ON CONFLICT clause:
ERROR: there is no unique or exclusion constraint matching the ON CONFLICT specification
```

## Potential Solution

Add a --use-partition-root flag (default true for backward compatibility):

--use-partition-root=true (default) | Current behavior: leaf events renamed to root, ON CONFLICT uses root PK
-- | --
--use-partition-root=false | New: leaf events keep their child partition names, ON CONFLICT uses child PK, data inserted directly into child partitions on target

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

makalaaneesh:

This PR adds support for live migration of partitioned tables where the root table has no primary key but child partitions do. This is a valid and common PostgreSQL schema pattern that was previously unsupported.

**Problem:**
- Live migration fails for partitioned tables when root has no PK (e.g., `CREATE TABLE orders (...) PARTITION BY LIST (region)` with no PK)
- Even if child partitions have PKs (e.g., `ALTER TABLE orders_us ADD PRIMARY KEY (id)`), Voyager rejects it
- If the guardrail is bypassed, import fails with `ERROR: there is no unique or exclusion constraint matching the ON CONFLICT specification` because the root table has no unique constraint

**Solution:**
Added `--use-partition-root` flag (default `true` for backward compatibility):
- When `true` (default): Current behavior - insert via root table, `ON CONFLICT` uses root PK
- When `false`: New behavior - SQL targets child partitions directly using their PKs

**Key changes:**

1. **Export side (cmd/export.go, cmd/exportData.go):**
   - Relax PK validation for root tables when all leaf partitions have PKs
   - Export behavior is unchanged - events are still renamed to root table

2. **Debezium layer (Java):**
   - Add `partitionTable` field to `Record.java` to store original partition info
   - `KafkaConnectRecordParser.java` stores original partition name before renaming
   - `QueueSegment.java` writes `partition_schema_name` and `partition_table_name` fields in CDC events

3. **Go Event struct (src/tgtdb/event.go):**
   - Add `partitionSchemaName`, `partitionTableName` (private), and `PartitionTableTup` fields
   - Add `IsPartitionEvent()`, `shouldUsePartitionTable()`, and `GetEffectiveTableNameForSQL()` methods
   - SQL generation uses partition table name when `usePartitionRoot=false`
   - `GetPreparedStmtName()` uses effective table name to ensure prepared statements are named correctly for each partition

4. **Import side (cmd/import.go, cmd/importData.go):**
   - Add `--use-partition-root` flag to import data command
   - Store `TargetUsePartitionRoot` in `ImportDataStatusRecord` for persistence
   - Add `checkPartitionConsistency()` guardrail to verify partitions exist on target before import
   - Validation ensures flag is only used for PostgreSQL to YugabyteDB live migrations

5. **Target DB layer (src/tgtdb/):**
   - Add `UsePartitionRoot` field to `tconf` for passing the flag to event processing
   - Updated `ExecuteBatch()` in PostgreSQL and YugabyteDB to pass `usePartitionRoot` to SQL generation

6. **Config templates:**
   - Updated live-migration YAML templates to include `use-partition-root` flag

**Implementation approach:**
- Export side always renames partition events to root table (preserves current behavior)
- CDC events contain both root table name AND original partition name as metadata
- Import side reads the flag and decides whether SQL targets root or partition
- Flag setting is persisted in `ImportDataStatusRecord` for consistency
- Backward compatible - default behavior unchanged

**New command-line flag for import data:**

```
--use-partition-root  [PostgreSQL/YugabyteDB only] For partitioned tables during live migration:
  - true (default): insert data via the root table using the root table's PK for ON CONFLICT
  - false: insert data directly into child partitions using their PKs. Useful when the root table
    has no PK but child partitions do.
```

**Usage example:**
```bash
# Import targeting partitions directly
yb-voyager import data \
  --use-partition-root false
```

**Config file support:**
The flag can also be set in YAML config files under `import-data` section.

1. **Unit tests:** All existing unit tests pass (`go test -tags unit ./...`)
2. **Static analysis:** No issues from `go vet`
3. **Debezium build:** Maven compile successful
4. **Integration tests:** New dedicated test file `live_migration_partitions_integrations_test.go` with 6 tests:
   - `TestLiveMigrationPartitionedTableWithChildPK` - single-level partitions with child PKs
   - `TestLiveMigrationWithMultiLevelPartitioningWithChildTablesHasPK` - multi-level partition hierarchies
   - `TestLiveMigrationPartitionedWithChildPKAndPartitionsAcrossDifferentSchemas` - partitions across different schemas with PK guardrail validation
   - `TestLiveMigrationWithIterationsOnPartitionedTableWithChildPK` - iterative cutover with partitioned tables

No changes to callhome/yugabyted payloads.

Yes - the following on-disk structures are affected:

1. **Export and import data queue:** CDC events now include optional `partition_table_name` and `partition_schema_name` fields
   - **Backward compatible**: Old events without these fields work fine (empty strings)
   - **Forward compatible**: Old Voyager versions will ignore the new fields

2. **Import Data State:** `ImportDataStatusRecord` has a new field `TargetUsePartitionRoot`
   - **Backward compatible**: Defaults to `true` (existing behavior) if not present

Fixes #3507

[Slack Thread](https://yugabyte.slack.com/archives/C04MC0N7B1A/p1776326652537849?thread_ts=1776326652.537849&cid=C04MC0N7B1A)

<div><a href="https://cursor.com/agents/bc-6982bf95-7c97-553b-a268-afb6fab84f33"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-6982bf95-7c97-553b-a268-afb6fab84f33"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
```diff
@@ -216,6 +226,14 @@ protected void parseTable(Struct value, Struct sourceNode, Record r) {
             es.updateTableSchema(t);
         }
         r.t = t;
+
+        // Store original partition table if this was a renamed partition
+        if (wasRenamed) {
```
makalaaneesh: One assumption we're making here is that every rename is a partition-related rename. This is error-prone. 
We should find a way to make it explicit. 

Maybe pass in another config with "partition-leaf:partition-root" and irrespective of whether we're renaming or not, we look at that config and then add in the partition table to the event? 

Alternatively, if there is no other use-case for rename, just edit the rename flag to make it specific for partitions.

yb-voyager/cmd/exportData.go
```diff
@@ -322,6 +322,10 @@ func startNextIterationImportDataToTarget() {
 		cmd = append(cmd, "--target-db-user", currentMsr.TargetDBConf.User)
 	}
 
+	if !currentMsr.ImportUsePartitionRoot {
```
makalaaneesh: as we were discussing, do we really need this? 
If it's part of the config file, then this will automatically be propagated. 
if it's overridden by the CLI flag, then shouldn't the above block (289) cover it?

makalaaneesh: Ah, the above block won't handle it because it's not a global flag?

yb-voyager/cmd/exportData.go
```diff
@@ -2022,29 +2033,80 @@ func finalizeTableAndColumnList(finalTableList []sqlname.NameTuple) ([]sqlname.N
...
+		rootTuple, ok := tableToUnQuotedMap[rootQualified]
+		if !ok {
+			return nil, goerrors.Errorf("lookup root partition %q", rootQualified)
+		}
+		leaves, ok := rootToLeafPartitions.Get(rootTuple)
+		if !ok {
+			leaves = []sqlname.NameTuple{}
+		}
+		leaves = append(leaves, leafTuple)
+		rootToLeafPartitions.Put(rootTuple, append(leaves, leafTuple))
```
makalaaneesh: cursor: Every leaf is stored twice.  (2014, 2015)?

yb-voyager/cmd/import.go
```diff
@@ -117,6 +119,28 @@ func validateImportDataFlags() error {
...
+		// Read the export type from MSR since importType may not be set yet in PreRun
+		msr, err := metaDB.GetMigrationStatusRecord()
+		if err != nil {
+			return goerrors.Errorf("failed to get migration status record: %w", err)
+		}
+		exportTypeFromSource := msr.ExportTypeFromSource
+		if !changeStreamingIsEnabled(exportTypeFromSource) {
+			return goerrors.Errorf("--use-partition-root=false is only valid for live migration")
+		}
+		if sourceDBType != POSTGRESQL && sourceDBType != YUGABYTEDB {
```
makalaaneesh: shouldn't it depend only on target. Why does it matter whether what the source is?

yb-voyager/cmd/import.go
```diff
@@ -117,6 +119,28 @@ func validateImportDataFlags() error {
 	return nil
 }
 
+func validateImportUsePartitionRootFlag() error {
+	// --use-partition-root flag is only valid for live migration with PostgreSQL or YugabyteDB source
```
makalaaneesh, line 123: so we're not using partition root in the snapshot phase? (or for offline migrations)? 
let's think about this a little more. I think ultimately, we should do this for snapshot phase as well. 
 We should document this properly in the flag if we're going with this behavior.

yb-voyager/cmd/import.go
```diff
@@ -117,6 +119,28 @@ func validateImportDataFlags() error {
...
+		exportTypeFromSource := msr.ExportTypeFromSource
+		if !changeStreamingIsEnabled(exportTypeFromSource) {
+			return goerrors.Errorf("--use-partition-root=false is only valid for live migration")
+		}
+		if sourceDBType != POSTGRESQL && sourceDBType != YUGABYTEDB {
+			return goerrors.Errorf("--use-partition-root flag is only valid for PostgreSQL and YugabyteDB source databases")
+		}
+	}
+	return metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
+		record.ImportUsePartitionRoot = bool(importUsePartitionRoot)
```
makalaaneesh: Are we using the same field for both Import-data-to-source and import-data-to-target? That seems error-prone. 
can we use `ImportDataStatusRecord` ? ( i believe that is only for import-data-to-target.) . Let's not populate more fields in MSR

yb-voyager/cmd/import.go
```diff
@@ -297,6 +321,11 @@ func registerImportDataCommonFlags(cmd *cobra.Command) {
 	BoolVar(cmd.Flags(), &truncateSplits, "truncate-splits", true,
 		"Truncate splits after importing")
 	cmd.Flags().MarkHidden("truncate-splits")
+	BoolVar(cmd.Flags(), &importUsePartitionRoot, "use-partition-root", true,
+		"[PostgreSQL/YugabyteDB only] For partitioned tables during live migration:\n"+
+			"  - true (default): insert data via the root table using the root table's PK for ON CONFLICT.\n"+
```
makalaaneesh: "Using the root table's PK for on conflict" this is hard to understand for a user reading the help msg. We should probably remove this.

yb-voyager/cmd/importData.go
```diff
@@ -419,6 +424,69 @@ func checkTablesPresentInTarget(tablesToImport []sqlname.NameTuple) {
 	}
 }
 
+// checkPartitionConsistency verifies that partitions are the same between source and target
+// when use-partition-root=false is used during import. This is required because CDC events
+// will contain partition table names that must exist on the target.
+func checkPartitionConsistency(msr *metadb.MigrationStatusRecord, importTableList []sqlname.NameTuple) {
+	if importerRole != TARGET_DB_IMPORTER_ROLE {
```
makalaaneesh, line 433: why only for target-db-importer? 
Shouldn't we do this for source-db-importer as well? (i.e. check target/source)? Nothing stops the user from using this only for import-data-to-source.

yb-voyager/cmd/importData.go
```diff
@@ -419,6 +424,69 @@ func checkTablesPresentInTarget(tablesToImport []sqlname.NameTuple) {
...
+		return tableTup.TargetTableAvailable()
+	}
+	// Check each source partition exists on target
+	missingRootToLeafPartitions := make(map[string][]string)
+	for root, leaves := range rootToLeafPartitions {
+		if lo.ContainsBy(importTableList, func(t sqlname.NameTuple) bool {
+			return t.ForKey() == root
+		}) {
+			continue
+		}
```
makalaaneesh: i don't understand this condition. why continue if the root exists in the importTableList? It's possible for the root to exist, but not the leaves, no? 
in fact, is it ever possible that we'll not have the root on target at this point in the code?

yb-voyager/cmd/importData.go
```diff
@@ -419,6 +424,69 @@ func checkTablesPresentInTarget(tablesToImport []sqlname.NameTuple) {
...
+	if len(missingRootToLeafPartitions) > 0 {
+		utils.PrintAndLogfInfo("\nWhen using --use-partition-root=false, CDC events will contain partition table names.")
+		utils.PrintAndLogfInfo("The following root table partitions are not present on the target database:")
+		for root, leaves := range missingRootToLeafPartitions {
+			utils.PrintAndLogfInfo("  - %s:", root)
+			for _, leaf := range leaves {
+				utils.PrintAndLogfInfo("    - %s", leaf)
+			}
+		}
+		utils.ErrExit("Ensure that all partitions from the source exist on the target, or use --use-partition-root=true (default).")
```
makalaaneesh: I think it might be a good idea to prompt rather than exit here. It allows for flexibility wherein we're only exporting a subset of the partitions from source (i.e source has 10 partitions, target only has latest 2)

yb-voyager/cmd/importData.go
```diff
@@ -917,8 +985,13 @@ func initialiseImportTableList(importFileTasks []*ImportFileTask, msr *metadb.Mi
 		//and if not, we need to exit with an error
 		//If the export type includes snapshot, then only use the importFileTasks to get the tables to import
 		//otherwise use the tables from msr
-		tablesToImport := lo.Ternary(importSnapshotRequired(), importFileTasksToTableNameTuples(importFileTasks), importTableList)
-		checkTablesPresentInTarget(tablesToImport)
+		// tablesToImport := lo.Ternary(importSnapshotRequired(), importFileTasksToTableNameTuples(importFileTasks), importTableList)
```
makalaaneesh: why is this commented out?

yb-voyager/src/tgtdb/event.go
```diff
@@ -33,13 +33,17 @@ import (
...
+	Op                  string
+	TableNameTup        sqlname.NameTuple
+	PartitionTableName  string // Original partition table name (before renaming to root)
+	PartitionSchemaName string // Original partition schema name (before renaming to root)
+	UsePartitionTable   bool   // If true and IsPartitionEvent(), use partition table for SQL instead of root
+	Key                 map[string]*string
+	Fields              map[string]*string //all the column values of the row - worst
+	BeforeFields        map[string]*string //all the column values of the row - worst
+	ExporterRole        string
+	ImporterRole        string
```
makalaaneesh: Hmm, I don't like the idea of putting importerROle in the event.
I could be importing the same event to target/source-replica. What will the importerRole then be? 
It's not a property of the event really. `ExporterRole` on the other hand is a property of the event because that talks about the origin of the event.

makalaaneesh: Same for `UsePartitionTable`. IMO, both of these fields should be passed down to the event->sql logic.

yb-voyager/src/tgtdb/event.go
```diff
@@ -141,6 +156,39 @@ func (e *Event) IsCutoverEvent() bool {
...
+	return true
+}
+
+// GetEffectiveTableNameForSQL returns the table name to use in SQL statements.
+// If UsePartitionTable is true and this is a partition event, returns the partition table name.
+// Otherwise, returns the root table name (TableNameTup).
+func (e *Event) GetEffectiveTableNameForSQL() string {
+	if e.UsePartitionTable && e.IsPartitionEvent() && e.supportsPartitionTableIngestion() {
+		// Quote the schema and table names for SQL safety
+		return fmt.Sprintf(`"%s"."%s"`, e.PartitionSchemaName, e.PartitionTableName)
```
makalaaneesh: we should keep things consistent here. Can we get a NameTuple here? and use ForUserQuery()? Are there any hurdles in doing that?

yb-voyager/cmd/import.go
```diff
@@ -297,6 +321,11 @@ func registerImportDataCommonFlags(cmd *cobra.Command) {
 	BoolVar(cmd.Flags(), &truncateSplits, "truncate-splits", true,
 		"Truncate splits after importing")
 	cmd.Flags().MarkHidden("truncate-splits")
+	BoolVar(cmd.Flags(), &importUsePartitionRoot, "use-partition-root", true,
```
makalaaneesh, line 335: we should add this to config-templates

yb-voyager/src/tgtdb/event.go
```diff
@@ -141,6 +156,39 @@ func (e *Event) IsCutoverEvent() bool {
...
+
+// GetPartitionQualifiedName returns the fully qualified partition table name (schema.table)
+func (e *Event) GetPartitionQualifiedName() string {
+	if !e.IsPartitionEvent() {
+		return ""
+	}
+	return fmt.Sprintf("%s.%s", e.PartitionSchemaName, e.PartitionTableName)
+}
+
+func (e *Event) supportsPartitionTableIngestion() bool {
```
makalaaneesh: the only case that does not support is import-data-to-target and "u". Right? Can we write the logic like that?

Pls add enough comments to document why

yb-voyager/src/tgtdb/event.go
```diff
@@ -33,13 +33,17 @@ import (
...
-	Op           string
-	TableNameTup sqlname.NameTuple
-	Key          map[string]*string
-	Fields       map[string]*string //all the column values of the row - worst
-	BeforeFields map[string]*string //all the column values of the row - worst
-	ExporterRole string
+	Vsn                 int64 // Voyager Sequence Number
+	Op                  string
+	TableNameTup        sqlname.NameTuple
+	PartitionTableName  string // Original partition table name (before renaming to root)
```
makalaaneesh: what about conflict detection? 
what if there are unique indexes on the child, not the root? Will they be detected properly? (Perhaps this is a problem irrespective of this change?)

yb-voyager/src/tgtdb/event.go
makalaaneesh: I realise that this might be in slight contradiction to how it's currently implemented. Let's discuss this once.

yb-voyager/src/tgtdb/event.go
priyanshi-yb: No the conflict detection works, as we have already identified that issue earlier and fixed it by combining all the unique indexes of leafs on the root table in the metadata so conflict detection works
refer this test we have - `TestLiveMigrationWithUniqueKeyConflictWithUniqueIndexOnlyOnLeafPartitions`

yb-voyager/cmd/importData.go
priyanshi-yb: yeah this is a miss. The condition is not, so if the root table is not present in the table list for some reason than skip it as its not being migrated

yb-voyager/cmd/importData.go
priyanshi-yb: makes sense

yb-voyager/cmd/import.go
priyanshi-yb: Yeah, we are not right now, and it will be some work to modify the DatafileDescriptor to also give out the partition table name, and then we will be able to use it. So not doing it right now as it's not important. Modifying the flag description and comments to explain that we can enhance it later if required

yb-voyager/cmd/exportData.go
priyanshi-yb: It won't be populated directly because the CLI overrides above are for the export data command's global params. So we don't have the information for the import level flag CLI overrides right now here

yb-voyager/src/tgtdb/event.go
priyanshi-yb: The problem in directly doing the lookup on the partition table on the import side is a bit tricky because of the support where we allow changing the partitions on the target. I tried using other `LookupIgnore*` functions, but it isn't working. The only thing I can think of now is to look up only when we are getting the partition table here, since we know at this point that we are in `use-partition-root false` and the guardrail is already done?  Lets discuss this

yb-voyager/cmd/import.go
makalaaneesh: ok, let's document this in the command help as well so that this is clear.

makalaaneesh: what about import-data-to-source-replica? 
Just to limit scope, let's not add this as a flag in import-data-to-source-replica?

yb-voyager/cmd/exportData.go
makalaaneesh: (let's this up as a separate PR): as discussed, let's think about the config flow a bit more:
- import-data-to-source: 
  - if value for use-partition-root is specified in config-file under this command, we should not override it. 
  
- import-data-to-source/import-data-to-target: 
   - if it's the default OR same as in config-file, no need to specify it again in the CLI.

yb-voyager/src/tgtdb/event.go
makalaaneesh: Discussion: 

Event 
1. SourceCDCEvent
2. Transform to TargetCompatibleCDCEvent (that is ingested)
    - nameTuple (name registry) in importerRole=Target ensures that the table name/schema name is that of the target. 
    - we use debezium's schema registry along with valueCOnverter to convert that event's fields to a target-compatible versions. 
3. Ingest TargetCompatibleCDCEvent


Event 
    - table (name tuple)
    - partition (name tuple)
        - delay lookup until we actually have to? (event -> SQL logic, at which point, all guardrails have been done, and lookup is expected to always be succesful)

importConfig
    - use-partition-root = true/false

targetDB.ImportEventBatch(eventBatch, importConfig)
    - sql = event.ToSQL(use-partition-root)
    - (Limitation of YB)
        - for updateEvents, ignore the use-partition-root config, just pass false.

yb-voyager/src/testlivemigration/live_migration_integration_test.go
makalaaneesh, line 1: let's add an iterative cutover test as well. 
Also, move these tests to a separate file (specific for partitions?)

yb-voyager/src/testlivemigration/live_migration_integration_test.go
```diff
@@ -4425,5 +4425,560 @@ func TestLiveMigrationWithFallbackWithIterationsAndArchiveChangesAndEndMigration
...
+		},
+		InitialDataSQL: []string{
+			// Insert initial data into different partitions
+			`INSERT INTO public.orders (id, region, amount) VALUES (1, 'US', 100);`,
+			`INSERT INTO public.orders (id, region, amount) VALUES (2, 'US', 200);`,
+			`INSERT INTO public.orders (id, region, amount) VALUES (3, 'EU', 150);`,
+			`INSERT INTO public.orders (id, region, amount) VALUES (4, 'EU', 250);`,
+			`INSERT INTO public.orders (id, region, amount) VALUES (5, 'APAC', 300);`,
+		},
+		SourceDeltaSQL: []string{
```
makalaaneesh: let's add one DELETE in SourceDelta as well as TargetDelta in all tests.

yb-voyager/src/testlivemigration/live_migration_integration_test.go
```diff
@@ -4425,5 +4425,560 @@ func TestLiveMigrationWithFallbackWithIterationsAndArchiveChangesAndEndMigration
...
+					INSERT INTO public.orders (id, region, amount) VALUES (i+105, 'EU', i * 100);
+					INSERT INTO public.orders (id, region, amount) VALUES (i+205, 'APAC', i * 100);
+
+					UPDATE public.orders SET amount = amount + 1 WHERE id = i;
+					UPDATE public.orders SET amount = amount + 1 WHERE id = i+105;
+					UPDATE public.orders SET amount = amount + 1 WHERE id = i+205;
+				END LOOP;
+			END $$;`,
+		},
+		CleanupSQL: []string{
```
makalaaneesh: orthogonal question: do we have source/target specific cleanups? Does this run on both currently, or only source?

yb-voyager/src/testlivemigration/live_migration_integration_test.go
```diff
@@ -4425,5 +4425,560 @@ func TestLiveMigrationWithFallbackWithIterationsAndArchiveChangesAndEndMigration
...
+	testutils.FatalIfError(t, err, "failed to validate data consistency")
+
+	// Execute CDC operations
+	err = lm.ExecuteSourceDelta()
+	testutils.FatalIfError(t, err, "failed to execute source delta")
+
+	// Wait for streaming to complete
+	// CDC events are still tracked by root table name (table_name in event)
+	// But SQL targets partitions using partition_table_name when '--use-partition-root false'
+	// SourceDeltaSQL: insert 3, update 1, delete 1
```
makalaaneesh: cursor: stale comment?

yb-voyager/src/testlivemigration/live_migration_integration_test.go
```diff
@@ -4425,5 +4425,560 @@ func TestLiveMigrationWithFallbackWithIterationsAndArchiveChangesAndEndMigration
...
+		},
+		`"public"."customers"`: {
+			Inserts: 3,
+			Updates: 1,
+			Deletes: 1,
+		},
+	}, 30, 1)
+	testutils.FatalIfError(t, err, "failed to wait for streaming complete")
+
+	err = lm.ValidateDataConsistency([]string{`"public"."orders"`}, "id")
```
makalaaneesh: cursor: The cross-schema test sets up two partitioned hierarchies (public.orders and public.customers) and asserts streaming counts for both, but ValidateDataConsistency is only ever called on orders:

yb-voyager/src/testlivemigration/live_migration_integration_test.go
```diff
@@ -4425,5 +4425,560 @@ func TestLiveMigrationWithFallbackWithIterationsAndArchiveChangesAndEndMigration
...
+	testutils.FatalIfError(t, err, "failed to validate data consistency")
+
+	err = lm.InitiateCutoverToSource(nil)
+	testutils.FatalIfError(t, err, "failed to initiate cutover to source")
+
+	err = lm.WaitForCutoverSourceComplete(0, 160)
+	testutils.FatalIfError(t, err, "failed to wait for cutover source complete")
+
```
makalaaneesh: cursor: The other two tests do a final ValidateDataConsistency after WaitForCutoverSourceComplete; this one stops at WaitForCutoverSourceComplete only.

yb-voyager/src/testlivemigration/live_migration_integration_test.go
```diff
@@ -4425,5 +4425,560 @@ func TestLiveMigrationWithFallbackWithIterationsAndArchiveChangesAndEndMigration
...
+			`ALTER TABLE public.customers_other REPLICA IDENTITY FULL;`,
+			`ALTER TABLE public.customers_active REPLICA IDENTITY FULL;`,
+			`ALTER TABLE public.customers_arr_small REPLICA IDENTITY FULL;`,
+			`ALTER TABLE public.customers_part11 REPLICA IDENTITY FULL;`,
+			`ALTER TABLE public.customers_part12 REPLICA IDENTITY FULL;`,
+			`ALTER TABLE public.customers_arr_large REPLICA IDENTITY FULL;`,
+			`ALTER TABLE public.customers_part21 REPLICA IDENTITY FULL;`,
+			`ALTER TABLE public.customers_part22 REPLICA IDENTITY FULL;`,
+		},
+		InitialDataSQL: []string{
```
makalaaneesh: cursor: There's no INSERT in InitialDataSQL or SourceDeltaSQL that lands in customers_other (status OTHER). The first such row only appears via the partition-key-changing UPDATE inside SourceDeltaSQL (UPDATE … status = 'OTHER' WHERE id = 4). That's fine functionally, but it means the snapshot path through a leaf in a different schema than the root is not exercised at all — the snapshot has zero rows for customers_other.

yb-voyager/src/testlivemigration/live_migration_integration_test.go
```diff
@@ -4425,5 +4425,560 @@ func TestLiveMigrationWithFallbackWithIterationsAndArchiveChangesAndEndMigration
...
+				END LOOP;
+			END $$;`,
+
+			`INSERT INTO public.customers (id, name, email, status, arr) VALUES (5, 'ABC', 'active@example.com', 'ACTIVE', 100);`,
+			`INSERT INTO public.customers (id, name, email, status, arr) VALUES (6, 'DEF', 'recurring@example.com', 'RECURRING', 105);`,
+			`UPDATE public.customers SET status = 'OTHER' WHERE id = 6 AND arr = 105;`,
+			`UPDATE public.customers SET status = 'REACTIVATED' WHERE id = 5 AND arr = 100;`,
+		},
+		CleanupSQL: []string{
+			`DROP TABLE IF EXISTS public.orders CASCADE;`,
```
makalaaneesh: cursor: customers cleanup is missing

yb-voyager/src/testlivemigration/live_migration_integration_test.go
```diff
@@ -4425,5 +4425,560 @@ func TestLiveMigrationWithFallbackWithIterationsAndArchiveChangesAndEndMigration
...
+	err = lm.InitiateCutoverToTarget(true, nil)
+	testutils.FatalIfError(t, err, "failed to initiate cutover to target")
+
+	err = lm.WaitForCutoverComplete(0, 30)
+	testutils.FatalIfError(t, err, "failed to wait for cutover complete")
+
+	err = lm.ExecuteTargetDelta()
+	testutils.FatalIfError(t, err, "failed to execute target delta")
+
+	err = lm.WaitForFallbackStreamingComplete(map[string]ChangesCount{
```
makalaaneesh: comments pls to explain how we DELETE/INSERTS because of change in partition-key in an UPDATE

yb-voyager/src/testlivemigration/live_migration_integration_test.go
makalaaneesh: test ideas for validation/guardrails: ( I think it would be useful because this is an important guardrail) 
- validation of use-partition-root failure in case partitions are not present on target. 
- test guardrail : A schema where the root has no PK and at least one leaf also has no PK

makalaaneesh (commented):
some comments on tests

yb-voyager/cmd/importData.go
```diff
@@ -419,6 +425,67 @@ func checkTablesPresentInTarget(tablesToImport []sqlname.NameTuple) {
...
+
+	// Get list of partitions from MSR (source partitions)
+	rootToLeafPartitions := make(map[string][]string)
+	for leaf, root := range msr.SourceRenameTablesMap {
+		rootToLeafPartitions[root] = append(rootToLeafPartitions[root], leaf)
+	}
+
+	checkIfTableExistsOnTarget := func(table string) bool {
+		// Try to lookup the partition in name registry
+		tableTup, err := namereg.NameReg.LookupTableNameAndIgnoreIfTargetNotFoundBasedOnRole(table)
```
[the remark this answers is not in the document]
priyanshi-yb: Not applicable

yb-voyager/src/testlivemigration/live_migration_integration_test.go
priyanshi-yb: Both

yb-voyager/src/testlivemigration/live_migration_integration_test.go
priyanshi-yb: It does have an INSERT for the `customers_other` partition with status `OTHER` 
```
INSERT INTO public.customers (id, name, email, status, arr) VALUES (4, 'OTHER', 'other@example.com', 'OTHER', 150);

```

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
```diff
@@ -69,6 +73,30 @@ private void retrieveRenameTablesFromConfig() {
                 renameTables.put(before, after);
             }
         }
+        if (!partitionToRootMappingsConfig.isEmpty()) {
+            for (String partitionToRootMappingConfig : partitionToRootMappingsConfig.split(",")) {
+                String[] beforeAndAfter = partitionToRootMappingConfig.split(":");
```
makalaaneesh: "beforeAndAfter" name is not really applicable here because this is not a rename. "partitionAndRoot" probably?

yb-voyager/src/testlivemigration/live_migration_integration_test.go
priyanshi-yb: Added iteration, validation of test guardrail for non-PK tables 
Will try to add the validation of use-partition-root failure

yb-voyager/cmd/import.go
```diff
@@ -117,6 +119,40 @@ func validateImportDataFlags() error {
...
+	}
+	tconf.UsePartitionRoot = bool(importUsePartitionRoot)
+	return metaDB.UpdateImportDataStatusRecord(func(record *metadb.ImportDataStatusRecord) {
+		switch importerRole {
+		case TARGET_DB_IMPORTER_ROLE:
+			record.TargetUsePartitionRoot = bool(importUsePartitionRoot)
+		case SOURCE_DB_IMPORTER_ROLE:
+			record.SourceUsePartitionRoot = bool(importUsePartitionRoot)
+		case SOURCE_REPLICA_DB_IMPORTER_ROLE:
+			record.SourceReplicaUsePartitionRoot = bool(importUsePartitionRoot)
```
makalaaneesh: Ideally, we should have separate status records for each command. This will soon get messy (like the MSR) if we proceed in this direction.

Is there a real reason we need to store this for SOURCE_DB_IMPORTER_ROLE and SOURCE_REPLICA_DB_IMPORTER_ROLE? Can we just store it for TARGET NOW?

yb-voyager/cmd/exportData.go
```diff
@@ -2022,29 +2042,84 @@ func finalizeTableAndColumnList(finalTableList []sqlname.NameTuple) ([]sqlname.N
...
 		utils.PrintAndLogf("Table names without a Primary key: %s", nonPKTables)
 		utils.ErrExit("Currently voyager does not support live-migration for tables without a primary key.\n" +
 			"You can exclude these tables using the --exclude-table-list argument.")
 	}
 }
 
+// buildRootToLeafPartitionsMap inverts the leaf->root rename map (qualified.Unquoted strings)
+// into a NameTuple-keyed map of root -> []leaf partitions, resolving each name through the name
+// registry so callers can compare against NameTuples coming from finalTableList.
+func buildRootToLeafPartitionsMap(partitionsToRootTableMap map[string]string, finalTableList []sqlname.NameTuple) (*utils.StructMap[sqlname.NameTuple, []sqlname.NameTuple], error) {
```
makalaaneesh, line 2098: One concern: Can this be problematic on resumption in the scenario wherein partition strategy is different across source/target? 
I guess not (because we're somehow getting the `finalTableList` (which as i understand contains both roots and leaves) so it should not be a problem. But just wanted to double-check

yb-voyager/cmd/exportData.go
```diff
@@ -2022,29 +2042,84 @@ func finalizeTableAndColumnList(finalTableList []sqlname.NameTuple) ([]sqlname.N
...
+	tableToUnQuotedMap := make(map[string]sqlname.NameTuple)
```
makalaaneesh: this name is a bit misleading, because the map is from string(unquoted) -> nametuple. 

Can we perhaps call it `finalTableListMap` or `finalTableListUnquotedMap`?

yb-voyager/cmd/importData.go
```diff
@@ -917,8 +984,12 @@ func initialiseImportTableList(importFileTasks []*ImportFileTask, msr *metadb.Mi
 		//and if not, we need to exit with an error
 		//If the export type includes snapshot, then only use the importFileTasks to get the tables to import
 		//otherwise use the tables from msr
-		tablesToImport := lo.Ternary(importSnapshotRequired(), importFileTasksToTableNameTuples(importFileTasks), importTableList)
-		checkTablesPresentInTarget(tablesToImport)
+		checkTablesPresentInTarget(importTableList) //to check whether tables exist or not we should use importTableList in live migration case
```
makalaaneesh: we seem to have removed this `tablesToImport`. I'm not clear on why

yb-voyager/cmd/importData.go
```diff
@@ -419,6 +425,67 @@ func checkTablesPresentInTarget(tablesToImport []sqlname.NameTuple) {
...
+		}
+		return tableTup.TargetTableAvailable()
+	}
+	// Check each source partition exists on target
+	missingRootToLeafPartitions := make(map[string][]string)
+	for root, leaves := range rootToLeafPartitions {
+		if !lo.ContainsBy(importTableList, func(t sqlname.NameTuple) bool {
+			return t.ForKey() == root
+		}) {
+			//if the root table is not in the import table list, then skip the check
```
makalaaneesh, line 474: so importTableList will by default have all roots, and no leaves?

yb-voyager/cmd/importData.go
```diff
@@ -419,6 +425,67 @@ func checkTablesPresentInTarget(tablesToImport []sqlname.NameTuple) {
 	}
 }
 
+// checkPartitionConsistency verifies that partitions are the same between source and target
+// when '--use-partition-root false' is used during import. This is required because CDC events
+// will contain partition table names that must exist on the target.
+func checkPartitionConsistency(msr *metadb.MigrationStatusRecord, importTableList []sqlname.NameTuple) {
```
makalaaneesh, line 432: what all roles do we want to run this for? 
for target_db_importer_role, for sure. 
for source_db_importer_role, do we need it? In this function, we seem to be assuming that we're only doing it for target_db_importer_role. If so, that should be the first check in this function, and return for other roles.

yb-voyager/src/tgtdb/event.go
```diff
@@ -141,39 +156,61 @@ func (e *Event) IsCutoverEvent() bool {
...
 
-func (e *Event) GetSQLStmt(tdb TargetDB) (string, error) {
+// IsPartitionEvent returns true if this event originated from a partition table that was renamed to root
+func (e *Event) IsPartitionEvent() bool {
+	return e.partitionSchemaName != "" && e.partitionTableName != ""
+}
+
+func (e *Event) GetSQLStmt(tdb TargetDB, usePartitionTable bool) (string, error) {
+	var err error
+	if usePartitionTable {
```
makalaaneesh: `if e.IsPartitionEvent() && usePartitionTable` ?

yb-voyager/src/tgtdb/event.go
```diff
@@ -141,39 +156,61 @@ func (e *Event) IsCutoverEvent() bool {
...
 	default:
 		panic("unknown op: " + e.Op)
 	}
 }
 
-func (e *Event) GetPreparedSQLStmt(tdb TargetDB, targetDBType string) (string, error) {
-	psName := e.GetPreparedStmtName()
+func (e *Event) GetPreparedSQLStmt(tdb TargetDB, targetDBType string, usePartitionTable bool) (string, error) {
+	var err error
+	if usePartitionTable {
```
makalaaneesh: `if e.IsPartitionEvent() && usePartitionTable` ?

yb-voyager/src/tgtdb/event.go
```diff
@@ -221,11 +258,17 @@ func (e *Event) GetParamsString() string {
...
-func (event *Event) GetPreparedStmtName() string {
+func (event *Event) GetPreparedStmtName(usePartitionTable bool) string {
 	// Build the base identifier (table name + columns for updates)
 	// This is what we'll hash if needed
 	var baseIdentifier strings.Builder
-	baseIdentifier.WriteString(event.TableNameTup.ForUserQuery())
+	// Use the effective table name for SQL to ensure prepared statement matches the target table
+	// This is important when UsePartitionTable is true - the prepared statement must be for the partition
+	if usePartitionTable {
+		baseIdentifier.WriteString(event.PartitionTableTup.ForUserQuery())
```
makalaaneesh: this could cause a panic if `PartitionTableTup` is nil (since it's lazily populated). we should validate and throw proper error. pls take care of this everywhere.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1055,6 +1055,20 @@ func (yb *TargetYugabyteDB) RestoreSequences(sequencesLastVal *utils.StructMap[s
...
+but with UPDATE <partition table> stmt on YB , there is a limiation that it errors out if the UPDATE statement doesn't include partition key so we are skipping
+ingestion of UPDATE events via partition table on Target DB. and in other importers we are ingesting data via partition table.
+*/
+func (yb *TargetYugabyteDB) shouldUsePartitionTable(event *Event) bool {
+	if yb.tconf.UsePartitionRoot || !event.IsPartitionEvent() {
+		//if the use-partition-root is true or the event is not a partition table event, we don't use the partition table
+		return false
+	}
+	return event.Op != "u"//
+}
```
makalaaneesh: I think it might be cleaner to pass `usePartitionLeafIfApplicable` (or something like that) to the Event.GetSQL* methods, and let it figure out whether it is applicable or not (i.e. let the event methods check for isPartitionEvent() first before even using the provided configuration . 


So, how about this: 
Event.GetSQL(usePartitionRoot) --- 
   - internally checks if it is even a partitioned table, and only then uses this config. 
   
For all cases, then, you simply pass on the config value provided by user. 
For YUgabyteDB, and UPDATES, you hard-code it to use partitionroot. 

Does that make sense?

yb-voyager/src/tgtdb/event.go
```diff
@@ -221,11 +258,17 @@ func (e *Event) GetParamsString() string {
 // NOTE: Prepared statements are currently used ONLY for INSERT and DELETE operations.
 // UPDATE operations use direct SQL (not prepared statements)
 // See ExecuteBatch() in yugabytedb.go/postgres.go where event.Op == "u" uses GetSQLStmt().
-func (event *Event) GetPreparedStmtName() string {
+func (event *Event) GetPreparedStmtName(usePartitionTable bool) string {
```
makalaaneesh: it might be a good idea to not flip the condition logic (to help with readability). 
I.e. keep it as UsePartitionRoot here as well instead of flipping it to UsePartitionTable. 

That makes it less error-prone.

yb-voyager/cmd/importData.go
makalaaneesh: >if the root table is not present in the table list for some reason

When is this possible? it seems like for live migration, it shouldn't be possible? isn't that correct?

yb-voyager/src/tgtdb/event.go
makalaaneesh: nice!

yb-voyager/cmd/exportData.go
priyanshi-yb: I don't think there will be any problems, as we are using the leaf-to-root map, which is stored in the first run, so we are not calling the DB for anything here.

yb-voyager/cmd/importData.go
makalaaneesh: let's add a TODO (not critical for the release, but good to have this validation even for import-data-to-source (where we check consistency between source and target).

yb-voyager/cmd/importData.go
makalaaneesh: pls comment - that this is not possible today in live migration (because we don't allow changing table-list), but just as a saftey measure, we have this logic. (could also just put a log)

yb-voyager/cmd/importData.go
makalaaneesh: let's remove the comment (stale) on 985

yb-voyager/cmd/importData.go
makalaaneesh: as discussed in the previous comment, let's restrict this function to target_db_importer_role for now, and then add a TODO for other roles later.

yb-voyager/src/tgtdb/event.go
makalaaneesh: you could use your previous method `GetEffectiveTableNameTup(usePartitionRoot) (nametuple, error)` to sort of have a common place for the check and the if-else

yb-voyager/cmd/importData.go
```diff
@@ -419,6 +425,67 @@ func checkTablesPresentInTarget(tablesToImport []sqlname.NameTuple) {
...
+			log.Warnf("Partition %s from source not found in name registry: %v", table, err)
+			return false
+		}
+		return tableTup.TargetTableAvailable()
+	}
+	// Check each source partition exists on target
+	missingRootToLeafPartitions := make(map[string][]string)
+	for root, leaves := range rootToLeafPartitions {
+		if !lo.ContainsBy(importTableList, func(t sqlname.NameTuple) bool {
+			return t.ForKey() == root
```
makalaaneesh: [cursor] **`checkPartitionConsistency` silently no-ops for case-sensitive partitioned tables.**

`t.ForKey()` returns the **Quoted** form of the qualified name (see `yb-voyager/src/utils/sqlname/nametuple.go:218-221`), but `root` here comes from `msr.SourceRenameTablesMap`, whose contract is documented as `table.Qualified.Unquoted` (see `yb-voyager/src/metadb/migrationStatus.go:94`).

For any partitioned table whose name needs quoting (e.g. `public."Orders"`), the comparison `t.ForKey() == root` will never match, so the loop hits `continue` and the per-leaf existence check is skipped entirely. The whole guardrail that's supposed to protect `--use-partition-root false` against a target with missing partitions becomes a no-op for case-sensitive identifiers.

Per the workspace `BUGBOT.md`: “all object name handling must go through the `sqlname` package” and “always test new table/column/schema handling with case-sensitive names”.

Suggested fix:

```go
if !lo.ContainsBy(importTableList, func(t sqlname.NameTuple) bool {
    return t.AsQualifiedCatalogName() == root  // Qualified.Unquoted, matches SourceRenameTablesMap
}) {
    continue
}
```

And add at least one integration test with a mixed-case schema/table partition, e.g. `"Public"."Orders_2024"`.

yb-voyager/src/metadb/importDataStatus.go
```diff
@@ -32,6 +32,10 @@ type ImportDataStatusRecord struct {
 		cdc partitioning strategy config for the import data auto, pk or table
 	*/
 	CdcPartitioningStrategyConfig string `json:"cdcPartitioningStrategyConfig"`
+
+	TargetUsePartitionRoot        bool `json:"TargetUsePartitionRoot"`        // false or true means use root table for partitions (default); false means insert via root table
```
makalaaneesh: [cursor] **Field comment is contradictory and the Go zero value contradicts the user-facing default.**

The comment `// false or true means use root table for partitions (default); false means insert via root table` says both `false` and `true` mean “use root” — a clear typo (replicated on all three fields).

More importantly, the bool zero value is `false`, while the user-facing default of `--use-partition-root` is `true`. Any read site that fires before `validateImportUsePartitionRootFlag` has been called for that role (which writes the field) will see `false` and incorrectly conclude that the user opted out of root-routing. Today this is masked because validation runs in PreRun before `exportData` reads it (see `cmd/exportData.go:330` and `:1928`), but the invariant is fragile — any future ordering change silently flips the propagated flag.

Per `yb-voyager/src/metadb/.cursor/BUGBOT.md`:

> When adding new fields, group related fields into sub-structs instead of adding flat fields at the top level... Add descriptive comments explaining their semantics and which process/role sets them.

Recommended:
1. Fix the comment to clearly state `true = use root (default), false = use leaf partition`, and identify which role writes each field.
2. Either invert the field (`UseLeafPartitionFor* bool`, default `false`), or store as `*bool` / a string enum so the unset state is distinguishable from explicit `false`.
3. Group all three into a sub-struct (`UsePartitionRoot UsePartitionRootByRole`).
4. Use lowerCamel JSON tags (`targetUsePartitionRoot`) for consistency with the rest of `ImportDataStatusRecord` (`errorPolicySnapshot`, `importDataStarted`, `tableToCDCPartitioningStrategyMap`).

yb-voyager/cmd/exportData.go
```diff
@@ -322,6 +323,14 @@ func startNextIterationImportDataToTarget() {
 		cmd = append(cmd, "--target-db-user", currentMsr.TargetDBConf.User)
 	}
 
+	importDataStatusRecord, err := metaDB.GetImportDataStatusRecord()
+	if err != nil {
+		utils.ErrExit("failed to get import data status record: %w", err)
+	}
+	if !importDataStatusRecord.TargetUsePartitionRoot {
```
makalaaneesh, line 330: [cursor] **Missing nil-check on `metaDB.GetImportDataStatusRecord()` result.**

Per `yb-voyager/.cursor/BUGBOT.md` and `yb-voyager/cmd/.cursor/BUGBOT.md`:

> metaDB record lookups can return `(nil, nil)` when a record is not found... When reading migration status records, always nil-check the result.

This call site dereferences `importDataStatusRecord.TargetUsePartitionRoot` immediately. The same issue exists at line **1928** in `startFallBackSetupIfRequired`. Even though `InitImportDataStatusRecord` runs at startup today, the contract requires the nil check; any future refactor that defers init will silently NPE here.

Related: this also propagates the flag by hand-appending `"--use-partition-root", "false"` instead of routing through `generateGlobalExportImportArguments` (called a few lines below), so the propagation logic is split. Per `yb-voyager/cmd/.cursor/BUGBOT.md`:

> Separate proper handling for config-file and CLI-only modes. Ensure CLI-overrides over config-files are handled properly.

If the iteration command is invoked via a config file rather than CLI, this code unconditionally adds the CLI flag, which may not be the right precedence. Consider centralizing in the helper.

Suggested fix:

```go
importDataStatusRecord, err := metaDB.GetImportDataStatusRecord()
if err != nil {
    utils.ErrExit("failed to get import data status record: %w", err)
}
if importDataStatusRecord == nil {
    utils.ErrExit("import data status record not found in meta db")
}
if !importDataStatusRecord.TargetUsePartitionRoot {
    cmd = append(cmd, "--use-partition-root", "false")
}
```

makalaaneesh (commented):
[cursor] Automated review — 12 findings (3 critical, 9 warning). See inline comments. Suggestions and nits were dropped. This is a non-blocking review; the human reviewer decides whether to block.

yb-voyager/cmd/exportData.go
```diff
@@ -1912,6 +1921,14 @@ func startFallBackSetupIfRequired() {
 		cmd = append(cmd, "--yes")
 	}
 
+	importDataStatusRecord, err := metaDB.GetImportDataStatusRecord()
+	if err != nil {
+		utils.ErrExit("failed to get import data status record: %w", err)
+	}
+	if !importDataStatusRecord.TargetUsePartitionRoot {
```
makalaaneesh, line 1928: [cursor] **Same nil-deref + manual CLI propagation issue as line 330.**

Mirror of the finding at `cmd/exportData.go:330`: this call site also dereferences `importDataStatusRecord.TargetUsePartitionRoot` without nil-checking the `(nil, nil)` not-found case, and propagates `--use-partition-root false` by manual append instead of using `generateGlobalExportImportArguments` (called immediately on the next line).

Apply the same fix here — ideally extracting a tiny helper so both spawn-importer paths share the propagation logic.

yb-voyager/src/tgtdb/event.go
```diff
@@ -141,39 +156,61 @@ func (e *Event) IsCutoverEvent() bool {
...
-func (e *Event) GetSQLStmt(tdb TargetDB) (string, error) {
+// IsPartitionEvent returns true if this event originated from a partition table that was renamed to root
+func (e *Event) IsPartitionEvent() bool {
+	return e.partitionSchemaName != "" && e.partitionTableName != ""
+}
+
+func (e *Event) GetSQLStmt(tdb TargetDB, usePartitionTable bool) (string, error) {
+	var err error
+	if usePartitionTable {
+		e.PartitionTableTup, err = namereg.NameReg.LookupTableName(fmt.Sprintf("%s.%s", e.partitionSchemaName, e.partitionTableName))
```
makalaaneesh, line 186: [cursor] **Mutating `e.PartitionTableTup` as a side-effect of the SQL builder is fragile.**

`GetSQLStmt` (here) and `GetPreparedSQLStmt` (line 189) both perform a side-effect on the event:

```go
e.PartitionTableTup, err = namereg.NameReg.LookupTableName(...)
```

Later, `GetPreparedStmtName` reads `event.PartitionTableTup`. In `cmd/yugabytedb.go:ExecuteBatch` and `cmd/postgres.go:ExecuteBatch`, today the order is `GetPreparedSQLStmt` → `GetPreparedStmtName`, so the side-effect happens to be observed. But:

1. Inside `GetPreparedSQLStmt` itself, `e.PartitionTableTup, ...` is set **before** `psName := e.GetPreparedStmtName(usePartitionTable)` on line 196. If anyone optimizes by computing `psName` first to short-circuit on the cache (a natural change — the cache lookup is the hot path), the cache key will be computed against an empty `PartitionTableTup` and silently mix prepared statements between root and leaf for the same partition event.
2. The `Copy()` method copies `PartitionTableTup`, so a stale value can leak across event reuse if anything ever resets `partitionSchemaName/Name` without clearing it.

Two cleaner fixes (pick one):

- Resolve `PartitionTableTup` once in `UnmarshalJSON` whenever `partitionSchemaName != ""`, so it's always populated for partition events.
- Or, pass the resolved tuple as a parameter to the builders rather than mutating the event.

Either one removes the implicit ordering contract.

yb-voyager/src/tgtdb/event.go
```diff
@@ -141,39 +156,61 @@ func (e *Event) IsCutoverEvent() bool {
...
+		if err != nil {
+			return "", fmt.Errorf("lookup partition table %s.%s in name registry: %w", e.partitionSchemaName, e.partitionTableName, err)
+		}
+		log.Infof("event %d: using partition table %s instead of root table %s",
```
makalaaneesh: [cursor] **`log.Infof` per CDC event will spam logs in `--use-partition-root false` mode.**

```go
log.Infof("event %d: using partition table %s instead of root table %s", ...)
```

This fires for every INSERT/DELETE in `--use-partition-root false` mode (and the duplicate at line 193 fires for every prepared-stmt path). At even modest event rates that's hundreds to thousands of Info-level lines per second, drowning out the actual progress messages and bloating disk usage.

Per `yb-voyager/.cursor/BUGBOT.md`:

> Use appropriate log levels: `log.Infof` for normal progress, `log.Warnf` for recoverable issues the user should know about. Do not log errors at Info level or progress at Warn level.

Per-event tracing belongs at `log.Debugf`. Convert both occurrences (line 171 and line 193) to `Debugf`. If you really want a single audit line, log it once when the importer starts (e.g. “using partition tables for CDC routing”) instead of per event.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1055,6 +1055,20 @@ func (yb *TargetYugabyteDB) RestoreSequences(sequencesLastVal *utils.StructMap[s
...
+partitioning strategy/names change on the target database. So with configuration '--use-partition-root false', we are ingesting data via partition table.
+but with UPDATE <partition table> stmt on YB , there is a limiation that it errors out if the UPDATE statement doesn't include partition key so we are skipping
+ingestion of UPDATE events via partition table on Target DB. and in other importers we are ingesting data via partition table.
+*/
+func (yb *TargetYugabyteDB) shouldUsePartitionTable(event *Event) bool {
+	if yb.tconf.UsePartitionRoot || !event.IsPartitionEvent() {
+		//if the use-partition-root is true or the event is not a partition table event, we don't use the partition table
+		return false
+	}
+	return event.Op != "u"//
```
makalaaneesh: [cursor] **`shouldUsePartitionTable` has a stray empty `//` and a confusing inline comment.**

```go
if yb.tconf.UsePartitionRoot || !event.IsPartitionEvent() {
    //if the use-partition-root is true or the event is not a partition table event, we don't use the partition table
    return false
}
return event.Op != "u"//   <-- stray empty `//`
```

Two concrete issues:

1. Line 1069 has a trailing empty `//` after the boolean expression. It's a lint smell and may trip up some linters/formatters. Drop it.
2. The comment at line 1066 reads as “we don’t use the partition table”, which is the **inverse** of what `return false` means in this method (the method tells the caller whether to use the partition table; `false` means “don’t use it, use root”). On first read this looks contradictory. Reword as e.g. `// fallback to root: --use-partition-root is enabled, or the event is not from a partitioned table`.

While here, the docblock above (lines 1058-1063) explains the YB UPDATE limitation but is hard to scan. Consider a more crisp form, e.g. cite the YB issue/docs link and tighten to:

> YB rejects an `UPDATE <leaf>` whose WHERE clause doesn't include the partition key. To avoid this, route UPDATE events through the root table (which lets PG/YB partition routing handle it), even when `--use-partition-root=false` is set.

yb-voyager/cmd/importData.go
makalaaneesh: [cursor] **Behavior change in `initialiseImportTableList` for live + snapshot is hidden behind a one-line comment.**

Diff:

```go
-tablesToImport := lo.Ternary(importSnapshotRequired(), importFileTasksToTableNameTuples(importFileTasks), importTableList)
-checkTablesPresentInTarget(tablesToImport)
+checkTablesPresentInTarget(importTableList) //to check whether tables exist or not we should use importTableList in live migration case
```

For `live + snapshot` migrations the “tables exist on target” guardrail used to derive its list from the actual data files. Now it always uses `importTableList`. If those two ever diverge (e.g. user adds a partition between snapshot and changes, uses `--include-table` on a subset, or `addLeafPartitionsInTableList` enriched the list), regressions in the table-existence check would only surface during streaming.

The one-line comment doesn’t justify the change. Either:

1. Keep the previous `lo.Ternary` for the snapshot branch and only switch to `importTableList` for the changes-only path, **or**
2. Expand the comment to explicitly state why `importTableList` is now correct for both phases (e.g. because in the live path it already includes all leaf partitions enriched by `addLeafPartitionsInTableList`, and `importFileTasks` may legitimately not yet contain CDC-only tables).

yb-voyager/cmd/importData.go
```diff
@@ -419,6 +425,67 @@ func checkTablesPresentInTarget(tablesToImport []sqlname.NameTuple) {
...
+		utils.PrintAndLogfInfo("\nWhen using '--use-partition-root false', CDC events will contain partition table names.")
+		utils.PrintAndLogfInfo("The following root table partitions are not present on the target database:")
+		for root, leaves := range missingRootToLeafPartitions {
+			utils.PrintAndLogfInfo("  - %s:", root)
+			for _, leaf := range leaves {
+				utils.PrintAndLogfInfo("    - %s", leaf)
+			}
+		}
+		utils.PrintAndLogfWarning("\nEnsure that all partitions from the source exist on the target, or use --use-partition-root true (default).")
+		if !utils.AskPrompt("\nDo you want to continue anyway") {
```
makalaaneesh, line 497: [cursor] **Partition-consistency report has nondeterministic ordering and demotes a hard error to a prompt without explanation.**

Two issues in this block:

1. **Nondeterministic output** — the `for root, leaves := range missingRootToLeafPartitions` loop on line 475 (and the inner leaf loop) iterates Go maps in random order. The same migration will print findings in different order across runs, which makes diffing diagnostic logs across attempts annoying. Sort `missingRootToLeafPartitions` keys (and each leaf slice) before printing.
2. **Prompt fallback** — if `AskPrompt("continue anyway?")` returns `true`, we proceed with potentially missing target partitions, but the ErrExit message at line 483 is very long and reads like an error rather than a deliberate user choice. Per `yb-voyager/cmd/.cursor/BUGBOT.md`:

   > When demoting a hard error to a warning... document *why* the error is non-fatal and ensure all error paths in the called function also behave non-fatally.

   Add a comment explaining that proceeding is allowed because the user has accepted responsibility, and shorten the no-path message (it’s already been printed above).

yb-voyager/cmd/exportData.go
```diff
@@ -2022,29 +2042,84 @@ func finalizeTableAndColumnList(finalTableList []sqlname.NameTuple) ([]sqlname.N
...
+// lacks a primary key. The check is partition-aware: a partitioned root table that has no PK of
+// its own is acceptable as long as every leaf partition under it carries a PK, because Debezium
+// streams change events from leaf partitions in PG/YB. Without this awareness, the root added by
+// addLeafPartitionsInTableList (so that catalog queries see it) would cause false-positive
+// failures for partition hierarchies whose PKs live only on the leaves.
+func reportUnsupportedTablesForLiveMigration(finalTableList []sqlname.NameTuple, partitionsToRootTableMap map[string]string) {
 	if !changeStreamingIsEnabled(exportType) {
 		return
 	}
+	rootToLeafPartitions, err := buildRootToLeafPartitionsMap(partitionsToRootTableMap, finalTableList)
```
makalaaneesh, line 2055: [cursor] **`buildRootToLeafPartitionsMap` hard-exits when a partition is filtered out of `finalTableList`.**

```go
rootToLeafPartitions, err := buildRootToLeafPartitionsMap(partitionsToRootTableMap, finalTableList)
if err != nil {
    utils.ErrExit("build root-to-leaf partitions map for non-pk check: %w", err)
}
```

Inside `buildRootToLeafPartitionsMap`, if any leaf or root in `partitionsToRootTableMap` is absent from `tableToUnQuotedMap` (i.e. not in `finalTableList`), the function returns `goerrors.Errorf("lookup leaf partition %q", leafQualified)` and the export hard-exits with a message that gives the user nothing actionable.

This is reachable in normal flows: a user can exclude an individual leaf partition via `--exclude-table`, or the root may have been added by `addLeafPartitionsInTableList` purely for catalog queries and intentionally dropped from the final list. In either case we shouldn't fail the export.

Suggested fix: skip silently when the root is not in `finalTableList` (the caller `reportUnsupportedTablesForLiveMigration` only cares about leaves of roots that *are* in the list). At minimum, include both names being looked up and the size of the table list in the error to make it diagnosable.

yb-voyager/cmd/import.go
```diff
@@ -117,6 +119,40 @@ func validateImportDataFlags() error {
...
+		}
+		if importerRole == SOURCE_REPLICA_DB_IMPORTER_ROLE {
+			return goerrors.Errorf("'--use-partition-root false' is not supported for source-replica")
+		}
+		if tconf.TargetDBType != POSTGRESQL && tconf.TargetDBType != YUGABYTEDB {
+			return goerrors.Errorf("'--use-partition-root' flag is only valid for PostgreSQL to YugabyteDB migrations")
+		}
+	}
+	tconf.UsePartitionRoot = bool(importUsePartitionRoot)
+	return metaDB.UpdateImportDataStatusRecord(func(record *metadb.ImportDataStatusRecord) {
```
makalaaneesh, line 147: [cursor] **`validateImportUsePartitionRootFlag` mixes validation with persistence and writes a meaningless value for source-replica.**

Two concerns:

1. **Naming/responsibility split** — the function is named `validate…` but mutates MSR via `UpdateImportDataStatusRecord`. Per the workspace style, separate concerns: `validateImportUsePartitionRootFlag()` should only validate; persist via a sibling `persistImportUsePartitionRootFlag()` (or fold persistence into the regular MSR-init path).
2. **Source-replica write is a no-op-but-misleading** — the validation rejects `false` for `SOURCE_REPLICA_DB_IMPORTER_ROLE` (line 137). But when `importUsePartitionRoot=true` (the default) on a source-replica run, the function still falls through and writes `record.SourceReplicaUsePartitionRoot = true` here. That field can therefore only ever be written as `true`, so it carries no information; future readers may believe it’s configurable per role and use it to gate behavior. Either skip the write entirely for source-replica (preferred), or remove the field from `ImportDataStatusRecord` and make the validation explicit.

Per `yb-voyager/cmd/.cursor/BUGBOT.md`:

> MSR boolean flags must be set only by the process/role they describe.

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/KafkaConnectRecordParser.java
```diff
@@ -69,6 +73,30 @@ private void retrieveRenameTablesFromConfig() {
...
+                if (beforeAndAfter.length != 2) {
+                    throw new RuntimeException(String.format(
```
makalaaneesh, line 80: [cursor] **Config-validation throws raw `RuntimeException`, and `partitionToRootMapping` stores a value that's never used.**

Two issues in this block:

1. **Exception type** — raw `RuntimeException` terminates the worker thread without being recognized as a user-input error. Use `IllegalArgumentException` (or a project-specific config-error type) so the failure mode is clear in logs/UIs. The `renameTables` block above does the same thing, so consider fixing both for parity.
2. **Unused root value** — `partitionToRootMapping` is declared `Map<String, String>` but the only consumer is `partitionToRootMapping.containsKey(qualifiedTableName)` later in `parseTable`. The root (“after” side) is never read. Either:

   - Replace the field with `Set<String> partitionTables` and rename it (clearer intent, fewer allocations), **or**
   - Actually use the root value to defensively cross-check against `renameTables.get(qualifiedTableName)` — today they should always agree, but a config mismatch would silently route events to the wrong table.

Bonus nit: the `partitionToRootMapping` validation logs only the offending inner token (`before` / `after`) on lines 89/94, while the rename-mapping branch logs the full mapping string. Make them consistent so user errors are diagnosable.

yb-voyager/src/testlivemigration/live_migration_integration_test.go
makalaaneesh: one more test (or modify existing): case-sensitive names for partition names AND schema names .

yb-voyager/cmd/exportData.go
priyanshi-yb: Not required

yb-voyager/src/tgtdb/event.go
priyanshi-yb: fixes proposed here are not valid as partition look up is not the right thing to do it blindly as it can error out in scenarios where partitions are different on target

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1086,18 +1072,28 @@ func (yb *TargetYugabyteDB) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
 	for i := 0; i < len(batch.Events); i++ {
 		event := batch.Events[i]
 		if event.Op == "u" {
-			stmt, err := event.GetSQLStmt(yb, yb.shouldUsePartitionTable(event))
+			/*
+			   Currently ingestion logic is to ingest cdc data via root table for the partitioned table by default to cases where
+			   partitioning strategy/names change on the target database. So with configuration '--use-partition-root false', we are ingesting data via partition table.
+			   but with UPDATE <partition table> stmt on YB , there is a limiation that it errors out if the UPDATE statement doesn't include partition key so we are skipping
+			   ingestion of UPDATE events via partition table on Target DB. and in other importers we are ingesting data via partition table.
+			   so use partition root table always for UPDATE events  in YB
```
makalaaneesh, line 1080: nit: add GH

yb-voyager/src/tgtdb/event.go
```diff
@@ -140,6 +141,16 @@ func (e *Event) Copy() *Event {
 	}
 }
 
+func (e *Event) GetEffectiveTableName(usePartitionRoot bool) (sqlname.NameTuple, error) {
+	if e.shouldUsePartitionTable(usePartitionRoot) {
+		if e.PartitionTableTup != sqlname.NIL_NAME_TUPLE {
```
makalaaneesh, line 146: hmm, i think it might be more idiomatic to define the field as a pointer because it's optional. 
But this is also fine 👍

yb-voyager/src/testlivemigration/live_migration_partitions_integrations_test.go
```diff
@@ -984,7 +959,7 @@ func TestLiveMigrationPartitionedWithChildPKAndPartitionsAcrossDifferentSchemas(
 	}, 30, 1)
 	testutils.FatalIfError(t, err, "failed to wait for streaming complete")
 
-	err = lm.ValidateDataConsistency([]string{`"public"."orders"`, `"public"."customers"`}, "id")
+	err = lm.ValidateDataConsistency([]string{`"TestSchemaCase"."orders"`, `"public"."customers"`}, "id")
```
makalaaneesh, line 1145: can we do per-leaf validations as well? 
Things could go to the wrong (default) partition in the worst case.

makalaaneesh:
@priyanshi-yb pls have AI update the PR description. It's a bit outdated. You can use the skill that Shivansh added.

makalaaneesh (commented):
LGTM! (I can't approve because i am the creator of the PR). 
@priyanshi-yb pls run jenkins tests (offline/live both) and then you can approve,merge

priyanshi-yb:
Jenkins runs -

PG/Oracle LIVE migrations with/without gRPC and oracle offline
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/7942/
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/7943/
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/7944/


Live migration resumption tests
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/491
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/493/

priyanshi-yb (approved):

PR #3515, opened 2026-04-24
Use ysql_terminate_and_drop_database in import-data-file test runner

Abhi-yb:

`migtests/tests/import-file/run-import-file-test` still uses the raw
`run_ysql "DROP DATABASE IF EXISTS ${TARGET_DB_NAME};"` pattern at both the
pre-test create step and the final cleanup step. #3503 introduced the
`ysql_terminate_and_drop_database` helper and converted the standard
orchestration scripts (`run-test.sh`, `run-schema-migration.sh`,
`live-migration-*.sh`, etc.) to use it, but this bespoke per-test runner
was missed. As a result, the cleanup step still occasionally fails with:

```
ERROR:  database "testdb" is being accessed by other users
DETAIL:  There is 1 other session using the database.
```

This PR converts both call-sites in `run-import-file-test` to
`ysql_terminate_and_drop_database`, matching the pattern now used by the
rest of the suite. No other tests are touched — the remaining
non-converted call-sites (the 3 `init-db` / `init-target-db` scripts under
`migtests/tests/resumption/`) are pre-test init and don't hit the
stale-session path, so they're intentionally left alone.

No user-facing changes.

No.

No.

### DIFF
--- a/migtests/tests/import-file/run-import-file-test
+++ b/migtests/tests/import-file/run-import-file-test
@@ -46,7 +46,7 @@ main() {
 	pushd ${TEST_DIR}
 
 	step "Create target database."
-	run_ysql yugabyte "DROP DATABASE IF EXISTS ${TARGET_DB_NAME};"
+	ysql_terminate_and_drop_database "${TARGET_DB_NAME}"
 	run_ysql yugabyte "CREATE DATABASE ${TARGET_DB_NAME}"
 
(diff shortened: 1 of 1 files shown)


shubham-yb (approved):
LGTM. Can you improve the commit message?
`Use ysql_terminate_and_drop_database in import-data-file test runner`

PR #3523, opened 2026-04-30
Export data guardrail if the PK is on the partitions only but its not same in all the partitions

priyanshi-yb:

Adds a live-migration export-data guardrail that catches a previously-silent
correctness hazard: when a partitioned root table has no primary key but its
leaf partitions do, those leaves must all agree on the same primary key columns
(in the same order). If they don't, CDC events from different leaves would
carry different key columns and silently corrupt downstream replication on the
target.

The check runs only for PostgreSQL and YugabyteDB sources (Oracle/MySQL skip
leaf expansion entirely) and only for roots that themselves lack a PK —
PostgreSQL already enforces PK propagation when the root has one, so those
cases cannot diverge. When mismatches are detected, export data fails fast,
prints the offending root tables along with the distinct PK column lists
observed across their leaves, and tells the user how to recover.

Implementation:
- New `SourceDB.GetPrimaryKeyColumns([]NameTuple) -> StructMap[NameTuple, []string]`
  interface method. PG/YB implementations issue a single bulk catalog query
  (`pg_index` + `pg_attribute`) for all leaves at once, ordered by
  `array_position(indkey, attnum)` so PK column order is preserved (this also
  addresses a review concern about per-leaf round-trips when a root has many
  partitions). Oracle/MySQL panic with "not implemented" since the caller never
  reaches them.
- New `reportLeafPartitionsWithMismatchedPrimaryKeys` in `cmd/exportData.go`,
  invoked at the end of the existing `reportUnsupportedTablesForLiveMigration`
  guardrail. It iterates the root→leaves map, fetches PKs in bulk per root,
  compares with `slices.Equal`, and aggregates distinct PK column-tuples per
  root for a deterministic, sorted error report.

- New error surface during `yb-voyager export data` for live migration. When a
  partitioned root has no PK and its leaves disagree on PK columns, the command
  now exits with:
  - stdout (info): `Partitioned tables with inconsistent primary keys across leaf partitions:` followed by `- <root>: (cols), (cols)` lines.
  - stderr (ErrExit): `Live migration requires all leaf partitions of a partitioned table to share the same primary key columns. Either align the leaves' primary keys, or exclude these tables using the --exclude-table-list argument.`
- No CLI flags, configuration, installation, or report changes.

- New unit tests behind the `integration` build tag, covering single-column PK,
  composite PK with order preservation, tables without a PK (absent from the
  result map), and case-sensitive (quoted) identifiers:
  - `TestPostgresGetPrimaryKeyColumns` in `yb-voyager/src/srcdb/postgres_test.go`
  - `TestYugabyteGetPrimaryKeyColumns` in `yb-voyager/src/srcdb/yugabytedb_test.go`
- Reworked integration test in `yb-voyager/src/testlivemigration/live_migration_partitions_integrations_test.go`:
  - `TestLiveMigrationPartitionedTableWithChildTablesHavingDifferentPK` no
    longer attempts the full CDC flow (which was commented out as not
    supported). It now asserts that `StartExportData` errors out, stderr
    contains the exact guardrail message, and stdout lists the offending roots
    (`"TestSchemaCase".orders` and `public.customers`) with their distinct PK
    sets `(id), (id1)`.
  - Test schema was extended to exercise multiple schemas (`public`,
    `test_schema`, `"TestSchemaCase"`), case-sensitive quoted identifiers, and
    multi-level partitioning (LIST → RANGE → HASH), so the guardrail and the
    bulk catalog query are validated end-to-end against realistic shapes.
- Manual run of `TestLiveMigrationPartitionedWithChildPKAndPartitionsAcrossDifferentSchemas`
  confirmed the existing supported-shape test still passes.

https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/7946/
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/7948

No.

No.

---

| Component        


yb-voyager/src/srcdb/mysql.go
```diff
@@ -502,6 +502,13 @@ func (ms *MySQL) ClearMigrationState(migrationUUID uuid.UUID, exportDir string)
 	return nil
 }
 
+// GetPrimaryKeyColumns is a no-op for MySQL: the partition-aware caller in
+// reportUnsupportedTablesForLiveMigration only iterates leaves built by
+// addLeafPartitionsInTableList, which is itself a no-op for non-PG/YB sources.
+func (ms *MySQL) GetPrimaryKeyColumns(table sqlname.NameTuple) ([]string, error) {
+	return nil, nil
```
makalaaneesh: better to panic(not implemented) rather then return nil, and rely on caller not calling it for Mysql/Oracle.

yb-voyager/cmd/exportData.go
```diff
@@ -2090,6 +2090,75 @@ func reportUnsupportedTablesForLiveMigration(finalTableList []sqlname.NameTuple,
...
+// reportLeafPartitionsWithMismatchedPrimaryKeys ensures that, for every
+// partitioned root whose own PK is missing but each leaf carries one, all
+// leaves share the *same* PK column list (in the same order). With CDC,
+// events from different leaves carry different key columns when their PKs
+// diverge, which would silently corrupt downstream replication.
+//
+// The check is intentionally restricted to roots without a PK: PostgreSQL
+// already enforces that a PK on a partitioned root is propagated to every
+// leaf, so those cases cannot diverge.
+func reportLeafPartitionsWithMismatchedPrimaryKeys(
```
makalaaneesh, line 2107: do we run this for export-data-from-target and export-data-from-source both?

priyanshi-yb: Yes it runs for both

yb-voyager/src/srcdb/postgres.go
```diff
@@ -1064,6 +1064,63 @@ func (pg *PostgreSQL) DropPublication(publicationName string) error {
...
+			log.Warnf("close rows for leaf primary-key query: %v", cerr)
+		}
+	}()
+
+	for rows.Next() {
+		var schema, table, col string
+		if err := rows.Scan(&schema, &table, &col); err != nil {
+			return nil, fmt.Errorf("scan PK column row for leaves: %w", err)
+		}
+		leaf, ok := catalogTableToTuple[fmt.Sprintf("%s.%s", schema, table)]
```
makalaaneesh: let's not use "leaf" as this is a generic func. 
Pls check error messages and other variables in this func

yb-voyager/src/srcdb/postgres.go
```diff
@@ -1064,6 +1064,63 @@ func (pg *PostgreSQL) DropPublication(publicationName string) error {
...
+		if !ok {
+			return nil, fmt.Errorf("leaf not found in catalog: %s.%s", schema, table)
+		}
+		cols, _ := result.Get(leaf)
```
makalaaneesh: don't we usually check ok or not and if not create an empty slice. Is this working fine without? does it return empty slice if not present (which will be the case for the first column)

yb-voyager/src/srcdb/postgres.go
```diff
@@ -1064,6 +1064,63 @@ func (pg *PostgreSQL) DropPublication(publicationName string) error {
...
+SELECT n.nspname, c.relname, a.attname
+FROM pg_index i
+JOIN pg_class      c ON c.oid = i.indrelid
+JOIN pg_namespace  n ON n.oid = c.relnamespace
+JOIN pg_attribute  a ON a.attrelid = c.oid AND a.attnum = ANY(i.indkey)
+WHERE (n.nspname, c.relname) IN (%s)
+  AND i.indisprimary
+ORDER BY array_position(i.indkey, a.attnum);`
+
+func (pg *PostgreSQL) GetPrimaryKeyColumns(tables []sqlname.NameTuple) (*utils.StructMap[sqlname.NameTuple, []string], error) {
```
makalaaneesh, line 1081: if possible, unit tests for this func pls

yb-voyager/src/testlivemigration/live_migration_partitions_integrations_test.go
```diff
@@ -645,63 +642,72 @@ func TestLiveMigrationPartitionedTableWithChildTablesHavingDifferentPK(t *testin
 	err = lm.SetupSchema()
 	testutils.FatalIfError(t, err, "failed to setup schema")
 
-	err = lm.StartExportData(true, nil)
-	testutils.FatalIfError(t, err, "failed to start export data")
+	err = lm.StartExportData(false, nil)
+	assert.Error(t, err)
```
makalaaneesh, line 646: nice

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: yeah it returns an empty slice, so we can append to it in case it doesn't exist in map

makalaaneesh (approved):

PR #3531, opened 2026-05-07
Add YB replica identity CHANGE guardrail

ShivanshGahlot:

This PR adds an export-data guardrail for YugabyteDB when it is used as the source database in the export-from-target flow. The guardrail checks the exported table list against `pg_class.relreplident` and reports any table whose replica identity is not `CHANGE`. This mirrors the existing PostgreSQL live migration guardrail that requires replica identity `FULL`, but uses YugabyteDB's expected `CHANGE` identity for the YB CDC path.

The check runs through `YugabyteDB.GetMissingExportDataPermissions`, so failures are surfaced through the existing export-data permissions/configuration guardrail path. If non-CHANGE tables are found, Voyager lists them under `Tables missing replica identity CHANGE`.

Users running `export data from target` with YugabyteDB logical CDC will now get a blocking guardrail error if any exported YB table does not have replica identity `CHANGE`.

There are no command-line, configuration, installation, or report format changes.

image

Locally

No new unit tests were added.

No.

No.

---

| Component        

### DIFF
--- a/yb-voyager/src/srcdb/yugabytedb.go
+++ b/yb-voyager/src/srcdb/yugabytedb.go
@@ -1343,20 +1343,22 @@ func (yb *YugabyteDB) GetMissingExportDataPermissions(exportType string, finalTa
 			return nil, false, fmt.Errorf("error in checking table replica identity: %w", err)
 		}
 		if len(tablesWithoutReplicaIdentityChange) > 0 {
-			combinedResult = append(combinedResult, fmt.Sprintf("\n%s[%s]", color.RedString("Tables not having replica identity CHANGE: "), strings.Join(tablesWithoutReplicaIdentityChange, ", ")))
+			combinedResult = append(combinedResult, fmt.Sprintf("\n%s[%s]", color.RedString("Tables not having replica identity CHANGE: "), strings.Join(lo.Map(tablesWithoutReplicaIdentityChange, func(t sqlname.NameTuple, _ int) string {
+				return t.ForOutput()
+			}), ", ")))
 		}
 	}
 
 	return combinedResult, len(combinedResult) > 0, nil
 }
 
-func (yb *YugabyteDB) listTablesMissingReplicaIdentityChange(tableList []sqlname.NameTuple) ([]string, error) {
+// TODO: migrate postgres.go listTablesMissingReplicaIdentityFull from queryTableList string to []NameTuple,
+// then consolidate both into a shared helper parameterized by expected relreplident char and label.
+func (yb *YugabyteDB) listTablesMissingReplicaIdentityChange(tableList []sqlname.NameTuple) ([]sqlname.NameTuple, error) {
 	var tableNamePairs []string
-	qualifiedNameByCatalogKey := make(map[string]string)
 	for _, table := range tableList {
 		sname, tname := table.ForCatalogQuery()
 		tableNamePairs = append(tableNamePairs, fmt.Sprintf("('%s','%s')", sname, tname))
-		qualifiedNameByCatalogKey[sname+"."+tname] = table.ForOutput()
 	}
 
 	checkTableReplicaIdentityQuery := fmt.Sprintf(`
@@ -1386,7 +1388,7 @@ func (yb *YugabyteDB) listTablesMissingReplicaIdentityChange(tableList []sqlname
 		}
 	}()
 
-	var tablesWithoutReplicaIdentityChange []string
+	tablesWithoutIdentityChangeKeys := make(map[string]bool)
 	var tableSchemaName, tableName, replicaIdentity, status string
 
 	for rows.Next() {
@@ -1395,7 +1397,7 @@ func (yb *YugabyteDB) listTablesMissingReplicaIdentityChange(tableList []sqlname
 			return nil, fmt.Errorf("error in scanning query rows for table names: %w", err)
 		}
 		if status == MISSING {
-			tablesWithoutReplicaIdentityChange = append(tablesWithoutReplicaIdentityChange, qualifiedNameByCatalogKey[tableSchemaName+"."+tableName])
+			tablesWithoutIdentityChangeKeys[tableSchemaName+"."+tableName] = true
 		}
 	}
 
@@ -1403,7 +1405,14 @@ func (yb *YugabyteDB) listTablesMissingReplicaIdentityChange(tableList []sqlname
 		return nil, fmt.Errorf("error iterating over query rows: %w", err)
 	}
 
-	return tablesWithoutReplicaIdentityChange, nil
+	var result []sqlname.NameTuple
+	for _, table := range tableList {
+		sname, tname := table.ForCatalogQuery()
+		if tablesWithoutIdentityChangeKeys[sname+"."+tname] {
+			result = append(result, table)
+		}
+	}
+	return result, nil
 }
 
 func (yb *YugabyteDB) GetMissingAssessMigrationPermissions() ([]string, bool, error) {
--- a/yb-voyager/src/srcdb/yugabytedb.go
+++ b/yb-voyager/src/srcdb/yugabytedb.go
@@ -1352,9 +1352,11 @@ func (yb *YugabyteDB) GetMissingExportDataPermissions(exportType string, finalTa
 
 func (yb *YugabyteDB) listTablesMissingReplicaIdentityChange(tableList []sqlname.NameTuple) ([]string, error) {
 	var tableNamePairs []string
+	qualifiedNameByCatalogKey := make(map[string]string)
 	for _, table := range tableList {
 		sname, tname := table.ForCatalogQuery()
 		tableNamePairs = append(tableNamePairs, fmt.Sprintf("('%s','%s')", sname, tname))
+		qualifiedNameByCatalogKey[sname+"."+tname] = table.ForOutput()
 	}
 
 	checkTableReplicaIdentityQuery := fmt.Sprintf(`
@@ -1393,7 +1395,7 @@ func (yb *YugabyteDB) listTablesMissingReplicaIdentityChange(tableList []sqlname
 			return nil, fmt.Errorf("error in scanning query rows for table names: %w", err)
 		}
 		if status == MISSING {
-			tablesWithoutReplicaIdentityChange = append(tablesWithoutReplicaIdentityChange, sqlname.NewSourceName(tableSchemaName, tableName).Qualified.MinQuoted)
+			tablesWithoutReplicaIdentityChange = append(tablesWithoutReplicaIdentityChange, qualifiedNameByCatalogKey[tableSchemaName+"."+tableName])
 		}
 	}
 
--- a/yb-voyager/src/srcdb/yugabytedb.go
+++ b/yb-voyager/src/srcdb/yugabytedb.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/fatih/color"
 	goerrors "github.com/go-errors/errors"
 	"github.com/google/uuid"
 	"github.com/jackc/pglogrepl"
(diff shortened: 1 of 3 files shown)


yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -1334,7 +1335,70 @@ func (yb *YugabyteDB) GetMissingExportSchemaPermissions(queryTableList string) (
 }
 
 func (yb *YugabyteDB) GetMissingExportDataPermissions(exportType string, finalTableList []sqlname.NameTuple) ([]string, bool, error) {
-	return nil, false, nil
```
makalaaneesh, line 1337: how is it that we did not have any checks till now? 🤔 I thought we added guardrails in the past even for export-data-from-target. no?

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -1334,7 +1335,70 @@ func (yb *YugabyteDB) GetMissingExportSchemaPermissions(queryTableList string) (
...
+	var combinedResult []string
+
+	qualifiedMinQuotedTableNames := lo.Map(finalTableList, func(table sqlname.NameTuple, _ int) string {
+		return table.ForOutput()
```
makalaaneesh: we should use `ForCatalogQuery`/ `AsQualifiedCatalogName` here. That will take care of case sensitive names as well.

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -1334,7 +1335,70 @@ func (yb *YugabyteDB) GetMissingExportSchemaPermissions(queryTableList string) (
...
+	})
+	queryTableList := fmt.Sprintf("'%s'", strings.Join(qualifiedMinQuotedTableNames, "','"))
+
+	missingTables, err := yb.listTablesMissingReplicaIdentityChange(queryTableList)
```
makalaaneesh: better variable name pls. `missingTables` is misleading. 
`tablesWithoutReplicaIdentityChange` (or something shorter)

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -1334,7 +1335,70 @@ func (yb *YugabyteDB) GetMissingExportSchemaPermissions(queryTableList string) (
...
+	if err != nil {
+		return nil, false, fmt.Errorf("error in checking table replica identity: %w", err)
+	}
+	if len(missingTables) > 0 {
+		combinedResult = append(combinedResult, fmt.Sprintf("\n%s[%s]", color.RedString("Tables missing replica identity CHANGE: "), strings.Join(missingTables, ", ")))
```
makalaaneesh: "missing" here is also a bit confusing IMO. How about "Tables not having replicable identity CHANGE"?

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -1334,7 +1335,70 @@ func (yb *YugabyteDB) GetMissingExportSchemaPermissions(queryTableList string) (
...
+		c.relname AS table_name,
+		c.relreplident AS replica_identity,
+		CASE
+			WHEN c.relreplident <> 'c'
+			THEN '%s'
+			ELSE '%s'
+		END AS status
+	FROM pg_class c
+	JOIN pg_namespace n ON c.relnamespace = n.oid
+	WHERE (quote_ident(n.nspname) || '.' || quote_ident(c.relname)) IN (%s)
```
makalaaneesh: `ForCatalogQuery` is a better candidate here. You don't need to concat in the query. 
(i.e. take []NameTuple as input args to this function

yb-voyager/src/srcdb/yugabytedb.go
makalaaneesh: pls add an integration test case for this

makalaaneesh: [cursor] **Critical: guardrail fires even when the YB gRPC connector is used, where replica identity is irrelevant.**

`GetMissingExportDataPermissions` runs `listTablesMissingReplicaIdentityChange` unconditionally for YB. The YB gRPC connector reads from the raft log via the CDC service and does **not** require `REPLICA IDENTITY CHANGE` — only the YB logical-replication path does.

The existing code at `cmd/exportData.go:834-835` already gates other live-replication checks on:

```go
(source.DBType == YUGABYTEDB && !bool(useYBgRPCConnector))
```

and `msr.UseYBgRPCConnector` is propagated into `source.IsYBGrpcConnector` at `cmd/exportData.go:549`, so the value is reachable from inside `YugabyteDB` via `yb.source.IsYBGrpcConnector`.

As written, any export-from-target run that uses the gRPC connector will block on a false-positive `Tables missing replica identity CHANGE` error if users haven't (unnecessarily) altered their tables.

**Suggested fix:** early-return when `yb.source.IsYBGrpcConnector` is true, or skip just the replica-identity probe in that branch.

Per the product-level BUGBOT (`.cursor/BUGBOT.md` — "Features gated by source type ... must have explicit validation or guardrails"), this needs an explicit connector-flavor gate.

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -1334,7 +1335,70 @@ func (yb *YugabyteDB) GetMissingExportSchemaPermissions(queryTableList string) (
 }
 
 func (yb *YugabyteDB) GetMissingExportDataPermissions(exportType string, finalTableList []sqlname.NameTuple) ([]string, bool, error) {
-	return nil, false, nil
+	var combinedResult []string
```
makalaaneesh, line 1338: [cursor] **Warning: `exportType` is accepted but ignored, unlike the PG sibling.**

The PG implementation at `postgres.go:1266` gates the replica-identity probe on:

```go
if exportType == utils.CHANGES_ONLY || exportType == utils.SNAPSHOT_AND_CHANGES { ... }
```

The YB implementation takes `exportType` as a parameter but never reads it, so the CHANGE probe also runs for `SNAPSHOT_ONLY`. Today YB-as-source is only used in live flows, but per the product invariants (`.cursor/BUGBOT.md` — "changes-only mode... differs from snapshot-and-changes path"), the gate should be explicit so we don't accidentally regress when the call site changes.

**Suggested fix:** mirror the PG structure — wrap the CHANGE probe in `if exportType == utils.CHANGES_ONLY || exportType == utils.SNAPSHOT_AND_CHANGES { ... }`.

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -1334,7 +1335,70 @@ func (yb *YugabyteDB) GetMissingExportSchemaPermissions(queryTableList string) (
...
+		return nil, false, fmt.Errorf("error in checking table replica identity: %w", err)
+	}
+	if len(missingTables) > 0 {
+		combinedResult = append(combinedResult, fmt.Sprintf("\n%s[%s]", color.RedString("Tables missing replica identity CHANGE: "), strings.Join(missingTables, ", ")))
+	}
+
+	return combinedResult, len(combinedResult) > 0, nil
+}
+
+func (yb *YugabyteDB) listTablesMissingReplicaIdentityChange(queryTableList string) ([]string, error) {
```
makalaaneesh: [cursor] **Warning: near-duplicate of `PostgreSQL.listTablesMissingReplicaIdentityFull` — violates the shared-SQL rule.**

Per `yb-voyager/src/.cursor/BUGBOT.md`:

> When the same SQL pattern is used for both PostgreSQL and YugabyteDB, extract it into a shared constant and document any version-specific differences.

`listTablesMissingReplicaIdentityChange` and `postgres.go:1674` `listTablesMissingReplicaIdentityFull` differ only in:
1. The expected `relreplident` char (`'c'` vs `'f'`)
2. The error/log prefix string

It also depends on the `MISSING`/`GRANTED` constants defined in `postgres.go:50-51` — the two implementations are coupled by symbol but not by structure, which is fragile.

**Suggested fix:** introduce a package-level helper, e.g.

```go
func listTablesMissingReplicaIdentity(db *sql.DB, expected rune, queryTableList, dbLabel string) ([]string, error)
```

and have both `postgres.go` and `yugabytedb.go` delegate.

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -1334,7 +1335,70 @@ func (yb *YugabyteDB) GetMissingExportSchemaPermissions(queryTableList string) (
...
+	var missingTables []string
+	var tableSchemaName, tableName, replicaIdentity, status string
+
+	for rows.Next() {
+		err = rows.Scan(&tableSchemaName, &tableName, &replicaIdentity, &status)
+		if err != nil {
+			return nil, fmt.Errorf("error in scanning query rows for table names: %w", err)
+		}
+		if status == MISSING {
+			missingTables = append(missingTables, fmt.Sprintf(`%s."%s"`, tableSchemaName, tableName))
```
makalaaneesh: [cursor] **Warning: manual `schema.table` concatenation — only quotes the table side.**

```go
missingTables = append(missingTables, fmt.Sprintf(`%s."%s"`, tableSchemaName, tableName))
```

This reconstructs a qualified name without going through `sqlname`, and quotes the table but not the schema. If the schema is case-sensitive (e.g. `"Sales"`), the output `Sales."orders"` is wrong and can mismatch a later display or lookup.

Per `yb-voyager/.cursor/BUGBOT.md` ("Object Names"):

> Use the `sqlname` package for all object name handling (table names especially). Do not construct qualified names via manual string concatenation.

The PG sibling at `postgres.go:1712` has the identical bug and is worth fixing in the same pass.

**Suggested fix:** either select `quote_ident(n.nspname) || '.' || quote_ident(c.relname)` server-side (the WHERE clause already uses this) and scan that single column, or build the qualified name via `NameTuple.ForOutput()` client-side.

makalaaneesh (commented):
[cursor] Automated review — 4 findings (1 critical, 3 warning). See inline comments. Filtered out 4 suggestions/nice-to-haves.

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Yeah there was a pre-existing check, called directly in `checkExportDataPermissions` (which also calls `GetMissingExportDataPermissions` for various DBs). The check is for whether replication slots are available to be created and it directly exits if they are not.

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Done. Changed `listTablesMissingReplicaIdentityChange` to accept `[]NameTuple` directly and use `ForCatalogQuery` to build the IN-list as `(n.nspname, c.relname) IN (...)` pairs, removing both the `lo.Map`+`ForOutput()` in the caller and the `quote_ident` concat in the WHERE clause. Note: the PG sibling `listTablesMissingReplicaIdentityFull` still uses the old approach — should we update that as well?

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Done. Renamed to `tablesWithoutReplicaIdentityChange`.

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Done. Changed to `"Tables not having replica identity CHANGE: "`.

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Done. Changed `listTablesMissingReplicaIdentityChange` to accept `[]NameTuple` directly and use `ForCatalogQuery` to build the IN-list as `(n.nspname, c.relname) IN (...)` pairs, removing both the `lo.Map`+`ForOutput()` in the caller and the `quote_ident` concat in the WHERE clause. Note: the PG sibling `listTablesMissingReplicaIdentityFull` still uses the old approach — should we update that as well?

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Fixed. Wrapped the replica identity check in `!yb.source.IsYBGrpcConnector` rather than doing an early return, so any future checks added to this function aren't inadvertently skipped for the gRPC connector path.

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Fixed. Added `exportType == utils.CHANGES_ONLY || exportType == utils.SNAPSHOT_AND_CHANGES` gate, mirroring the PG sibling. YB-as-source is always `CHANGES_ONLY` today but making this explicit prevents regressions if the call site changes.

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Can address in a follow-up. The PG sibling `listTablesMissingReplicaIdentityFull` still uses the old `queryTableList string` signature — consolidating the two would be cleaner once that function is also migrated to `[]NameTuple`.

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Fixed. Using `sqlname.NewSourceName(tableSchemaName, tableName).Qualified.MinQuoted` instead of manual concatenation — this correctly quotes both schema and table sides.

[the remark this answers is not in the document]
ShivanshGahlot: Fixed. Using `sqlname.NewSourceName(tableSchemaName, tableName).Qualified.MinQuoted` instead of manual concatenation — this correctly quotes both schema and table sides.

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Done. Added `TestYugabyteReplicaIdentityGuardrail` in `yugabytedb_test.go` with 7 sub-tests: table with CHANGE identity (not flagged), table with FULL identity (flagged), mixed list (only FULL appears), `snapshot-only` skips the check, `changes-only` and `snapshot-and-changes` detect the missing identity, and `IsYBGrpcConnector=true` skips the check entirely.

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -1334,7 +1335,75 @@ func (yb *YugabyteDB) GetMissingExportSchemaPermissions(queryTableList string) (
...
+	var tablesWithoutReplicaIdentityChange []string
+	var tableSchemaName, tableName, replicaIdentity, status string
+
+	for rows.Next() {
+		err = rows.Scan(&tableSchemaName, &tableName, &replicaIdentity, &status)
+		if err != nil {
+			return nil, fmt.Errorf("error in scanning query rows for table names: %w", err)
+		}
+		if status == MISSING {
+			tablesWithoutReplicaIdentityChange = append(tablesWithoutReplicaIdentityChange, qualifiedNameByCatalogKey[tableSchemaName+"."+tableName])
```
makalaaneesh: What is the purpose of `qualifiedNameByCatalogKey` ? 
I see two options: 
1. We return []string. In which case, why do you need `qualifiedNameByCatalogKey` ? We can just return what we read from the row (i.e. just schemaName.tableName) since we're just returning strings. 
2. We return []NameTuple. In which case, we need to map the rows we read back to NameTuples. In this case we need a map to get back the name tuple. 

I'd prefer option2. all table names are dealt with in []NameTuple. Less error-prone that way.

cc: @priyanshi-yb in case this is not aligned with how we were doing it in other places

yb-voyager/src/srcdb/yugabytedb.go
makalaaneesh: This seems wrong? We aren't using sqlname.NewSourceName anywhere from what i see

yb-voyager/src/srcdb/yugabytedb.go
makalaaneesh: makes sense. Pls put a TODO

makalaaneesh (commented):
High level Lgtm. one important comment

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Added a TODO.

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: The fix is in place — just implemented differently than I described. Rather than calling `sqlname.NewSourceName` directly, the function pre-builds a `qualifiedNameByCatalogKey` map from the input `[]NameTuple` list:

    qualifiedNameByCatalogKey[sname+"."+tname] = table.ForOutput()

When scanning query results, line 1400 looks up that map by the raw catalog key (`tableSchemaName+"."+tableName`) to get the properly-quoted display name. So both schema and table sides are correctly quoted via `ForOutput()` — no manual concatenation on the output path.

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Good point. Changed the return type to `[]sqlname.NameTuple`. The map is now a `map[string]bool` set (`tablesWithoutIdentityChangeKeys`) — the query collects missing catalog keys, then we filter the original `tableList` to return the matching `NameTuple`s.

yb-voyager/src/srcdb/yugabytedb.go
ShivanshGahlot: Update: the implementation was further simplified. The `qualifiedNameByCatalogKey` map no longer exists. Instead, we collect missing catalog keys into a `map[string]bool` (`tablesWithoutIdentityChangeKeys`) while scanning results, then filter the original `tableList` at the end to return matching `NameTuple`s. The return type is now `[]sqlname.NameTuple`.

makalaaneesh (approved):
LGTM!

PR #3538, opened 2026-05-12
Fix import data --truncate-tables failure when FK child is empty

### LINKED ISSUE #3561: Import data --start-clean fails on TRUNCATE when fk child is empty
Jira Link: [DB-21556](https://yugabyte.atlassian.net/browse/DB-21556)

sincfaio:

`yb-voyager import data --truncate-tables true --start-clean true` (and the equivalent `yb-voyager import data file ...` flow) fails on PostgreSQL/YugabyteDB targets when the target has a non-empty parent table whose FK-dependent child happens to be empty. Root cause: `cleanImportState` filtered the import-scope table list through `tdb.GetNonEmptyTables` before issuing TRUNCATE, so the empty FK-dependent was excluded from the statement and the database rejected it with `cannot truncate a table referenced in a foreign key constraint`.

Fix: in the `--truncate-tables` branch of `cleanImportState`, pass the full import-scope table list (`tableNames`) to `tdb.TruncateTables` instead of the non-empty subset (`nonEmptyNts`). FK-dependent siblings that are part of the migration are now included in the same TRUNCATE statement, eliminating the FK-constraint rejection. The else branch (no `--truncate-tables`, warn + prompt) is untouched and still uses `GetNonEmptyTables`. Both `import data` and `import data file` share this code path, so the fix applies to both commands.

TRUNCATE is intentionally still issued without `CASCADE` — this avoids silently wiping data in tables outside the import scope when running a subset migration (`--table-list`). The narrower limitation that remains: subset migrations where an FK-dependent table is excluded from the migration entirely will still fail, and that case requires a separate fix (or operator intervention).

- **Command line / flags:** No changes. `--truncate-tables true` now succeeds in the FK-empty-child scenario where it previously failed, for both `import data` and `import data file`.
- **Configuration:** No changes.
- **Installation:** No changes.
- **Reports:** No changes.
- **Log output:** the previous single line `Truncating non-empty tables on DB: <list>` is replaced by two lines: `Non-empty tables on DB: <list>` followed by `Truncating all tables in import scope on DB to keep FK-dependents consistent`.

Two new regression tests under build tag `integration_voyager_command`, both using real PostgreSQL + YugabyteDB containers via testcontainers:

- `TestImportData_TruncateTables_FKEmptyChild` (`yb-voyager/cmd/importData_test.go`) — end-to-end PG → YB flow: creates `projects` (parent) + `tasks` (FK child) in both, populates both on PG but only the parent on YB (bug repro), runs `export data` then `import data --start-clean true --truncate-tables true`, and asserts the command succeeds and source/target data match for both tables.
- `TestImportDataFile_TruncateTables_FKEmptyChild` (`yb-voyager/cmd/importDataFile_test.go`) — file-based flow: creates the same FK schema on YB, pre-populates the parent (child empty), generates CSVs for both, and runs `import data file --start-clean true --truncate-tables true`. Asserts the import succeeds and both tables hold the expected row counts after the truncate.

Verified locally:
- `go test -tags integration_voyager_command -timeout 30m -run ^TestImportData_TruncateTables_FKEmptyChild$ ./cmd/` — PASS (~27s)
- `go test -tags integration_voyager_command -timeout 30m -run ^TestImportDataFile_TruncateTables_FKEmptyChild$ ./cmd/` — PASS (~18s)
- The `import data` test fails against the pre-fix binary (TRUNCATE rejected with the FK error) and passes against the fixed binary.
- Existing unit suite (`go test -tags unit ./...`) unaffected.

No.

No.

### DIFF
--- a/yb-voyager/cmd/importData_test.go
+++ b/yb-voyager/cmd/importData_test.go
@@ -2301,14 +2301,12 @@ func TestImportData_TruncateTables_FKEmptyChild(t *testing.T) {
 	defer testutils.RemoveTempExportDir(exportDir)
 
 	postgresContainer := testcontainers.NewTestContainer("postgresql", nil)
-	if err := postgresContainer.Start(ctx); err != nil {
-		utils.ErrExit("Failed to start Postgres container: %v", err)
-	}
+	err := postgresContainer.Start(ctx)
+	testutils.FatalIfError(t, err, "Failed to start Postgres container")
 
 	yugabytedbContainer := testcontainers.NewTestContainer("yugabytedb", nil)
-	if err := yugabytedbContainer.Start(ctx); err != nil {
-		utils.ErrExit("Failed to start YugabyteDB container: %v", err)
-	}
+	err = yugabytedbContainer.Start(ctx)
+	testutils.FatalIfError(t, err, "Failed to start YugabyteDB container")
 
 	createSchemaSQL := `CREATE SCHEMA IF NOT EXISTS test_schema;`
 	dropSchemaSQL := `DROP SCHEMA IF EXISTS test_schema CASCADE;`
@@ -2335,7 +2333,7 @@ CREATE TABLE test_schema.tasks (
 	yugabytedbContainer.ExecuteSqls(createSchemaSQL, createParentSQL, createChildSQL, insertParent)
 	defer yugabytedbContainer.ExecuteSqls(dropSchemaSQL)
 
-	err := testutils.NewVoyagerCommandRunner(postgresContainer, "export data", []string{
+	err = testutils.NewVoyagerCommandRunner(postgresContainer, "export data", []string{
 		"--export-dir", exportDir,
 		"--source-db-schema", "test_schema",
 		"--disable-pb", "true",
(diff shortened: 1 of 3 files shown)


yb-voyager/cmd/importData_test.go
```diff
@@ -2288,3 +2288,79 @@ SELECT 'name_' || g FROM generate_series(1, 100) AS g;`
...
+	testutils.FatalIfError(t, err, "connecting to YugabyteDB")
+	defer ybConn.Close()
+
+	if err := testutils.CompareTableData(ctx, pgConn, ybConn, "test_schema.projects", "project_id"); err != nil {
+		t.Errorf("projects mismatch: %v", err)
+	}
+	if err := testutils.CompareTableData(ctx, pgConn, ybConn, "test_schema.tasks", "task_id"); err != nil {
+		t.Errorf("tasks mismatch: %v", err)
+	}
+}
```
priyanshi-yb, line 2371: can you add a test for the `import data file` command also

priyanshi-yb (approved):
LGTM

PR #3546, opened 2026-05-14
Retry TRUNCATE on YB read-restart (SQLSTATE 40001)

sincfaio:
## Summary

Handle YugabyteDB's SQLSTATE 40001 transient error that occurs during multi-table TRUNCATE operations under hybrid-time clock contention.

## Problem

When multiple tests run against a shared YugabyteDB container (as in CI), catalog mutations accumulate, creating a collision window during distributed TRUNCATE. YB signals this with SQLSTATE 40001 ("Restart read required"), which is transient but was not retried, causing test flakes and deployment failures.

## Solution

Implement linear-backoff retry logic in `yugabytedb.TruncateTables()`:
- Retry up to 5 attempts on SQLSTATE 40001
- Linear backoff: 2s, 4s, 6s, 8s
- Final attempt fails through (user sees the error)

Also fixes the original FK constraint bug from PR #3538: pass all import-scope tables to TRUNCATE (not just non-empty) so FK-dependents are included in the same statement.

## Testing

- Add regression test `TestImportData_TruncateTables_FKEmptyChild` (importData_test.go)
- Add regression test `TestImportDataFile_TruncateTables_FKEmptyChild` (importDataFile_test.go)
- Both verify that TRUNCATE succeeds when a non-empty parent table has an empty FK-child table in the import scope

## Files Changed

- `yb-voyager/src/tgtdb/yugabytedb.go`: Add SQLSTATE 40001 retry logic to `TruncateTables()`
- `yb-voyager/cmd/importData.go`: Adjust logging in `cleanImportState()` (clarifies behavior change)
- `yb-voyager/cmd/importData_test.go`: Add FK regression test
- `yb-voyager/cmd/importDataFile_test.go`: Add FK regression test

### DIFF
--- a/yb-voyager/cmd/importDataFile_test.go
+++ b/yb-voyager/cmd/importDataFile_test.go
@@ -25,7 +25,6 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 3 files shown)


priyanshi-yb (approved):
LGTM

PR #3547, opened 2026-05-14
Fix intermittent numeric overflow in partition-iteration-test

Abhi-yb:
## Summary
- The `source_dml.sql` in `partition-iteration-test` uses `salary = salary * 1.10` which can overflow `NUMERIC(10,2)` when the event generator concurrently sets a high salary value
- Replaced with a fixed value assignment to avoid compounding across iterations

## Test plan
- [x] Ran 3-iteration local test — passed
- Pipeline runs confirmed the overflow at iteration 10 (intermittent)


priyanshi-yb (approved):
LGTM

PR #3548, opened 2026-05-14
De-flake CDC batch-failure tests (TestCDCMultiTableBatchFailureAndResume + 2 related)

makalaaneesh:
## Summary

- Fixes the flake in `TestCDCMultiTableBatchFailureAndResume` (and the same latent flake in `TestCDCBatchFailureAndResume` and `TestCDCMultipleBatchFailures`) by gating the Byteman exception injection on actual queue writes rather than `handleBatch` invocation count.
- Adds `BytemanHelper.AddCDCBatchFailAfterWritesRule` — a reusable helper that installs a pair of rules: one counts events that pass `checkIfEventNeedsToBeWritten`, the other throws at the next `before-batch-streaming` only when at least one prior batch produced writes.
- Loosens post-failure count assertions from exact equality to bounded ranges; end-to-end correctness still pinned by `finalEventCount == 60`, `VerifyNoEventIDDuplicates`, `ValidateRowCount`, and per-operation `ChangesCount`.

## Root cause

`incrementCounter("cdc_batch") == 2` at the `before-batch-streaming` marker advances on every streaming `handleBatch`, including ones whose records all get filtered out by `checkIfEventNeedsToBeWritten` (transaction-boundary records, heartbeats, etc.). When Debezium's first streaming batch happened to contain only such records, the counter still went 1 → 2 and the throw fired before any user data was written, leaving `eventCountAfterFailure == 0` when the test expected `20`. Observed in CI: [run 25863423824](https://github.com/yugabyte/yb-voyager/actions/runs/25863423824/job/75999070716).

## Fix

The new helper installs two rules:

```
RULE <name>_count_writes
... AT entry of cdc("before-write-record")
DO incrementCounter("<name>_events_written")

RULE <name>
... AT entry of cdc("before-batch-streaming")
IF  incrementCounter("<name>_cdc_batch") > 1
    && readCounter("<name>_events_written") > 0
DO  throw
```

The throw fires at the start of a streaming `handleBatch` only after some earlier batch has actually committed events. Filtered-empty first batches no longer trip it.

Loosened assertions:
| Test | Was | Now |
|---|---|---|
| `TestCDCBatchFailureAndResume` | `Equal(20)` | `>= 1 && < 60` |
| `TestCDCMultipleBatchFailures` run 1 | `Equal(20)` | `>= 1 && < 40` |
| `TestCDCMultipleBatchFailures` run 2 | `Equal(40)` | `> run1 && < 60` |
| `TestCDCMultiTableBatchFailureAndResume` | `Equal(20)` | `>= 1 && < 60` |

`TestFirstCDCBatchFailure` (which intentionally asserts `eventCount == 0` by failing on the very first batch entry) is deterministic by construction and left unchanged.

## Test plan

- [x] `TestCDCMultiTableBatchFailureAndResume` — 13 consecutive local runs, 13/13 pass (66–75 s, post-fail=20, final=60)
- [x] `TestCDCBatchFailureAndResume` — 3 consecutive local runs, 3/3 pass (71–72 s)
- [x] `TestCDCMultipleBatchFailures` — 3 consecutive local runs, 3/3 pass (95–96 s, counts `[20, 40, 60]`)
- [ ] CI Failpoint: Export job is green on this branch
- [ ] CI runs the same three tests across multiple iterations (or back-to-back PRs) without re-introducing the flake

### DIFF
--- a/yb-voyager/src/testlivemigration/export_data_cdc_failures_test.go
+++ b/yb-voyager/src/testlivemigration/export_data_cdc_failures_test.go
@@ -113,15 +113,13 @@ func TestCDCBatchFailureAndResume(t *testing.T) {
 	bytemanHelper, err := testutils.NewBytemanHelper(exportDir)
 	require.NoError(t, err, "Failed to create Byteman helper")
 
-	bytemanHelper.AddRuleFromBuilder(
-		testutils.NewRule("fail_cdc_batch_2").
-			AtMarker(testutils.MarkerCDC, "before-batch-streaming").
-			If("incrementCounter(\"cdc_batch\") == 2").
-			ThrowException("java.lang.RuntimeException", "TEST: Simulated batch processing failure on batch 2"),
+	bytemanHelper.AddCDCBatchFailAfterWritesRule(
+		"fail_cdc_batch_after_writes",
+		"TEST: Simulated batch processing failure at batch start after writes",
 	)
 	require.NoError(t, bytemanHelper.WriteRules())
 
-	// Run 1: export with Byteman injection - should fail on 2nd CDC batch
+	// Run 1: export with Byteman injection - should fail on a streaming batch after some events have been written.
 	err = lm.StartExportDataWithEnv(true, nil, bytemanHelper.GetEnv())
 	require.NoError(t, err, "Failed to start export")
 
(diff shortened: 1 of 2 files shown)


makalaaneesh:
image

ShivanshGahlot (approved):
The analysis makes sense. I guess its fine to confirm that "some events have landed but not all" in the verification part instead of the older way where we matched the exact number of events. 
LGTM

PR #3550, opened 2026-05-21
Fix import-data hang when --adaptive-parallelism-max is below computed default parallel-jobs

makalaaneesh:

Fix a silent hang during `import data` when the user passes
`--adaptive-parallelism-max N` without `--parallel-jobs`, and the
auto-computed default parallel-jobs (`clusterCores / 4`) exceeds `N`.

The hang reproduces on clusters where total cluster cores ÷ 4 is larger
than the user's `--adaptive-parallelism-max` (e.g. 9 tservers × 32 cores
= 288 cores → default Parallelism = 72; user supplies
`--adaptive-parallelism-max 12`). `setDefaultParallelism` was not
reconciling the two, and `NewConnectionPool` then blocked forever on
the init loop because it tried to drain 72 elements from a channel of
capacity 12.

Two fixes:
1. `setDefaultParallelism` (yb-voyager/src/tgtdb/yugabytedb.go) now
   caps `Parallelism` to `MaxParallelism` when adaptive parallelism is
   enabled and the auto-computed default exceeds the user-supplied max,
   logging a `Warnf` so the cap is visible in the log.
2. `NewConnectionPool` (yb-voyager/src/tgtdb/conn_pool.go) now returns
   `(*ConnectionPool, error)` and refuses to start when
   `NumConnections > NumMaxConnections` (or sizes are otherwise
   invalid), so any future invariant violation fails fast instead of
   silently deadlocking. Existing callers in `yugabytedb.go` and
   `postgres.go` are updated to propagate the error.

No change to `validateParallelismFlags` — voyager already rejects
`--parallel-jobs` together with adaptive parallelism, so the only
path to a mismatch is the auto-computed default, which is now handled.

- A run that previously hung silently now either starts successfully
  with a warning log (`Computed default parallel-jobs (X) exceeds
  --adaptive-parallelism-max (Y); capping initial parallelism to Y`),
  or, in the unlikely case of any future invariant violation, exits
  with `creating connection pool: invalid pool config: NumConnections
  (X) > NumMaxConnections (Y)` instead of hanging.
- No changes to CLI flags, configuration, installation, or reports.

New `unit`-tagged tests in `yb-voyager/src/tgtdb/parallelism_unit_test.go`:
- `TestSetDefaultParallelism_AdaptiveMax_CapsComputedParallelism` —
  the regression case (Parallelism=72, MaxParallelism=12 → capped to 12).
- `TestSetDefaultParallelism_AdaptiveNoUserMax_DefaultsMaxTo4xParallelism`
  — existing default-max path preserved.
- `TestSetDefaultParallelism_AdaptiveDisabled_MaxEqualsParallelism` —
  non-adaptive path preserved.
- `TestSetDefaultParallelism_AdaptiveMaxAboveParallelism_LeavesAsIs` —
  no-cap-needed case.
- `TestNewConnectionPool_RejectsInvertedSizes` — uses a 2s timeout to
  prove there's no deadlock and an error is returned.
- `TestNewConnectionPool_AcceptsValidSizes`,
  `TestNewConnectionPool_RejectsZeroMax` — boundary cases.

Verified locally: `go build ./...`, `go vet ./...`, and
`go test -tags unit -count=1 ./...` all pass, including `./cmd/...`
and `./src/tgtdb/...`. Existing integration-tagged tests in
`conn_pool_test.go` were updated to the new `NewConnectionPool`
signature (compile-verified; not run locally — they need a live YB
testcontainer).

Not yet exercised: end-to-end import-data run reproducing the original
user scenario (9-tserver cluster + `--adaptive-parallelism-max 12`).
The unit tests cover the deterministic logic; an e2e run would only
re-confirm what the unit tests already verify.

No.

No.

### DIFF
--- a/yb-voyager/src/tgtdb/conn_pool.go
+++ b/yb-voyager/src/tgtdb/conn_pool.go
@@ -71,7 +71,18 @@ type ConnectionPool struct {
 	pendingConnsToCloseLock sync.Mutex
 }
 
-func NewConnectionPool(params *ConnectionParams) *ConnectionPool {
+func NewConnectionPool(params *ConnectionParams) (*ConnectionPool, error) {
+	// NumConnections > NumMaxConnections would deadlock the init loop below
+	// (idleConns has only NumMaxConnections slots, but the loop tries to drain
+	// NumConnections from it). Refuse to start instead of hanging silently.
+	if params.NumConnections > params.NumMaxConnections {
+		return nil, fmt.Errorf("invalid pool config: NumConnections (%d) > NumMaxConnections (%d)",
+			params.NumConnections, params.NumMaxConnections)
+	}
+	if params.NumConnections < 0 || params.NumMaxConnections < 1 {
+		return nil, fmt.Errorf("invalid pool config: NumConnections=%d NumMaxConnections=%d (must be >=0 and >=1)",
+			params.NumConnections, params.NumMaxConnections)
+	}
 	pool := &ConnectionPool{
 		params:                    params,
 		conns:                     make(chan *pgx.Conn, params.NumMaxConnections),
@@ -91,7 +102,7 @@ func NewConnectionPool(params *ConnectionParams) *ConnectionPool {
 	if pool.params.SessionInitScript == nil {
 		pool.params.SessionInitScript = defaultSessionVars
 	}
-	return pool
+	return pool, nil
 }
 
 func (pool *ConnectionPool) GetNumConnections() int {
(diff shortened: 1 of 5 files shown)


makalaaneesh:
jenkins live migration - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/8086/stages/

jenkins live migration resumption - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/599/stages/

offline resumption tests : 
- https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-resumption-tests/115/
- https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-resumption-tests/114/
- https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-resumption-tests/117/

priyanshi-yb (approved):
LGTM

PR #3556, opened 2026-05-21
Handle missing pg_stat_statements on PostgreSQL read replicas during assessment

AmeyaKasture:
This PR fixes PostgreSQL multi-node migration assessment when `pg_stat_statements` is unavailable on one or more read replicas because it is not loaded through `shared_preload_libraries`. Previously, Voyager checked PGSS availability during preflight, but the metadata collection path used the primary node’s PGSS status for every node. As a result, if the primary had PGSS enabled but a replica did not, Voyager could still try to query `pg_stat_statements` on that replica and fail the assessment.
The change preserves PGSS availability per validated replica and passes each node’s own PGSS status into metadata collection. This lets Voyager continue collecting other assessment metadata from replicas while skipping query-summary collection only on nodes where PGSS is unavailable.
`assess-migration` no longer fails during replica metadata collection when a PostgreSQL read replica does not have `pg_stat_statements` loaded via `shared_preload_libraries`. There are no command-line, configuration, or installation changes. Assessment output can now be limited for query-level analysis on replicas where PGSS is unavailable, while the rest of replica assessment continues.
Ran focused package tests:
`go test ./src/migassessment ./src/srcdb`
No new unit tests were added. Manual/integration coverage with a PostgreSQL primary and read replica where PGSS is disabled on the replica would be useful to validate the end-to-end behavior.
No.
No.
---
| Component        


yb-voyager/src/migassessment/permissions.go
ShivanshGahlot: We want to solve the problem that currently we just return the pgss status of the primary node and we just blindly apply it to all the nodes. A few pointers regarding the current approach:

- We are adding a new field `PgssEnabled` to the `ReplicaEndpoint` struct. I feel that struct should be purely used for connection related things.
- In the current approach we are passing around `validatedReplicas` and are lazily mutating the properties inside them. This approach currently works because we are passing the same copy around and it is mutating the same underlying array. What happens if we at some middle step clone the array and use different arrays at different places? There is no documentation or check that tells us we should clone or copy this array only after `CheckAssessmentPermissionsOnAllNodes` has filled it with the pgss info. And even with such documentation it becomes very difficult to track such things. We are passing around an array and the general idea a developer will get from that is that we are just going to read from it.
- The function signature for `CheckAssessmentPermissionsOnAllNodes` still returns `pgssEnabled`. Now this is misleading. What is the `pgssEnabled` for? It's actually just the primary's status, but the name reads like an assessment-wide answer.

So I feel a better approach might be to make the signature of this function:

```
func CheckAssessmentPermissionsOnAllNodes(
    source *srcdb.Source,
    validatedReplicas []srcdb.ReplicaEndpoint,
) (pgssByNode map[string]bool, err error)
```

Keyed by the same `nodeName` the gather phase already uses ("primary" / "<host>:<port>"). Built from the results `[]NodePermissionResult` this function already constructs. `GatherAssessmentMetadataFromPG` looks up `pgssByNode[n.nodeName]` when building each `collectionNode`.

This drops the new fields on `ReplicaEndpoint` and `collectionNode`, eliminates the mutation, and makes the return type honest.

ShivanshGahlot (approved):
LGTM. Thanks!

PR #3564, opened 2026-05-25
Fix non-deterministic step count in assess-migration progress display

AmeyaKasture:

  The `assess-migration` progress indicator showed an inconsistent number of steps
  depending on which metadata collection scripts were skipped for a given source
  database.
 This happened because `detectStageFromOutput` in `gather.go` counted
  "Skipping ..." lines and the "Assessment metadata collection started" preamble as
  The `assess-migration` progress indicator showed an inconsistent number of steps
  depending on which metadata collection scripts were skipped for a given source
  database. 
  Since the set of skipped scripts varies by source type and
  configuration, the total step count was non-deterministic and differed across
  workflows.

  This PR narrows `detectStageFromOutput` to only recognize actual data-collection
  steps — lines containing "Collecting" or "Executing" — and the final "Complete"

"Skipping ..." lines (which are informational follow-ups to a
  "Collecting ..." line for the same script) and the startup preamble are no longer
  counted as steps. 
This makes the progress step count deterministic and consistent
  steps — lines containing "Collecting" or "Executing" — and the final "Complete"
  message. "Skipping ..." lines (which are informational follow-ups to a
  "Collecting ..." line for the same script) and the startup preamble are no longer
  counted as steps. This makes the progress step count deterministic and consistent
  across all source database types.

  The progress bar / step counter shown during `assess-migration` metadata collection
  will now display a stable, predictable step count. Previously, skipped scripts
  inflated the count and caused the progress to appear inconsistent. No changes to
  the command line interface, configuration, installation, or reports.

Manually checking the workflows 

  No.

  No.


ShivanshGahlot:
@AmeyaKasture can you fix the PR description. It seems like it got messed up.

yb-voyager/src/migassessment/gather.go
```diff
@@ -531,12 +531,6 @@ func runGatherAssessmentMetadataScriptBuffered(
 		return fmt.Errorf("error starting gather assessment metadata script: %w", err)
 	}
 
-	// Report starting status
```
makalaaneesh, line 534: why are we removing this? 
is it safe to do so? cc: @ShivanshGahlot i think you initially added this for read-replicas?

AmeyaKasture: This is done to make the replica case be in accordance with the without replica case, it encounters the same issue that this is treated as a seperate stage and thus end up incrementing the count,

We can keep this too , but for it to work seamlessly we would have to make the changes in the filtering part of the stage

Both approaches work

makalaaneesh: sorry i should have been more clear; i understand the change but the question i have is: is this "starting collection..." stage progress being used anywhere other than for incrementing stages? (current use-case)

ShivanshGahlot: I guess it should not break anything as @AmeyaKasture explained.

AmeyaKasture: Other than the increment stage, its not used anywhere because its passed in the progress channel which itself is only used in tracking the progress and incrementing the stage

ShivanshGahlot: It just provides more or less an instant visual feedback on process start. So it should be safe to remove. I don't think there is any other place it is being used other than the step counter.

makalaaneesh (approved):
LGTM. 
Pls look into why the tests are failing

PR #3566, opened 2026-05-29
Send source schema OIDs to the callhome in the source db details

priyanshi-yb:
Adds a field in the source DB details to also send the schema OIDs for unique identity, two migrations in callhome on the same DB ID and with the same schema OIDs.

manually
```
callhome=# select migration_phase,source_db_details from diagnostics where migration_uuid='b1097a26-57ed-4678-9842-3dc9611eef21';
-[ RECORD 1 ]-----+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
migration_phase   | assess-migration
source_db_details | {"host": "", "db_id": 1632131, "db_name": "db_1fb007232479fbe7", "db_type": "postgresql", "db_version": "17.2 (Debian 17.2-1.pgdg120+1)", "schema_oids": [2200, 1632132], "schema_names": ["schema_5f50868529bea433", "schema_103d6939255641b0"], "payload_version": "1.2", "total_db_size_bytes": 3006464, "db_system_identifier": 7449422312135041063}
-[ RECORD 3 ]-----+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
migration_phase   | export-data
source_db_details | {"host": "", "db_id": 1632131, "db_name": "db_1fb007232479fbe7", "db_type": "postgresql", "db_version": "17.2 (Debian 17.2-1.pgdg120+1)", "schema_oids": [2200, 1632132], "schema_names": ["schema_5f50868529bea433", "schema_103d6939255641b0"], "payload_version": "1.2", "total_db_size_bytes": 3006464, "db_system_identifier": 7449422312135041063}
-[ RECORD 4 ]-----+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
migration_phase   | export-schema
source_db_details | {"host": "", "db_id": 1632131, "db_name": "db_1fb007232479fbe7", "db_type": "postgresql", "db_version": "17.2 (Debian 17.2-1.pgdg120+1)", "schema_oids": [2200, 1632132], "schema_names": ["schema_5f50868529bea433", "schema_103d6939255641b0"], "payload_version": "1.2", "total_db_size_bytes": 3006464, "db_system_identifier": 7449422312135041063}

callhome=# 

```

yes added a new field in source-db-details json but increased the payload version to 1.2

---

| Component        

### DIFF
--- a/migtests/scripts/functions.sh
+++ b/migtests/scripts/functions.sh
@@ -1525,6 +1525,7 @@ normalize_callhome_json() {
             .db_version = "IGNORED" |
             .db_system_identifier = "IGNORED" |
             .db_id = "IGNORED" |
+            .schema_oids = "IGNORED" |
             .target_db_details = "IGNORED" |
             .total_db_size_bytes = "IGNORED"
         elif type == "array" then
--- a/migtests/tests/pg/datatypes/expected_callhome_payloads/import_data_callhome.json
+++ b/migtests/tests/pg/datatypes/expected_callhome_payloads/import_data_callhome.json
@@ -12,6 +12,7 @@
       "enable_yb_adaptive_parallelism": true,
       "adaptive_parallelism_max": 4,
       "error_policy_snapshot": "abort",
+      "iterative_cutover_enabled": false,
       "start_clean": false,
       "data_metrics": {
         "migration_snapshot_total_rows": 101,
--- a/migtests/tests/pg/datatypes/expected_callhome_payloads/import_data_callhome.json
+++ b/migtests/tests/pg/datatypes/expected_callhome_payloads/import_data_callhome.json
@@ -4,7 +4,7 @@
     "migration_type": "offline",
     "migration_uuid": "507f2869-f4bb-4459-b776-79c2f85f8d7e",
     "phase_payload": {
-      "payload_version": 1.5,
+      "payload_version": 1.6,
       "batch_size": 20000,
       "control_plane_type": "",
       "parallel_jobs": 1,
--- a/yb-voyager/cmd/callhome_payload_builder.go
+++ b/yb-voyager/cmd/callhome_payload_builder.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/samber/lo"
@@ -184,6 +185,8 @@ func anonymizeQualifiedTableNames(tableNames []string) []string {
 
 // anonymizeSourceDBDetails creates anonymized source DB details for callhome
 func anonymizeSourceDBDetails(source *srcdb.Source) callhome.SourceDBDetails {
+	// Sort schema oids to ensure consistent order
+	slices.Sort(source.SchemaOids)
 	details := callhome.SourceDBDetails{
 		PayloadVersion:     callhome.SOURCE_DB_DETAILS_PAYLOAD_VERSION,
 		DBType:             source.DBType,
--- a/yb-voyager/cmd/callhome_payload_builder.go
+++ b/yb-voyager/cmd/callhome_payload_builder.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"sort"
 	"strings"
 
 	"github.com/samber/lo"
--- a/yb-voyager/cmd/callhome_payload_builder.go
+++ b/yb-voyager/cmd/callhome_payload_builder.go
@@ -191,6 +191,7 @@ func anonymizeSourceDBDetails(source *srcdb.Source) callhome.SourceDBDetails {
 		DBSize:             source.DBSize,
 		DBSystemIdentifier: source.DBSystemIdentifier,
 		DBID:               source.DBID,
+		SchemaOids:         source.SchemaOids,
 	}
 
 	// Anonymize database name
--- a/yb-voyager/src/callhome/diagnostics.go
+++ b/yb-voyager/src/callhome/diagnostics.go
@@ -93,8 +93,9 @@ type Payload struct {
 Version History
 1.0: Introduced DBName and SchemaNames fields
 1.1: Added db_id (PostgreSQL/YugabyteDB: pg_database.oid; Oracle: v$database.dbid; MySQL: 0)
+1.2: Added schema_oids field (PostgreSQL/YugabyteDB: pg_namespace.oid)
 */
-var SOURCE_DB_DETAILS_PAYLOAD_VERSION = "1.1"
+var SOURCE_DB_DETAILS_PAYLOAD_VERSION = "1.2"
 
 type SourceDBDetails struct {
 	PayloadVersion     string   `json:"payload_version"`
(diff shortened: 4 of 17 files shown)


yb-voyager/cmd/callhome_payload_builder.go
```diff
@@ -191,6 +191,7 @@ func anonymizeSourceDBDetails(source *srcdb.Source) callhome.SourceDBDetails {
 		DBSize:             source.DBSize,
 		DBSystemIdentifier: source.DBSystemIdentifier,
 		DBID:               source.DBID,
+		SchemaOids:         source.SchemaOids,
```
makalaaneesh, line 196: should we sort them? what if the order of schemas in the pg_catalog is non-deterministic?

yb-voyager/src/callhome/diagnostics.go
AmeyaKasture: We are dropping db id in the struct below 
`DBID               int64    `json:"db_id,omitempty"``
I checked the python file , and it doesnt seem to be used so maybe we should remove this from the documentation that this cannot be dropped

yb-voyager/src/srcdb/postgres.go
```diff
@@ -653,6 +653,35 @@ func (pg *PostgreSQL) FetchDBID() error {
 	return nil
 }
 
+func (pg *PostgreSQL) FetchSchemaOids() error {
+	var oids []int64
+	schemaList := sqlname.JoinIdentifiersUnquoted(pg.source.Schemas, "','")
+	query := fmt.Sprintf(`SELECT oid FROM pg_namespace WHERE nspname IN ('%s')`, schemaList)
```
AmeyaKasture, line 659: A schema name containing a single quote will break the query, and the pattern violates the cmd/Go BUGBOT SQL rule to use parameterized queries for user-supplied values. It also leaves the new field untested for case-sensitive or unusual identifiers

Bugbot Rules:
`- Use parameterized queries or prepared statements rather than `fmt.Sprintf` with user-supplied values.
`

It would break for schemas like cutomer's_name

migtests/scripts/functions.sh
```diff
@@ -1525,6 +1525,7 @@ normalize_callhome_json() {
             .db_version = "IGNORED" |
             .db_system_identifier = "IGNORED" |
             .db_id = "IGNORED" |
+            .schema_oids = "IGNORED" |
```
AmeyaKasture, line 1528: Most of the tests ignore the value, so can we add focussed unit tests for the same

migtests/tests/pg/datatypes/expected_callhome_payloads/import_data_callhome.json
```diff
@@ -4,14 +4,15 @@
...
-      "payload_version": 1.5,
+      "payload_version": 1.6,
       "batch_size": 20000,
       "control_plane_type": "",
       "parallel_jobs": 1,
       "on_primary_key_conflict_action": "ERROR-POLICY",
       "enable_yb_adaptive_parallelism": true,
       "adaptive_parallelism_max": 4,
       "error_policy_snapshot": "abort",
+      "iterative_cutover_enabled": false,
```
AmeyaKasture, line 15: Just checking: is this iterative_cutover_enabled fixture update needed for the schema OID change, or is it from another call-home update that this branch includes?

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: We don't support schema names with these special characters, so this is not a valid case, and we use this format of query in all most all source related queries.

migtests/scripts/functions.sh
priyanshi-yb: Currently, we don't have any unit tests for explicit callhome validations. Anyways, this schema oid thing can't be validated properly every time, as these will change on every run.

migtests/tests/pg/datatypes/expected_callhome_payloads/import_data_callhome.json
priyanshi-yb: This is a miss in the earlier commit for callhome changes

yb-voyager/src/callhome/diagnostics.go
priyanshi-yb: yeah, this isn't required, cursor made this change. will revert

yb-voyager/cmd/callhome_payload_builder.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
```
makalaaneesh: [AI] **Unused `sort` import breaks the build.**

This PR adds `"sort"` to the import block, but nothing in the file uses it — the new sort call on line 189 is `slices.Sort(source.SchemaOids)`, which comes from the already-imported `golang.org/x/exp/slices`. Go treats an unused import as a compile error, so the entire `cmd` package fails to build:

```
cmd/callhome_payload_builder.go:23:2: "sort" imported and not used
```

Confirmed with `go build ./...` from `yb-voyager/`.

Suggested fix: remove the `"sort"` import line.

makalaaneesh (commented):
[AI] Automated review — 2 findings (1 critical, 1 warning). See inline comments.

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -104,9 +105,10 @@ type SourceDBDetails struct {
 	DBSize             int64    `json:"total_db_size_bytes"`            //bytes
 	Role               string   `json:"role,omitempty"`                 //for differentiating replica details
 	DBSystemIdentifier int64    `json:"db_system_identifier,omitempty"` //Database system identifier for unique instance identification (currently only implemented for PostgreSQL)
-	DBID               int64    `json:"db_id"`                          // postgresql/yugabytedb: pg_database.oid;
+	DBID               int64    `json:"db_id,omitempty"`                // postgresql/yugabytedb: pg_database.oid;
```
makalaaneesh, line 108: [AI] **`db_id` changed to `omitempty` alters payload shape for db_id=0 sources.**

The `DBID` JSON tag changed from `json:"db_id"` to `json:"db_id,omitempty"`, and `db_id` was simultaneously dropped from the "SHOULD NOT REMOVE" comment that points at the yugabyte-growth parser (`diagnostics_v2/main.py#L549`).

With `omitempty`, any source where `DBID == 0` will now omit `db_id` from the payload entirely. Per the v1.1 version-history note in this same file, MySQL deliberately reports `db_id = 0`, so MySQL migrations will silently stop emitting the field rather than sending `0`. Per the callhome BUGBOT rule ("When changing any struct serialized into callhome payloads ... verify downstream parsing still works"), please confirm the growth parser tolerates a missing `db_id` key for these sources before merge.

If the intent is only to suppress `db_id` when unset, consider whether 0 is actually a meaningful value worth keeping for MySQL, or document that MySQL no longer reports it.

AmeyaKasture (approved):
LGTM!

PR #3567, opened 2026-05-29
Fix: Do not allow running different command names but same underlying command together 

priyanshi-yb:
Example - 
1. export data and export data from source 
2. import data and import data to target 

yes, now running `yb-voyager export data` and `yb-voyager export data from source` commands together will error out
```
➜ ~ yb-voyager export data from source -c ~/export-dir-hackathon/config.yaml
Using config file: /Users/priyanshigupta/export-dir-hackathon/config.yaml
Using export-dir: /Users/priyanshigupta/export-dir
migrationID: fe5b8528-d141-41a2-b4c5-06985b595b89
....
export of data for source type as 'postgresql'
Continuing streaming from where we left off...
num tables to export: 6
table list for data export: [public.customers public.sales_region public.test_partitions_sequences public.range_columns_partition_test public.sales public.emp]
streaming changes to a local queue file...

➜  ~ yb-voyager export data from source -c ~/export-dir-hackathon/config.yaml 
Using config file: /Users/priyanshigupta/export-dir-hackathon/config.yaml
Using export-dir: /Users/priyanshigupta/export-dir
Another instance of yb-voyager 'export data' is running for this migration
```

Manually 

No
No
---

| Component        

### DIFF
--- a/yb-voyager/cmd/config.go
+++ b/yb-voyager/cmd/config.go
@@ -172,7 +172,7 @@ var allowedImportDataToSourceConfigKeys = mapset.NewThreadUnsafeSet[string](
 var allowedImportDataToSourceReplicaConfigKeys = mapset.NewThreadUnsafeSet[string](
 	"log-level", "run-guardrails-checks",
 	"batch-size", "parallel-jobs", "truncate-tables", "disable-pb", "max-retries-streaming",
-	"prometheus-metrics-port",	
+	"prometheus-metrics-port",
 	// environment variables keys
 	"ybvoyager-max-colocated-batches-in-progress", "num-event-channels",
 	"event-channel-size", "max-events-per-batch", "max-interval-between-batches",
@@ -240,9 +240,9 @@ var allowedConfigSections = map[string]mapset.Set[string]{
 }
 
 // Define mutually exclusive section groups
-var aliasCommandsPrefixes = [][]string{
-	{"export-data", "export-data-from-source"},
-	{"import-data", "import-data-to-target"},
+var aliasCommandsPrefixes = map[string][]string{
+	"export-data": {"export-data", "export-data-from-source"},
+	"import-data": {"import-data", "import-data-to-target"},
 }
 
 // ConfigParam represents a CLI flag/Config/EnvVar whose value was set by the user depending on the mode of configuration.
--- a/yb-voyager/cmd/root.go
+++ b/yb-voyager/cmd/root.go
@@ -297,7 +297,7 @@ func resolveToActiveIterationIfRequired(cmd *cobra.Command) error {
 	*/
 	cmdPath := cmd.CommandPath()
 	isFallbackCmd := slices.Contains(fallbackPhaseCommands, cmdPath)
-	//setting the metaDB to the iteration metaDB as the getCutoverStatus is using the global metaDB   
+	//setting the metaDB to the iteration metaDB as the getCutoverStatus is using the global metaDB
 	//and we are fetching the cutover status for the current iteration from this metaDB.
 	latestIterInForwardPhase := (GetCutoverStatus(iterationMetaDB) == NOT_INITIATED)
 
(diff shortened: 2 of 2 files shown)


yb-voyager/cmd/root.go
```diff
@@ -504,7 +504,20 @@ func validateExportDirFlag() {
 	}
 }
 
+var commandAliases = map[string]string{
```
makalaaneesh: [AI] **Alias source-of-truth is duplicated — risk of drift with `config.go`.**

`config.go` already encodes the exact same command-equivalence pairs in `aliasCommandsPrefixes`:

```go
var aliasCommandsPrefixes = [][]string{
    {"export-data", "export-data-from-source"},
    {"import-data", "import-data-to-target"},
}
```

This PR introduces a second, independent declaration of the same knowledge (`commandAliases`) in a different shape (`map[alias]canonical`). If a future command rename adds a third alias pair, a maintainer must remember to update both structures; updating only one silently breaks either config-key validation or the lockfile/command-ID collapse.

Per `.cursor/BUGBOT.md` ("Consolidate duplicate logic into shared helpers rather than copy-pasting") and `yb-voyager/.cursor/BUGBOT.md` ("Avoid introducing new global variables"), prefer deriving `commandAliases` from `aliasCommandsPrefixes` (or vice-versa) so there is a single source of truth — e.g. build the alias→canonical map from `aliasCommandsPrefixes` in an `init()` or a shared helper.

yb-voyager/cmd/root.go
```diff
@@ -504,7 +504,20 @@ func validateExportDirFlag() {
...
+	"export-data-from-source": "export-data",
+	"import-data-to-target":   "import-data",
+}
+
 func GetCommandID(c *cobra.Command) string {
+	commandID := buildCommandID(c)
```
makalaaneesh, line 508: [AI] **Aliasing `GetCommandID` also silently changes the log file name, not just the lockfile.**

`GetCommandID` is used for two distinct purposes:
- the lockfile path: `.<id>Lockfile.lck` (root.go:127, 158) — this is the intended target of the fix
- the per-command log file name: `yb-voyager-<id>.log` (logging.go:57)

With this change, running `export data from source` now writes to `yb-voyager-export-data.log` instead of `yb-voyager-export-data-from-source.log` (same for `import data to target` → `yb-voyager-import-data.log`). This is a behavioral/observability change that existing user scripts, support tooling, and docs referencing the legacy log file names will not expect.

If merging the logs of the two equivalent commands is intended, please call it out in the PR description / changelog. If only lockfile sharing is intended, consider keeping the command-ID-to-log-name mapping distinct from the lockfile-collision key (e.g. resolve the alias only at the lockfile call sites in `lockExportDir`, rather than inside `GetCommandID` which feeds both paths).

Note also: no unit test was added for the new alias resolution. Per `yb-voyager/.cursor/BUGBOT.md` (Testing), a small table-driven `unit` test asserting `GetCommandID` for both the legacy and new forms — and confirming the fall-back/fall-forward forms (`import-data-to-source`, `import-data-to-source-replica`) are NOT collapsed — would guard against accidental over-collapsing.

makalaaneesh (commented):

yb-voyager/cmd/config.go
```diff
@@ -240,9 +240,9 @@ var allowedConfigSections = map[string]mapset.Set[string]{
 }
 
 // Define mutually exclusive section groups
-var aliasCommandsPrefixes = [][]string{
-	{"export-data", "export-data-from-source"},
-	{"import-data", "import-data-to-target"},
+var aliasCommandsPrefixes = map[string][]string{
```
makalaaneesh, line 243: Seems like a structural change. does this not require changes in config.go (current usage)?

yb-voyager/cmd/root.go
makalaaneesh: @priyanshi-yb I think this is worth thinking about. What do we want our log files to be named? is it "export-data.log" or "export-data-from-source.log"?

yb-voyager/cmd/config.go
priyanshi-yb: No, previously it was an array of arrays of strings, and we accessed it via `for _, aliases:= range aliasCommandsPrefixes,` and now I have changed it to map[string][]string. So the previous loop will work as is as we only need a list of aliases there, but in the new use case, I am accessing a key, which is a fixed command name for both, and then checking the aliases, so that will just be accessed using `for alias, aliases := range aliasCommandsPrefixes`

yb-voyager/cmd/root.go
priyanshi-yb: I think since the name has been `export-data` from the starting, I feel it is okay to keep the log file as `export-data.log` for any command? what do you think?

makalaaneesh (approved):
LGTM

PR #3568, opened 2026-06-01
Fix prepared-statement cache PID collision on multi-node import targets

makalaaneesh:

Fixes a prepared-statement cache collision that caused `import data` to fail against multi-node YugabyteDB targets.
https://yugabyte.atlassian.net/browse/DB-19670
https://github.com/yugabyte/yb-voyager/issues/3266

The import connection pool (`ConnectionPool`) caches the set of prepared statements per connection so it can skip re-preparing a statement that is already prepared on a given connection. That cache was keyed by the connection's **backend PID** (`conn.PgConn().PID()`). Backend PIDs are only unique *within a single node* — two connections to different nodes of a multi-node cluster (or behind a load balancer) can report the same PID. On such a collision, voyager believed a statement was already prepared on a connection where it had never been prepared, skipped the `Prepare` call, and the batch later failed when it referenced the (non-existent) prepared statement — surfacing as a confusing `syntax error at or near ...` for the prepared-statement name.

This only affected multi-node targets. Single-node PostgreSQL targets used in `import-data-to-source` / `import-data-to-source-replica` (fall-back / fall-forward) never collide, which is why it was not seen there.

The fix keys the cache by the `*pgx.Conn` object itself instead of the PID. The connection object is unique per live connection regardless of which node it lands on or whether a load balancer is in front, so it cannot collide (a composite `(host, PID)` key would *not* have fixed the load-balancer case, where every connection reports the same host). A new `removePreparedStmtCacheForConn` helper is called on every connection-close path; this includes `RemoveConnectionsForHosts`, which previously closed connections without evicting their cache entry — harmless under PID keying but a leak once entries are keyed by a pointer that keeps the connection object alive.

No user-facing changes. No changes to the command line, configuration, installation, or reports. This is an internal correctness fix; the visible effect is that `import data` to multi-node targets no longer intermittently fails with the prepared-statement error.

- Added `unit`-tagged tests in `conn_pool_unit_test.go` (run in CI's default `unit` tier, no DB required):
  - `TestPreparedStmtCache_NoCrossConnCollision` — the regression test: a statement cached on one connection must not be reported as prepared on a different connection (the exact failure the PID key caused).
  - `TestPreparedStmtCache_PerStatement` — distinct statement names on the same connection are tracked independently.
  - `TestPreparedStmtCache_RemoveOnClose` — closing a connection evicts its cache entry.
- These exercise the cache helpers directly using bare `&pgx.Conn{}` values, since the helpers only use the connection as a map key and never dereference it — so no mocks or live database are needed.
- `go build ./...`, `go vet -tags unit ./src/tgtdb/`, and `go test -tags unit -run TestPreparedStmtCache ./src/tgtdb/` all pass.
- End-to-end verification against a real multi-node YB target was not added as an integration test; the unit tests cover the keying logic where the bug lived.

No.

No. The cache is in-memory only; no on-disk structures are affected.

### DIFF
--- a/yb-voyager/src/tgtdb/conn_pool_unit_test.go
+++ b/yb-voyager/src/tgtdb/conn_pool_unit_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/jackc/pgx/v4"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // The prepared-statement cache helpers key only on the *pgx.Conn pointer; they
@@ -34,7 +35,7 @@ func newTestPool(t *testing.T) *ConnectionPool {
 		NumMaxConnections: 1,
 		ConnUriList:       []string{"postgres://localhost/test"},
 	})
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	return pool
 }
 
--- a/yb-voyager/src/tgtdb/conn_pool_unit_test.go
+++ b/yb-voyager/src/tgtdb/conn_pool_unit_test.go
@@ -1,0 +1,92 @@
+//go:build unit
+
+/*
+Copyright (c) YugabyteDB, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package tgtdb
+
+import (
+	"testing"
+
+	"github.com/jackc/pgx/v4"
+	"github.com/stretchr/testify/assert"
+)
+
+// The prepared-statement cache helpers key only on the *pgx.Conn pointer; they
+// never dereference the connection. So we can exercise the collision logic with
+// bare conn objects and no real database.
+func newTestPool(t *testing.T) *ConnectionPool {
+	t.Helper()
+	pool, err := NewConnectionPool(&ConnectionParams{
+		NumConnections:    1,
+		NumMaxConnections: 1,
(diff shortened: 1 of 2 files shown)


yb-voyager/src/tgtdb/conn_pool_unit_test.go
```diff
@@ -0,0 +1,92 @@
...
+// never dereference the connection. So we can exercise the collision logic with
+// bare conn objects and no real database.
+func newTestPool(t *testing.T) *ConnectionPool {
+	t.Helper()
+	pool, err := NewConnectionPool(&ConnectionParams{
+		NumConnections:    1,
+		NumMaxConnections: 1,
+		ConnUriList:       []string{"postgres://localhost/test"},
+	})
+	assert.NoError(t, err)
```
makalaaneesh: [cursor] **Use `require.NoError` for must-succeed setup, not `assert.NoError`.**

Per `yb-voyager/src/tgtdb/.cursor/BUGBOT.md` (and the repo `BUGBOT.md` testing rules): "Use `testify/require` for setup steps that must succeed for the test to be meaningful."

If `NewConnectionPool` ever returns an error here, `assert.NoError` records the failure but execution continues with a `nil` pool, so the very next line (`return pool`) hands back `nil` and the calling test panics with an opaque nil-deref that hides the real cause. `require.NoError(t, err)` aborts the test cleanly at the setup point.

Suggested fix:
```go
import "github.com/stretchr/testify/require"
...
pool, err := NewConnectionPool(&ConnectionParams{...})
require.NoError(t, err)
return pool
```

In practice the current params never error, so this is defensive — but it's a named BUGBOT rule and cheap to honor.

---

makalaaneesh (commented):
[cursor] Automated review — 1 finding (0 critical, 1 warning). See inline comment. (2 suggestions dropped.)

---

makalaaneesh:
live migration tests - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/8176/stages 

live migration resumption tests - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/669/stages/?selected-node=210

makalaaneesh:

priyanshi-yb (approved):
LGTM!

PR #3571, opened 2026-06-01
Fix import-schema failure for PL/pgSQL functions using unqualified %TYPE

### LINKED ISSUE #3522: Referencing %TYPE syntax errors out in import schema
Jira Link: [DB-21195](https://yugabyte.atlassian.net/browse/DB-21195)

sincfaio:

`yb-voyager import schema` fails on PL/pgSQL functions whose body uses unqualified `%TYPE` refs (e.g., `employees.salary%TYPE`). The function imports cleanly via `ysqlsh` — it's a voyager-specific bug.

Root cause: `pg_dump` emits `SELECT pg_catalog.set_config('search_path', '', false);` in every per-object `.sql` preamble. With `check_function_bodies = true` (default) on YugabyteDB, the PL/pgSQL validator runs at `CREATE FUNCTION` time and can't resolve the unqualified ref against an empty path — fails with SQLSTATE 42601. The existing `"does not exist"` defer classifier doesn't match this error wording, so the import bails outright.

Fix is two-pronged:

- **Export-side** (`pg_dump_extract_schema.go`): inject `SET search_path = <source schemas>;` into `function.sql` and `procedure.sql` only, immediately after pg_dump's preamble. The injected SET is a regular SET statement so the existing capture loop picks it into `sessionVariables` and any deferred retry inherits it without extra plumbing. Other per-object files (table.sql, view.sql, etc.) are unchanged.
- **Import-side** (`importSchema.go`, `importSchemaYugabyteDB.go`): add `isPercentTypeResolutionError` peer to `missingRequiredSchemaObject` — matches SQLSTATE 42601 plus `%TYPE`/`%ROWTYPE` in the failing stmt. OR'd at the defer site and the post-loop error-swallow branch. This handles the secondary case where a function refers to a VIEW (`v_sal.salary%TYPE`) that voyager imports *after* functions in its object order — first pass defers, retry after view import succeeds.

No command-line, configuration, or installation changes. The exported `function.sql`/`procedure.sql` files now contain one extra `SET search_path = <schemas>;` line after pg_dump's preamble — internal to the migration flow.

Functions that previously failed import (and required manual fixes via `replacement_dir` in tests) now import cleanly on the first pass.

- **New integration test**: `TestImportSchemaWithPercentType` (`yb-voyager/cmd/importSchema_test.go`, `integration_voyager_command` tag) — uses real PG 17 and YB 2025.2 containers. Covers both Case A (function before table, unqualified `%TYPE`) and Case B (function before view, forward ref). Asserts the export-side injection happens (function.sql contains the SET, table.sql does not), the expected import path is taken (no defer log for Case A; defer-and-retry log for Case B), and both functions return correct values when invoked under the right search_path. Run: `go test -tags integration_voyager_command -run TestImportSchemaWithPercentType ./cmd/`.
- **mgi migtest fixture**: removed the 17 stale `invalid type name "...%TYPE"` expected-failure blocks. Verified by running the real mgi migration locally — produced `failed.sql` is exactly 372 lines (only the 62 ALTER CLUSTER blocks), matching the updated fixture.
- `go vet ./...` clean. `go test -tags unit ./cmd/...` passes (unrelated batch-producer flake aside).
- Audited other PG migtests with `%TYPE` in their source (omnibus, assessment-report-test, multiple-schemas, misc-objects-2) — they either use qualified refs or have `%TYPE` only in comments, so they're unaffected.

No.

No.


sincfaio:
@gemini review

yb-voyager/cmd/importSchema.go
```diff
@@ -583,6 +584,21 @@ func missingRequiredSchemaObject(err error) bool {
...
 }
 
+// Matches SQL block comments (/* ... */, including multi-line) and line comments (-- to end-of-line).
+var sqlCommentRegex = regexp.MustCompile(`(?s)/\*.*?\*/|--[^\n]*`)
+
+// 42601 wording for failed %TYPE / %ROWTYPE resolution varies across PG and
+// YB versions, so we key on the syntactic construct in the failing stmt and
+// strip comments to avoid documentation refs in headers triggering.
+func isPercentTypeResolutionError(err error, stmt string) bool {
+	if !strings.Contains(err.Error(), "(SQLSTATE 42601)") {
```
priyanshi-yb: There is a way you can fetch the error code from pg error and have a check on that directly, instead of a contains check.

yb-voyager/cmd/importSchema.go
```diff
@@ -583,6 +584,21 @@ func missingRequiredSchemaObject(err error) bool {
 	return strings.Contains(err.Error(), "does not exist")
 }
 
+// Matches SQL block comments (/* ... */, including multi-line) and line comments (-- to end-of-line).
+var sqlCommentRegex = regexp.MustCompile(`(?s)/\*.*?\*/|--[^\n]*`)
+
+// 42601 wording for failed %TYPE / %ROWTYPE resolution varies across PG and
+// YB versions, so we key on the syntactic construct in the failing stmt and
+// strip comments to avoid documentation refs in headers triggering.
```
priyanshi-yb: I didn't understand why the error varies in various versions, and even if it changes, we need just to run the defer logic for the error `invalid type name "<anything>%TYPE"`

yb-voyager/src/srcdb/pg_dump_extract_schema.go
```diff
@@ -221,13 +221,26 @@ func parseSchemaFile(exportDir string, schemaDir string, exportObjectTypesList [
...
 			continue
 		}
 		filePath := utils.GetObjectFilePath(schemaDir, objType)
 		log.Infof("creating schema objects file for %q at %q", objType, filePath)
-		dataBytes := []byte(setSessionVariables.String() + sqlStmts.String())
+		prefix := setSessionVariables.String()
+		if (objType == "FUNCTION" || objType == "PROCEDURE") && searchPathOverride != "" {
+			prefix += searchPathOverride
+		}
+		dataBytes := []byte(prefix + sqlStmts.String())
```
priyanshi-yb: Hmm, interesting, I was assuming you would modify the import schema code to set the search path on the connection for these.
I think that might be a better approach as that will cover %TYPE syntax coming from PG/ORACLE both.
There is already code in import schema where we are setting the search path, but I believe since we are running the SQL file's SET search path and the schema one is getting overridden.
```
func newTargetConn() *ImportSchemaTargetConn {
	.....
	conn, err := pgx.ConnectConfig(context.Background(), conf)
	errExit(err)

	setTargetSchema(conn) <------ this function is setting the search path to target schemas

	if sourceDBType == ORACLE && enableOrafce {
		setOrafceSearchPath(conn)
	}

	return &ImportSchemaTargetConn{conn: &conn}
}
```

yb-voyager/cmd/importSchema.go
```diff
@@ -583,6 +584,21 @@ func missingRequiredSchemaObject(err error) bool {
...
+func isPercentTypeResolutionError(err error, stmt string) bool {
+	if !strings.Contains(err.Error(), "(SQLSTATE 42601)") {
+		return false
+	}
+	stripped := sqlCommentRegex.ReplaceAllString(stmt, "")
+	upper := strings.ToUpper(stripped)
+	return strings.Contains(upper, "%TYPE") || strings.Contains(upper, "%ROWTYPE")
```
makalaaneesh: [AI] **`isPercentTypeResolutionError` keys on the generic 42601 SQLSTATE, so genuinely-broken `%TYPE` functions get misrouted through the defer path.**

`42601` is PostgreSQL/YB's generic `syntax_error` class, not a dedicated "could not resolve `%TYPE`" code. This classifier returns `true` for *any* statement that (a) fails with `(SQLSTATE 42601)` and (b) contains the substring `%TYPE`/`%ROWTYPE` anywhere in the comment-stripped text. Two consequences:

1. A function whose body has a real, unrelated syntax error but also happens to use `%TYPE` will be classified as a resolution error and pushed onto `deferredSqlStmts` instead of being reported immediately. The inline `color.Red(err)` print (and, under `--continue-on-error=false`, the immediate `NewExecuteDDLError` return) is suppressed; the failure now only surfaces at the very end after the deferred-retry loop exhausts. That is a diagnostic regression for a genuinely-broken statement.
2. Comment stripping (`sqlCommentRegex`) removes `--`/`/* */` comments but not dollar-quoted string literals, so a function body containing the literal text `%TYPE` inside a quoted string would also trip the heuristic and trigger a spurious deferral.

Neither case is data loss (failures still land in `finalFailedSqlStmts`), so this is a Warning, not a blocker. But the false-positive surface is wider than the comment claims.

Suggested tightening: in addition to the SQLSTATE + `%TYPE` check, require the error message to reference the type-resolution wording (the PG/YB variants of `invalid type name "...%TYPE"` / `"...%ROWTYPE"`), e.g. match on `%TYPE`/`%ROWTYPE` appearing in `err.Error()` itself rather than only in the statement text. That keeps the version-agnostic intent while excluding unrelated 42601 syntax errors.

makalaaneesh (commented):

yb-voyager/cmd/importSchema_test.go
```diff
@@ -129,3 +130,127 @@ func TestImportSchemaWithYBSpecificSyntax(t *testing.T) {
...
+	importLog, err := os.ReadFile(filepath.Join(tempExportDir, "logs", "yb-voyager-import-schema.log"))
+	if err != nil {
+		t.Fatalf("read import-schema log: %v", err)
+	}
+	if strings.Contains(string(importLog), "deffering execution of SQL: CREATE FUNCTION pct.get_employee_salary") {
+		t.Fatalf("get_employee_salary was deferred — search_path override skip didn't take effect")
+	}
+	if !strings.Contains(string(importLog), "deffering execution of SQL: CREATE FUNCTION pct.get_view_salary") {
+		t.Fatalf("get_view_salary was NOT deferred — isPercentTypeResolutionError didn't catch the 42601+%%TYPE forward-ref case")
+	}
```
priyanshi-yb: I feel this is a bit overkill to check the log, as logs can change at any time. We should just verify if the statement is eventually getting created in YB or not.

priyanshi-yb (approved):
LGTM with a comment

PR #3572, opened 2026-06-02
Fix duplicate error output for failures before logging initialization

AmeyaKasture:

  Errors raised during early command bootstrap — before `InitLogging()` runs —
  were printed to the console twice. `utils.ErrExit` both writes the message to
  stderr via `fmt.Fprintf` and logs it via `log.Errorf`. Once logging is
  initialized those go to two different destinations (console + log file), but
  before initialization logrus is still on its default stderr output, so the
  message lands on the terminal twice (the second copy carrying logrus's default
  `ERRO[0000]` prefix).

  This affects the pre-init validation/bootstrap window for every command, e.g.
  the `export-dir doesn't exist` error. The fix follows the convention already
  used elsewhere in this window (`initConfig`, `ValidateLogLevel`): pre-init
  failures print to the console directly and call `atexit.Exit(1)` instead of
  `utils.ErrExit`. Converted spots: `validateExportDirFlag`,
  `validateBulkAssessmentDirFlag`, and the three `InitLogging`-failure exits in
  `root.go`. `ErrExit` itself and all post-init call sites are unchanged, so
  behavior elsewhere is untouched.

  Yes — minor console output only. Early-bootstrap errors (bad `--export-dir` /
  `--bulk-assessment-dir`, logging init failure) now print once instead of twice.
  No changes to command line flags, configuration, installation, or reports.

  `go build ./cmd/...` and `go vet ./cmd/` pass clean. Manually reasoned through
  the pre-init code path. No new automated tests added — the change only affects
  console-output duplication during early exit, which existing tests don't assert
  on. Recommend a quick manual check: run any command with a non-existent
  `--export-dir` and confirm the error appears only once.

  No.

  No.

### DIFF
--- a/yb-voyager/cmd/assessMigrationBulkCommand.go
+++ b/yb-voyager/cmd/assessMigrationBulkCommand.go
@@ -31,6 +31,7 @@ import (
 	goerrors "github.com/go-errors/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/tebeka/atexit"
 
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/callhome"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/config"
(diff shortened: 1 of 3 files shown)


yb-voyager/cmd/assessMigrationBulkCommand.go
```diff
@@ -455,19 +456,25 @@ func isMigrationAssessmentDoneForConfig(dbConfig AssessMigrationDBConfig) bool {
 }
 
 func validateBulkAssessmentDirFlag() {
+	// This runs before InitLogging(), so logging is not yet redirected to the log
+	// file. Using utils.ErrExit here would print the message twice (once to stderr
+	// and again via logrus's default stderr output), so print to console directly.
 	if bulkAssessmentDir == "" {
-		utils.ErrExit(`ERROR required flag "bulk-assessment-dir" not set`)
+		fmt.Fprintf(os.Stderr, "ERROR required flag \"bulk-assessment-dir\" not set\n")
```
makalaaneesh: how about a helper func similar to ErrExit? `ErrExitBeforeLoggerInitialized` (or something more shorter)

makalaaneesh:

priyanshi-yb (approved):
LGTM

PR #3573, opened 2026-06-03
Fix pg/datatypes callhome compare-performance test baseline

Abhi-yb:

The `pg/datatypes` callhome test was failing at the `compare-performance` step due to query count mismatches. Instead of hardcoding exact counts in the baseline JSON, this PR normalizes `source_only_queries`, `target_only_queries`, and `total_queries` to `"GT_ZERO"` in `normalize_callhome_json` when their value is > 0. This makes the test resilient to count changes from new voyager queries or YugabyteDB version bumps, so we don't need to keep updating the baseline on every code change.

No user-facing changes.

Ran `bash migtests/scripts/callhome/run-callhome-test.sh pg/datatypes` locally end-to-end. All phases passed including compare-performance.

No payload shape changes.

No.

### DIFF
--- a/migtests/scripts/functions.sh
+++ b/migtests/scripts/functions.sh
@@ -1526,7 +1526,10 @@ normalize_callhome_json() {
             .db_system_identifier = "IGNORED" |
             .db_id = "IGNORED" |
             .target_db_details = "IGNORED" |
-            .total_db_size_bytes = "IGNORED"
+            .total_db_size_bytes = "IGNORED" |
+            .source_only_queries? |= (if type == "number" and . > 0 then "GT_ZERO" else . end) |
+            .target_only_queries? |= (if type == "number" and . > 0 then "GT_ZERO" else . end) |
+            .total_queries? |= (if type == "number" and . > 0 then "GT_ZERO" else . end)
         elif type == "array" then
 			sort_by(tostring)
         elif type == "string" and (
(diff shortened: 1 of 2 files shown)


makalaaneesh:
>due to a new voyager query added to the source collection in a recent code change.
@Abhi-yb  what was the query exactly ? 

@priyanshi-yb i remember you had a better suggestion for dealing with such changes because they can be very frequent, and we shouldn't necessarily have to keep coming and updating this test so frequently.

priyanshi-yb:
>@priyanshi-yb i remember you had a better suggestion for dealing with such changes because they can be very frequent, and we shouldn't necessarily have to keep coming and updating this test so frequently.

I was suggesting not to validate the exact number of queries altogether, and just do a validation whether it is greater than 0 or not, since the change in queries is frequent with any code change or the DB-related change (auto-analyse), and easy to miss, as this is just a call-home test.

makalaaneesh:

Abhi-yb:
> > @priyanshi-yb i remember you had a better suggestion for dealing with such changes because they can be very frequent, and we shouldn't necessarily have to keep coming and updating this test so frequently.
> 
> I was suggesting not to validate the exact number of queries altogether, and just do a validation whether it is greater than 0 or not, since the change in queries is frequent with any code change or the DB-related change (auto-analyse), and easy to miss, as this is just a call-home test.

>> Will change accordingly as suggested

makalaaneesh (approved):

PR #3586, opened 2026-06-04
Detect pg_stat_statements availability even when guardrails checks are disabled

AmeyaKasture:
When migration assessment is run with `--run-guardrails-checks=false`, the
permission-check step was skipped entirely and returned a `nil`
pg_stat_statements availability map. The downstream gather step interpreted this
as "`pg_stat_statements` disabled" on every node, so Unsupported Query Constructs
detection was silently skipped even when the extension was fully installed and
enabled.
This PR decouples `pg_stat_statements` availability detection from the heavier
guardrail permission validations:
- Adds `PostgreSQL.IsPgStatStatementsAvailable()` in `srcdb`, a lightweight check
  that reuses the existing `checkPgStatStatementsSetup()` logic but only reports
  whether pgss is usable (without the detailed guardrail reason).
- Adds `DetectPgssAvailabilityOnAllNodes()` in `migassessment`, which probes the
  primary and each validated replica (via short-lived connections) and returns a
  per-node availability map. It is PostgreSQL-specific and treats detection
  failures as non-fatal (recording the node as pgss-unavailable), since the user
  has explicitly opted out of guardrail checks and should not be blocked.
- Wires this into `checkAssessmentPermissions()` so that when guardrails are
  disabled, pgss availability is still detected instead of defaulting to disabled.
No new commands, flags, configuration, or installation changes. Behavioral change
only: when running assessment with guardrails checks disabled on a PostgreSQL
source where `pg_stat_statements` is enabled, Unsupported Query Constructs are now
detected and reported (previously they were silently omitted).
Verified locally against a PostgreSQL source with two databases — one with the
`pg_stat_statements` extension installed and loaded (via `shared_preload_libraries`),
and one without it:
- `PostgreSQL.IsPgStatStatementsAvailable()` returned `true` for the database with
  the extension and `false` for the one without it.
- `DetectPgssAvailabilityOnAllNodes()` correctly populated the per-node map,
  reporting `true`/`false` for the `primary` node matching the database under test.
No.
No.


yb-voyager/src/migassessment/permissions.go
```diff
@@ -57,6 +58,76 @@ func CheckAssessmentPermissionsOnAllNodes(source *srcdb.Source, validatedReplica
...
+		pgssByNode[nodeKey] = replicaPgss
+	}
+
+	return pgssByNode, nil
+}
+
+// detectPgssOnReplicaNode opens a short-lived connection to a replica and reports
+// whether pg_stat_statements is available on it.
+func detectPgssOnReplicaNode(source *srcdb.Source, replica srcdb.ReplicaEndpoint) (bool, error) {
+	replicaSource := srcdb.Source{
```
makalaaneesh, line 139: [AI] **Duplicated replica `srcdb.Source` construction — extract a shared helper.**

The `replicaSource := srcdb.Source{...}` block here (lines 106-120) is a verbatim copy of the field-by-field construction already in `checkPermissionsOnReplicaNode` (lines 215-229), including the `replicaSource.DB().(*srcdb.PostgreSQL)` assertion and `Connect()`/`defer Disconnect()` dance.

This violates the root `.cursor/BUGBOT.md` rule: *"Consolidate duplicate logic into shared helpers rather than copy-pasting."* It is also a correctness hazard: the replica `Source` deliberately carries the SSL/auth fields (`SSLMode`, `SSLCertPath`, `SSLKey`, `SSLRootCert`, `SSLCRL`, `User`, `Password`, `NumConnections`). If any of those connection fields are added or changed in the future, a maintainer must remember to update both copies; missing one would silently make replica pgss-detection connect with the wrong credentials/TLS settings.

Suggested fix: extract a small helper, e.g. `buildReplicaSource(source *srcdb.Source, replica srcdb.ReplicaEndpoint) srcdb.Source`, and call it from both `detectPgssOnReplicaNode` and `checkPermissionsOnReplicaNode`.

yb-voyager/src/migassessment/permissions.go
```diff
@@ -57,6 +58,76 @@ func CheckAssessmentPermissionsOnAllNodes(source *srcdb.Source, validatedReplica
...
+// DetectPgssAvailabilityOnAllNodes determines pg_stat_statements availability on the
+// primary and each replica node WITHOUT running the full guardrails permission checks.
+//
+// This is used when guardrails checks are disabled (--run-guardrails-checks=false), so
+// that query-level metadata collection still behaves correctly instead of defaulting to
+// "pgss disabled" and falsely reporting that pg_stat_statements is not enabled.
+//
+// Detection failures are treated as non-fatal (the node is recorded as pgss-unavailable),
+// since the user has explicitly opted out of guardrail checks and should not be blocked.
+func DetectPgssAvailabilityOnAllNodes(source *srcdb.Source, validatedReplicas []srcdb.ReplicaEndpoint) (map[string]bool, error) {
```
makalaaneesh, line 73: [AI] **New pgss-availability detection path has no tests.**

`DetectPgssAvailabilityOnAllNodes` / `detectPgssOnReplicaNode` (and the underlying `PostgreSQL.IsPgStatStatementsAvailable`) introduce the behavior that, when `--run-guardrails-checks=false`, pgss availability is still probed so Unsupported Query Constructs detection is not silently skipped. No unit or integration test accompanies this change.

This path is correctness-sensitive and fails silently: the entire bug being fixed is that a wrong `false` here makes the gather step report "pgss disabled" and skip a whole detection feature with no error. A regression would be invisible without a test. Per `yb-voyager/src/migassessment/.cursor/BUGBOT.md` (Testing) and the repo Go testing rules, new logic should have coverage.

Suggested fix: add a testcontainers-backed integration test (under the `integration` tag) asserting that with guardrails disabled, `pgssByNode["primary"]` is `true` when `pg_stat_statements` is installed/loaded and `false` when it is not — covering both the extension-present and extension-absent cases.

makalaaneesh (commented):

yb-voyager/src/migassessment/permissions.go
AmeyaKasture: The helper doesnt really help much, as most of the code is gone in initialising , while if we created a helper we have to import the source db as  a parameter.
As the code is used only twice and both in completely different execution paths, the helper is not really required

ShivanshGahlot (commented):
I still feel since pgss is more like a capability probe rather than a permission check, we should extract it completely out of the guardrails check path. Basically, we can use the `DetectPgssAvailabilityOnAllNode` function you have created here always and remove the older code path and the return types associated with it. This function will be called regardless of whether guardrails are enabled or disabled. What do you think? cc @makalaaneesh 

Also in this case too I think it makes sense to anyways extract out the srcdb.Source{...} block into a helper.

Also can you share the screenshots of how the UI is looking currently when guardrails are enabled and pgss is not available in a single node and multi node case. And similarly when guardrails is disabled and pgss is not available.

yb-voyager/src/migassessment/permissions.go
```diff
@@ -35,36 +36,132 @@ type NodePermissionResult struct {
...
+			} else {
+				utils.PrintAndLogfWarning("\n⚠ pg_stat_statements not available (query-level analysis will be limited)")
+			}
+		}
+
+		// If some nodes have pg_stat_statements and some don't, inform the user.
+		if len(nodesWithoutPgss) < len(pgssByNode) {
+			utils.PrintAndLogfInfo("\nNote: Query-level analysis (Unsupported Query Constructs) will only include data from nodes with pg_stat_statements.")
+		}
+	}
```
ShivanshGahlot, line 131: Nit:

```
if len(nodesWithoutPgss) > 0 {
    if len(pgssByNode) == 1 {
        utils.PrintAndLogfWarning("\n⚠ pg_stat_statements not available (query-level analysis will be limited)")
    } else {
        utils.PrintAndLogfWarning("\n⚠ pg_stat_statements not available on %s (query-level analysis will be limited)",
            strings.Join(nodesWithoutPgss, ", "))
        if len(nodesWithoutPgss) < len(pgssByNode) {
            utils.PrintAndLogfInfo("\nNote: Query-level analysis (Unsupported Query Constructs) will only include data from nodes with pg_stat_statements.")
        }
    }
}
```

yb-voyager/src/migassessment/permissions.go
```diff
@@ -35,36 +36,132 @@ type NodePermissionResult struct {
...
+		Schemas:        source.Schemas,
+		SSLMode:        source.SSLMode,
+		SSLCertPath:    source.SSLCertPath,
+		SSLKey:         source.SSLKey,
+		SSLRootCert:    source.SSLRootCert,
+		SSLCRL:         source.SSLCRL,
+		NumConnections: source.NumConnections,
+	}
+
+	replicaDB := replicaSource.DB().(*srcdb.PostgreSQL)
```
ShivanshGahlot: Nit: Always better to use the comma-ok idiom as the gemini comment says.

yb-voyager/src/migassessment/permissions.go
```diff
@@ -35,36 +36,132 @@ type NodePermissionResult struct {
...
+
+	primaryPgss, err := pg.IsPgStatStatementsAvailable()
+	if err != nil {
+		log.Warnf("failed to detect pg_stat_statements availability on primary: %v", err)
+		primaryPgss = false
+	}
+	pgssByNode["primary"] = primaryPgss
+	if !primaryPgss {
+		nodesWithoutPgss = append(nodesWithoutPgss, "primary")
+	}
```
ShivanshGahlot, line 95: Currently we are swallowing any connection errors and they are indistinguishable from genuine pgss missing.

We could do this instead:
```
primaryPgss, err := pg.IsPgStatStatementsAvailable()
switch {
case err != nil:
    utils.PrintAndLogfWarning("\n⚠ could not verify pg_stat_statements on primary: %v (query-level analysis will be limited)", err)
case !primaryPgss:
    nodesWithoutPgss = append(nodesWithoutPgss, "primary")
}
pgssByNode["primary"] = primaryPgss
```

yb-voyager/src/migassessment/permissions.go
```diff
@@ -35,36 +36,132 @@ type NodePermissionResult struct {
...
+		nodeKey := fmt.Sprintf("%s:%d", replica.Host, replica.Port)
+		replicaPgss, err := detectPgssOnReplicaNode(source, replica)
+		if err != nil {
+			log.Warnf("failed to detect pg_stat_statements availability on replica %s: %v", nodeKey, err)
+			replicaPgss = false
+		}
+		pgssByNode[nodeKey] = replicaPgss
+		if !replicaPgss {
+			nodesWithoutPgss = append(nodesWithoutPgss, nodeKey)
+		}
```
ShivanshGahlot, line 107: Similar to the primary case without swallowing any connection errors:

```
replicaPgss, err := detectPgssOnReplicaNode(source, replica)
switch {
case err != nil:
    utils.PrintAndLogfWarning("\n⚠ could not verify pg_stat_statements on %s: %v (query-level analysis will be limited)", nodeKey, err)
case !replicaPgss:
    nodesWithoutPgss = append(nodesWithoutPgss, nodeKey)
}
pgssByNode[nodeKey] = replicaPgss
```

yb-voyager/src/migassessment/permissions.go
```diff
@@ -35,36 +36,132 @@ type NodePermissionResult struct {
...
+			}
+		}
+
+		// If some nodes have pg_stat_statements and some don't, inform the user.
+		if len(nodesWithoutPgss) < len(pgssByNode) {
+			utils.PrintAndLogfInfo("\nNote: Query-level analysis (Unsupported Query Constructs) will only include data from nodes with pg_stat_statements.")
+		}
+	}
+
+	return pgssByNode, nil
```
ShivanshGahlot, line 133: Did we remove the prompt entirely from here which asked the users to confirm if they want to continue in case pgss is missing?

ShivanshGahlot (commented):
Few minor comments

yb-voyager/src/migassessment/permissions.go
```diff
@@ -35,36 +35,148 @@ type NodePermissionResult struct {
...
+		for _, node := range nodesWithoutPgss {
+			if hasMultipleNodes {
+				utils.PrintAndLogfWarning("\n⚠ pg_stat_statements not available on %s (query-level analysis will be limited)", node)
+			} else {
+				utils.PrintAndLogfWarning("\n⚠ pg_stat_statements not available (query-level analysis will be limited)")
+			}
+		}
+
+		// If some nodes have pg_stat_statements and some don't, inform the user.
+		if len(nodesWithoutPgss) > 0 {
```
ShivanshGahlot: Seems like this is old code/duplicate? We might be printing twice because of this. We probably just need the code from line 124.

ShivanshGahlot (approved):

PR #3588, opened 2026-06-04
Improve error handling UX

AmeyaKasture:

yb-voyager surfaces all fatal errors through `utils.ErrExit`, which previously wrote the error message to the console as plain, unstyled white text. In commands with structured UX — most visibly `assess-migration` — this produced an inconsistent experience: a failed step would print a red ✗ marker (e.g. ✗ Schema resolution), immediately followed by the actual error message in plain white, visually disconnected from the failure it described.

This PR fixes the problem globally at the single point where errors are rendered. `ErrExit` now prints the console (stderr) message in red via the existing `ErrorColor`, while the log-file line is kept as plain text (no ANSI escape codes). Because the fix lives in `ErrExit` itself, every command's fatal error — including the entire `assess-migration` preflight and pipeline flow — is now consistently styled without per-call-site changes. `fatih/color` automatically disables escape codes for non-TTY output and when `NO_COLOR` is set, so piped output and CI logs degrade gracefully to plain text.

Yes, a cosmetic console change: fatal error messages printed by yb-voyager now appear in red on the terminal (when stdout is a TTY and `NO_COLOR` is not set), matching the existing red ✗ failure markers. There are no changes to command-line flags, configuration, installation, or report contents. Log files are unchanged and remain plain text. When output is redirected/piped or `NO_COLOR` is set, error text is plain as before.

- `go build ./...` for the affected packages passes.
- Verified the exact console line `ErrExit` now emits with a throwaway demo, inspecting raw bytes with `cat -v`: the console (stderr) line is wrapped in red ANSI codes (`\033[31m … \033[0m`) while the corresponding log line contains no escape codes — confirming console styling without log pollution.
- No existing tests assert the old plain `ErrExit` stderr format (tests monkey-patch `ErrExit` via `MonkeyPatchUtilsErrExit`, so the real implementation isn't exercised), so no test changes were required.
- A broader manual run of `assess-migration` against a failing preflight (e.g. invalid `--source-db-schema`) is recommended to eyeball the end-to-end styling in a real terminal.

No.

No. The change only affects how error messages are rendered to the console; no on-disk structures, serialized state, or log formats are modified.

### DIFF
--- a/yb-voyager/go.mod
+++ b/yb-voyager/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/aws/aws-sdk-go-v2 v1.17.5
 	github.com/aws/aws-sdk-go-v2/config v1.18.15
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.30.5
+	github.com/creack/pty v1.1.18
 	github.com/davecgh/go-spew v1.1.1
 	github.com/deckarep/golang-set/v2 v2.7.0
 	github.com/docker/go-connections v0.5.0
@@ -163,7 +164,7 @@ require (
 	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/magiconair/properties v1.8.7
 	github.com/mattn/go-colorable v0.1.13 // indirect
-	github.com/mattn/go-isatty v0.0.20 // indirect
+	github.com/mattn/go-isatty v0.0.20
 	github.com/mattn/go-runewidth v0.0.14 // indirect
 	github.com/mitchellh/mapstructure v1.5.0 // indirect
 	github.com/pavlo-v-chernykh/keystore-go/v4 v4.4.1
--- a/yb-voyager/go.mod
+++ b/yb-voyager/go.mod
@@ -10,7 +10,6 @@ require (
 	github.com/aws/aws-sdk-go-v2 v1.17.5
 	github.com/aws/aws-sdk-go-v2/config v1.18.15
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.30.5
-	github.com/creack/pty v1.1.18
 	github.com/davecgh/go-spew v1.1.1
 	github.com/deckarep/golang-set/v2 v2.7.0
 	github.com/docker/go-connections v0.5.0
--- a/yb-voyager/src/utils/logging.go
+++ b/yb-voyager/src/utils/logging.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/fatih/color"
 	goerrors "github.com/go-errors/errors"
+	"github.com/mattn/go-isatty"
 	log "github.com/sirupsen/logrus"
 	"github.com/tebeka/atexit"
 )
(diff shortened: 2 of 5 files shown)


makalaaneesh:

yb-voyager/src/utils/logging.go
```diff
@@ -33,8 +33,14 @@ var ErrExitErr error
 var ErrExit = func(formatString string, args ...interface{}) {
 	ErrExitErr = goerrors.Errorf(formatString, args...)
 	formatString = strings.Replace(formatString, "%w", "%s", -1)
-	fmt.Fprintf(os.Stderr, formatString+"\n", args...)
-	log.Errorf(formatString+"\n", args...)
+	message := fmt.Sprintf(formatString, args...)
```
makalaaneesh, line 50: previous code had `+"\n"`. 

I see that we're using `Fprintln` for stderr, but there would be no "\n" for the `log.Error()` ? I guess logger will anyway add a `\n`; is that why it's not necessary?

yb-voyager/src/utils/logging.go
```diff
@@ -33,8 +33,14 @@ var ErrExitErr error
...
+	// Console: render the error in red so failures stand out and visually match
+	// the red "✗" markers printed by the progress/preflight UX. ErrorColor
+	// auto-disables escape codes for non-TTY output / NO_COLOR, so piped output
+	// degrades gracefully to plain text.
+	ErrorColor.Fprintln(os.Stderr, message)
+	// Log file: keep plain text only (never embed ANSI escape codes in logs).
+	log.Error(message)
```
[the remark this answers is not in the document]
makalaaneesh: @AmeyaKasture this seems relevant; can you check?

yb-voyager/src/utils/logging.go
AmeyaKasture: Yes Exactly

yb-voyager/src/utils/logging.go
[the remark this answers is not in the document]
AmeyaKasture: Implemented the fix

yb-voyager/src/utils/logging.go
```diff
@@ -30,11 +31,32 @@ var originalErrExit func(formatString string, args ...interface{})
...
 var ErrExitErr error
 
+// errorColor renders fatal errors in red on stderr. fatih/color's global
+// NoColor is derived from stdout, so it cannot decide coloring for stderr on its
+// own: when stdout is a TTY but stderr is redirected (e.g. `yb-voyager ... 2>
+// err.log`), it would still emit ANSI escape codes into the file. We therefore
+// disable color when stderr itself is not a TTY. When stderr is a TTY the
+// object's per-instance flag stays unset, so it falls back to the global NoColor
+// and NO_COLOR / TERM=dumb are still honored.
+var errorColor = func() *color.Color {
```
makalaaneesh, line 39: nit: func() ... () seems a bit unnecessary. 
the idiomatic way to do this in go would be to do this in `init`?

yb-voyager/go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/aws/aws-sdk-go-v2 v1.17.5
 	github.com/aws/aws-sdk-go-v2/config v1.18.15
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.30.5
+	github.com/creack/pty v1.1.18
```
makalaaneesh: run go mod tidy? I don't see a direct dependency use

yb-voyager/src/utils/logging.go
```diff
@@ -30,11 +31,32 @@ var originalErrExit func(formatString string, args ...interface{})
 
 var ErrExitErr error
 
+// errorColor renders fatal errors in red on stderr. fatih/color's global
+// NoColor is derived from stdout, so it cannot decide coloring for stderr on its
+// own: when stdout is a TTY but stderr is redirected (e.g. `yb-voyager ... 2>
+// err.log`), it would still emit ANSI escape codes into the file. We therefore
+// disable color when stderr itself is not a TTY. When stderr is a TTY the
```
makalaaneesh: nit: remove the last bit of comment; it's confusing

makalaaneesh (approved):
LGTM

yb-voyager/src/utils/logging.go
AmeyaKasture: This immediate function is intentional: it keeps errorColor fully initialized at declaration time and keeps the stderr-specific color setup colocated with the variable. 
An init() would work too, but it would split the declaration from the initialization and add additional package-level setup for a single computed value; a small constructor/IIFE is idiomatic for this kind of local initialization in Go

PR #3600, opened 2026-06-10
Do not detect unique key conflicts for the tables have cdc partitioning strategy as table

priyanshi-yb:

During live migration, the conflict detection cache holds in-flight events and checks incoming events against cached ones that share unique-key values, in order to preserve correctness when events for the same table are processed out-of-order across multiple event channels.

When the CDC partitioning strategy for a table is `table` (`PARTITION_BY_TABLE`), all events for that table are hashed solely by table name and therefore routed to the same event channel, where they are applied serially in order. In that case the out-of-order scenario that conflict detection guards against cannot happen, so the conflict check is redundant.

This PR threads the `tableToPartitioningStrategyMap` into the `ConflictDetectionCache` and short-circuits `eventsConfict` to return `false` for tables whose strategy is `PARTITION_BY_TABLE`, avoiding unnecessary conflict detection and the associated cache stalls for those tables.

No user-facing changes. The behavior change only affects internal live-migration event processing for tables using the (hidden) `--cdc-partitioning-strategy table` option; no command line, configuration, installation, or report changes.

- Existing unit tests in `live_migration_test.go` were updated to pass the new `tableToPartitioningStrategyMap` argument (passing `nil`, which preserves the prior conflict-detection behavior).
- Added a new integration test `TestLiveMigrationWithUniqueKeyConflictWithTablePartitioning` in `live_migration_integration_test.go` that runs a live migration with `--cdc-partitioning-strategy=table` over a unique-key workload (1500 inserts, 2500 updates, 1000 deletes), then validates snapshot and post-streaming data consistency through cutover. The shared setup was refactored into a `getPartialPredicateTestForUniqueConflictDetection` helper reused by the existing partial-predicate test.

No.

No.

### DIFF
--- a/yb-voyager/cmd/conflictDetectionCache.go
+++ b/yb-voyager/cmd/conflictDetectionCache.go
@@ -122,20 +122,23 @@ type ConflictDetectionCache struct {
 		Worst event size can be 7kb for 30-50 columns in the table
 		so for the 500000 events (100 channels * 500 events per channel) at worst in the cache it will be 500000 * 7kb = 3.5GB
 	*/
-	m                       map[int64]*tgtdb.Event
-	cond                    *sync.Cond
-	tableToUniqueKeyColumns *utils.StructMap[sqlname.NameTuple, []string]
-	evChans                 []chan *tgtdb.Event
-	sourceDBType            string
+	m                              map[int64]*tgtdb.Event
+	cond                           *sync.Cond
+	tableToUniqueKeyColumns        *utils.StructMap[sqlname.NameTuple, []string]
+	evChans                        []chan *tgtdb.Event
+	sourceDBType                   string
+	tableToPartitioningStrategyMap *utils.StructMap[sqlname.NameTuple, string]
 }
 
-func NewConflictDetectionCache(tableToUniqueKeyColumns *utils.StructMap[sqlname.NameTuple, []string], evChans []chan *tgtdb.Event, sourceDBType string) *ConflictDetectionCache {
+func NewConflictDetectionCache(tableToUniqueKeyColumns *utils.StructMap[sqlname.NameTuple, []string], evChans []chan *tgtdb.Event, sourceDBType string,
+	tableToPartitioningStrategyMap *utils.StructMap[sqlname.NameTuple, string]) *ConflictDetectionCache {
 	c := &ConflictDetectionCache{}
 	c.m = make(map[int64]*tgtdb.Event)
 	c.cond = sync.NewCond(&c.Mutex)
 	c.tableToUniqueKeyColumns = tableToUniqueKeyColumns
 	c.sourceDBType = sourceDBType
 	c.evChans = evChans
+	c.tableToPartitioningStrategyMap = tableToPartitioningStrategyMap
 	return c
 }
 
@@ -313,6 +316,13 @@ func (c *ConflictDetectionCache) eventsConfict(cachedEvent *tgtdb.Event, incomin
 	}
 
 	if c.sameTableEventsHaveSamePK(cachedEvent, incomingEvent) {
+		return false
+	}
+
+	strategy, ok := c.tableToPartitioningStrategyMap.Get(cachedEvent.TableNameTup)
+	if ok && strategy == PARTITION_BY_TABLE {
+		// if the table is partitioned by table, then we don't need to check for conflicts
+		// because all the events for a table will be processed in the same channel
 		return false
 	}
 
--- a/yb-voyager/cmd/conflictDetectionCache.go
+++ b/yb-voyager/cmd/conflictDetectionCache.go
@@ -319,11 +319,13 @@ func (c *ConflictDetectionCache) eventsConfict(cachedEvent *tgtdb.Event, incomin
 		return false
 	}
 
-	strategy, ok := c.tableToPartitioningStrategyMap.Get(cachedEvent.TableNameTup)
-	if ok && strategy == PARTITION_BY_TABLE {
-		// if the table is partitioned by table, then we don't need to check for conflicts
-		// because all the events for a table will be processed in the same channel
-		return false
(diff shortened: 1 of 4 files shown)


yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -316,6 +319,15 @@ func (c *ConflictDetectionCache) eventsConfict(cachedEvent *tgtdb.Event, incomin
 		return false
 	}
 
+	if c.tableToPartitioningStrategyMap != nil {
```
makalaaneesh: I feel like this is too late a check. 

If an event is partitioned by table, we should NOT
- add it to conflict detection cache 
- check if it conflicts with any other event. 

It seems to me that "partitioning strategy" is a concern outside the conflict detection cache altogether. It's the concern of the live migration pipeline that orchestrates it (i.e the component that hashes, checks for conflicts, puts it into a channel, converts, etc etc) 

Basically, what I'm proposing is: 
Avoid all calls to the conflict detection cache in `live_migration.go` in case the event.table's partitioning strategy is table. 

Does that make sense?

priyanshi-yb: yeah makes sense

yb-voyager/src/testlivemigration/live_migration_integration_test.go
```diff
@@ -1237,6 +1241,55 @@ FROM generate_series(1, 20) as i;`,
 
 }
 
+func TestLiveMigrationWithUniqueKeyConflictWithTablePartitioning(t *testing.T) {
```
makalaaneesh, line 1244: [AI] **Parallel test reuses database name `test7`, colliding with the existing test on shared containers.**

`TestLiveMigrationWithUniqueKeyConflictWithTablePartitioning` calls `t.Parallel()` and builds its fixture via `getPartialPredicateTestForUniqueConflictDetection`, which hardcodes `DatabaseName: "test7"` for both source and target — the same DB name used by `TestLiveMigrationWithUniqueKeyValuesWithPartialPredicateConflictDetectionCases`, which is also `t.Parallel()`.

`NewTestContainer` keys its container registry only by `dbType-[live-]version` (see `buildContainerName` in `yb-voyager/test/containers/testcontainers.go`), so both tests share the same source/target container instances. `CreateDatabase` is idempotent (returns early if the DB already exists), so both tests end up operating on the *same physical* `test7` database.

Running concurrently, the two tests will execute schema setup, initial-data inserts, the source-delta loop, and `DropDatabase("test7")` cleanup against the same database — corrupting each other's data and producing flaky failures. Every other `t.Parallel()` test in this file deliberately uses a unique DB name (test1, test2, …, test8, test10) for exactly this reason; the only one reusing a name (`TestLiveMigrationResumptionWithChangeInCDCPartitioningStrategy`, test6) is notably *not* parallel.

Suggested fix: give this test its own database name (e.g. `test_table_partitioning_conflict`) — either by parameterizing `getPartialPredicateTestForUniqueConflictDetection` to take a DB name, or by overriding `lm.config` DB names after constructing the fixture.

makalaaneesh (commented):
[AI] Automated review — 1 finding (0 critical, 1 warning). See inline comment. Note: the production change (skipping conflict detection for `PARTITION_BY_TABLE`) is correct — with the table strategy all events for a table hash to a single channel and are applied serially, so the conflict cache is unnecessary.

yb-voyager/cmd/live_migration.go
```diff
@@ -425,8 +438,7 @@ func handleEvent(event *tgtdb.Event,
 		Checking for all possible conflicts among events
 		For more details about ConflictDetectionCache see the related comment in [conflictDetectionCache.go](../conflictDetectionCache.go)
 	*/
-	uniqueKeyCols, _ := conflictDetectionCache.tableToUniqueKeyColumns.Get(event.TableNameTup)
-	if len(uniqueKeyCols) > 0 {
+	if shouldDetectConflicts(event, conflictDetectionCache.tableToUniqueKeyColumns, tableToPartitioningStrategyMap) {
```
makalaaneesh: this block is not just about detecting conflicts but also about adding events to the conflict detection cache, right? 

maybe a better name here? `shouldHandleConflicts` ?

yb-voyager/cmd/live_migration.go
```diff
@@ -402,6 +402,19 @@ func shouldFormatValues(event *tgtdb.Event) bool {
 	}
 	return false
 }
+
+func shouldDetectConflicts(event *tgtdb.Event, tableToUniqueKeyColumns *utils.StructMap[sqlname.NameTuple, []string], tableToPartitioningStrategyMap *utils.StructMap[sqlname.NameTuple, string]) bool {
+	if tableToUniqueKeyColumns == nil || tableToPartitioningStrategyMap == nil {
```
makalaaneesh: I think we might have to have stricter guardrails here in this func. 

When is it possible that: 
- tableToUniqueKeyColumns = nil
- tableToPartitioningStrategyMap == nil
- table is not present in either of the maps. 

If any of this is not possible/not expected, we should error out by doing checks first in this func

yb-voyager/cmd/live_migration.go
```diff
@@ -402,6 +402,19 @@ func shouldFormatValues(event *tgtdb.Event) bool {
...
+		return false
+	}
+	uniqueKeyCols, _ := tableToUniqueKeyColumns.Get(event.TableNameTup)
+	partitioningStrategy, _ := tableToPartitioningStrategyMap.Get(event.TableNameTup)
+	if len(uniqueKeyCols) == 0 || partitioningStrategy == PARTITION_BY_TABLE {
```
makalaaneesh: comments here pls to explain why `PARTITION_BY_TABLE` does not need conflict handling

makalaaneesh (approved):
LGTM

PR #3603, opened 2026-06-11
Add guardrail for start-clean import after queue segments are archived

AmeyaKasture:

For change-streaming imports (live migration), `import data --start-clean` resets the per-importer `imported_by` flags and re-streams the change-event queue from the earliest segment. If `archive changes` has already archived/deleted that earliest segment from disk, the queue can no longer be re-streamed from the beginning — leaving the migration in a broken, partially-reset state.

This PR adds a guardrail in `clearMigrationStateForImportDataStartClean` that detects this condition **before** mutating any metaDB state and fails fast with a clear error message. A new `MetaDB.IsResumeSegmentDeleted` helper checks whether the earliest queue segment (lowest `segment_no` for the relevant exporter role) has been marked deleted in the queue-segment metadata table. For fall-back imports (`SOURCE_DB_IMPORTER_ROLE`), the check is scoped to segments exported by the target DB (`TARGET_DB_EXPORTER_FB_ROLE`), since that is the only queue a fall-back import consumes.

The metaDB-mutating steps (clearing `OnPrimaryKeyConflictAction` and `TableToCDCPartitioningStrategyMap`) were moved out of the `else` branch to after the guardrail, so they only run once the start-clean is confirmed safe.

`import data --start-clean` now aborts with an explanatory error when the change-event queue's earliest segment has been archived/deleted by `archive changes`. Previously this scenario would proceed and leave the migration in an inconsistent state. No command-line, configuration, installation, or report changes.

No.

No. The change only adds a read-only query against the existing queue-segment metadata table; no on-disk structures are modified.

---

| Component        


yb-voyager/cmd/importData.go
```diff
@@ -2369,15 +2369,35 @@ func clearMigrationStateForImportDataStartClean(state *ImportDataState, importFi
...
+			return goerrors.Errorf("failed to check for archived/deleted queue segments: %s", err)
+		}
+		if resumeSegmentDeleted {
+			utils.ErrExit("cannot perform import data with --start-clean: some queue segments have already been archived/deleted by 'archive changes'. The change-event queue can no longer be re-streamed from the beginning, so a clean restart is not possible.")
+		}
+	}
+
+	metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
+		msr.OnPrimaryKeyConflictAction = ""
+	})
```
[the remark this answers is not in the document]
AmeyaKasture: @priyanshi-yb , I think this is from the iterative change, can you please look into this

yb-voyager/cmd/importData.go
```diff
@@ -2369,15 +2369,35 @@ func clearMigrationStateForImportDataStartClean(state *ImportDataState, importFi
...
+		if importerRole == SOURCE_DB_IMPORTER_ROLE {
+			// fall-back import only consumes segments exported from the target db.
+			segmentsExporterRole = TARGET_DB_EXPORTER_FB_ROLE
+		}
+		resumeSegmentDeleted, err := metaDB.IsResumeSegmentDeleted(segmentsExporterRole)
+		if err != nil {
+			return goerrors.Errorf("failed to check for archived/deleted queue segments: %s", err)
+		}
+		if resumeSegmentDeleted {
+			utils.ErrExit("cannot perform import data with --start-clean: some queue segments have already been archived/deleted by 'archive changes'. The change-event queue can no longer be re-streamed from the beginning, so a clean restart is not possible.")
```
makalaaneesh: [AI] **`utils.ErrExit` inside an error-returning function bypasses caller error handling.**

`clearMigrationStateForImportDataStartClean` returns an `error`, and its caller already handles it:

```go
err = clearMigrationStateForImportDataStartClean(...)
if err != nil {
    utils.ErrExit("Failed to clean MigrationStatusRecord ...: %s", err)
}
```

Calling `utils.ErrExit` here (line 2390) terminates the process directly, bypassing the caller's error path, error wrapping, and any deferred cleanup. Per `yb-voyager/cmd/.cursor/BUGBOT.md`: "Command-level functions should return errors, not call `utils.ErrExit` directly (unless they are the top-level entry point for a command)."

Suggested fix: `return goerrors.Errorf("cannot perform import data with --start-clean: ...")` so the existing caller-level `ErrExit` reports it. (The neighboring pre-existing `ErrExit` calls at the snapshot/queue-reset steps have the same smell, but this one is newly added.)

yb-voyager/cmd/importData.go
```diff
@@ -2369,15 +2369,35 @@ func clearMigrationStateForImportDataStartClean(state *ImportDataState, importFi
...
+		}
+	}
+
+	metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
```
makalaaneesh, line 2389: [AI] **MSR update closure mutates the captured `msr` copy instead of the persisted `record`, so `OnPrimaryKeyConflictAction` is never actually cleared.**

```go
    msr.OnPrimaryKeyConflictAction = ""   // writes to the outer captured copy
})
```

`UpdateMigrationStatusRecord` -> `UpdateJsonObjectInMetaDB` fetches a *fresh* `record` from the metaDB, calls `updateFn(record)`, then marshals `record` back. This closure modifies the previously-fetched `msr` rather than the `record` argument, so the freshly-loaded record is written back unchanged and `OnPrimaryKeyConflictAction` is never reset in the DB. The returned error is also ignored.

This is pre-existing logic that the PR carried forward verbatim while removing the `else` block, but it sits on a changed line and defeats the start-clean reset of this field.

Suggested fix:
```go
err = metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
    record.OnPrimaryKeyConflictAction = ""
})
if err != nil { return goerrors.Errorf("reset OnPrimaryKeyConflictAction: %s", err) }
```
The `UpdateImportDataStatusRecord` call just below it also has its returned `err` left unchecked before `handleStartCleanForSnapshot`.

makalaaneesh (commented):
[AI] Automated review — 3 findings (0 critical, 3 warning). See inline comments. The core guardrail logic is sound: it consistently mirrors the resume path in eventQueue.go (same exporter-role filtering and earliest-segment selection), and the `deleted` flag is set in lockstep with file removal in archivechanges/cleanup.go, so it is a reliable proxy for a missing segment file.

yb-voyager/src/metadb/metadataDB.go
```diff
@@ -398,6 +398,31 @@ func (m *MetaDB) GetMinSegmentExportedByAndNotImportedBy(importerRole string, ex
...
 }
 
+// IsResumeSegmentDeleted reports whether the earliest queue segment (the lowest
+// segment_no for the given exporter role) has been deleted from disk by the
+// `archive changes` workflow. This is the segment that `import data --start-clean`
+// would reset to and attempt to re-stream from after clearing the per-importer
+// imported_by flags. If it has been deleted, the change-event queue can no longer
+// be re-streamed from the beginning and start-clean cannot proceed safely.
+// Returns false when there are no queue segments yet.
+func (m *MetaDB) IsResumeSegmentDeleted(exporterRole string) (bool, error) {
```
makalaaneesh: [AI] **New guardrail logic has no unit-test coverage.**

`IsResumeSegmentDeleted` and the start-clean guardrail it backs are new behavior with non-trivial branching (empty vs. fall-back exporter role, no-rows case, earliest-segment selection), but no tests were added even though `yb-voyager/src/metadb/metadataDB_test.go` already exists.

Per `yb-voyager/.cursor/BUGBOT.md`: new logic should have tests, and error/edge paths should verify the specific result — not just that the function ran. Worth covering at minimum: (a) no queue segments -> returns `false, nil`; (b) earliest segment `deleted=1` -> `true`; (c) earliest segment `deleted=0` with a later segment deleted -> `false`; (d) fall-back role filtering picks the earliest `target_db_exporter_fb` segment rather than a globally-earlier `source_db_exporter` one.

These cases also document the invariant that the guardrail stays in sync with `resolveSegmentToResumeFrom` in `eventQueue.go`.

yb-voyager/cmd/importData.go
```diff
@@ -2369,15 +2369,35 @@ func clearMigrationStateForImportDataStartClean(state *ImportDataState, importFi
...
 	}
 
+	// Guardrail: for change-streaming imports, start-clean resets the per-importer
+	// imported_by flags and re-streams the queue from the earliest segment. If that
+	// segment has already been archived/deleted by `archive changes`, the queue can
+	// no longer be re-streamed from the beginning. Detect this and fail before
+	// mutating any metaDB state.
+	if changeStreamingIsEnabled(importType) {
+		segmentsExporterRole := ""
+		if importerRole == SOURCE_DB_IMPORTER_ROLE {
```
makalaaneesh: start-clean is not supported for import-data-to-source. So this should not be required?

yb-voyager/src/metadb/metadataDB.go
makalaaneesh: "Resume" name is a bit misleading no? 
It's not resume that we want to check, it's start-clean. 

So the check is more about `AnySegmentsDeletedOrArchived`  perhaps?

yb-voyager/src/metadb/metadataDB.go
```diff
@@ -398,6 +398,31 @@ func (m *MetaDB) GetMinSegmentExportedByAndNotImportedBy(importerRole string, ex
...
+	query := fmt.Sprintf("SELECT deleted FROM %s", QUEUE_SEGMENT_META_TABLE_NAME)
+	if exporterRole != "" {
+		query = fmt.Sprintf("%s WHERE exporter_role = '%s'", query, exporterRole)
+	}
+	query = fmt.Sprintf("%s ORDER BY segment_no ASC LIMIT 1;", query)
```
makalaaneesh: I think it might be a good idea NOT to assume that archive changes is archiving/deleting in ASC order. 

We can just check: Have any segments been deleted? And if the answer is yes, we should prevent start-clean.  

Tomorrow, for whatever reason, if we decide to change the order in which archive-changes goes about deleting/archiving, then this code will break. 

Does that make sense?

yb-voyager/cmd/importData.go
AmeyaKasture: After cutover if import data gets interrupted, the guardrail wouid be essential too

yb-voyager/src/metadb/metadataDB.go
AmeyaKasture: Wont we always have them in ascending order as data would be consumed sequentially , thus it only makes sense to archive them sequnetially

yb-voyager/cmd/importData.go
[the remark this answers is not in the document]
priyanshi-yb: I don't think its from anything recent as I can see its some old code and I feel the suggestion from gemini is correct so we should modify it if it has come up

yb-voyager/src/metadb/metadataDB.go
```diff
@@ -398,6 +398,30 @@ func (m *MetaDB) GetMinSegmentExportedByAndNotImportedBy(importerRole string, ex
...
 }
 
+// AnySegmentsDeletedOrArchived reports whether any queue segment (the lowest
+// segment_no for the given exporter role) has been deleted or archived by the
+// `archive changes` workflow. `import data --start-clean` resets per-importer
+// imported_by flags and attempts to re-stream the queue from the beginning, so a
+// missing segment means it cannot proceed safely. Returns false when there are no
+// queue segments yet.
+func (m *MetaDB) AnySegmentsDeletedOrArchived(exporterRole string) (bool, error) {
+	query := fmt.Sprintf("SELECT deleted FROM %s", QUEUE_SEGMENT_META_TABLE_NAME)
```
priyanshi-yb: why only `deleted` here? we should also check `archived` right?

yb-voyager/cmd/importData.go
priyanshi-yb: The flow before and after cutover looks like this 
1. Before cutover - `export data from source `(source_db_exporter) and `import data` (target_db_importer) or optionally in fall-forward workflow `import data to source-replica` (source_replica_db_importer)
2. After cutover - `export data from target` (target_db_Exporter_fb/ff) and ( in case of fallback `import data to soruce` (source_db_importer)  or import data to source-replica in case of fall-forward(source_replica_db_importer) )

`import data` and `import data to source-replica` support `start-clean` so we need to have this guardrail for them. For import data to source, it doesn't allow start-clean, so this isn't required

Two things
1. We should have a guardrail for the import data to source-replica (fall-forward) also.
2. How is this working? I don't see setting `segmentsExporterRole` to `source_db_exporter`.

yb-voyager/src/metadb/metadataDB.go
AmeyaKasture: Archived and Deleted are set and unset together as an atomic operation, wont make any difference

yb-voyager/cmd/importData.go
AmeyaKasture: The default case is source db exporter , if we dont need for the case of target db exporter we can simply remove the parameter altogether and it would work for both imports (target and source replica) as in both cases source db will be exporting from source and thus only need to check if they are deleted

priyanshi-yb: I missed mentioning above - 
After cutover - export data from target (target_db_Exporter_fb/ff) and ( in case of fallback import data to soruce (source_db_importer) or import data to source-replica in case of fall-forward(source_replica_db_importer) )

so the source-replica-db-importer is present event after cutover and taking the changes from target-db-exporter-ff so we need to handle for target exporter for taht case

yb-voyager/cmd/importData.go
```diff
@@ -2369,15 +2369,35 @@ func clearMigrationStateForImportDataStartClean(state *ImportDataState, importFi
...
+	// imported_by flags and re-streams the queue from the earliest segment. If that
+	// segment has already been archived/deleted by `archive changes`, the queue can
+	// no longer be re-streamed from the beginning. Detect this and fail before
+	// mutating any metaDB state.
+	if changeStreamingIsEnabled(importType) {
+		segmentsExporterRole := ""
+		if importerRole == SOURCE_REPLICA_DB_IMPORTER_ROLE {
+			// fall-forward import only consumes segments exported from the target db.
+			segmentsExporterRole = TARGET_DB_EXPORTER_FF_ROLE
+		}
```
priyanshi-yb: source-replica importer not only imports from the target db, it also imports from the source db, so we need to keep a check for both. If the segment from any of them is archived/deleted, we should error out. Let's also reorganise this a bit better

```
var segmentsExporterRole []string
switch importerRole {
case TARGET_DB_IMPORTER_ROLE:
          segmentsExporterRole = append(segmentsExporterRole, SOURCE_DB_EXPORTER)
case SOURCE_REPLICA_DB_IMPORTER_ROLE:
          segmentsExporterRole = append(segmentsExporterRole, SOURCE_DB_EXPORTER)
          segmentsExporterRole = append(segmentsExporterRole,     TARGET_DB_EXPORTER_FF_ROLE)

case SOURCE_DB_IMPORTER_ROLE:
         return fmt.Errorf("--start-clean is not supported for import data to source")
}
```

yb-voyager/src/metadb/metadataDB.go
```diff
@@ -398,6 +398,31 @@ func (m *MetaDB) GetMinSegmentExportedByAndNotImportedBy(importerRole string, ex
 	return segmentNum.Int64, nil
 }
 
+// AnySegmentsDeletedOrArchived reports whether any queue segment (the lowest
+// segment_no for the given exporter role) has been deleted or archived by the
+// `archive changes` workflow. `import data --start-clean` resets per-importer
+// imported_by flags and attempts to re-stream the queue from the beginning, so a
+// missing segment means it cannot proceed safely. Returns false when there are no
+// queue segments yet.
+func (m *MetaDB) AnySegmentsDeletedOrArchived(exporterRole string) (bool, error) {
```
priyanshi-yb: Modify this to handle list of exporter roles as per what I mentioned above for source-replica

priyanshi-yb (approved):
LGTM

yb-voyager/src/metadb/metadataDB.go
makalaaneesh: sorry missed replying to this. 

While we do it that way today, there's no strong reason we need to archive them sequentially. Once a segment is imported, it can be archived. Tomorrow, if we change that logic, then this logic will be buggy. 

I'd still recommend we do the proper check here: is ANY segment deleted OR archived. (instead of just checking the first one)

makalaaneesh (approved):

PR #3604, opened 2026-06-11
Fix: handle multi column unique indexes in the conflict detection cache

priyanshi-yb:

The live-migration conflict-detection cache previously modeled each table's unique keys as a flat list of columns (`tableToUniqueKeyColumns: map[table][]string`) and checked for conflicts one column at a time. A conflict was flagged whenever *any single* unique-key column value matched between a cached and an incoming event. For composite (multi-column) unique indexes this produced large numbers of false-positive conflicts — two events that collide on just one column of a composite index were serialized even though their full composite keys differ — which unnecessarily throttles import throughput.

This PR remodels the per-table unique-key metadata as a list of unique *indexes*, where each index is its ordered set of columns (`tableToUniqueIndexes: map[table][][]string`). Conflict detection now evaluates each unique index as a unit: a before-after or before-before conflict is reported only when *all* columns of that index match between the two events (with NULL treated as equal to NULL, preserving prior semantics). The source-side extraction queries for PostgreSQL, YugabyteDB, and Oracle were rewritten to group columns by constraint/index name and order them by ordinal position so composite keys are reconstructed correctly. Leaf-partition indexes are merged into the root table via `MergeUniqueIndexes`, and a new metaDB key `table_to_unique_indexes_key` stores the richer structure.

Limitation - 
Currently we are using the source side unique indexes for detecting the unique conflicts but ideally we should use the target db indexes as those are the actual indexes that will fail with unique violation if the events are not exectued in order. Ticket for tracking it - #3608 

No user-facing changes to the command line, configuration, installation, or reports. The behavior change is internal to live-migration event processing: composite unique indexes are now handled correctly, which eliminates false positive conflicts and reduces unnecessary serialization of events during `import data` for live migrations.

- Added a new unit-test file `conflictDetectionCache_test.go` (build tag `unit`) covering composite-index true conflicts, the composite false-positive fix (match on one column only ⇒ no conflict), multiple composite indexes per table, missing-column cases, same-PK no-conflict, and NULL handling for both before-after and before-before paths (`TestIndexTupleConflicts_*`, `TestUniqueIndexConflicts_*`, `TestEventsConfict_*`).
- Updated existing source-driver integration tests to assert composite-index extraction: `TestPostgresGetTableToUniqueKeyColumnsMap` and `TestYugabyteGetTableToUniqueKeyColumnsMap` now include a `composite_unique_table` (`UNIQUE (first_name, last_name)` plus a single-column `phone` unique), and `TestOracleGetTableToUniqueKeyColumnsMap` expects `{EMAIL, PHONE}` + `{ADDRESS}`. A shared `assertEqualUniqueIndexes` helper compares index sets order-independently.
- Updated `live_migration_test.go` constructor call sites for the new `[][]string` map type.
- Existing live-migration integration tests exercise the end-to-end path. Running a real composite-unique-index live workload to confirm the throughput improvement on a representative dataset is still recommended before release.

No.

Yes — **MetaDB**. A new key `table_to_unique_indexes_key` is added alongside the legacy `table_to_unique_key_columns_key`. The change is designed to be upgrade-safe: when reading, the new code first looks for the new key and, if absent, falls back to the legacy flat-column key and upgrades each column into a single-column index (`[]string{col}` → `[][]string{{col}}`). This covers the common case of upgrading the binary mid-migration after export was done on an older version. Downgrading to an older binary after this binary has written the new key is not supported (the old binary would not find the legacy key), which is consistent with the project's general no-downgrade-mid-migration stance.

---

| Component        


yb-voyager/src/testlivemigration/live_migration_integration_test.go
```diff
@@ -1521,6 +1521,282 @@ FROM generate_series(1, 20) as i;`,
...
+	err = liveMigrationTest.InitiateCutoverToTarget(true, nil)
+	testutils.FatalIfError(t, err, "failed to initiate cutover to target")
+
+	err = liveMigrationTest.WaitForCutoverComplete(0, 30)
+	testutils.FatalIfError(t, err, "failed to wait for cutover complete")
+
+	err = liveMigrationTest.ExecuteTargetDelta()
+	testutils.FatalIfError(t, err, "failed to execute target delta")
+
+	err = liveMigrationTest.WaitForFallbackStreamingComplete(map[string]ChangesCount{
```
makalaaneesh: [AI] **Swallowed error from `WaitForFallbackStreamingComplete`.**

The return value is assigned to `err` but never checked — the very next statement overwrites `err` with the `ValidateDataConsistency` result:

```go
err = liveMigrationTest.WaitForFallbackStreamingComplete(...)   // no FatalIfError

err = liveMigrationTest.ValidateDataConsistency(...)
testutils.FatalIfError(t, err, "failed to validate data consistency")
```

Every other step in this test calls `testutils.FatalIfError`. If fall-back streaming times out or fails, that error is silently lost and the test reports a less specific failure (or none at all).

Suggested fix: add `testutils.FatalIfError(t, err, "failed to wait for fallback streaming complete")` immediately after the call, matching the surrounding pattern.

yb-voyager/src/srcdb/yugabytedb_test.go
```diff
@@ -269,36 +269,82 @@ func TestYugabyteGetTableToUniqueKeyColumnsMap(t *testing.T) {
...
+			val2 text
+		) PARTITION BY LIST (data);
+		CREATE TABLE test_expression_indexes.table_partitioned_l5 PARTITION OF test_expression_indexes_cross.table_partitioned5 FOR VALUES IN ('London');
+		CREATE TABLE test_expression_indexes.table_partitioned_s5 PARTITION OF test_expression_indexes_cross.table_partitioned5 FOR VALUES IN ('Sydney');
+		CREATE TABLE test_expression_indexes.table_partitioned_b5 PARTITION OF test_expression_indexes_cross.table_partitioned5 FOR VALUES IN ('Boston');
+
+		CREATE UNIQUE INDEX idx_expression_indexes_17 ON test_expression_indexes.table_partitioned_l5 (val) WHERE val2<>'';`)
 	defer testYugabyteDBSource.TestContainer.ExecuteSqls(`DROP SCHEMA test_schema CASCADE;`)
 
+	testYugabyteDBSource.Schemas = []sqlname.Identifier{
```
makalaaneesh, line 302: [AI] **Test leaks two schemas and mutates shared `Schemas` without restoring.**

This test now creates `test_expression_indexes_cross` and `test_expression_indexes`, and reassigns `testYugabyteDBSource.Schemas`, but the deferred cleanup only drops `test_schema`:

```go
defer testYugabyteDBSource.TestContainer.ExecuteSqls(`DROP SCHEMA test_schema CASCADE;`)
```

The analogous `postgres_test.go` (`TestPostgresGetTableToUniqueKeyColumnsMap`) correctly drops all three schemas. Since the testcontainer is shared across the package's tests, the two extra schemas and the unrestored `Schemas` field leak into subsequent tests.

Per `yb-voyager/src/srcdb/.cursor/BUGBOT.md`: "Each test should set up its own schema, run queries, assert results, and clean up."

Suggested fix: drop `test_expression_indexes_cross` and `test_expression_indexes` in the deferred cleanup (mirroring `postgres_test.go`), and restore the original `Schemas` value.

yb-voyager/cmd/live_migration.go
```diff
@@ -585,36 +585,59 @@ func processEvents(chanNo int, evChan chan *tgtdb.Event, lastAppliedVsn int64, d
...
-	conflictDetectionCache = NewConflictDetectionCache(tableToUniqueKeyColumns, evChans, sourceDBTypeForConflictCache)
+	conflictDetectionCache = NewConflictDetectionCache(tableToUniqueIndexes, evChans, sourceDBTypeForConflictCache)
 	return nil
 }
 
-func getTableToUniqueKeyColumnsMapFromMetaDB(exporterRole string) (*utils.StructMap[sqlname.NameTuple, []string], error) {
-	log.Infof("fetching table to unique key columns map from metaDB")
-	var metaDbData map[string][]string
-	res := utils.NewStructMap[sqlname.NameTuple, []string]()
+func getTableToUniqueIndexesMapFromMetaDB(exporterRole string) (*utils.StructMap[sqlname.NameTuple, [][]string], error) {
```
makalaaneesh, line 623: [AI] **Backward-compat upgrade path is untested.**

`getTableToUniqueIndexesMapFromMetaDB` holds the load-bearing logic for users upgrading voyager mid-migration: it reads the new `TABLE_TO_UNIQUE_INDEXES_KEY`, and otherwise falls back to the legacy `TABLE_TO_UNIQUE_KEY_COLUMNS_KEY`, converting each flat column into a single-column index. No unit test exercises this fallback or the new-key-takes-precedence behavior.

Per `yb-voyager/src/metadb/.cursor/BUGBOT.md`: "Test the upgrade path: verify that a metaDB created by an older voyager version still works with new code." This is a serialized-state backward-compatibility surface, so a regression here would silently break in-flight migrations on upgrade.

Suggested fix: add a unit test that (a) writes only the legacy key and asserts the converted `[][]string` (one column per single-column index), and (b) writes both keys and asserts the new key wins.

makalaaneesh (commented):
[AI] Automated review — 3 findings (0 critical, 3 warning). See inline comments. (6 suggestions were filtered out and not posted.)

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -316,7 +317,11 @@ func (c *ConflictDetectionCache) eventsConfict(cachedEvent *tgtdb.Event, incomin
 		return false
 	}
 
-	uniqueKeyColumns, _ := c.tableToUniqueKeyColumns.Get(cachedEvent.TableNameTup)
+	uniqueIndexes, _ := c.tableToUniqueIndexes.Get(cachedEvent.TableNameTup)
+	uniqueKeyColumns := make([]string, 0) // flattening the unique indexes to get the unique key columns
+	for _, indexColumns := range uniqueIndexes {
+		uniqueKeyColumns = append(uniqueKeyColumns, indexColumns...)
```
makalaaneesh: we still need this for the export-data-from-source side? 
but by default, we use partition-by-table there, so i guess this is also just a fallback for that, correct?  
Can we put this block inside the `if isTargetDBExporter(incomingEvent.ExporterRole)` ? 

Also pls update the comments properly to indicate what the current state is and when this block would be invoked.

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -344,143 +349,181 @@ func (c *ConflictDetectionCache) eventsConfict(cachedEvent *tgtdb.Event, incomin
...
+func uniqueIndexColumnValuesEqual(leftFields, rightFields map[string]*string, indexColumns []string) bool {
+	for _, column := range indexColumns {
+		if !uniqueKeyColumnValuesEqual(leftFields[column], rightFields[column]) {
+			return false
 		}
+	}
+	return true
+}
 
+func formatUniqueIndexColumnValuesForLog(fields map[string]*string, indexColumns []string) string {
```
makalaaneesh, line 517: this is in the critical path. We might want to use string builder here instead of creating and appending multiple strings.

yb-voyager/cmd/conflictDetectionCache_test.go
```diff
@@ -0,0 +1,358 @@
...
+		BeforeFields: map[string]*string{"a": strPtr("1"), "b": strPtr("2")},
+	}
+	incoming := &tgtdb.Event{
+		Vsn:          2,
+		Op:           "c",
+		TableNameTup: testTableTuple(),
+		Key:          map[string]*string{"id": strPtr("2")},
+		Fields:       map[string]*string{"a": strPtr("1"), "b": strPtr("2")},
+	}
+	assert.True(t, cache.uniqueIndexConflicts(cached, incoming, []string{"a", "b"}))
```
makalaaneesh: I think it would be better to call `.eventsConflict()` rather than an internal function `uniqueIndexConflicts` , no?

yb-voyager/cmd/exportData.go
```diff
@@ -2303,55 +2303,54 @@ func createUpdateExportedRowCountEventList(tableNames []string) []*cp.UpdateExpo
...
-		log.Infof("no table to unique key columns map found, saving nil to metaDB")
-		key := fmt.Sprintf("%s_%s", metadb.TABLE_TO_UNIQUE_KEY_COLUMNS_KEY, exporterRole)
-		err = metadb.UpdateJsonObjectInMetaDB(metaDB, key, func(record *map[string][]string) {
+		log.Infof("no table to unique indexes map found, saving nil to metaDB")
+		err = metadb.UpdateJsonObjectInMetaDB(metaDB, key, func(record *map[string][][]string) {
 			*record = nil
 		})
 		if err != nil {
-			utils.ErrExit("insert table to unique key columns map: %w", err)
+			utils.ErrExit("insert table to unique indexes map: %w", err)
```
makalaaneesh: let's fix this up. no ErrExit here. pls check other places in this func as well.

yb-voyager/cmd/exportData.go
```diff
@@ -2303,55 +2303,54 @@ func createUpdateExportedRowCountEventList(tableNames []string) []*cp.UpdateExpo
...
-	res, err := source.DB().GetTableToUniqueKeyColumnsMap(tableList)
+func saveTableToUniqueIndexesMapInMetaDB(tableList []sqlname.NameTuple, leafPartitions *utils.StructMap[sqlname.NameTuple, []sqlname.NameTuple]) {
+	res, err := source.DB().GetTableToUniqueIndexesMap(tableList)
 	if err != nil {
-		utils.ErrExit("get table to unique key columns map: %w", err)
+		utils.ErrExit("get table to unique indexes map: %w", err)
 	}
 
+	key := fmt.Sprintf("%s_%s", metadb.TABLE_TO_UNIQUE_INDEXES_KEY, exporterRole)
 	if res == nil {
```
makalaaneesh, line 2429: why/when would this be nil? 
Is it when there are no unique indexes at all?

yb-voyager/cmd/exportData.go
```diff
@@ -2303,55 +2303,54 @@ func createUpdateExportedRowCountEventList(tableNames []string) []*cp.UpdateExpo
...
+			res.Put(rootTable, rootUniqueIndexes)
 		}
 		return true, nil
 	})
 
-	metaDbData := make(map[string][]string)
-	res.IterKV(func(k sqlname.NameTuple, v []string) (bool, error) {
+	metaDbData := make(map[string][][]string)
+	res.IterKV(func(k sqlname.NameTuple, v [][]string) (bool, error) {
 		metaDbData[k.AsQualifiedCatalogName()] = v
```
makalaaneesh, line 2461: we should be using AsKey here no?

yb-voyager/cmd/live_migration.go
```diff
@@ -585,36 +585,59 @@ func processEvents(chanNo int, evChan chan *tgtdb.Event, lastAppliedVsn int64, d
...
+				return nil, goerrors.Errorf("lookup table %s in name registry: %v", tableNameRaw, err)
+			}
+			res.Put(tableName, indexes)
+		}
+		return res, nil
 	}
-	log.Infof("fetched table to unique key columns map: %v", metaDbData)
 
-	for tableNameRaw, columns := range metaDbData {
+	// Backward compatibility: upgrade legacy flat column lists to single-column indexes.
```
makalaaneesh: let's discuss if we really need this

yb-voyager/src/srcdb/oracle.go
```diff
@@ -629,11 +630,13 @@ func (ora *Oracle) GetPartitions(tableName sqlname.NameTuple) []string {
 	panic("not implemented")
 }
 
-var oraQueryTmplForUniqCols = `
+const oraQueryTmplForUniqIndexes = `
```
makalaaneesh, line 633: Not really reviewing this oracle code. 
Have you manually verified this?

yb-voyager/src/srcdb/postgres_test.go
```diff
@@ -307,6 +308,13 @@ func TestPostgresGetTableToUniqueKeyColumnsMap(t *testing.T) {
             age INT
         );`,
 		`CREATE UNIQUE INDEX idx_age ON test_schema.another_unique_table(age);`,
+		`CREATE TABLE test_schema.composite_unique_table (
```
makalaaneesh, line 311: do we want to also include partitioned tables here?

yb-voyager/src/srcdb/postgres.go
```diff
@@ -955,57 +956,106 @@ WHERE parent.relname='%s' AND nmsp_parent.nspname = '%s' `, tname, sname)
...
+	return result
+}
+
+// MergeUniqueIndexes merges two index lists, deduplicating by column signature.
+func MergeUniqueIndexes(existing, additional [][]string) [][]string {
+	return dedupeUniqueIndexes(append(existing, additional...))
+}
+
+func buildUniqueIndexesMapFromPGRows(rows *sql.Rows, tableStrToNameTupleMap map[string]sqlname.NameTuple) (*utils.StructMap[sqlname.NameTuple, [][]string], error) {
+	indexColumnMaps := make(map[tableIndexKey]map[int]string)
```
makalaaneesh: ~~i don't think you can use a struct as a key for a map in golang? Isn't that why we have StructMap? 🤔~~

Ok just saw our struct map documentation, it says "In golang, by default, you can use structs as keys as long as the fields of the structs themselves are comparable."

yb-voyager/src/srcdb/postgres.go
```diff
@@ -955,57 +956,106 @@ WHERE parent.relname='%s' AND nmsp_parent.nspname = '%s' `, tname, sname)
...
+
+	result := utils.NewStructMap[sqlname.NameTuple, [][]string]()
+	for key, ordinalToColumn := range indexColumnMaps {
+		tableNameTuple, ok := tableStrToNameTupleMap[key.tableCatalogName]
+		if !ok {
+			return nil, goerrors.Errorf("table %s not found in table list", key.tableCatalogName)
+		}
+
+		ordinals := lo.Keys(ordinalToColumn)
+		sort.Ints(ordinals)
```
makalaaneesh: is there a way you can push this ordinal sorting and agg into a single list logic into SQL itself?
We wouldn't have to write all this go code then

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -1055,88 +1056,44 @@ WITH unique_constraints AS (
...
+        AND c.contype IS NULL
         AND n.nspname = ANY('{%s}')
         AND t.relname = ANY('{%s}')
 )
--- UNION will remove duplicate rows between unique constraints and unique indexes
-SELECT table_schema, table_name, column_name FROM unique_constraints
-UNION
-SELECT table_schema, table_name, column_name FROM unique_indexes;
+SELECT table_schema, table_name, index_key, column_name, ordinal_position FROM unique_constraints
+UNION ALL
```
makalaaneesh: what's the significance of UNION vs UNION ALL change. Is that relevant?

yb-voyager/src/srcdb/oracle.go
priyanshi-yb: Not yet, I am relying on the unique conflict live migration test case of Oracle that we have. Will try to do that before closing the PR.

yb-voyager/src/srcdb/postgres_test.go
priyanshi-yb: we have partitioned case for expression based one, but not for this. Will add one

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: yes, this struct is basic and easily comparable

yb-voyager/cmd/exportData.go
priyanshi-yb: yes

yb-voyager/cmd/exportData.go
priyanshi-yb: Yeah, we can, I can't think of any reason for using QualifiedCatalogName as we are already doing lookup in import while reading this. I think its just because of clean metadata without quotes etc..

yb-voyager/src/srcdb/yugabytedb.go
priyanshi-yb: Not really, cursor made this change for some better output but there is no difference in result as such

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -344,143 +354,181 @@ func (c *ConflictDetectionCache) eventsConfict(cachedEvent *tgtdb.Event, incomin
...
+	var logStr strings.Builder
+	for _, column := range indexColumns {
+		val, ok := fields[column]
+		if !ok {
+			logStr.WriteString(column + "=<missing>")
+			continue
+		}
+		if val == nil {
+			logStr.WriteString(column + "=nil")
```
makalaaneesh, line 526: We're using a strings.Builder, but we're still creating temporary strings with + before passing them to WriteString, which reduces some of the performance benefit.

suggestion: 
```
       logStr.WriteString(column)
        logStr.WriteString("=<missing>")
```

yb-voyager/cmd/exportData.go
makalaaneesh: most likely this is already tested, but i hope the read of this field during import side does not fail in case it's nil.

yb-voyager/cmd/live_migration.go
```diff
@@ -425,8 +425,8 @@ func handleEvent(event *tgtdb.Event,
 		Checking for all possible conflicts among events
 		For more details about ConflictDetectionCache see the related comment in [conflictDetectionCache.go](../conflictDetectionCache.go)
 	*/
-	uniqueKeyCols, _ := conflictDetectionCache.tableToUniqueKeyColumns.Get(event.TableNameTup)
-	if len(uniqueKeyCols) > 0 {
+	uniqueIndexes, _ := conflictDetectionCache.tableToUniqueIndexes.Get(event.TableNameTup)
```
makalaaneesh: this will have merge conflicts, right?

yb-voyager/src/metadb/metadataDB.go
```diff
@@ -45,6 +45,8 @@ var (
 	SOURCE_DB_IDENTITY_COLUMNS_KEY             = "source_db_identity_columns_key"
 	SOURCE_INDEXES_INFO_KEY                    = "source_indexes_info_key"
 	TABLE_TO_UNIQUE_KEY_COLUMNS_KEY            = "table_to_unique_key_columns_key"
+	TABLE_TO_UNIQUE_INDEXES_KEY                = "table_to_unique_indexes_key"
+	CONFLICT_DETECTION_STATS_KEY               = "conflict_detection_stats"
```
makalaaneesh: subsequent PR? don't see usages in this

yb-voyager/src/srcdb/yugabytedb.go
```diff
@@ -1055,88 +1056,62 @@ WITH unique_constraints AS (
...
 )
--- UNION will remove duplicate rows between unique constraints and unique indexes
-SELECT table_schema, table_name, column_name FROM unique_constraints
-UNION
-SELECT table_schema, table_name, column_name FROM unique_indexes;
+SELECT
+    table_schema,
+    table_name,
+    index_key,
+    array_agg(column_name ORDER BY ordinal_position) AS columns
```
makalaaneesh, line 1102: nice, this is what i was hoping would be possible :)

yb-voyager/src/srcdb/postgres.go
```diff
@@ -955,57 +956,92 @@ WHERE parent.relname='%s' AND nmsp_parent.nspname = '%s' `, tname, sname)
...
+
+		tableCatalogName := fmt.Sprintf("%s.%s", schemaName, tableName)
+		tableNameTuple, ok := tableStrToNameTupleMap[tableCatalogName]
+		if !ok {
+			return nil, goerrors.Errorf("table %s not found in table list", tableCatalogName)
+		}
+
+		indexes, _ := result.Get(tableNameTuple)
+		indexes = append(indexes, columns)
+		result.Put(tableNameTuple, dedupeUniqueIndexes(indexes))
```
makalaaneesh: why do we need dedupe here? The query that returns rows will not have duplicates, right?

yb-voyager/src/testlivemigration/multi_column_uk_conflict_sql.go
```diff
@@ -0,0 +1,140 @@
+//go:build integration_live_migration
```
makalaaneesh: why separate file here? 
IF we want let's create a separate test file for conflicts and put both the sql and the tests in there?

yb-voyager/src/srcdb/postgres.go
priyanshi-yb: This is possible in case there are multiple indexes with the same column set, like redundant indexes, so we are just deduplicating that.

yb-voyager/src/testlivemigration/multi_column_uk_conflict_sql.go
priyanshi-yb: Yeah, agree, cursor separated out the SQL thing in this, I will move the unique conflict tests in this file

yb-voyager/cmd/live_migration.go
makalaaneesh: not doing backward-compat; we will mark it as a breaking release

yb-voyager/src/srcdb/postgres.go
makalaaneesh: can we do this at the end instead of while appending each entry?

yb-voyager/cmd/live_migration.go
priyanshi-yb: removed the backward compatibility

yb-voyager/cmd/live_migration.go
```diff
@@ -611,36 +611,35 @@ func processEvents(chanNo int, evChan chan *tgtdb.Event, lastAppliedVsn int64, d
...
-	found, err := metaDB.GetJsonObject(nil, key, &metaDbData)
+	indexesKey := fmt.Sprintf("%s_%s", metadb.TABLE_TO_UNIQUE_INDEXES_KEY, exporterRole)
+	var indexesMetaDbData map[string][][]string
+	found, err := metaDB.GetJsonObject(nil, indexesKey, &indexesMetaDbData)
 	if err != nil {
 		return nil, err
 	}
-	if !found {
-		return nil, goerrors.Errorf("table to unique key columns map not found in metaDB")
+	if found {
```
makalaaneesh: !found? 
Seems like a mistake

makalaaneesh (approved):
LGTM

PR #3611, opened 2026-06-16
Single source of truth for the CDC connector version + freshness check

ShivanshGahlot:

The bundled YugabyteDB CDC connector versions were hardcoded as full GitHub download URLs in `installer_scripts/install-yb-voyager`, with the version literal duplicated within each URL and no machine-readable declaration of "which connector do we ship". Because the CDC team does **not** guarantee forward-compatibility, the connector must be upgraded regularly to track new YugabyteDB series — which needs (a) one canonical place to bump and (b) an automated way to notice when we fall behind.

This PR introduces that single source of truth and a freshness check:

- **`yb-voyager/versions/connector-versions.json`** — the canonical declaration of the logical and gRPC connector release tags. Connector compatibility is per YB series (`YEAR.TRACK`): a `2025.2` connector works with all `2025.2.x` servers; the trailing `.3` in `dz.2.5.2.yb.2025.2.3` is the connector's own release counter. The supported series is embedded in the tag itself, so it is not stored separately.
- **`versions.go`** embeds the JSON (same `//go:embed` pattern as `yb-versions.json`) and exposes `GetLogicalConnectorTag()` and `GetGRPCConnectorTag()`.
- **`install-yb-voyager`** now derives the two connector download URLs from the JSON (inside `package_debezium_server_local`, via `jq`). The hardcoded URLs and their duplicated version literals are gone; the constructed URLs are byte-for-byte identical to before. The script fails fast with a clear message if the JSON is missing or a tag is empty/`null`, rather than building a broken URL that 404s later. `jq` was previously installed only for release installs; since `package_debezium_server_local` runs only in local-source builds, the jq install is moved out of the release-only gate so it is available in local builds too (`fetch_release_data` stays release-only).
- **Freshness test** (`connector_latest_test.go`, build tag `connector_latest_stable`) queries the `yugabyte/debezium` releases API and fails when a newer logical-connector release exists than the bundled one (series-first comparison, then connector counter). It runs as a `Test Latest Connector Version` step in `.github/workflows/issue-tests.yml`, alongside the YB latest-stable check and guarded to the latest-stable matrix entry, so a stale connector is surfaced as a failing check on the PR.
- The `update-yb-latest-stable` skill gains a connector-version-bump section documenting the source of truth, how to bump, and the freshness check.

No functional user-facing changes. The installer downloads the exact same connector jars from the exact same URLs as before — the URLs are now derived from `connector-versions.json` instead of being hardcoded. No CLI, configuration, or report changes.

- New unit tests in `versions/versions_test.go` cover the loader and the three accessors. They assert tag/series **format and non-emptiness** (not pinned literals), so a future connector bump that edits only `connector-versions.json` does not require touching the tests. `go test -tags unit ./versions/...` passes.
- The new freshness test was run live against the GitHub API: `go test -v -tags connector_latest_stable ./versions/...` → `PASS` (bundled `dz.2.5.2.yb.2025.2.3` is current). It fails cleanly (not a nil-panic) on a network error, skips on HTTP 403/429 (rate limit), and authenticates with `GITHUB_TOKEN` in CI.
- `go build ./...` and `go vet -tags connector_latest_stable ./versions/...` pass.
- Installer change verified with `bash -n` (syntax) and by confirming the derived URLs reconstruct the original URLs exactly; the connector download path is reached only in local-build mode (`package_debezium_server_local`), and release installs (`install_debezium_server_from_release`) are unaffected.

No.

No.


yb-voyager/versions/connector_latest_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// logicalConnectorTagRe matches logical connector tags of the form:
+//
+//	dz.<DZMAJOR>.<DZMINOR>.<DZPATCH>.yb.<YBYEAR>.<YBTRACK>.<COUNTER>
+//
+// It intentionally rejects gRPC tags (which contain ".yb.grpc.") and any other non-logical tags.
+var logicalConnectorTagRe = regexp.MustCompile(`^dz\.\d+\.\d+\.\d+\.yb\.(\d+)\.(\d+)\.(\d+)$`)
```
makalaaneesh: [AI] **Freshness regex misses counter-less series-GA tags — the exact "new YB series" case this check targets.**

`logicalConnectorTagRe` requires a trailing `.<COUNTER>` group: `^dz\.\d+\.\d+\.\d+\.yb\.(\d+)\.(\d+)\.(\d+)$`. But the actual `yugabyte/debezium` release history shows the *first* GA of a series has **no** counter, e.g. `dz.2.5.2.yb.2025.2`, `dz.2.5.2.yb.2025.1`, `dz.2.5.2.yb.2024.2`, `dz.2.5.2.yb.2024.1`. A 2026.1 series is already in flight (`dz.2.5.2.yb.2026.1.SNAPSHOT.1`), and its GA tag will almost certainly be `dz.2.5.2.yb.2026.1`.

Two consequences from this one limitation:

1. **The freshness check won't fire when a brand-new YB series ships its first (counter-less) connector.** `parseLogicalConnectorTag` returns `ok=false` for `dz.2.5.2.yb.2026.1`, so it is skipped in the release scan and `TestConnectorLatestStable` stays green — defeating the stated goal of "the bundled connector must be upgraded regularly to track new YugabyteDB series."
2. **If the bundled tag is ever set to a counter-less GA tag**, `parseLogicalConnectorTag(bundledTag)` returns `ok=false` and the test hits the `t.Fatalf("bundled logical connector tag ... does not match expected format")` path on a perfectly valid `connector-versions.json` — a false failure.

Suggested fix: make the counter optional and treat its absence as counter `0` (a series GA is the baseline, lower than `.1`, `.2`, ...). For example:

```go
var logicalConnectorTagRe = regexp.MustCompile(`^dz\.\d+\.\d+\.\d+\.yb\.(\d+)\.(\d+)(?:\.(\d+))?$`)
```

and in `parseLogicalConnectorTag`, default `counter` to 0 when `m[3] == ""`. (`.SNAPSHOT.N` pre-release tags still won't match because `SNAPSHOT` is non-numeric, which is the desired behavior.)

makalaaneesh (commented):

yb-voyager/versions/yb-cdc-connector-versions.json
```diff
@@ -0,0 +1,8 @@
+{
```
makalaaneesh, line 1: better name for the file? yb-cdc-connector-versions.json?

yb-voyager/versions/connector_latest_test.go
makalaaneesh: pls take a look if this makes sense

yb-voyager/versions/connector_latest_test.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// isNewer reports whether the candidate (ybYear2, ybTrack2, counter2) is strictly newer
+// than the reference (ybYear1, ybTrack1, counter1).
+//
+// Comparison order:
+//  1. YB series (ybYear, ybTrack) — the most significant dimension; a connector for a newer
+//     YB series is always considered newer regardless of counter.
+//  2. counter — within the same YB series, a higher counter is a newer patch release.
+func isNewer(ybYear1, ybTrack1, counter1, ybYear2, ybTrack2, counter2 int) bool {
```
makalaaneesh, line 71: ideally we should probably be using some version library (that we already use internally for other purposes) to compare versions, but it might not be such a big deal since this is just a test.

yb-voyager/versions/connector_latest_test.go
ShivanshGahlot: This was a good find. Fixed it. The latest 2026.1 connector has no trailing connector version attached. Older code would have missed this. Any subsequent releases in the 2026.1 connector series will probably have a version attached. Made the counter optional and returning it as 0 if it is missing.

ShivanshGahlot: 2026.1 connector got released yesterday. So now the test is failing on that too. Also I have modified the tests in the latest stable path so that both the Test Latest Stable YB version and Test Latest Connector tests run irrespective of whether the previous tests in that file failed or not.

yb-voyager/versions/connector_latest_test.go
ShivanshGahlot: I would just keep it as it is since its a test. If we ever require it in production code we can add to the library

yb-voyager/versions/yb-cdc-connector-versions.json
ShivanshGahlot: Done

makalaaneesh (approved):
LGTM

PR #3614, opened 2026-06-18
Add schema-snapshot capture library (schemasnapshot)

ShivanshGahlot:

This PR adds the `schemasnapshot` library — the capture-and-persist backend for the upcoming `schema drift-analysis` feature. It introduces a neutral, multi-engine data model (`SchemaSnapshot` with flat `Table`/`Column` lists keyed by `ObjectRef`), a provider-registry seam (`databases/<dbtype>` self-registering via `init()`) so engine-specific capture lives behind a common `SnapshotProvider` interface, and a PostgreSQL provider that captures tables (ordinary/partitioned/foreign) and columns — including partition parent/child links and legacy `INHERITS` linkage.

Captured snapshots are persisted to a new MetaDB `schema_snapshots` table via `SaveSnapshot`, with metadata-only **placeholder** rows recorded for failed capture attempts so a lifecycle moment still appears on the timeline. Capture runs inside a read-only `REPEATABLE READ` transaction so every catalog query sees one consistent point-in-time view, and the orchestrator (`Capture` / `CaptureAndSaveSnapshot`) stamps all header fields (version, source identity, captured-at, stable-identity) so providers only ever produce schema content.

This PR is **capture-only by design** — the diff/drift engine is deferred to a follow-up PR. An `Attrs` extension seam is declared on the model but intentionally left empty and uncompared in v1.

No user-facing changes. This is internal library scaffolding not yet wired to any command — no changes to the command line, configuration, installation process, or reports.

- **Unit tests** (`go test -tags unit ./src/schemasnapshot/...`): cover the data model and JSON round-trip, label/reason vocabulary validation (`ValidateLabelReason`), the persist flow against a mocked MetaDB (`SaveSnapshot`/`ListSnapshots`/`LoadSnapshotByName`, placeholder handling, version gating), and the `Capture`/`CaptureAndSaveSnapshot` orchestration using `sqlmock` (success, and capture-failure with placeholder on/off).
- **Integration tests** (`go test -tags integration ./src/schemasnapshot/databases/postgres/...`): run against a real `postgres:17` testcontainer — `TestCaptureAgainstLivePostgres` exercises ordinary/partitioned/foreign tables, columns, a dropped column, multi-level partitioning, and multiple inheritance; `TestCapturePersistRoundTrip` does a full Capture → Save → List → Load field-by-field round trip through MetaDB.

All unit and integration tests pass locally. No new end-to-end migtests are needed at this stage since the library is not yet wired to a command.

No.

Yes, but purely additive: it adds a new **MetaDB** table `schema_snapshots`. The table is created lazily (`CREATE TABLE IF NOT EXISTS` on first write) rather than in `initMetaDB`, so it appears regardless of which binary first created the metadb — including migrations started on an older binary and upgraded mid-flight. No existing table is altered, and older binaries simply ignore the new table, so there is no upgrade/downgrade compatibility issue.


yb-voyager/src/schemasnapshot/databases/postgres/loader.go
```diff
@@ -0,0 +1,285 @@
...
+	"strings"
+
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/constants"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/schemasnapshot"
+)
+
+// buildInPlaceholders returns a comma-separated list of n PostgreSQL positional
+// placeholders, e.g. n=3 → "$1,$2,$3", and a corresponding []interface{} of the
+// schema names. PostgreSQL (pgx / lib/pq) uses $N placeholders, not "?".
+func buildInPlaceholders(schemas []string) (string, []interface{}) {
```
makalaaneesh: [AI] **Empty `schemas` produces invalid `IN ()` SQL.**

`buildInPlaceholders(nil)` / `buildInPlaceholders([]string{})` returns an empty placeholder string, so the catalog templates render `... n.nspname IN () ...` (see `sqlLoadTablesFmt`, `sqlLoadTableLinksFmt`, `sqlLoadColumnsFmt` in `catalog_sql.go`). PostgreSQL rejects `IN ()` with a syntax error.

There is no guard in `Capture` / `TakeSnapshot` / `CaptureAndSaveSnapshot`, so a caller that passes an empty schema list (easy to do once this gets wired into a command) gets an opaque `syntax error at or near ")"` instead of a clear validation error. Per `yb-voyager/.cursor/BUGBOT.md` (Nil and Boundary Checks) and the root rules on early-return guards.

Failure case: `Capture(ctx, db, source, nil)` → SQL syntax error from `loadTables`.

Suggested fix: validate up front in `Capture` (or `TakeSnapshot`) and return a descriptive error, e.g.

```go
if len(schemas) == 0 {
    return nil, goerrors.Errorf("schemasnapshot: no schemas in scope for capture")
}
```

---

yb-voyager/src/schemasnapshot/persist.go
```diff
@@ -0,0 +1,205 @@
...
+	IsPlaceholder   bool      // true when snapshot_json is NULL (capture-attempt marker, no schema content)
+}
+
+// deriveName builds the primary-key name "{label}_{timestamp-at-second-precision}".
+func deriveName(label string, capturedAt time.Time) string {
+	return fmt.Sprintf("%s_%s", label, capturedAt.UTC().Format("20060102T150405Z"))
+}
+
+// schemasToString joins a string slice to a comma-separated string.
+func schemasToString(schemas []string) string {
```
makalaaneesh: [AI] **Comma-join of schema names corrupts schemas containing commas.**

`schemasToString` joins the schema list with `,` and `schemasFromString` splits on `,`. PostgreSQL allows quoted identifiers that contain commas (e.g. a schema literally named `a,b`), so a round-trip of `[]string{"a,b"}` comes back as `[]string{"a", "b"}` — two schemas instead of one. This is exactly the case-sensitive / special-character identifier hazard called out in the root `.cursor/BUGBOT.md` (Case-Sensitive Identifiers: object-name handling must not rely on ad-hoc string concatenation/splitting).

The corrupted value lands in the `schemas` column read back by `ListSnapshots`/`SnapshotMetadata.Schemas`. (Today this column is display/metadata only — the canonical list is also inside the snapshot JSON — which is why this is a warning rather than critical, but it should be fixed before the package is wired into a command.)

Failure case: a schema named `"a,b"` is persisted and listed as the two schemas `a` and `b`.

Suggested fix: store the schema list with a delimiter-safe encoding instead of naive comma join/split — e.g. JSON-encode the slice into the `schemas` column:

```go
func schemasToString(schemas []string) string {
    b, _ := json.Marshal(schemas)
    return string(b)
}

func schemasFromString(s string) []string {
    if s == "" {
        return nil
    }
    var out []string
    if err := json.Unmarshal([]byte(s), &out); err != nil {
        // fall back to legacy comma-split for rows written before this change
        return strings.Split(s, ",")
    }
    return out
}
```

---

makalaaneesh (commented):
[AI] Automated review of PR #3614 (schemasnapshot library) — 2 findings (0 critical, 2 warning). See inline comments. Both are latent: the package is not yet wired into any command, so these concern the library's public API robustness before it gets a caller.

---

yb-voyager/src/schemasnapshot/snapshot.go
```diff
@@ -0,0 +1,93 @@
...
+import "time"
+
+// SchemaSnapshot is a point-in-time capture of a database schema.
+// The per-object lists are flat; each record keys back to its parent via an ObjectRef field.
+// v1 carries tables and columns only.
+type SchemaSnapshot struct {
+	Version         int           `json:"version"`                    // snapshot JSON format version; gates parse-compatibility on load. Currently 1.
+	DatabaseType    string        `json:"database_type"`              // source engine; selects the provider that produced this snapshot, e.g. "postgresql".
+	DatabaseVersion string        `json:"database_version,omitempty"` // server version, truncated at the first space (e.g. "16.4"); display-only, never diffed.
+	StableIdentity  bool          `json:"stable_identity"`            // true when object IDs are stable enough for rename matching (PostgreSQL OIDs: always true).
```
makalaaneesh: - StableIdentity - seems unnecessary at this point. IF/when we start to support a database that does not give out object IDs that are stable, we can consider. 
    - let's verify again for postgres that this is true. I remember Karthik saying that OIDs might not be stable in pg. 
    - Curious: what happens if i drop/re-create a table with the same name?

yb-voyager/src/schemasnapshot/snapshot.go
```diff
@@ -0,0 +1,93 @@
...
+	CapturedAt      time.Time     `json:"captured_at"`                // time the snapshot was taken, in UTC.
+	CaptureSource   CaptureSource `json:"capture_source"`             // descriptive source coordinates (host/port/db/user/role) for display.
+	Schemas         []string      `json:"schemas"`                    // schemas in scope at capture time; fixed once captured and never widened/narrowed.
+	Series          string        `json:"series,omitempty"`           // the capture series (== persist label); empty until SaveSnapshot stamps it.
+	Reason          string        `json:"reason,omitempty"`           // the capture reason where the series carries one; empty otherwise.
```
makalaaneesh: can you give examples for Series/Reason? What is a label? HOw are they related? 
Unclear what purpose they solve.

yb-voyager/src/schemasnapshot/snapshot.go
```diff
@@ -0,0 +1,93 @@
...
+
+// CaptureSource holds the descriptive source coordinates for a snapshot.
+// It is display-only identity — never connection secrets.
+type CaptureSource struct {
+	DatabaseType string `json:"database_type"` // source engine; selects the provider, e.g. "postgresql".
+	Host         string `json:"host"`          // source host, for display.
+	Port         int    `json:"port"`          // source port, for display.
+	Database     string `json:"database"`      // source database name, for display.
+	User         string `json:"user"`          // connecting user, for display.
+	Role         string `json:"role"`          // logical role of this source; "source" in v1.
```
makalaaneesh: this could get confusing with user "role" in postgres. We should think of a better name. MigrationRole?

yb-voyager/src/schemasnapshot/snapshot.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// ObjectRef is the (schema, name) identity embedded in every per-object struct.
+type ObjectRef struct {
```
makalaaneesh, line 91: We should think about case sensitivity here. Can you discuss with @priyanshi-yb ? 
We have SchemaName/ObjectName for this purpose, perhaps we can reuse those?

yb-voyager/src/schemasnapshot/snapshot.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+// Table represents a table (ordinary, partitioned, or foreign).
+type Table struct {
+	ObjectRef                     // embedded (schema, name) identity of the table.
+	ID                string      `json:"id"`                           // pg_class.oid as a decimal string; used to match the table across snapshots (rename detection).
+	Kind              TableKind   `json:"kind"`                         // relation kind: ordinary, partitioned, or foreign.
+	PartitionParent   *ObjectRef  `json:"partition_parent,omitempty"`   // set on a partition child: the table it is a partition of; nil for non-partitions
+	PartitionChildren []ObjectRef `json:"partition_children,omitempty"` // set on a partitioned parent: its immediate partition children; nil otherwise
+	InheritsFrom      []ObjectRef `json:"inherits_from,omitempty"`      // legacy table inheritance (INHERITS): parent table(s) this table inherits from; can be multiple. Distinct from declarative partitioning.
```
makalaaneesh, line 115: If i remember correctly, partitioned tables and inherited tables share some implementation details in PG. i.e. partitioned tables also show up as inherited tables. Pls keep that in mind while populating this.

yb-voyager/src/schemasnapshot/snapshot.go
```diff
@@ -0,0 +1,93 @@
...
+	Attrs             Attrs       `json:"attrs,omitempty"`              // engine-specific extension attributes; declared as the seam but empty/uncompared in v1.
+}
+
+// Column represents a single column within a table.
+// ID encodes the parent table OID and the column attnum as "{parentTableOID}:{attnum}".
+type Column struct {
+	Table    ObjectRef `json:"table"`             // identity of the parent table this column belongs to.
+	ID       string    `json:"id,omitempty"`      // "{parentTableOID}:{attnum}"; matches the column across snapshots even after a rename.
+	Name     string    `json:"name"`              // column name.
+	DataType string    `json:"data_type"`         // rendered type via format_type(), e.g. "integer", "character varying(255)".
```
makalaaneesh: should we normalize this? (some common enums that we define?) today it's just source v source comparison, but in future we will have source v target, where the data types might need to be normalized before being compared.

yb-voyager/src/schemasnapshot/snapshot.go
```diff
@@ -0,0 +1,93 @@
...
+	NotNull  bool      `json:"not_null"`          // true when the column has a NOT NULL constraint.
+	Default  string    `json:"default,omitempty"` // default expression text (pg_get_expr); "" when the column has no default.
+	Attrs    Attrs     `json:"attrs,omitempty"`   // engine-specific extension attributes; declared as the seam but empty/uncompared in v1.
```
makalaaneesh: I'm ignoring the Attrs for now, because as i understand we're not using them at all. Will review it when we actually start using it. 

If it's not too much trouble, i'd recommend pulling out this code from the PR

yb-voyager/src/schemasnapshot/provider.go
```diff
@@ -0,0 +1,173 @@
...
+	// TakeSnapshot captures the schema for the given schemas via db and returns
+	// a populated *SchemaSnapshot. The header fields (CapturedAt, CaptureSource,
+	// StableIdentity, etc.) are stamped by the Capture orchestrator after this
+	// call returns, so the provider must not set them.
+	//
+	// Why: providers produce schema content only; the headers are capture-event
+	// metadata computed identically for every engine. Stamping them once in the
+	// orchestrator keeps them consistent (one Version/clock/source) and the
+	// provider never even receives the CaptureSource, so it can't set them wrong.
+	TakeSnapshot(ctx context.Context, db QueryExecutor, schemas []string) (*SchemaSnapshot, error)
```
makalaaneesh: Why do we need both a `QueryExecutor` and a `SnapshotProvider` ? 

why not just `TakeSnapshot(ctx, schemas) (*SchemaSnapshot, error) ` Let Postgres internally populate the snapshot however it wants.

yb-voyager/src/schemasnapshot/provider.go
```diff
@@ -0,0 +1,173 @@
...
+	providerRegistryMu.RLock()
+	factory, ok := providerRegistry[databaseType]
+	providerRegistryMu.RUnlock()
+	if !ok {
+		return nil, goerrors.Errorf("schemasnapshot: no provider registered for database type %q (import the matching databases/<dbtype> package)", databaseType)
+	}
+	return factory(), nil
+}
+
+// ─── Capture orchestrator ─────────────────────────────────────────────────────
```
makalaaneesh: let's put this in a separate file, because it's the entrypoint

yb-voyager/src/schemasnapshot/provider.go
```diff
@@ -0,0 +1,173 @@
...
+func RegisterProvider(databaseType string, factory func() SnapshotProvider) {
+	providerRegistryMu.Lock()
+	defer providerRegistryMu.Unlock()
+	providerRegistry[databaseType] = factory
+}
+
+// NewSnapshotProvider looks up the factory registered for databaseType and
+// returns a fresh provider. If no factory has been registered for the given
+// type it returns a clear error.
+func NewSnapshotProvider(databaseType string) (SnapshotProvider, error) {
```
makalaaneesh: do we really need the whole registry abstraction? why not a simple switch case that returns the provider based on the database type?

yb-voyager/src/schemasnapshot/provider.go
```diff
@@ -0,0 +1,173 @@
...
+//  2. Opens a read-only REPEATABLE READ transaction on db.
+//  3. Calls provider.TakeSnapshot inside that transaction.
+//  4. Commits on success (rollback on any error) — atomic, never partial.
+//  5. Stamps all header fields onto the returned snapshot.
+//
+// db and source are both needed and not derivable from each other: db is the
+// connection we query against; source is the descriptive identity (host/role/
+// type) recorded into the snapshot — database/sql can't be introspected for it,
+// and source.DatabaseType also selects the provider.
+func Capture(ctx context.Context, db *sql.DB, source CaptureSource, schemas []string) (*SchemaSnapshot, error) {
```
makalaaneesh: I'm a little iffy about passing `*sql.DB`. It seems like too much work for the caller. 
Let me think a little bit more and get back. We can discuss this on call. cc: @priyanshi-yb

makalaaneesh (commented):
First high-level pass. Yet to review some files (like the persist/lables/databases/metadb)

yb-voyager/src/schemasnapshot/snapshot.go
ShivanshGahlot: Yeah thats correct it does seem unnecessary at this point. The consumer code for this field will be in the diff part of the library. Currently since we only plan on working on YB and PG we can think of removing it. 

When we try to compare things across YB and PG, this anyways won't help us as OIDs won't be maintained across PG to YB. Over there we will anyways fall back to name based matching. We can remove this for now as long as we don't support anything other than PG/YB.

On the OID question: PG OIDs are stable for an object across RENAME/ALTER, but they do change on drop+recreate and on dump/restore / pg_upgrade. So "drop & recreate a table with the same name" gets a new OID and therefore the diff reports it as drop + add (which I feel can be considered correct as it is  genuinely a different object).

I'll verify this further and see if we need the StableIdentity field right now at all.

yb-voyager/src/schemasnapshot/snapshot.go
ShivanshGahlot: Label and Series are essentially the same value. We can unify the names here. 
Examples of labels:  export_schema, export_data_from_source_start, export_data_from_source_periodic, export_data_from_source_exit. These are the moments the tool captures a schema snapshot.  Series is that label stamped onto the snapshot at save time: SaveSnapshot does snap.Series = label. 

Reason is an optional sub-qualifier within a label
- export_schema → no reason
- export_data_from_source_start → initial | resume | clean_restart
- export_data_from_source_periodic → no reason
- export_data_from_source_exit → cutover | complete | interrupt | error

makalaaneesh: let's discuss this on call. I understand the need, but i want to explore just keeping one field - label.

yb-voyager/src/schemasnapshot/postgres.go
```diff
@@ -0,0 +1,285 @@
+// Copyright (c) YugabyteDB, Inc.
```
makalaaneesh, line 1: @priyanshi-yb need your help in reviewing the queries and the partition logic here

yb-voyager/src/schemasnapshot/persist.go
```diff
@@ -0,0 +1,205 @@
...
+// SaveSnapshot stamps label/reason into snap.Series/Reason, serializes snap to JSON,
+// and persists it under the name "{label}_{second-precision-timestamp}".
+// It mutates snap in place.
+// Returns the derived name on success.
+func SaveSnapshot(mdb *metadb.MetaDB, snap *SchemaSnapshot, label, reason string) (string, error) {
+	if err := ValidateLabelReason(label, reason); err != nil {
+		return "", err
+	}
+
+	// Stamp Series and Reason onto the snapshot before serializing.
```
makalaaneesh: i don't quite understand why this happens at the persist layer. 
persist should just be about storing it/loading it from metadb, no?

yb-voyager/src/schemasnapshot/persist.go
```diff
@@ -0,0 +1,205 @@
...
+
+// Sentinel errors returned by LoadSnapshotByName.
+var (
+	ErrSnapshotNotFound    = errors.New("snapshot not found")
+	ErrPlaceholderSnapshot = errors.New("snapshot is a placeholder: no schema content captured")
+)
+
+// SnapshotMetadata holds the header columns for a persisted snapshot.
+// It is returned by ListSnapshots and never contains the full schema content.
+type SnapshotMetadata struct {
```
makalaaneesh: this also seems like it should be generated outside the "persist" layer. 
You seem to have a `SchemaSnapshot` and `SchemaSnapshotMetadata`. These two are independent of "how and where you persist" and IMO should be populated before you enter the "persist" layer.

yb-voyager/src/metadb/schemaSnapshot.go
```diff
@@ -0,0 +1,188 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+// schemaSnapshotsTableName is the SQLite table that persists schema snapshots.
+const schemaSnapshotsTableName = "schema_snapshots"
+
+// SchemaSnapshotRow is the primitive row stored in schema_snapshots.
+// snapshot_json is NULLable: NULL means the row is a placeholder (no schema content captured).
+// This struct intentionally contains no schemadiff types — the metadb package must not import schemadiff.
+type SchemaSnapshotRow struct {
```
makalaaneesh, line 38: this separation seems clean IMO. 
you deal with the row structs in the metadb layer. 
you convert them to proper schemaSnapshot/schemaSnapshotMetadata in the `schemasnapshots` layer.

yb-voyager/src/schemasnapshot/snapshot.go
ShivanshGahlot: Yep agreed this can be ambiguous. There is a 'Side' field in SnapshotMetadata, so probably I'll just unify them to use the same word 'Side' i.e. rename here.

yb-voyager/src/schemasnapshot/snapshot.go
ShivanshGahlot: Yeah. Makes sense. Not sure if we can reuse SchemaName/ObjectName. Let me see what can be done here.

yb-voyager/src/schemasnapshot/snapshot.go
ShivanshGahlot: The loader handles this explicitly. 
linkTableHierarchy (loader.go:148) routes each pg_inherits row by relispartition (is_partition in the catalog query), so declarative partitions populate PartitionParent/PartitionChildren and legacy INHERITS populates InheritsFrom/InheritedBy. This way the two never cross.

There are unit tests too for this: TestTakeSnapshotPartitionLinkage, TestTakeSnapshotInheritanceLinkage

yb-voyager/src/schemasnapshot/snapshot.go
ShivanshGahlot: Yep this can be done. Will add a normalization step around this. We can store this DataType as it is and normalize at diff time probably.

yb-voyager/src/schemasnapshot/snapshot.go
ShivanshGahlot: Yeah let me see if this can be removed rn.

yb-voyager/src/schemasnapshot/provider.go
ShivanshGahlot: Do you mean like we pass a sql.Tx or sql.DB directly to TakeSnapshot and let it execute? If so then I have added a comment why I made it such. QueryExecutor can sure be removed. The main benefits are these as mentioned in the comment:

```
// Why an interface and not *sql.Tx directly:
//   - Least privilege: providers get read-only access; they cannot Commit,
//     Rollback, Exec, or Close. Capture stays the sole owner of the tx lifecycle.
//   - Trivially fakeable in tests (sqlmock/stub) — no real DB or tx needed.
```

Or do you mean that TakeSnapshot will create its own connection as a whole internally for each DB type?

yb-voyager/src/schemasnapshot/provider.go
ShivanshGahlot: I have added this to avoid having an import cycle between databas and schemasnapshot as we had discussed. 

A switch that returns postgres.New() has to import the provider packages, and those already import the core to implement SnapshotProvider so the switch can't live in core. 

It could live in databases/all, but Capture (in core) calls the resolver, and core importing all reintroduces the cycle.

yb-voyager/src/schemasnapshot/provider.go
ShivanshGahlot: Yep makes sense

yb-voyager/src/schemasnapshot/provider.go
ShivanshGahlot: We can pass the connection params I guess and then open a connection here itself.

ShivanshGahlot: We might need to duplicate the connection logic or do connection construction. It might be easier to just expose the db connection we are already creating in our source layer and pass it to this Capture function.

yb-voyager/src/schemasnapshot/persist.go
ShivanshGahlot: Hmm the idea was the label/series is available with the caller. Sometimes the caller just wants to capture the snapshot and not attach a label to it like in capturing the snapshot of the live current database. So I had kept this stamping in the persist layer given that we will only call persist when we have a label.

But I guess the easier way is to have a label defined for all the caller sites(even for the live capture). And the Capture itself can stamp these values in the snapshot. And then the only job the persist layer has is to store it.

yb-voyager/src/schemasnapshot/persist.go
ShivanshGahlot: This is just a struct defined to return values from "ListSnapshots" . We are not generating it while saving the snapshot. Its only used for returning a list of snapshots without exposing the entire schema content they have. More like giving a header only view of all the snapshots available in the metadb

yb-voyager/src/metadb/schemaSnapshot.go
ShivanshGahlot: Yep

ShivanshGahlot: Though I now feel it can be made a bit better. Right now the queries are fetching the entire snapshot(I meant he actual snapshot content of tables/columns etc) irrespective of whether we just need to list the snapshots(not load the actual snapshot) or actually load the content of a paritcular snapshot. In case of list snapshot we check whether the snapshot content returned is null or not and using that we populate isPlaceHolder int the metadata struct.

Instead of that I feel like it might be a better design to have a separate struct for when we just want to get the list of snapshots. There we will populate isPlaceholder directly using the query and the query wont need to fetch the entire snapshot captured content.

What do you think?

yb-voyager/src/metadb/schemaSnapshot.go
```diff
@@ -0,0 +1,188 @@
...
+	return nil
+}
+
+// InsertSchemaSnapshot inserts a full snapshot row (snapshot_json is set).
+// Returns an error if the name (primary key) already exists.
+func (m *MetaDB) InsertSchemaSnapshot(row SchemaSnapshotRow) error {
+	if err := m.createSchemaSnapshotsTable(); err != nil {
+		return err
+	}
+	capturedAtStr := row.CapturedAt.UTC().Format(time.RFC3339Nano)
```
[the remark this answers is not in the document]
ShivanshGahlot: Fixed

yb-voyager/src/schemasnapshot/provider.go
```diff
@@ -0,0 +1,172 @@
...
+	provider, err := NewSnapshotProvider(source.DatabaseType)
```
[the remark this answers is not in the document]
ShivanshGahlot: Fixed

yb-voyager/src/schemasnapshot/databases/postgres/loader.go
ShivanshGahlot: Fixed

yb-voyager/src/schemasnapshot/persist.go
ShivanshGahlot: Fixed

yb-voyager/src/schemasnapshot/snapshot.go
ShivanshGahlot: Add a TODO around it

yb-voyager/src/schemasnapshot/provider.go
ShivanshGahlot: Moved the database related files out of the database package into the schemasnapshot package and using the switch case there

yb-voyager/src/schemasnapshot/snapshot.go
ShivanshGahlot: Added a TODO for this. Picking it up in a subsequent PR.

yb-voyager/src/schemasnapshot/snapshot.go
ShivanshGahlot: Removed StableIdentity

makalaaneesh (commented):
[AI] Automated review — 2 findings (0 critical, 2 warning). See inline comments. (3 suggestions dropped: stale field comment, per-insert CREATE TABLE logging, silent relkind default.)

---

yb-voyager/src/schemasnapshot/capture.go
```diff
@@ -0,0 +1,144 @@
...
+// if false it returns the capture error without writing anything.
+func CaptureAndSaveSnapshot(ctx context.Context, db *sql.DB, mdb *metadb.MetaDB,
+	req CaptureRequest) (name string, err error) {
+
+	snap, captureErr := Capture(ctx, db, req.CaptureParams)
+	if captureErr != nil {
+		if req.PlaceholderOnFailure {
+			// placeholder dbVersion is "" (the version probe was part of the failed capture).
+			h := newHeader(req.CaptureParams, time.Now().UTC(), "", true)
+			_, _ = SavePlaceholder(mdb, h)
```
makalaaneesh: [AI] **`SavePlaceholder` error is silently swallowed.**

`CaptureAndSaveSnapshot` discards both return values of `SavePlaceholder` here. Returning the original capture error is correct, but if the placeholder insert *also* fails (e.g. a primary-key collision on a same-second placeholder, or a disk/IO error), that failure is invisible — the timeline marker is silently lost.

`yb-voyager/.cursor/BUGBOT.md` requires: "Never silently swallow errors … log it with sufficient context."

Suggested fix:
```go
if _, perr := SavePlaceholder(mdb, h); perr != nil {
    log.Warnf("schemasnapshot: failed to write placeholder marker for label %q: %v", req.Label, perr)
}
```

---

yb-voyager/src/metadb/schemaSnapshot.go
```diff
@@ -0,0 +1,218 @@
...
+func (m *MetaDB) createSchemaSnapshotsTable() error {
+	query := fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s (
+		name             TEXT    PRIMARY KEY,
+		label            TEXT    NOT NULL,
+		reason           TEXT    NOT NULL DEFAULT '',
+		side             TEXT    NOT NULL,
+		captured_at      TEXT    NOT NULL,
+		database_version TEXT    NOT NULL DEFAULT '',
+		schemas          TEXT    NOT NULL DEFAULT '',
+		is_placeholder   INTEGER NOT NULL DEFAULT 0,
```
makalaaneesh, line 88: [AI] **`is_placeholder` column duplicates `snapshot_json IS NULL` (two sources of truth).**

Placeholder-ness is now persisted twice: implicitly via `snapshot_json` nullness, and explicitly in this `is_placeholder` column. `LoadSnapshotByName` (persist.go) branches solely on the column, so the two can drift:

- `is_placeholder=1` with a non-NULL blob → `ErrPlaceholderSnapshot` is returned, hiding real content.
- `is_placeholder=0` with a NULL blob → `DecodeSnapshot("")` fails with "unexpected end of JSON input".

The list-query goal (avoid materializing the blob) does not require a second column — `snapshot_json IS NULL` evaluates in SQL without returning the TEXT. Note the test comment at `schemaSnapshot_test.go:111-114` actually describes the `snapshot_json IS NULL` approach, which no longer matches the code — evidence of the drift risk.

Suggested fix: drop the column and compute the flag in both `ListSchemaSnapshots` and `GetSchemaSnapshotByName`:
```sql
SELECT name, label, reason, side, captured_at, database_version, schemas,
       (snapshot_json IS NULL) AS is_placeholder
FROM   schema_snapshots ORDER BY captured_at ASC;
```
and fix the stale test comment. If the column is intentional, document that `snapshot_json IS NULL` is the invariant it must uphold.

---

yb-voyager/src/metadb/schemaSnapshot.go
```diff
@@ -0,0 +1,218 @@
...
+// capturedAtLayout is a fixed-width UTC timestamp layout so that captured_at TEXT
+// values sort lexicographically == chronologically (RFC3339Nano strips trailing zeros
+// producing variable-length strings that break lexicographic ordering).
+const capturedAtLayout = "2006-01-02T15:04:05.000000000Z"
+
+// SchemaSnapshotRow is the primitive row stored in schema_snapshots.
+// snapshot_json is NULLable: NULL means the row is a placeholder (no schema content captured).
+// This struct intentionally contains no schemadiff types — the metadb package must not import schemadiff.
+type SchemaSnapshotRow struct {
+	Name            string         // PRIMARY KEY: "{label}_{timestamp-at-second-precision}"
```
makalaaneesh: Slightly unsure about the PK. It feels like almost all the header fields make up the PK.  Let's discuss this.

makalaaneesh: Can we define a composite PK ? `(side, label, captured_at)` ?

ShivanshGahlot: Done, made this composite key and added UNIQUE to name

yb-voyager/src/metadb/schemaSnapshot.go
ShivanshGahlot: Ignoring this

yb-voyager/src/schemasnapshot/capture.go
ShivanshGahlot: Fixed

yb-voyager/src/schemasnapshot/postgres.go
```diff
@@ -0,0 +1,341 @@
...
+FROM   pg_attribute  a
+JOIN   pg_class      c ON c.oid = a.attrelid
+JOIN   pg_namespace  n ON n.oid = c.relnamespace
+LEFT   JOIN pg_attrdef d ON d.adrelid = a.attrelid AND d.adnum = a.attnum
+WHERE  a.attnum > 0
+  AND  NOT a.attisdropped
+  AND  c.relkind IN ('r','p','f')
+  AND  n.nspname IN (%s)
+ORDER  BY a.attrelid, a.attnum
+`
```
priyanshi-yb, line 91: `sqlLoadTableLinksFmt` and `sqlLoadColumnsFmt`, why don't we have these queries filtering on the tables fetched from the first query? I think not right now, but in future it will be required if we have table-list fiter etc..

yb-voyager/src/schemasnapshot/postgres.go
```diff
@@ -0,0 +1,341 @@
...
+					ObjectRef{Schema: lnk.parentSchema, Name: lnk.parentName},
+				)
+			}
+			if parentInScope {
+				tables[parentIdx].InheritedBy = append(
+					tables[parentIdx].InheritedBy,
+					ObjectRef{Schema: lnk.childSchema, Name: lnk.childName},
+				)
+			}
+		}
```
priyanshi-yb, line 252: With this logic, I feel the `tables` will only store the immediate parent of the partition and immediate children of the partitioned table.
Shouldn't we have recursive calls to update all the root/partitions in case of multi-level partitions?

ShivanshGahlot: You are correct. We only store the immediate parent/children here. But we capture all the tables in a partition tree separately as there own Table struct with their immediate parent/children. So we don't really lose the tree. We can construct it any time by just traversing through the Immediate children and then their children etc.

yb-voyager/src/schemasnapshot/postgres.go
ShivanshGahlot: So the idea is to capture the entire schema snapshot with all the tables regardless. If we filter on tables here, we wont be able to capture any newly added tables or table renames etc. So I have delegated this filtering responsibility to the Diff engine in the next PR. Over there in the future we will also have ignore rules etc so it made sense to keep the logic for filtering out the table list there itself. The command will have the table list flag from the get go. Just that the filtering will be done in the Diff library. We will find out all the diffs on all the tables. Same goes for any new objects that might need to be suppressed when there table is not in the table-list like Indexes etc. We will filter those objects out too over there(in the future).

makalaaneesh (approved):
LGTM

priyanshi-yb (approved):
LGTM

PR #3623, opened 2026-06-26
Add yugabytedb-amp target-db-type for offline migration into YugabyteDB AMP

makalaaneesh:

This PR adds a new target database type **`yugabytedb-amp`** (`--target-db-type yugabytedb-amp`) enabling offline migration (export-schema → import-schema → export-data → import-data, plus finalize-schema-post-data-import) from PostgreSQL into **YugabyteDB AMP (yb-amp)** — a stateless, patched PostgreSQL 17 compute backed by YugabyteDB storage that is PostgreSQL-compatible on the wire.

Because yb-amp speaks plain PostgreSQL (standard heap DDL, `COPY`, standard session GUCs) and exposes none of YugabyteDB's cluster control surface (`yb_servers()`, tserver metrics, colocation/tablets, adaptive parallelism), the new `TargetYugabyteDBAmp` driver **embeds the existing `TargetPostgreSQL` driver** and layers on only AMP-specific identity validation; the import code paths treat `yugabytedb-amp` as PostgreSQL-compatible and skip the YugabyteDB-only machinery.

Key design decisions:
- **Engine identity verified at driver `Init()`**, mutually exclusive among the three PG-wire-compatible engines: real YB (`yb_servers()`), yb-amp (`yb_amp.*` GUCs), and plain PostgreSQL. A mistyped `--target-db-type` fails fast naming the correct value.
- **YugabyteDB-only machinery gated by target type** — colocation-aware task picker, table-type/colocation lookup, target health monitoring, the `yb_is_database_colocated()` colocation prompt, and the YB-version analyze report all run only for a real YugabyteDB target; yb-amp uses the sequential importer and plain-PG paths.
- **Adaptive parallelism default is per-target** — Balanced for YugabyteDB, Disabled for every other target (yb-amp and the PostgreSQL fall-forward/fall-back targets, none of which have the cluster control API it needs). This lets `--parallel-jobs` be used for those targets without explicitly disabling adaptive parallelism.
- **Pristine schema for amp via consistent backups.** export-schema transforms the exported DDL for a YB target (colocation, `SET yb_*` sharding), which is invalid on yb-amp's PG17. A single helper `EnsurePlainBackup` is now called by every step that mutates a schema file (the colocation recommender and the file transformers) *before* it mutates, producing one consistent, skip-if-exists `backup_<file>` that always holds the pristine pg_dump DDL. import-schema imports that pristine `backup_<file>` for a yb-amp target (falling back to `object.sql` when none exists). This replaces an earlier `.orig`/MetaDB-flag scheme and is robust across all `--skip-colocation-recommendations` / `--skip-performance-recommendations` combinations.
- **Strict guardrails** reject flags that don't apply to yb-amp (`--target-endpoints`, `--use-public-ip`, `--enable-upsert`, and `--on-primary-key-conflict` other than `ERROR-POLICY`), and restrict a yb-amp target to a PostgreSQL source.

Live migration, fall-forward/fall-back, and assess-migration remain explicit non-goals.

- **Command line:** New `--target-db-type` flag on `import schema`, `import data` / `import data to-target`, and `finalize-schema-post-data-import`; values `yugabytedb` (default) and `yugabytedb-amp`. For `yugabytedb-amp`: `--target-db-port` is required (no canonical default); the source must be PostgreSQL; and `--target-endpoints`, `--use-public-ip`, `--enable-upsert`, `--on-primary-key-conflict <non-ERROR-POLICY>`, `--adaptive-parallelism balanced|aggressive`, and `--adaptive-parallelism-max` are rejected with actionable errors. A target type that doesn't match the actual endpoint fails fast. The flag is intentionally NOT exposed on `import data file` or `compare-performance` (always real YugabyteDB). TLS works via the existing `--target-ssl-*` flags.
- **Configuration:** `--target-db-type` is settable via the config file (`target: db-type:`); parallelism flags behave identically whether set via CLI or config.
- **Installation / reports:** No changes.

- **Unit:** `go build ./...`, `go vet`, and `go test -tags unit ./cmd/... ./src/tgtdb/... ./src/query/sqltransformer/... ./src/metadb/... ./src/utils/sqlname/...` all pass. The colocation task-picker unit tests were updated to set the YugabyteDB target type (they exercise the YB-only colocation path). No new dedicated unit tests were added for the amp paths.
- **Manual end-to-end (PostgreSQL → yb-amp):** full export-schema → import-schema → export-data → import-data verified — row counts match source, 0 FK orphans, view + indexes + constraints + sequences created, sequences correctly resumed; finalize-schema-post-data-import verified.
- **Backup/pristine correctness:** export→import-into-amp across the matrix {no assess-migration, assess-migration} × {default, `--skip-performance-recommendations`, `--skip-colocation-recommendations`, both}. In all 6 runs the file imported into yb-amp is byte-identical (the pristine pg_dump), free of colocation/`SET yb_*`, and imports cleanly; the YugabyteDB path is unaffected (transformed `object.sql` still carries colocation/sharding).
- **Guardrails** exercised via CLI and config file (wrong target type vs endpoint; the rejected amp flags; `--parallel-jobs` works for amp without disabling adaptive).
- **TLS:** validated a full PG → yb-amp migration over `--target-ssl-mode require` (TLSv1.3).
- **Gap / follow-up:** no automated integration tests for the `yugabytedb-amp` path yet; `integration_voyager_command`-style coverage would be a valuable follow-up.

No payload shape changes. Callhome reuses the PostgreSQL-shaped target info; the existing target-db-type field can now carry the value `yugabytedb-amp`. Payload version not incremented (no structural change).

No. export-schema now writes a `backup_<file>` sidecar in the schema directory (replacing the previous `<file>.orig`) before transforming; these are additive, informational backups that older binaries ignore (import reads the primary `object.sql`), so there is no upgrade concern. No MetaDB / name-registry / data-file / callhome / status-json shapes changed.

---

| Component        

---

### DIFF
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -681,21 +681,20 @@ func validateFFDBSchemaFlag() {
 }
 
 // defaultAdaptiveParallelismMode returns the adaptive-parallelism mode to use when the
-// user did NOT pass --adaptive-parallelism. It is per-target:
-//   - YUGABYTEDB: Balanced — adaptive parallelism uses the YB cluster control API
-//     (yb_servers(), tserver metrics) and is the recommended default.
-//   - YUGABYTEDB_AMP: Disabled — yb-amp is a stateless PG17 compute with no YB cluster
-//     control API, so adaptive parallelism is not available; --parallel-jobs controls
-//     import parallelism instead.
+// user did NOT pass --adaptive-parallelism. Adaptive parallelism relies on the YugabyteDB
+// cluster control API (yb_servers(), tserver metrics), so it is the recommended default
+// (Balanced) ONLY for a real YugabyteDB target. Every other target — yb-amp (stateless
+// PG17 compute) and the PostgreSQL fall-forward/fall-back targets — has no such API, so it
+// defaults to Disabled; --parallel-jobs controls import parallelism there.
 //
-// Defaulting amp to Disabled is also what lets a user pass --parallel-jobs for amp without
-// having to also pass --adaptive-parallelism disabled (validateParallelismFlags only
-// conflicts --parallel-jobs with an *enabled* adaptive mode).
+// Defaulting non-YB targets to Disabled is also what lets a user pass --parallel-jobs for
+// them without having to also pass --adaptive-parallelism disabled (validateParallelismFlags
+// only conflicts --parallel-jobs with an *enabled* adaptive mode).
 func defaultAdaptiveParallelismMode(targetDBType string) types.AdaptiveParallelismMode {
-	if targetDBType == YUGABYTEDB_AMP {
-		return types.DisabledAdaptiveParallelismMode
+	if targetDBType == YUGABYTEDB {
+		return types.BalancedAdaptiveParallelismMode
 	}
-	return types.BalancedAdaptiveParallelismMode
+	return types.DisabledAdaptiveParallelismMode
 }
 
 func validateParallelismFlags(cmd *cobra.Command) {
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -198,11 +198,19 @@ func registerCommonImportFlags(cmd *cobra.Command) {
 	BoolVar(cmd.Flags(), &tconf.RunGuardrailsChecks, "run-guardrails-checks", true, "Run guardrails checks during import")
 }
 
-func registerTargetDBConnFlags(cmd *cobra.Command) {
+// registerTargetDBTypeFlag registers --target-db-type. It is intentionally
+// NOT part of registerTargetDBConnFlags: the choice of target engine is only
+// meaningful for the commands that import schema/data into the target
+// (import schema, import data / ...toTarget, finalize-schema-post-data-import).
+// Commands like import-data-file and compare-performance always target a real
+// YugabyteDB, so they don't expose it.
+func registerTargetDBTypeFlag(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&tconf.TargetDBType, "target-db-type", "",
 		fmt.Sprintf("type of the target database to import into. Supported values: %s (default), %s (YugabyteDB AMP — a PostgreSQL-compatible compute over YugabyteDB storage)",
 			YUGABYTEDB, YUGABYTEDB_AMP))
+}
 
+func registerTargetDBConnFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&tconf.Host, "target-db-host", "127.0.0.1",
 		"host on which the YugabyteDB server is running")
 
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -102,7 +102,7 @@ func validateImportFlags(cmd *cobra.Command, importerRole string) error {
 	case SOURCE_DB_IMPORTER_ROLE:
 		getSourceDBPassword(cmd)
 	}
-	validateParallelismFlags()
+	validateParallelismFlags(cmd)
 
 	return nil
 }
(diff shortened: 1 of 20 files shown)


yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -75,6 +96,12 @@ func importSchemaInternal(exportDir string, importObjectList []string,
 	return nil
 }
 
+// originalSchemaFilePath returns the backup_<base> sibling that export-schema
+// writes alongside any schema file it transforms (tables, indexes, mviews).
+func originalSchemaFilePath(filePath string) string {
+	return filepath.Join(filepath.Dir(filePath), "backup_"+filepath.Base(filePath))
```
makalaaneesh, line 95: [AI] **`originalSchemaFilePath` resolves the wrong backup when colocation recommendations were applied — yb-amp will import YugabyteDB-specific DDL.**

This function returns `backup_<file>`, and `useOriginalSchemaFiles` is gated on `SchemaOptimizationsApplied`, which `exportSchema.go:240` sets to `assessmentRecommendationsApplied || !skipPerfOptimizations`. But export-schema uses **two different backup conventions**, and they are layered:

1. `applyMigrationAssessmentRecommendations()` (colocation) runs first (`exportSchema.go:199`). For the colocation path, `applyShardedTablesRecommendation` renames the true plain-PG original to `<file>.orig` (`exportSchema.go:578`, e.g. `table.sql.orig`) and writes the colocated DDL (`WITH (colocation = ...)`) into `table.sql`. It sets `assessmentRecommendationsApplied = true`.
2. `applyTableFileTransformations()` runs next (`exportSchema.go:211`). `TableFileTransformer.Transform` copies the **current** `table.sql` — which now contains the colocated DDL — into `backup_table.sql` (`sqltransformer/sqlfiletransformer.go:198-203`).

So when colocation recommendations are applied, `backup_table.sql` still carries the YB-only `WITH (colocation = ...)` clauses; the genuine plain-PG original lives in `table.sql.orig`. Reading `backup_table.sql` for a yb-amp (PostgreSQL-compatible) target hands PostgreSQL DDL it cannot parse.

MVIEW makes it worse: `MviewFileTransformer.Transform` is a no-op and never creates `backup_mview.sql`, so for mviews `originalSchemaFilePath` points at a non-existent file, `FileOrFolderExists` is false, and import silently falls back to the optimized `mview.sql`.

Failure case: offline migration with assessment colocation recommendations applied (the default — `--skip-colocation-recommendations` not set), then `import schema --target-db-type yugabytedb-amp`. yb-amp rejects the colocation clause / silently imports the wrong DDL.

Suggested fix: make the original-file resolution mirror the actual backup conventions — prefer `<file>.orig` (colocation original) when present, then `backup_<file>` (perf-transform original), and only treat a file as "has a plain original" when one of those exists. Consider recording the authoritative pre-transformation path in the MSR at export time instead of reconstructing the name at import time, so the two conventions can't drift.

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -1238,3 +1238,69 @@ func (pg *TargetPostgreSQL) GetEnabledTriggersAndFks() (enabledTriggers []string
...
+		schemaNames = append(schemaNames, schemaName)
+	}
+	return schemaNames, rows.Err()
+}
+
+func (pg *TargetPostgreSQL) GetAllTableNamesRaw(schemaName string) ([]string, error) {
+	query := fmt.Sprintf(`SELECT table_name
+			  FROM information_schema.tables
+			  WHERE table_type = 'BASE TABLE' AND
+			        table_schema = '%s';`, schemaName)
```
makalaaneesh, line 1271: [AI] **New `GetAll*Raw` methods interpolate `schemaName` into SQL via `fmt.Sprintf` instead of using a parameterized query.**

Per `yb-voyager/.cursor/BUGBOT.md` (SQL Queries): "Use parameterized queries or prepared statements rather than `fmt.Sprintf` with user-supplied values." `GetAllTableNamesRaw` and `GetAllSequencesRaw` build `... table_schema = '%s'` / `... schemaname = '%s'` by string-formatting the caller-supplied `schemaName`.

The schema name ultimately derives from `--target-db-schema`, so a value containing a single quote breaks the query and is an injection vector. These methods are newly added to the PostgreSQL driver by this PR (now reused by the yb-amp driver via embedding), so they are worth fixing at the point of introduction even though `TargetYugabyteDB` has the same legacy pattern.

Suggested fix: bind the schema name as a parameter, e.g.

```go
query := `SELECT table_name FROM information_schema.tables
          WHERE table_type = 'BASE TABLE' AND table_schema = $1`
rows, err := pg.QueryParameterized(query, schemaName) // or the driver's parameterized Query equivalent
```

or, at minimum, quote/escape the literal with the pgx identifier/literal helpers. Apply the same to `GetAllSequencesRaw`.

makalaaneesh (commented):
[AI] Automated review of PR #3623 (yugabytedb-amp target-db-type) — 2 findings (1 critical, 1 warning). See inline comments. Generated by Claude Code (Opus 4.8); a human reviewer should confirm.

yb-voyager/cmd/importSchemaYugabyteDB.go
makalaaneesh: fixed

yb-voyager/src/tgtdb/yugabytedb_amp.go
```diff
@@ -0,0 +1,80 @@
...
+		return err
+	}
+	return amp.validateAmpTarget()
+}
+
+// validateAmpTarget confirms the connected compute is a yb-amp endpoint
+// rather than a vanilla PostgreSQL or YugabyteDB server, so a user who
+// mistypes --target-db-type gets a clear, actionable error instead of a
+// subtly wrong migration.
+func (amp *TargetYugabyteDBAmp) validateAmpTarget() error {
```
priyanshi-yb, line 59: If we can use these functions to directly set the DB type internally, whether it's AMP or not, it will be a better UX where the user doesn't have to pass anything. Voyager will decide how to do the import based on the YB setup.

yb-voyager/cmd/import.go
```diff
@@ -432,20 +506,39 @@ func validateTargetSchemaFlag() {
...
 		}
 		return
-	} else if tconf.TargetDBType != POSTGRESQL {
+	} else if tconf.TargetDBType != POSTGRESQL && tconf.TargetDBType != YUGABYTEDB_AMP {
 		splits := strings.Split(tconf.SchemaConfig, ",")
 		if len(splits) > 1 {
 			utils.ErrExit("Error --target-db-schema flag can only contain one schema name. Got: %s", tconf.SchemaConfig)
 		}
 	}
 }
```
priyanshi-yb, line 524: couple more config params that I think are not covered 
1. batch-size
2. error-policy

priyanshi-yb (approved):
LGTM, can we please run oracle tests once if not already done.

yb-voyager/src/tgtdb/yugabytedb_amp.go
makalaaneesh: will address this in a subsequent PR after discussion

makalaaneesh:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/8379/

PR #3629, opened 2026-06-30
Add root-PK partitioned coverage to iteration resumption test; rename partition-iteration-test to use-partition-root-false-test

Abhi-yb:

The live-migration iteration resumption test (`iteration-test`) previously ran only flat (non-partitioned) tables via the shared `common-sql/init.sql`. This PR makes it cover **"normal" partitioned tables with the PRIMARY KEY on the partition root**, imported with `--use-partition-root=true` so CDC routes through the root. Together with the renamed `use-partition-root=false` test, this gives the resumption suite coverage of both partitioned import modes — the scenario that surfaces the read-restart behaviour observed on YB 2025.2.4 / 2026.1.

`iteration-test` now uses a dedicated `init.sql` (RANGE / LIST / HASH / multilevel / multi-column-range / shared-sequence families, all root-PK) instead of the shared flat schema, with generator `column_overrides` so random values route across the real partitions, and DEFAULT partitions as a safety net. `customers.arr` is `numeric(10,2)` (fixed scale) to avoid row-hash false-positives from CDC trailing-zero normalization on unbounded numeric. Flat-table coverage is unchanged — it still lives in the four other resumption tests that share `common-sql/init.sql`.

`partition-iteration-test` is renamed to `use-partition-root-false-test` to make its purpose explicit (no PK on root, PK on children, `--use-partition-root=false`). The two tests now form a matched pair: root-PK / `true` vs child-PK / `false`.

No user-facing changes. This PR only touches end-to-end migration test scenarios under `migtests/` (no command line, configuration, installation, or report changes).

Ran the `iteration-test` scenario end-to-end via `scripts/live-migration-resumption/orchestrator.py` against **YugabyteDB 2026.1.0.0** (PostgreSQL source). Both resumption iterations completed, both cutover directions (to-target and to-source), and all `row_count_validations`, `row_hash_validations`, and `validate_archive_changes` stages passed. The new partitioned DDL was also validated against PostgreSQL 17. No Go unit/integration code is touched — changes are test-scenario files only.

No.

No.

### DIFF
--- a/migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
+++ b/migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
@@ -45,7 +45,7 @@ generator_source:
     generator:
       schema_name: public
       manual_table_list: []
-      exclude_table_list: [cutover_table]
+      exclude_table_list: [cutover_table, cust_active, cust_arr_small, cust_arr_large, cust_part11, cust_part12, cust_part21, cust_part22, cust_other]
       num_iterations: -1
       wait_after_operations: 1000
       wait_duration_seconds: 1
@@ -58,6 +58,30 @@ generator_source:
       delete_rows: 1
       insert_max_retries: 5
       update_max_retries: 1
+      # Pin region/sale_date/status to valid partition values so generated rows route
+      # across the real partitions (DEFAULT partitions in partitioned-init.sql catch any strays).
+      column_overrides:
+        sales_region:
+          region:
+            type: choice
+            values: ["London", "Sydney", "Boston"]
+        test_partitions_sequences:
+          region:
+            type: choice
+            values: ["London", "Sydney", "Boston"]
+        sales:
+          sale_date:
+            type: timestamp_range
+            min: "2019-10-01 00:00:00"
+            max: "2020-06-30 23:59:59"
+        customers:
+          statuses:
+            type: choice
+            values: ["ACTIVE", "RECURRING", "REACTIVATED"]
+          arr:
+            type: int_range
+            min: 0
+            max: 200
 
 # Fallback (target) event generator configuration.
 # This will be used later when we start a generator on the target side.
@@ -73,7 +97,7 @@ generator_target:
     generator:
       schema_name: public
       manual_table_list: []
-      exclude_table_list: [cutover_table]
+      exclude_table_list: [cutover_table, cust_active, cust_arr_small, cust_arr_large, cust_part11, cust_part12, cust_part21, cust_part22, cust_other]
       num_iterations: -1
       wait_after_operations: 5000
       wait_duration_seconds: 1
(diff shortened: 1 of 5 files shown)


shubham-yb (changes requested):
We still need to have the normal schema present here not just the partitions tables. They are an addition to the test not a replacement. Also we can directly call the pg/partitions schema along with the existing common init.sql

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -45,7 +45,7 @@ generator_source:
     generator:
       schema_name: public
       manual_table_list: []
-      exclude_table_list: [cutover_table]
+      exclude_table_list: [cutover_table, cust_active, cust_arr_small, cust_arr_large, cust_part11, cust_part12, cust_part21, cust_part22, cust_other]
```
makalaaneesh, line 48: [AI] **Only the `customers` hierarchy leaves are excluded; the LIST/RANGE leaves of the other three partitioned tables are not, so a large share of generated INSERTs silently fail.**

`exclude_table_list` here drops the `customers` sub-partitions (`cust_active`, `cust_arr_small`, `cust_part11`, ...), but the leaf partitions of `sales_region` (`London`, `Sydney`, `Boston`), `sales` (`sales_2019_Q4`, `sales_2020_Q1`, `sales_2020_Q2`) and `test_partitions_sequences` (`test_partitions_sequences_l/s/b`) are **not** excluded.

The generator discovers tables via `information_schema.tables WHERE table_type = 'BASE TABLE'` (`migtests/scripts/event-generator/utils.py:287`), which returns both partition roots **and** leaf partitions. `get_column_override` (`utils.py:215`) keys overrides by exact table name, and the overrides here are keyed only to the roots (`sales_region`, `sales`, `test_partitions_sequences`). So when the generator picks a leaf like `London` directly, it inserts a *random* `region`/`sale_date` value with no override. For a single-value LIST leaf (`FOR VALUES IN ('London')`) a random text can never satisfy the partition constraint, so every such INSERT fails; `execute_with_retry` regenerates equally-random values, exhausts the 5 retries, and the operation is silently rolled back (`generator.py:180`, `:255`). The result is a meaningful fraction of no-op iterations and reduced effective coverage on exactly the partitioned tables this PR is trying to exercise.

Suggested fix: also add the LIST/RANGE leaf partitions to `exclude_table_list` (mirroring how the `customers` leaves are excluded) so the generator only writes to the roots and lets PG route:
```yaml
exclude_table_list: [cutover_table,
  cust_active, cust_arr_small, cust_arr_large, cust_part11, cust_part12, cust_part21, cust_part22, cust_other,
  London, Sydney, Boston,
  sales_2019_q4, sales_2020_q1, sales_2020_q2,
  test_partitions_sequences_l, test_partitions_sequences_s, test_partitions_sequences_b]
```
(Same edit needed on the `generator_target` block at line 100.) Note the existing `partition-iteration-test` has the same latent pattern but hides it with `insert_max_retries: 50` vs the `5` used here — a smaller retry budget makes the wasted-iteration effect worse.

migtests/tests/resumption/live-migration/iteration-test/scenario.yaml.template
```diff
@@ -58,6 +58,30 @@ generator_source:
       delete_rows: 1
       insert_max_retries: 5
       update_max_retries: 1
+      # Pin region/sale_date/status to valid partition values so generated rows route
+      # across the real partitions (DEFAULT partitions in partitioned-init.sql catch any strays).
```
makalaaneesh, line 62: [AI] **Comment references a non-existent `partitioned-init.sql`; the DEFAULT partitions are actually created in `partitioned-extras.sql`.**

The comment says "DEFAULT partitions in partitioned-init.sql catch any strays", but there is no `partitioned-init.sql` in this PR — the DEFAULT partitions for `sales_region`, `sales`, and `test_partitions_sequences` are added by the new `partitioned-extras.sql` (referenced by the `create_partitioned_extras` stage at line 170). Per `.cursor/BUGBOT.md` ("Comments should explain *why*... workarounds and known limitations deserve comments") an accurate filename matters here since it points readers at the safety-net for stray values.

This same stale reference appears in the `generator_target` block comment at line 114.

Suggested fix: replace `partitioned-init.sql` with `partitioned-extras.sql` in both comments.

makalaaneesh (commented):

PR #3630, opened 2026-06-30
Run fallback-unique-conflict-test conflicts continuously during streaming

Abhi-yb:

Enhances fallback-unique-conflict-test to exercise the streaming-phase unique-key conflict-detection cache (conflictDetectionCache.go) continuously and under resumption stress, on both the forward (PG→YB) and fallback (YB→PG) legs.

Adds

Continuous conflict generator — ConflictGenerator (helpers.py, wired via conflict_generator_start/stop in orchestrator.py) re-applies the conflict DML on a loop, dynamic per cycle (-v cycle=N), in parallel with the random event generator, so conflicts are produced continuously instead of once.
Conflict coverage — all four transitions (DELETE-INSERT, DELETE-UPDATE, UPDATE-INSERT, UPDATE-UPDATE) across every unique-index shape: single/multi unique constraint & index, subset, partial, expression, NULLS NOT DISTINCT, NULLS DISTINCT, and a partitioned table with a unique key that includes the partition column.
validate_conflicts_detected — asserts the forward import-data log recorded conflict detections (fails if none), so the test proves the cache fired

Jenkins run : https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/945/stages/?selected-node=210


migtests/tests/resumption/live-migration/fallback-unique-conflict-test/source_dml.sql
```diff
@@ -55,6 +59,7 @@ COMMIT;
 -- 2. multi_unique_constraint (id PK, UNIQUE(first_name, last_name))
 -- ============================================================
 BEGIN;
+DELETE FROM multi_unique_constraint WHERE id >= 900000000;
 INSERT INTO multi_unique_constraint (id, first_name, last_name) VALUES
     (900010001, 'SrcJohn',  'Doe'),
     (900010002, 'SrcJane',  'Smith'),
```
priyanshi-yb: This is not the right way to generate; here you are essentially just doing the same set of conflicts by deleting the rows before running every time. Can you please try to just take inspiration from this set of conflicts on what types of conflicts you should generate, and then see if you can make it more dynamic by doing conflicts on a different set of rows everytime

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -755,6 +756,131 @@ def stop_generator(proc: subprocess.Popen | None, graceful_timeout_sec: int) ->
...
+            name=self.label,
+            daemon=True,
+        )
+        self._thread.start()
+        log(
+            f"{self.label}: started (sql={os.path.basename(self.sql_path)}, "
+            f"interval={self.interval_sec}s, {self._cases_per_cycle} conflict-cases/cycle)"
+        )
+
+    def stop(self, timeout_sec: int = 60) -> None:
```
makalaaneesh, line 835: [AI] **`ConflictGenerator.stop()` does not kill an in-flight `psql` child, so a cycle can commit conflict DML after stop() returns.**

`_run_once()` calls a blocking `subprocess.run(["psql", ...])` inside the daemon thread. `stop()` only sets `stop_flag` and `join(timeout_sec)`. If a cycle is mid-`psql` when `stop()` is called, `join` waits up to `timeout_sec`; if the child is still running when the join times out (or completes the COMMIT during the join window), the psql process is never terminated and its conflict-DML transaction can commit *after* `stop()` has returned.

The stated design goal is that "the generators stop before each backlog marker so the drain stays deterministic" (PR description). That guarantee is not airtight here: a still-running child can inject one more conflict batch after the stop marker. Contrast with `stop_generator()` (helpers.py:755), which calls `kill(proc, ...)` on the child process itself.

In practice the risk is low (each cycle is 9 tiny transactions, well under the 60s default timeout), but the invariant the drain relies on is enforced only by timing, not by the code.

Suggested fix: hold a reference to the current `subprocess.Popen`/process in `_run_once` (e.g. `self._proc`) and, in `stop()`, terminate/kill it after setting `stop_flag` and before/after `join` so an in-flight cycle cannot outlive the stop. Alternatively, if a timed-out join is observed, log a warning and kill the child explicitly.

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.yaml.template
```diff
@@ -240,7 +286,12 @@ stages:
 
   - name: soak_fallback_with_resumptions
     action: sleep
-    seconds: 3000
+    seconds: 400
```
makalaaneesh: [AI] **Fallback soak cut from 3000s to 400s — a 7.5x reduction in fallback-leg resumption coverage.**

`soak_fallback_with_resumptions` drops from `seconds: 3000` to `seconds: 400`. The forward soak was simultaneously raised (30s -> 400s), so the intent — balancing the two legs at the iteration-test's 400s — is reasonable and documented. But the fallback leg previously ran resumption restarts for 3000s; at `min/max_restart_wait` of 5–30s that was on the order of ~100–600 restart cycles, now reduced to roughly a seventh of that.

Per `migtests/.cursor/BUGBOT.md` (Live Migration Tests): "Wait intervals for streaming phases should be long enough to allow meaningful data flow." The fallback leg is where the #3604 YugabyteDB-source multi-column-unique-index path is exercised, so this is precisely the path this test is meant to stress under resumption.

Suggested change: confirm 400s still yields enough resumption iterations on the fallback leg to reliably reproduce the conflict-cache regressions this test guards against (the PR notes an occasional ~0.5% transient row-count drift on random-generator rows under aggressive resumption — a shorter soak gives fewer chances to catch a real conflict-cache bug, not just the transient). If 400s was chosen only for CI-time budget, note that rationale in a comment next to the `seconds:` value so a future editor doesn't assume the two legs must stay symmetric.

makalaaneesh (commented):
[AI] Automated review of PR #3630 — 2 findings (0 critical, 2 warning). Test-infrastructure-only change (migtests/); no yb-voyager CLI code touched. See inline comments.

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/source_dml.sql
Abhi-yb: reworked to be dynamic. the generator passes -v cycle=N and every id/value is derived from that cycle (ids :base + offset, unique values suffixed with the cycle), so each cycle runs the conflicts on a fresh set of rows instead of recycling. dropped the leading DELETE. verified locally

migtests/scripts/live-migration-resumption/helpers.py
Abhi-yb: now tracks the psql child and kills it in stop() via the shared kill() helper (same as stop_generator), so an in-flight cycle can't commit after stop returns. verified locally — it interrupts a running cycle immediately instead of waiting it out.

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.yaml.template
Abhi-yb: kept it 400s to stay symmetric for both fallback and forward case

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/source_dml.sql
```diff
@@ -1,170 +1,178 @@
...
---    the deterministic rows never collide with generator-inserted rows.
---  * On the source side every statement is individually valid (no unique
---    violation); the violation only *would* occur on the target if the events
---    were applied out of order, which is exactly what the cache prevents.
+-- value: a "freeing" event (DELETE/UPDATE the holder) and a "reusing" event
+-- (INSERT/UPDATE a different PK to that value). They route to different import
+-- channels, so the cache must serialize them; otherwise the reusing event hits a
+-- unique violation. Every statement is individually valid on the source; the
+-- violation would only occur on the target if events were applied out of order.
+-- Each table's seed + conflicts run in ONE transaction.
```
priyanshi-yb, line 25: Let's add a few more cases to cover the NULL-related cases - Unique indexes with NULLS NOT DISTINCT- and then test whether it properly handles conflicts and whether the data is correctly imported.

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.yaml.template
```diff
@@ -238,9 +284,18 @@ stages:
...
+  # here; raise it if a longer fallback resumption soak is needed.
   - name: soak_fallback_with_resumptions
     action: sleep
-    seconds: 3000
+    seconds: 400
+
+  - name: stop_target_conflict_generator
+    action: conflict_generator_stop
+    generator_key: conflict_generator_target
+    graceful_timeout_sec: 60
```
priyanshi-yb, line 294: update the `scenario.local.yml` with these changes for consistency

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/source_dml.sql
priyanshi-yb, line 25: Lets add one partition table cases in this schema also

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/init.sql
```diff
@@ -81,6 +81,35 @@ CREATE TABLE test_partial_unique_index (
...
+    id INT,
+    region VARCHAR(50),
+    email VARCHAR(255),
+    PRIMARY KEY (id, region)
+) PARTITION BY LIST (region);
+CREATE TABLE partitioned_unique_conflict_east PARTITION OF partitioned_unique_conflict FOR VALUES IN ('east');
+CREATE TABLE partitioned_unique_conflict_west PARTITION OF partitioned_unique_conflict FOR VALUES IN ('west');
+CREATE TABLE partitioned_unique_conflict_default PARTITION OF partitioned_unique_conflict DEFAULT;
+CREATE UNIQUE INDEX idx_partitioned_unique_email ON partitioned_unique_conflict (email, region);
+
```
priyanshi-yb: should we have one more index for partition case with NULLS NOT DISTINCT?

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.local.yaml
```diff
@@ -22,9 +23,9 @@ target:
   type: yugabytedb
   host: 127.0.0.1
   port: 5433
-  database:  test_db
-  user: ybvoyager
-  password: Test@123#$%^&*()!
+  database: test_db
+  user: yugabyte
+  password: yugabyte
```
priyanshi-yb: is this a test change?

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.local.yaml
```diff
@@ -155,15 +193,20 @@ stages:
...
       export_data:
-        max_restarts: 1000
-        min_interrupt_seconds: 30
+        max_restarts: 25
+        min_interrupt_seconds: 15
         max_interrupt_seconds: 60
-        min_restart_wait_seconds: 30
-        max_restart_wait_seconds: 60
+        min_restart_wait_seconds: 5
+        max_restart_wait_seconds: 30
```
priyanshi-yb: Initially I have added the good number of restarts so that test can run for atleast an hour, so if you can keep that

priyanshi-yb (commented):
mostly LGTM but please wait for @shubham-yb 's review on the conflict cases

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/source_dml.sql
```diff
@@ -266,26 +274,82 @@ COMMIT;
...
+-- Under NULLS NOT DISTINCT two NULLs are equal, so a NULL free->reuse across PKs is a
+-- real conflict. Only one NULL can exist at a time, so the NULL is moved off at the end
+-- of the block, leaving none behind for the next cycle.
+-- ============================================================
+BEGIN;
+-- UPDATE-INSERT (non-null value)
+INSERT INTO single_unique_index_nulls_not_distinct (id, email) VALUES (:base + 85001, ('nnd1_user1@conflict.test' || :cycle));
+UPDATE single_unique_index_nulls_not_distinct SET email = ('nnd1_user1_moved@conflict.test' || :cycle) WHERE id = :base + 85001;
+INSERT INTO single_unique_index_nulls_not_distinct (id, email) VALUES (:base + 85101, ('nnd1_user1@conflict.test' || :cycle));
+
```
priyanshi-yb, line 312: NULL conflicts cases, lets add it on the previous table, which has index with default property i.e. NULLS DISTINCT

shubham-yb:
```(Note: row_count_validations can occasionally show a small ~0.5% transient drift on random event-generator rows under aggressive resumption restarts — a drain-timing property of the harness, not the conflict logic; conflict rows always match — and a re-run clears it.)```

Can you explain this a bit more? Can this cause intermittent issues?

**Comment by shubham-yb** [comment]
Have we checked the Voyager logs while running this script to ensure that we really are hitting the expected number of conflicts?
Also please attach some Jenkins runs to the PR.

File: migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.local.yaml
Comment (shubham-yb) [inline migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.local.yaml:1]: We don't need this file. Let's only keep the template.

File: migtests/tests/resumption/live-migration/fallback-unique-conflict-test/init.sql [hunk 10]
```diff
@@ -92,3 +131,9 @@ ALTER TABLE different_columns_unique_constraint_and_index REPLICA IDENTITY FULL;
 ALTER TABLE subset_columns_unique_constraint_and_index REPLICA IDENTITY FULL;
 ALTER TABLE expression_based_unique_index REPLICA IDENTITY FULL;
 ALTER TABLE test_partial_unique_index REPLICA IDENTITY FULL;
+ALTER TABLE single_unique_index_nulls_not_distinct REPLICA IDENTITY FULL;
+ALTER TABLE multi_unique_index_nulls_not_distinct REPLICA IDENTITY FULL;
+ALTER TABLE single_unique_index_nulls_distinct REPLICA IDENTITY FULL;
+ALTER TABLE partitioned_unique_conflict_east REPLICA IDENTITY FULL;
+ALTER TABLE partitioned_unique_conflict_west REPLICA IDENTITY FULL;
+ALTER TABLE partitioned_unique_conflict_default REPLICA IDENTITY FULL;
```
shubham-yb: The scripts must already be handling this. Please check and remove.

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.local.yaml
```diff
@@ -56,6 +57,20 @@ generator_source:
...
+# Forward (source) conflict generator: re-applies this test's own dynamic
+# source_dml.sql on a loop throughout streaming, in parallel with generator_source.
+conflict_generator_source:
+  config_inline:
+    connection:
+      host: 127.0.0.1
+      port: 5432
+      database: test_db
+      user: postgres
+      password: postgres
```
shubham-yb: Since this is part of the same ecosystem, why can't it directly take the source connection variables from the top level key? It will simplify things for us.

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.local.yaml
```diff
@@ -75,16 +90,33 @@ generator_target:
...
+# Fallback (target) conflict generator: re-applies this test's own dynamic
+# target_dml.sql on a loop throughout fallback streaming, in parallel with generator_target.
+conflict_generator_target:
+  config_inline:
+    connection:
+      host: 127.0.0.1
+      port: 5433
+      database: test_db
+      user: yugabyte
+      password: yugabyte
```
shubham-yb: Same comment as above.

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.local.yaml
```diff
@@ -75,16 +90,33 @@ generator_target:
...
+    conflict:
+      sql_path: ./target_dml.sql
+      interval_seconds: 10
+
 voyager:
+  export_schema:
+    flags:
+      skip-performance-recommendations: "true"
```
shubham-yb: Why are we doing this?

Abhi-yb: Local-run assessement will fail

migtests/scripts/live-migration-resumption/orchestrator.py
```diff
@@ -71,6 +72,74 @@ def generator_stop_action(stage: Dict[str, Any], ctx: Any) -> None:
...
+    """
+    key = stage.get("generator_key", "conflict_generator")
+    cfg_block = ctx.cfg.get(key) or {}
+    inline = cfg_block.get("config_inline") or cfg_block.get("config") or {}
+    connection = inline.get("connection") or {}
+    conflict_cfg = inline.get("conflict") or {}
+
+    # If no explicit connection is given, reuse the top-level source/target
+    # connection (admin creds) so the block need not duplicate it --
+    # conflict_generator_source -> source, conflict_generator_target -> target.
```
shubham-yb: Let's not have explicit connection option at all. Always pick from the parent.

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.yaml.template
```diff
@@ -180,6 +208,13 @@ stages:
     condition: remaining_events_eq_0
     timeout_sec: 1200
 
+  # Confirm the conflict-detection cache actually fired on the forward (PG->YB)
+  # import, not just that row_count/row_hash matched.
+  - name: validate_conflicts_detected
+    action: validate_conflicts_detected
+    log: yb-voyager-import-data.log
+    min_count: 1
```
shubham-yb, line 216: Why don't we check it for YB -> PG?

Abhi-yb:
> `(Note: row_count_validations can occasionally show a small ~0.5% transient drift on random event-generator rows under aggressive resumption restarts — a drain-timing property of the harness, not the conflict logic; conflict rows always match — and a re-run clears it.)`
> 
> Can you explain this a bit more? Can this cause intermittent issues?

Abhi-yb:
> Have we checked the Voyager logs while running this script to ensure that we really are hitting the expected number of conflicts? Also please attach some Jenkins runs to the PR.

sure

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/init.sql
```diff
@@ -3,10 +3,10 @@
...
 -- (yb-voyager/cmd/conflictDetectionCache.go) reasons about.
 --
--- REPLICA IDENTITY FULL is set on every table so that Debezium captures the
--- before-image of UPDATE/DELETE rows. The conflict-detection cache compares the
--- before-values of unique-key columns (incl. partial-index predicate columns),
--- so without full before-images the DELETE-*/UPDATE-* conflicts cannot be detected.
+-- REPLICA IDENTITY FULL (so Debezium captures before-images of UPDATE/DELETE rows,
+-- which the conflict-detection cache needs) is set on all tables by the grant step
+-- (yb-voyager-pg-grant-migration-permissions.sql run by grant_source_permissions),
+-- so it is not set here.
```
shubham-yb: Let's remove this comment.

shubham-yb (approved):
LGTM.

PR #3634, opened 2026-07-01
Enable plain live migration (snapshot + CDC) for YugabyteDB AMP target

makalaaneesh:

Builds on the offline PostgreSQL → YugabyteDB AMP (yb-amp) POC (base branch `ybamp-target-poc`) to support **plain live migration** — snapshot-and-changes streaming into the amp compute followed by cutover to target. The offline POC had already made the shared import-to-target path amp-aware (colocation / health / adaptive-parallelism are gated to real YugabyteDB, so amp takes the sequential / PG code path); the CDC event-apply itself (`TargetPostgreSQL.ExecuteBatch`, the event-metadata schema, PK/unique-key conflict detection) is inherited by `TargetYugabyteDBAmp` via embedding and is already exercised by the existing PG fall-back path — so no new event-apply code was needed.

What remained were a few target-type branches in the live path that enumerated YugabyteDB/PostgreSQL but omitted `yugabytedb-amp`, plus a guardrail. Specifically: the Debezium value-converter suite now routes amp to the YB/PG suite (it previously fell through to `default` and errored `no converter suite found` for both the live-snapshot and CDC-streaming converters — a hard blocker); `shouldFormatValues` treats amp like PG/YB (format values on UPDATE events); and the default `"auto"` CDC-partitioning strategy routes amp to `PARTITION_BY_TABLE` (sequential per-table apply), mirroring the existing single-node PG/Oracle rationale in the same function and sidestepping `getExpressionUniqueIndexTables()`, which is implemented only on the `TargetYugabyteDB` driver.

**Fall-back / fall-forward are deliberately NOT supported for amp** — they require streaming changes *out of* the target via YugabyteDB CDC, which a stateless PG-wire amp compute does not provide. The cutover-to-target command now fails fast with a clear message if fall-back/fall-forward is requested against an amp target.

- **Command line:** `export data --export-type snapshot-and-changes` (PG source) followed by `import data to target --target-db-type yugabytedb-amp` and `initiate cutover to target --prepare-for-fall-back no` now performs a plain live migration into a yb-amp compute. No new flags. `--use-partition-root` is now accepted for an amp target (PG partition semantics). `initiate cutover to target` fails fast with an actionable error if `--prepare-for-fall-back yes` (or a fall-forward-enabled migration) is used with an amp target.
- **Configuration:** No new config keys.
- **Installation:** No change.
- **Reports:** No change.

- Existing unit tests: full `cmd` unit suite (`go test -tags unit ./cmd/...`) and `src/dbzm` unit tests pass; `go build ./...` and `go vet` clean.
- **Manual end-to-end run (PG → yb-amp, plain live migration):** export schema → import schema → `export data --export-type snapshot-and-changes` → `import data to target` → applied INSERT/UPDATE/DELETE on the source across two tables → verified all 6 CDC events replicated correctly to the amp target → `initiate cutover to target --prepare-for-fall-back no`. Final source↔target reconciliation matched on all tables (customers 501/501, products 201/201, orders 2000/2000, order_items 2000/2000), and the cutover completed cleanly ("Completed streaming all relevant changes to yugabytedb-amp"). All three code changes were exercised — the migration hard-fails without each.
- Not added: automated live-migration integration test coverage for the amp target (would need the Debezium + logical-replication harness); the plain-live path reuses the existing PG/YB live-migration code, which is covered by existing integration tests.

No. `BuildCallhomeYBClusterMetrics` now returns empty for non-YugabyteDB targets instead of a misleading type-assertion error, but the payload shape/version is unchanged.

No. The amp `"auto"` strategy writes `PARTITION_BY_TABLE` into the existing `ImportDataStatusRecord.TableToCDCPartitioningStrategyMap` (Import Data State) — same structure, amp just takes a different code path to populate it. No new or altered on-disk structure.

### DIFF
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -681,21 +681,20 @@ func validateFFDBSchemaFlag() {
 }
 
 // defaultAdaptiveParallelismMode returns the adaptive-parallelism mode to use when the
-// user did NOT pass --adaptive-parallelism. It is per-target:
-//   - YUGABYTEDB: Balanced — adaptive parallelism uses the YB cluster control API
-//     (yb_servers(), tserver metrics) and is the recommended default.
-//   - YUGABYTEDB_AMP: Disabled — yb-amp is a stateless PG17 compute with no YB cluster
-//     control API, so adaptive parallelism is not available; --parallel-jobs controls
-//     import parallelism instead.
+// user did NOT pass --adaptive-parallelism. Adaptive parallelism relies on the YugabyteDB
+// cluster control API (yb_servers(), tserver metrics), so it is the recommended default
+// (Balanced) ONLY for a real YugabyteDB target. Every other target — yb-amp (stateless
+// PG17 compute) and the PostgreSQL fall-forward/fall-back targets — has no such API, so it
+// defaults to Disabled; --parallel-jobs controls import parallelism there.
 //
-// Defaulting amp to Disabled is also what lets a user pass --parallel-jobs for amp without
-// having to also pass --adaptive-parallelism disabled (validateParallelismFlags only
-// conflicts --parallel-jobs with an *enabled* adaptive mode).
+// Defaulting non-YB targets to Disabled is also what lets a user pass --parallel-jobs for
+// them without having to also pass --adaptive-parallelism disabled (validateParallelismFlags
+// only conflicts --parallel-jobs with an *enabled* adaptive mode).
 func defaultAdaptiveParallelismMode(targetDBType string) types.AdaptiveParallelismMode {
-	if targetDBType == YUGABYTEDB_AMP {
-		return types.DisabledAdaptiveParallelismMode
+	if targetDBType == YUGABYTEDB {
+		return types.BalancedAdaptiveParallelismMode
 	}
-	return types.BalancedAdaptiveParallelismMode
+	return types.DisabledAdaptiveParallelismMode
 }
 
 func validateParallelismFlags(cmd *cobra.Command) {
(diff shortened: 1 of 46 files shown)


yb-voyager/cmd/live_migration.go
```diff
@@ -217,6 +217,19 @@ func getCdcPartitioningStrategyPerTable(tableNames []sqlname.NameTuple) (*utils.
...
+			// source/source-replica case above — PARTITION_BY_TABLE has no real
+			// throughput downside and is safer for correctness. It also sidesteps
+			// getExpressionUniqueIndexTables(), which is implemented only on the
+			// TargetYugabyteDB driver. (Explicit --cdc-partitioning-strategy
+			// pk/table still flows through the default branch below.)
+			for _, t := range tableNames {
+				tableToPartitioningStrategyMap.Put(t, PARTITION_BY_TABLE)
+			}
+			break
+		}
```
priyanshi-yb, line 232: Do we want to allow changing this config for the YB_AMP?
As I feel we should just set it to PARTITION_BY_TABLE always and disallow this config

yb-voyager/cmd/import.go
```diff
@@ -194,7 +194,7 @@ func validateImportUsePartitionRootFlag() error {
 		if importerRole == SOURCE_REPLICA_DB_IMPORTER_ROLE {
 			return goerrors.Errorf("'--use-partition-root false' is not supported for source-replica")
 		}
-		if tconf.TargetDBType != POSTGRESQL && tconf.TargetDBType != YUGABYTEDB {
+		if tconf.TargetDBType != POSTGRESQL && tconf.TargetDBType != YUGABYTEDB && tconf.TargetDBType != YUGABYTEDB_AMP {
```
makalaaneesh, line 197: only thing we need to check here is source Db type == postgres. if not, we error out .

yb-voyager/cmd/live_migration.go
makalaaneesh: as discussed, no harm in allowing this

priyanshi-yb (approved):
LGTM

makalaaneesh:
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/8378/stages/

PR #3636, opened 2026-07-01
schemasnapshot: case-sensitive / quoted-identifier–aware ObjectRef identity

ShivanshGahlot:
**Stacked on #3614** (`shivansh/schemadiff-scaffolding`). Review/merge after that lands; the base here is the scaffolding branch, not `main`.

### What & why
`ObjectRef.String()` joined `schema` and `name` with a plain `"."`, which:
- loses case-sensitivity intent for quoted identifiers (`Orders` vs `orders` are distinct tables in PG), and
- collides for names that themselves contain a dot (schema `a.b`/name `c` vs schema `a`/name `b.c` both → `a.b.c`).

This delegates rendering and keying to the existing `sqlname` helpers instead of hand-joining identifiers (per the repo rule to never concatenate schema/table names).

### Changes
- `ObjectRef` keeps its raw `{Schema, Name}` catalog values (**storage/JSON unchanged**) and gains:
  - `ForDisplay(dbType)` — minimally-quoted, always-qualified, for reports/logs/user-facing SQL (`public.orders`, `public."My Table"`).
  - `ForKey(dbType)` — per-part-quoted, case-sensitive, collision-safe canonical map key (`"public"."orders"`).
  - The naive `String()` is **removed**; the struct stays comparable for direct use as a Go map key.
- `Column` gets composite `ForKey`/`ForDisplay` so case-sensitive **column** names are quoted correctly too.

### Scope
Same-engine (PG/YB) only. Cross-engine identity reconciliation (Oracle/MySQL source, source-vs-target) is a matching-layer concern handled elsewhere and is out of scope here.

### Tests
`snapshot_test.go` adds coverage for lowercase, mixed-case, reserved-word (`user`), and dotted identifiers; case-sensitivity distinctness; dot-collision safety; column composites; and a pinned KNOWN-LIMITATION for embedded double-quotes (not escaped yet). Verified: `go build ./...`, `go vet`, `go test -tags unit`, integration-compile, `gofmt` all clean.

---

### DIFF
--- a/yb-voyager/src/schemasnapshot/snapshot.go
+++ b/yb-voyager/src/schemasnapshot/snapshot.go
@@ -152,16 +152,16 @@ type Column struct {
 	Default  string `json:"default,omitempty"` // default expression text (pg_get_expr); "" when the column has no default.
 }
 
-// ForKey returns a collision-safe canonical key for the column, combining the
-// parent table's ForKey with the quoted column name. For example:
-// "public"."orders"."Col"
-func (c Column) ForKey(dbType string) string {
-	return c.Table.ForKey(dbType) + "." + sqlname.NewIdentifier(dbType, c.Name).Quoted
+// sqlName builds the sqlname view of this column's fully-qualified (schema, table,
+// column) identity. defaultSchema="" keeps it always fully qualified.
+func (c Column) sqlName(dbType string) *sqlname.ObjectNameQualifiedWithTableName {
+	return sqlname.NewObjectNameQualifiedWithTableName(dbType, "", c.Name, c.Table.Schema, c.Table.Name)
 }
 
-// ForDisplay returns the minimally-quoted, always-fully-qualified rendering of
-// the column for reports, logs, and user-facing SQL. For example:
-// public.orders."Col"
-func (c Column) ForDisplay(dbType string) string {
-	return c.Table.ForDisplay(dbType) + "." + sqlname.NewIdentifier(dbType, c.Name).MinQuoted
-}
+// ForKey returns a collision-safe, per-part-quoted canonical key for the column:
+// "public"."orders"."Col".
+func (c Column) ForKey(dbType string) string { return c.sqlName(dbType).Qualified.Quoted }
+
+// ForDisplay returns the minimally-quoted, always-fully-qualified rendering of the
+// column for reports, logs, and user-facing SQL: public.orders."Col".
+func (c Column) ForDisplay(dbType string) string { return c.sqlName(dbType).MinQualified.MinQuoted }
--- a/yb-voyager/src/schemasnapshot/snapshot.go
+++ b/yb-voyager/src/schemasnapshot/snapshot.go
@@ -87,8 +87,8 @@ type DBMetadata struct {
 //
 // Schema and Name hold the raw, unquoted catalog names exactly as returned by the
 // database catalog (e.g. pg_class.relname). Rendering and matching are engine-aware:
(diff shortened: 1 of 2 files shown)


yb-voyager/src/schemasnapshot/snapshot.go
```diff
@@ -83,18 +85,34 @@ type DBMetadata struct {
...
+// public."Orders".
+func (o ObjectRef) ForDisplay(dbType string) string { return o.sqlName(dbType).String() }
+
+// ForKey returns the case-sensitive, per-part-quoted, collision-safe canonical key
+// suitable for use as a string map key. For example: "public"."orders".
+// Two ObjectRefs that differ only in case produce different ForKey values, and two
+// refs that would naively produce the same "schema.name" join (e.g. schema "a.b" /
+// name "c" vs schema "a" / name "b.c") produce different ForKey values because each
+// part is independently double-quoted.
+func (o ObjectRef) ForKey(dbType string) string { return o.sqlName(dbType).Qualified.Quoted }
```
makalaaneesh: [AI] **`ForKey` is not collision-safe for catalog names that themselves begin and end with a double-quote.**

The docstring guarantees `ForKey` produces a "collision-safe canonical key" and that distinct objects yield distinct keys. That guarantee does not hold for a class of legal PostgreSQL names.

`sqlName` → `sqlname.NewObjectName` (and `Column.ForKey` → `sqlname.NewIdentifier`) call `IsQuoted(part)` and strip the outer double-quotes whenever a part starts and ends with `"`. `quote2` then re-wraps the stripped name **without escaping**. So a raw catalog `relname` of `x` and a raw `relname` of `"x"` (a legal 3-character name, e.g. from `CREATE TABLE public."""x"""`) both collapse to the same key:

```
ObjectRef{Schema:"public", Name:`x`}.ForKey(pg)   == `"public"."x"`
ObjectRef{Schema:"public", Name:`"x"`}.ForKey(pg) == `"public"."x"`   // collision
```

(Confirmed by running the code against this branch.) This is the same underlying defect that `TestColumnForKeyKnownLimitationEmbeddedQuote` pins for embedded quotes, but here it breaks the **collision-safety** property that is the entire purpose of `ForKey` — two distinct catalog objects map to one key, so schemadiff would treat them as identical and miss an add/drop. The pinned test frames this only as "invalid SQL", which undersells the impact.

Suggested fix: for `ForKey`, build the key by SQL-escaping each raw part directly (double any inner `"` and wrap), e.g. `\"` + strings.ReplaceAll(part, `\"`, `\"\"`) + `\"`, rather than routing through `NewObjectName`/`NewIdentifier`, which strip already-present quotes. That makes both the embedded-quote and quoted-looking-name cases round-trip losslessly and restores true collision safety. At minimum, tighten the docstring to state that raw names containing/looking-like quotes are not yet handled, so callers don't rely on a guarantee that doesn't hold.

---

makalaaneesh (commented):

priyanshi-yb (approved):
LGTM

yb-voyager/src/schemasnapshot/snapshot.go
priyanshi-yb: nit: but for consistency, we can probably use SQLObjectName's Key() directly, as it keeps the Key as Unquoted, which is readable. There are limitations, but those limitations are there in every usage of the SQL name, so it should be fine. Whenever we support those, we will do it and that will automatically reflect here.

ShivanshGahlot: Using Key() directly. When fixes land in sqlname for any limitations, they will land here too automatically.

PR #3637, opened 2026-07-02
Surface target-db-type in config templates and guardrail error-policy-snapshot for yb-amp

makalaaneesh:

A few config-template and guardrail improvements for the YugabyteDB AMP (yb-amp) target, following the offline (#3623) and live (#3634) yb-amp support now in `main`:

- Adds an active `target: db-type` field (values `yugabytedb` / `yugabytedb-amp`, with help) to the **offline** and **live** migration config templates, so users can select the target engine from the config file.
- Documents the now target-type-dependent **adaptive-parallelism default** — `balanced` for YugabyteDB, `disabled` for YugabyteDB AMP — in both the `--adaptive-parallelism` CLI help and the offline/live templates.
- Stops the templates from hardcoding `adaptive-parallelism: balanced`: it is commented out in **all five** templates (keeping `balanced` visible in the comment) so the dynamic per-target default takes effect; `parallel-jobs` stays commented for the same reason. `disable-pb: false` is uncommented in the same section so it remains a valid non-empty YAML map.
- Adds a guardrail rejecting `--error-policy-snapshot stash-and-continue` for `--target-db-type yugabytedb-amp` (not yet validated on amp's plain-COPY import path); the default `abort` continues to work. `batch-size` remains supported for amp.
- Fixes `--use-partition-root false` validation to gate on the **source** DB type (must be PostgreSQL) instead of the target engine (addresses review feedback on #3634). The flag governs how PostgreSQL declarative-partitioned tables are streamed during CDC — a source-schema concept — so the old target-based check incorrectly let a non-PostgreSQL source (e.g. Oracle/MySQL → YugabyteDB) pass.

This branch is based on the latest `main` (including the just-merged #3634), so `target: db-type: yugabytedb-amp` is now real for both offline and live flows.

Yes.
1. **Command line:**
   - `--adaptive-parallelism` help updated to describe the target-type-dependent default.
   - New guardrail: `--error-policy-snapshot stash-and-continue` is rejected for `--target-db-type yugabytedb-amp`.
   - `--use-partition-root false` is now rejected when the source database is not PostgreSQL (previously gated on the target engine).
2. **Configuration:** offline/live templates gain a `target: db-type` field; all templates no longer set `adaptive-parallelism: balanced` by default (commented out) and instead have `disable-pb: false` active in the import section. No config keys were added or removed at the schema level (`target.db-type` was already an accepted key).
3. **Installation:** no change.
4. **Reports:** no change.

- `go build ./...`, `go vet ./...`, `gofmt` — all clean.
- `go test -tags unit ./cmd/...` — passes (includes config-template parsing and error-policy config tests).
- Validated all five templates parse as YAML; each import section is a non-empty map with `adaptive-parallelism` commented out and `disable-pb: false` active; `target.db-type` reads back as `yugabytedb` on the offline/live templates only.
- For the `--use-partition-root` change: confirmed `sourceDBType` is populated (from the MSR) before the validation runs; no existing unit test encoded the old target-based behavior.
- No new unit tests added: the error-policy guardrail (`validateAmpUnsupportedFlags`) is `utils.ErrExit`-based (process exit), and `validateImportUsePartitionRootFlag` depends on the metaDB migration-status record + package globals — neither is covered by the current cmd unit harness without additional fixtures.

No.

No. Config templates are sample files shipped with the CLI; the guardrail, CLI-help, and validation changes don't touch any on-disk structure.

### DIFF
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -621,7 +621,7 @@ func registerFlagsForTarget(cmd *cobra.Command) {
 		"Adapt parallelism based on the resource usage (CPU, memory) of the target YugabyteDB cluster."+
 			"\n"+
 			"Specify the mode for adaptive parallelism behavior: disabled, balanced, aggressive "+
-			"(default: balanced for YugabyteDB targets; disabled for other target types e.g. yugabytedb-amp, since they have no cluster control API — use --parallel-jobs there)"+
+			"(default: balanced for YugabyteDB, disabled for YugabyteDB AMP)"+
 			"\n"+
 			"\tbalanced: Operate with moderate thresholds. Recommended to be used when there are other workloads running on the cluster.\n"+
 			"\taggressive: Operate with aggressive max-CPU thresholds for better performance. Recommended to be used when there are no other workloads running on the cluster.\n"+
--- a/yb-voyager/cmd/import.go
+++ b/yb-voyager/cmd/import.go
@@ -186,7 +186,7 @@ func validateAmpUnsupportedFlags(cmd *cobra.Command) {
 }
 
 func validateImportUsePartitionRootFlag() error {
-	// --use-partition-root flag is only valid for live migration with PostgreSQL or YugabyteDB source
+	// --use-partition-root flag is only valid for live migration with a PostgreSQL source
 	//and only for the CDC streaming phase and snapshot part isn't supported right now.
 	if !importUsePartitionRoot {
 		// Only validate when flag is explicitly set to false (non-default)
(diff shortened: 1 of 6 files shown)


yb-voyager/cmd/import.go
```diff
@@ -175,6 +176,13 @@ func validateAmpUnsupportedFlags(cmd *cobra.Command) {
 	if strings.ToUpper(tconf.OnPrimaryKeyConflictAction) != constants.PRIMARY_KEY_CONFLICT_ACTION_ERROR_POLICY {
 		notApplicable("on-primary-key-conflict")
 	}
+	// Only the default `abort` error policy is validated for amp. `stash-and-continue`
+	// (stashing errored snapshot rows and continuing) has not been tested on amp's
+	// plain-COPY path, so reject it explicitly rather than silently allowing it.
+	if errorPolicySnapshotFlag == importdata.StashAndContinueErrorPolicy {
```
makalaaneesh, line 182: [AI] **Amp guardrails (including this new `stash-and-continue` reject) are bypassed on the `import data file` path.**

`validateAmpUnsupportedFlags` is only invoked from the `import data` / `import data to target` PreRun (`importData.go`). The `import data file` command, however, also reaches an amp target:

- It registers `--target-db-type` via `registerFlagsForTarget(importDataFileCmd)` (`importDataFileCommand.go:489`) and `--error-policy`, which is bound to the same `errorPolicySnapshotFlag` (`importDataFileCommand.go:545`).
- `validateTargetDBTypeFlag` explicitly whitelists `YUGABYTEDB_AMP` for `IMPORT_FILE_ROLE` (`import.go:542`), so amp is an intended target for bulk load.
- Its PreRun (`checkImportDataFileFlags`) never calls `validateAmpUnsupportedFlags`.

So `yb-voyager import data file --target-db-type yugabytedb-amp --error-policy stash-and-continue` reaches amp's plain-COPY snapshot path with exactly the policy this PR intends to reject — the guardrail is silently skipped there.

Per `.cursor/BUGBOT.md` ("Features gated by source type ... must have explicit validation or guardrails"), the gate should cover every reachable path.

Suggested fix: invoke the amp guardrails from the import-data-file PreRun (e.g. in `checkImportDataFileFlags`), or — if amp is intentionally unsupported for bulk load — reject it explicitly on that path.

---

makalaaneesh (commented):

yb-voyager/cmd/import.go
makalaaneesh: `target-db-type` NOT registered for import data file

priyanshi-yb (approved):
LGTM

PR #3640, opened 2026-07-02
Add pluggable CDC ingest benchmark framework (test/cdcbench)

makalaaneesh:

The live-migration import streaming path (`streamChangesFromSegment` and everything under it: NDJSON decode, name-registry lookups, Debezium value conversion, the conflict-detection cache, event channels, batching) had no repeatable way to measure ingest throughput or conflict-detection behavior. This PR adds a benchmark framework that replays **real Debezium-generated change events** through the **unmodified production import code**, with exactly one mock: `TargetDB.ExecuteBatch` succeeds without a database.

Design:
- **Workloads are pluggable**: a workload = 3 SQL files (schema/seed/DML) + a ~10-line `Register()` call in `test/cdcbench`. Four founding workloads at 20k events each span the behavior space: `inserts-no-uk` (control), `updates-uk-no-conflict`, `mixed-uk-no-conflict` (60/30/10 op mix), `updates-uk-conflict-pairs` (10k engineered unique-key conflicts).
- **Artifacts are real and cached**: events are generated once by an actual `yb-voyager export data` (snapshot-and-changes) run against a testcontainers PostgreSQL source (`wal_level=logical`), then cached keyed by a content hash of the workload definition. Editing a workload's SQL auto-invalidates only its artifact; cached artifacts replay with no Docker or voyager-binary dependency.
- **Clean layering**: `test/cdcbench` has zero knowledge of the `cmd` package. The three pieces needing cmd internals are injected as closures (`Bootstrap`, `StreamAll`, `CacheDepth`) by a thin shim `cmd/cdc_ingest_bench_test.go` behind the `cdc_benchmark` build tag. **No production code changes.**
- **Measured, not assumed**: per run the framework reports events/s, batches, and conflict-cache depth (100ms sampler), and *asserts* the exact processed-event count plus the workload's conflict expectation (zero vs non-zero, counted via a logrus hook on the cache's conflict logs). Output is both benchstat-compatible `Benchmark` lines and a human summary table.

Motivation: recent conflict-detection work needed before/after throughput numbers; this framework reproduces that methodology as a first-class, reusable tool. Baseline on current main (Apple M1 Pro): `inserts-no-uk` ~129k events/s vs `updates-uk-no-conflict` ~1.1k events/s (conflict-cache scans; depth ~1.4k) — the ~100× gap this framework exists to track.

No user-facing changes. No CLI, configuration, installation, or report changes. The framework is a test-support library plus a build-tag-gated test file; released binaries are unaffected (`go build ./...` compiles identically).

- `go vet` and `staticcheck` clean on `test/cdcbench/` and on `cmd/` with `-tags cdc_benchmark`; `go build ./...` unaffected.
- End-to-end on Apple M1 Pro (Docker + installed yb-voyager with debezium-server): artifact generation validated for all four workloads (~1 min each, one-time), then
  `go test -tags cdc_benchmark -run '^$' -bench CDCIngest -benchtime 1x ./cmd/`
  runs the full suite from cache in ~31s. All per-run assertions pass: exact event counts (20,000 per workload) and conflict expectations (0 conflicts for the three no-conflict workloads; ~14k detections for `updates-uk-conflict-pairs`).
- Graceful skip verified: without a `yb-voyager` binary on PATH (and no cached artifact), the benchmark skips with instructions rather than failing.
- Not wired into CI (needs Docker + installed voyager/Debezium for generation); documented as a local tier in CLAUDE.md and `test/cdcbench/README.md`.

No.

No. The framework only reads/writes its own generated benchmark artifacts (gitignored `test/cdcbench/artifacts/`); it never touches user migration state. Within its own artifacts it writes both historical variants of the unique-key metaDB key so artifacts replay identically across voyager versions — a benchmark-local concern, not an on-disk format change.

### DIFF
--- a/yb-voyager/cmd/cdc_ingest_bench_test.go
+++ b/yb-voyager/cmd/cdc_ingest_bench_test.go
@@ -1,0 +1,141 @@
+//go:build cdc_benchmark
+
+/*
+Copyright (c) YugabyteDB, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+/*
+CDC ingest benchmark: replays real export-data queue segments through the
+real import streaming path with only TargetDB.ExecuteBatch mocked.
+
+All orchestration (workloads, artifact generation/caching, metrics,
+assertions) lives in test/cdcbench; this file only injects the closures that
+need cmd-package internals. Workloads are sub-benchmarks:
+
+	go test -tags cdc_benchmark -bench CDCIngest -benchtime 1x -count 5 ./cmd/
+	go test -tags cdc_benchmark -bench 'CDCIngest/updates-uk-no-conflict' -benchtime 1x ./cmd/
+
+See test/cdcbench/README.md for workload authoring and knobs.
+*/
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/callhome"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/dbzm"
+	reporterstats "github.com/yugabyte/yb-voyager/yb-voyager/src/reporter/stats"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/tgtdb"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils"
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/utils/sqlname"
+	"github.com/yugabyte/yb-voyager/yb-voyager/test/cdcbench"
+)
+
+func BenchmarkCDCIngest(b *testing.B) {
+	// state shared between Bootstrap and StreamAll within one run
+	var run struct {
+		converter dbzm.StreamingPhaseValueConverter
+		partMap   *utils.StructMap[sqlname.NameTuple, string]
+		state     *ImportDataState
+	}
+
+	cdcbench.Run(b, cdcbench.Hooks{
+		Bootstrap: func(artifactDir string, mock tgtdb.TargetDB) error {
+			// mirrors the importData command's bootstrap, minus the target DB:
+			// channel metadata and vsn state are fresh-migration values, and the
+			// only target interaction left in the streaming path is the mocked
+			// ExecuteBatch.
+			exportDir = artifactDir
+			metaDB = initMetaDB(exportDir)
+			if err := retrieveMigrationUUID(); err != nil {
+				return fmt.Errorf("retrieve migration uuid: %w", err)
+			}
+			sourceDBType = GetSourceDBTypeFromMSR()
+			sqlname.SourceDBType = sourceDBType
+			importerRole = TARGET_DB_IMPORTER_ROLE
+			callhome.SendDiagnostics = false
+			// production default resolution for tables without expression-based
+			// unique indexes; avoids the target-DB query of the "auto" path
+			cdcPartitioningStrategy = "pk"
+			tconf = tgtdb.TargetConf{TargetDBType: YUGABYTEDB, SchemaConfig: "public"}
+			tconf.Schemas = sqlname.ParseIdentifiersFromString(tconf.TargetDBType, tconf.SchemaConfig, ",")
+
+			if err := InitNameRegistry(exportDir, importerRole, nil, nil, &tconf, nil, false); err != nil {
+				return fmt.Errorf("init name registry: %w", err)
+			}
+			msr, err := metaDB.GetMigrationStatusRecord()
+			if err != nil {
+				return fmt.Errorf("get migration status record: %w", err)
+			}
+			tableList, err := getInitialImportTableListForLive(msr.TableListExportedFromSource)
+			if err != nil {
+				return fmt.Errorf("get import table list: %w", err)
+			}
+			run.converter, err = dbzm.NewStreamingPhaseDebeziumValueConverter(tableList, exportDir, tconf, importerRole, sourceDBType)
+			if err != nil {
+				return fmt.Errorf("create streaming value converter: %w", err)
+			}
+			run.partMap, err = getCdcPartitioningStrategyPerTable(tableList)
+			if err != nil {
+				return fmt.Errorf("get cdc partitioning strategy: %w", err)
+			}
+			run.state = NewImportDataState(exportDir)
+
+			tdb = mock
+			prevExporterRole = "" // force conflict-cache re-init on the first event
+			return nil
+		},
+
+		StreamAll: func() error {
+			// the segment loop from streamChanges (live_migration.go), with
+			// fresh-migration channel metadata (what InitLiveMigrationState
+			// inserts and GetEventChannelsMetaInfo returns on a first run)
+			var evChans []chan *tgtdb.Event
+			var doneChans []chan bool
+			chanMeta := map[int]EventChannelMetaInfo{}
+			for i := 0; i < NUM_EVENT_CHANNELS; i++ {
+				evChans = append(evChans, make(chan *tgtdb.Event, EVENT_CHANNEL_SIZE))
+				doneChans = append(doneChans, make(chan bool, 1))
+				chanMeta[i] = EventChannelMetaInfo{ChanNo: i, LastAppliedVsn: -1}
+			}
+			statsReporter := &reporterstats.StreamImportStatsReporter{}
+
+			eventQueue = NewEventQueue(exportDir)
+			for !eventQueue.EndOfQueue {
+				segment, err := eventQueue.GetNextSegment()
+				if err != nil {
+					return fmt.Errorf("get next queue segment: %w", err)
+				}
+				err = streamChangesFromSegment(segment, evChans, doneChans, chanMeta, statsReporter, run.state, run.converter, run.partMap)
+				if err != nil {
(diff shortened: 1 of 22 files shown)


makalaaneesh (commented):
[AI] Automated review — 3 findings (0 critical, 3 warning). See inline comments. Nice, self-contained benchmark framework with a clean cmd/test split and content-hash-keyed artifact caching.

yb-voyager/cmd/cdc_ingest_bench_test.go
```diff
@@ -0,0 +1,141 @@
...
+				err = streamChangesFromSegment(segment, evChans, doneChans, chanMeta, statsReporter, run.state, run.converter, run.partMap)
+				if err != nil {
+					return fmt.Errorf("stream changes from segment %s: %w", segment.FilePath, err)
+				}
+			}
+			return nil
+		},
+
+		CacheDepth: func() int {
+			c := conflictDetectionCache
```
makalaaneesh, line 109: [AI] **Data race on the `conflictDetectionCache` global between the sampler goroutine and `StreamAll`.**

The `CacheDepth` closure reads the package-global pointer `conflictDetectionCache` from the sampler goroutine (spawned in `runWorkload`). Meanwhile `StreamAll` runs `streamChangesFromSegment`, which on the first event assigns that same global inside `initializeConflictDetectionCache` (`cmd/live_migration.go:632`). That is an unsynchronized concurrent read/write of a shared pointer variable.

The `c.Lock()`/`c.Unlock()` here only guard the cache's internal map — they do not guard the pointer read `c := conflictDetectionCache` itself. So `go test -race` on this benchmark will report a race, and strictly it is undefined behavior even without `-race`.

Per `yb-voyager/.cursor/BUGBOT.md`: "Shared mutable state must be protected by a mutex or channel."

Suggested fix: don't read the global from the sampler at all. Publish the cache pointer through an `atomic.Pointer[ConflictDetectionCache]` set in `Bootstrap`/at init time, or have `CacheDepth` read a value that `StreamAll` stores atomically. At minimum, wrap both the write (in the shim) and the sampler read behind a small mutex so the pointer access is synchronized.

yb-voyager/test/cdcbench/artifact.go
```diff
@@ -0,0 +1,438 @@
...
+func ensureUniqueKeyCompatKeys(exportDir string) error {
+	mdb, err := metadb.NewMetaDB(exportDir)
+	if err != nil {
+		return err
+	}
+	oldKey := fmt.Sprintf("%s_%s", metadb.TABLE_TO_UNIQUE_KEY_COLUMNS_KEY, sourceDBExporterRole)
+	newKey := fmt.Sprintf("%s_%s", metadb.TABLE_TO_UNIQUE_INDEXES_KEY, sourceDBExporterRole)
+
+	var flat map[string][]string
+	oldFound, _ := mdb.GetJsonObject(nil, oldKey, &flat)
```
makalaaneesh: [AI] **Swallowed error from `GetJsonObject` misreports real metaDB failures as "key not found".**

```go
oldFound, _ := mdb.GetJsonObject(nil, oldKey, &flat)
...
newFound, _ := mdb.GetJsonObject(nil, newKey, &indexes)
```

`GetJsonObject` returns `(false, nil)` only for a genuinely missing key (`sql.ErrNoRows`); for a query failure or a JSON-unmarshal failure it returns a non-nil error (see `src/metadb/metadataDB.go:292`). Discarding that error means a corrupt/failed read is silently treated as "not found," which then either takes the wrong derive branch or falls through to the `default` case with the misleading message "neither ... found in artifact metaDB."

Per `yb-voyager/.cursor/BUGBOT.md`: "Never silently swallow errors."

Suggested fix: capture and propagate the errors, e.g.

```go
oldFound, err := mdb.GetJsonObject(nil, oldKey, &flat)
if err != nil {
    return fmt.Errorf("read %q from artifact metaDB: %w", oldKey, err)
}
newFound, err := mdb.GetJsonObject(nil, newKey, &indexes)
if err != nil {
    return fmt.Errorf("read %q from artifact metaDB: %w", newKey, err)
}
```

yb-voyager/test/cdcbench/artifact.go
```diff
@@ -0,0 +1,438 @@
...
+	pgStartErr  error
+)
+
+func sourceContainer(b *testing.B) testcontainers.TestContainer {
+	b.Helper()
+	pgOnce.Do(func() {
+		pgContainer = testcontainers.NewTestContainer("postgresql", &testcontainers.ContainerConfig{
+			ForLive: true, // wal_level=logical for CDC
+		})
+		pgStartErr = pgContainer.Start(context.Background())
```
makalaaneesh: [AI] **Shared source-PG container is started but never terminated.**

`sourceContainer` starts `pgContainer` lazily via `sync.Once` and nothing ever calls `pgContainer.Terminate(...)`. The rest of the repo's testcontainer suites clean up explicitly in a `TestMain` (e.g. `src/srcdb/main_test.go:88`, `src/tgtdb/main_test.go`); this framework relies solely on the Ryuk reaper, which only fires after the whole process exits.

Per `yb-voyager/.cursor/BUGBOT.md`: "Integration tests that use testcontainers should clean up their own resources."

Suggested fix: add a `TestMain(m *testing.M)` in the `cdcbench` (or `cmd` shim) package that runs the benchmarks and then terminates `pgContainer` if it was started, so the container is torn down deterministically rather than left to the reaper. Alternatively register cleanup via `b.Cleanup` on the container-owning benchmark, guarded so it only terminates once.

makalaaneesh:
Addressed all review findings in 37e04d801 (propagated to the stacked branches #3641/#3642 via merge):

- **Data race on `conflictDetectionCache`** (gemini + [AI] review): fixed **without production-code changes** — `Bootstrap` now pre-initializes the cache via the real `initializeConflictDetectionCache` and pins `prevExporterRole`, so the global pointer is written before the depth sampler starts and never reassigned mid-stream (benchmark artifacts carry a single exporter role by construction, so the re-init-on-role-change path is unreachable for them). Verified: the full suite passes under `go test -race` with **zero warnings**. The suggested `atomic.Pointer` in production remains a reasonable hardening if the global ever needs concurrent readers in production paths, but this PR deliberately ships no production changes.
- **Source-PG container never terminated** (gemini + [AI] review): the shared container is now terminated via `b.Cleanup` when the suite finishes, with a restartable mutex-based lifecycle (so `-count=N` runs that regenerate artifacts still work) and a sticky start-error so a broken Docker skips instead of re-dialing. Verified: zero containers remain after a generation run.
- **Global logrus left on a closed file** (gemini): the run-log closer now points the logger at `io.Discard` before closing, and `Run` restores the caller's original output at the end.
- **Swallowed `GetJsonObject` errors** ([AI] review): errors are now propagated with the key in context instead of being conflated with "key not found".
- **`time.After` in the poll loop** (gemini): replaced with a reused `Ticker`.

---

yb-voyager/test/cdcbench/artifact.go
```diff
@@ -0,0 +1,467 @@
...
+		return err
+	}
+	return os.WriteFile(path, out, 0644)
+}
+
+// ensureUniqueKeyCompatKeys makes the artifact readable regardless of which
+// unique-key metaDB key the exporting binary wrote (it changed across voyager
+// versions): whichever of the flattened-columns / per-index keys is present,
+// derive the other. Exact for single-column unique keys.
+func ensureUniqueKeyCompatKeys(exportDir string) error {
```
makalaaneesh: i think we should remove this. the fix has landed in master, we don't need this anymore.

yb-voyager/cmd/cdc_ingest_bench_test.go
```diff
@@ -0,0 +1,157 @@
...
+			// lazily on the first event) and pin prevExporterRole so the
+			// streaming loop never reassigns the package-global cache pointer
+			// mid-stream: the framework's depth sampler reads that pointer
+			// concurrently, and writing it during StreamAll would be a data
+			// race. Benchmark artifacts carry a single exporter role by
+			// construction, so the re-init-on-role-change path is unreachable
+			// for them anyway.
+			run.evChans = nil
+			run.doneChans = nil
+			for i := 0; i < NUM_EVENT_CHANNELS; i++ {
```
makalaaneesh: can we just directly call streamChanges after initializing instead of all this duplicate code?

yb-voyager/test/cdcbench/artifact.go
makalaaneesh: Removed in b2b6edc57 — `ensureUniqueKeyCompatKeys` is gone along with its metadb dependency. Artifacts generated by pre-fix binaries now fail loudly at conflict-cache initialization ("unique indexes map not found") instead of being silently adapted; the README and the composite canary's comment document the regenerate-with-current-binary remedy. Validated: full suite passes against artifacts generated by a current binary with no key derivation.

---

yb-voyager/cmd/cdc_ingest_bench_test.go
makalaaneesh: Done in b2b6edc57 — `StreamAll` is now literally `return streamChanges(run.state, run.tableList)`; the duplicated bootstrap (converter, channel metadata, conflict cache, stats reporter, segment loop) is deleted and the real entrypoint runs end to end, including `waitForDebeziumStartIfRequired`, the cutover-detected check, and `InitLiveMigrationState`.

What made this possible: the streaming bootstrap keeps live-migration bookkeeping in target-DB tables via `tdb.Query/QueryRow/Exec/WithTx` (returning concrete `*sql.Rows`/`*sql.Row`), so the mock now answers those from an in-memory store holding the two `ybvoyager_metadata` tables with fresh-migration values (`mock_metadata.go`); `ExecuteBatch` stays a no-op.

Per our discussion, the depth sampler stays on its 100ms ticker; its one-time unsynchronized read of the cache pointer during first-event init is accepted (single-exporter-role artifacts — the pointer is written once and never changes mid-run) and documented in the shim + README, including the expectation that `go test -race` flags it.

Validated: suite metrics match the previous baseline within noise (132k / 1.9k / 31.7k / 1.1k events/s, conflict counts unchanged); propagated to #3641/#3642, where the POC also passes through the `streamChanges` path.

---

priyanshi-yb (approved):
LGTM

PR #3641, opened 2026-07-02
cdcbench: 28-workload OLTP catalog for the CDC ingest benchmark

makalaaneesh:
Stacked on #3640 (benchmark framework, now merged into main).

### Describe the changes in this pull request

Grows the CDC ingest benchmark from its 4 founding workloads to a **28-workload catalog** modeling common and corner-case customer OLTP shapes:

- **Taxonomy, enforced at registration**: `oltp-` (realistic patterns), `schema-` (shape probes: index count, row width, composite/partial/NULL key structure), `edge-` (degenerate op mixes), `conflict-` (engineered real conflicts). A workload's registry name, its `testdata/<name>/` SQL directory, and the benchmark sub-name (`-bench 'CDCIngest/<name>'`) are the same string by construction.
- **Highlights**: order lifecycle, session churn, job queue, ledger, multi-table checkout, UK+plain co-streaming (collateral-damage measurement), skewed hot rows, state-machine transition log (composite + partial unique index), wide rows, all four conflict types documented in `conflictDetectionCache.go`, `NULLS NOT DISTINCT` slot-handoff.
- **Detection-semantics probes with asserted expectations**: three workloads assert today's known NULL false positives (`oltp-insert-null-then-fill`, `schema-composite-uk-null-component`, `schema-partial-index-inactive-rows`) and flip to zero-conflict expectations when the semantics are fixed, while `conflict-nulls-not-distinct` pins the behavior such a fix must preserve. The composite pair (`shared-tenants` vs `unique-tenants`, same DML, different seed) discriminates per-column vs tuple comparison — shared-tenants asserts zero conflicts, verifying the multi-column unique index support end to end.
- **Adapts to #3652** (unique-index discovery moved to a live target-DB query): artifact generation captures the unique-index map from the source schema into `manifest.json`, and the benchmark's mock serves it via `GetTableToUniqueIndexesMap`.

Every run asserts its workload's exact event count and conflict expectation; metrics reported per workload are events/s, conflicts, batches, and conflict-cache depth (sampled).

### Sample output

Apple M1 Pro (10 cores, 16GB RAM), current main-based import code, 20k events/workload, `ExecuteBatch` mocked:

```
--- cdcbench summary ---
WORKLOAD                               RUNS  EVENTS/S  CONFLICTS/RUN  BATCHES/RUN  CACHE DEPTH AVG/MAX  TIME/RUN
oltp-checkout-multi-table              1     5,899     0              200          860 / 1,361          3.39s
[...]

PR #3642, opened 2026-07-02
POC: unique-key-value-indexed conflict detection cache + targeted flush

makalaaneesh:
Stacked on #3641 (workload catalog) → #3640 (benchmark framework). **Draft: proof of concept**, opened to share the design and benchmark evidence; needs productionization review before being a real candidate.

Replaces the conflict detection cache's O(in-flight events) per-event scan with an inverted index keyed by `(table, unique index, encoded before-image UK tuple)`, making the conflict check O(#unique indexes) hash lookups — independent of how many events are in flight. On detection, only the blocking events' channels are flushed (gathered across all blockers so multiple blockers resolve in one batch round-trip), instead of guillotining all 100 channels' batches.

Semantics deliberately preserved bug-for-bug: all-index-columns-present requirement, nil==nil equality (sentinel-encoded, length-prefixed tuples), same-PK exclusion, and the op-based export-from-target (fall-back/fall-forward) path per table. Cache entries shrink from full event copies (~7KB worst case) to slim entries (vsn, chanNo, PK, index keys). The failpoint instrumentation for conflict count/validation tests is preserved.

Why: the 25-workload benchmark catalog (#3641) shows every ordinary customer OLTP pattern on a unique-key table throttled to ~1–2k events/s by the scan — the cache depth is set by the 2s batch-flush timer, and every update/insert pays a full scan under one mutex to prove the absence of conflicts. Results on the full catalog (Apple M1 Pro, 20k events/workload, ExecuteBatch mocked):

| workload group | main | POC |
| --- | --- | --- |
| oltp-order-lifecycle | 1,244 ev/s | 128,276 ev/s (103x) |
| oltp-job-queue | 1,506 | 154,449 (103x) |
| oltp-session-churn | 2,334 | 166,260 (71x) |
| oltp-mixed-uk-and-plain-tables | 5,571 | 199,964 (36x) — UK table no longer throttles co-streamed plain tables |
| schema-two-unique-indexes | 1,058 | 110,438 (104x) |
| schema-composite-uk-no-conflict | 1,094 | 125,329 (115x) |
| edge-all-updates | 1,073 | 96,266 (90x) |
| conflict-* (5 real-conflict workloads) | 27k–44k | 77k–98k (2.2–3.0x), and every conflict count becomes exact (one detection per engineered conflict; no rescan re-detections) |
| no-cache paths (ledger, all-inserts, all-deletes, no-UK control) | unchanged | unchanged |

With a simulated 100ms target batch-commit latency, main's rates are unchanged (pipeline-bound) while POC rates respond to the target (DB-bound) — i.e., the import pipeline stops being the bottleneck.

Known TODOs before this graduates from POC:
- skewed-hot-rows runs at 36k (vs ~100k peers): a hot row's in-flight updates share one UK tuple, so lookups iterate a large same-PK hit set only to exclude it — needs same-PK-aware indexing
- fall-back/fall-forward path kept as an O(per-table in-flight) scan; needs its own review
- a differential correctness test (scan vs index must agree on every conflict over the same artifact)
- NULL and partial-predicate false positives are intentionally preserved (tracked by canary-null-fill and canary-partial-index-inactive-rows in #3641); fixing them is separate, semantic work

No interface changes. Live-migration import throughput on tables with unique keys improves substantially; conflict logging remains but each real conflict is now logged once instead of once per rescan.

- Full `cmd` unit suite passes (including the multi-column unique index validation tests and conflict-cache white-box tests).
- Full 25-workload cdcbench catalog passes with all per-run assertions (exact event counts; conflict expectations); conflict counts are exact per engineered conflict.
- Delayed-target runs (CDCBENCH_EXEC_DELAY_MS=100) confirm behavior under realistic commit latency.
- Not yet run: live-migration integration/failpoint test tiers; fall-back/fall-forward end-to-end.

No.

No. The cache is in-memory only; metaDB keys and event queue formats are unchanged.


makalaaneesh:
closing; separate implementation already merged to main

PR #3648, opened 2026-07-06
Add the schema diff engine for tables and columns (schemadiff) (tables + columns)

ShivanshGahlot:

Adds the schema **diff engine** (`schemadiff` package) for schema-drift detection:
given two captured schema snapshots (from the `schemasnapshot` library already in
main), it computes their structural differences as a sorted list of typed
`Difference` values.

- **Scope of findings** — table-level and column-level changes, the 15 v1
  `DiffType`s (add / drop / rename / schema-move / type / nullability / default /
  partition / inheritance).
- **Structured findings** — each `Difference` carries the changed object's rich
  identity on each side (`ObjectA`/`ObjectB`, either a schema-level `ObjectRef` or
  a table-scoped `TableScopedObjectRef`) rather than a pre-rendered string, plus
  decomposed classification facets — `Operation` (ADDED/DROPPED/CHANGED),
  `ObjectType` (TABLE/COLUMN), and `Attribute` (NAME/TYPE/NULLABILITY/…) — with
  `SideAValue`/`SideBValue` holding the before/after payloads. `DiffType` is kept
  as the single-string form. Rendering is deferred and engine-aware
  (`ForKey`/`ForDisplay(dbType)`), which the future source-vs-target comparison
  needs.
- **ID vs name matching** — objects are matched by their stable catalog ID (PG
  `pg_class.oid`; column `{tableOID}:{attnum}`) when the two snapshots are the
  **same engine** and that engine exposes stable IDs (**PostgreSQL / YugabyteDB**),
  so a rename surfaces as a single `*_NAME_CHANGED` instead of add+drop. Otherwise
  — a cross-engine diff, or an engine without stable per-object IDs (e.g. MySQL) —
  matching falls back to the case-sensitive, unquoted dot-joined name key
  (`ForKey`; see its doc for the dot-in-identifier collision limitation, an
  accepted v1 edge). ID presence is a property of the engine, not the individual
  object, so this is a single capability check rather than a per-object scan.
  _(TODO tracked: tighten the "same engine" gate to "same instance" before
  enabling cross-instance source↔target diffs — different servers of the same
  engine have independent OID spaces.)_
- **Scope filter (`FilterByScope`)** — narrows results by table and object type
  (the include/exclude lists a future `--table-list` / `--object-type-list` resolve
  into). `COLUMN` is a first-class object-type selector (a column change is picked
  by `--object-type-list=COLUMN`, not swept in under `TABLE`); table-scoping for
  `--table-list` is orthogonal — a finding's anchor table is derived from its
  identity. A rename/move **alias** map (matching a finding anchored to a renamed
  table on either its old or new identity) is implemented but **temporarily
  disabled**, pending a decision on cross-window alias behaviour; with it off, a
  renamed table's findings match only their as-emitted anchor (list both the old
  and new name to see the full history). The alias code is preserved in place for
  re-enable.
- **`Differ` façade** — a config-struct wrapper over the pure diff functions; the
  extension point for future post-diff knobs (e.g. ignore-rules).
- **Noise suppression** — per-column findings for a wholly added/dropped table are
  suppressed; the `TABLE_ADDED`/`TABLE_DROPPED` finding carries the whole table
  payload instead (columns plus kind / partition / inheritance metadata).
- **`schemasnapshot` identity type** — adds `TableScopedObjectRef{Table, Name}` for
  table-scoped objects (columns today, indexes/constraints later); `Column` embeds
  it via named composition, so field access and rendering stay unchanged and the
  persisted snapshot JSON shape is preserved.
- **Library-only** — no command wiring here; the capture hooks and the
  `detect-drift` command land in follow-up branches. Builds on the `schemasnapshot`
  capture library (#3614) and case-sensitive identity work (#3636), both merged.

No user-facing changes — internal library only:

- No command-line changes.
- No configuration changes.
- No installation changes.
- No report changes — the command and drift report that consume this engine land
  in follow-up PRs.

- **Unit tests** (`-tags unit`) across the package — `diff_test.go`,
  `columns_test.go`, `tables_test.go`, `filter_test.go`, `differ_test.go` —
  covering:
  - ID / name matching, rename/move detection, drop-and-recreate, and same-engine
    fallback to name matching for an ID-less engine (MySQL);
  - lifecycle suppression and whole-table add/drop payloads;
  - deterministic ordering;
  - scope include/exclude (incl. `COLUMN` as a first-class object type) and
    no-anchor handling;
  - `Difference` facet decomposition — one scenario emits all 15 `DiffType`s and
    verifies each finding's `(Operation, ObjectType, Attribute)` against a
    `Type`-string oracle.
  - _(The rename/move alias tests are `t.Skip`'d while the alias is disabled; they
    re-enable together with the alias.)_
- **Integration test** (`integration_test.go`, `-tags integration`) — runs the
  engine end-to-end against a live PostgreSQL via testcontainers.
- `go test -tags unit ./src/schemadiff/... ./src/schemasnapshot/...` passes;
  `go vet` is clean under both the `unit` and `integration` tags.

No.

No. The diff engine is a pure in-memory computation over already-captured
snapshots. The one `schemasnapshot` change — `Column` embedding the new
`TableScopedObjectRef` — preserves the on-disk snapshot JSON shape (same keys;
verified by round-trip), so it is not an upgrade concern. (Snapshot storage landed
with #3614.)

---


yb-voyager/src/schemadiff/tables.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	// Pass 2: name-based reconciliation of the residue.
+	diffs = append(diffs, diffTablesByName(residueA, residueB, matchByID, a.DatabaseType, b.DatabaseType, colsByTableA, colsByTableB)...)
+
+	return diffs
+}
+
+// diffTablesByName reconciles the tables left unmatched by the ID pass, keyed on
+// the collision-safe, case-sensitive, per-part-quoted canonical key returned by
```
makalaaneesh: [AI] **Comments describe `ForKey` as "collision-safe" and "per-part-quoted", contradicting its documented behavior.**

The name-match pass keys tables and columns on `ObjectRef.ForKey(dbType)` / `Column.ForKey(dbType)`, and the surrounding comments justify the design by calling this "the collision-safe, case-sensitive, per-part-quoted canonical key" (also at `tables.go:29`, `tables.go:118`, and `columns.go:20-21/85/93`).

But `ForKey`'s own doc in `schemasnapshot/snapshot.go` says the opposite: the key is *unquoted, dot-joined*, and explicitly **not** collision-safe — "an identifier that itself contains a dot can collide with a different (schema, name) split, and reserved words / embedded quotes are not escaped."

Concrete failure: `("public", "a.b")` and `("public.a", "b")` both produce the key `public.a.b`, so they collide in `byNameA`/`byNameB` and one table is silently dropped from the diff. The limitation may well be acceptable for v1, but the comment asserts the inverse as fact in exactly the spot where a maintainer decides whether name collisions need guarding.

Suggested fix: reword to match reality, e.g. "keyed on the case-sensitive, unquoted dot-joined canonical key `ForKey(dbType)`; see its doc for the dot-collision limitation" — and drop the "collision-safe" / "per-part-quoted" claims at `tables.go:29/118` and `columns.go:20-21/85/93`.

---

makalaaneesh (commented):
[AI] Automated review — 1 finding (0 critical, 1 warning). See inline comment. New `schemadiff` library; matching/diffing logic and unit coverage are solid (rename detection, drop-and-recreate, hybrid ID→name residue, cross-engine gating, nil anchors, deterministic ordering). The one warning is a comment/behavior mismatch; a few scope/YAGNI observations on the not-yet-consumed `AnchorTable` field, `Differ` façade, and object-type filter were left off as they're questions for the author rather than defects.

---

yb-voyager/src/schemadiff/columns.go
```diff
@@ -0,0 +1,153 @@
...
+// dropped/added immediately avoids a spurious drop+add when an ID is missing on
+// one side. The name pass guards the inverse case: two same-named columns with
+// real but DIFFERENT IDs are a genuine drop-and-recreate and stay an add+drop
+// (see nameMatchAllowed).
+func diffColumns(a, b *schemasnapshot.SnapshotContent) []Difference {
+	var diffs []Difference
+
+	// Same engine ⇒ IDs are stable/comparable; match by ID. Different engine ⇒
+	// fall back to name matching.
+	matchByID := a.DatabaseType == b.DatabaseType
```
makalaaneesh: what ID are we using here for columns? 
From a quick check, columns don't have IDs, they only have attnums

makalaaneesh: should this also include a.side = b.side? what if we migrate PG(source) -> PG(target)? 
database type is equal but they will have different IDs because they are different systems

yb-voyager/src/schemadiff/columns.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// Pass 1: ID-based matching for columns that carry a usable stable ID.
+	// Columns without one start life in the name-pass residue.
+	byIDA := make(map[string]schemasnapshot.Column)
+	byIDB := make(map[string]schemasnapshot.Column)
+	var residueA, residueB []schemasnapshot.Column
```
makalaaneesh: when would we need residueA/B? Why/when will ID not be filled?

yb-voyager/src/schemadiff/columns.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, c := range residueA {
+		byNameA[keyA(c)] = c
+	}
+	for _, c := range residueB {
+		byNameB[keyB(c)] = c
+	}
+
+	matchedB := make(map[string]bool, len(byNameB))
+	for key, cA := range byNameA {
+		if cB, ok := byNameB[key]; ok && nameMatchAllowed(matchByID, cA.ID, cB.ID) {
```
makalaaneesh: why do we need `nameMatchAllowed`? we're already inside the function `diffColumnsByName` ?

yb-voyager/src/schemadiff/columns.go
```diff
@@ -0,0 +1,153 @@
...
+//     within the same engine. Same engine ⇒ match by ID, which lets a rename
+//     surface as COLUMN_NAME_CHANGED instead of an add+drop pair; different
+//     engine ⇒ fall back to name matching.
+//  2. Name pass — columns left unmatched by the ID pass (no usable ID, or an ID
+//     present on only one side) are reconciled by the composite key
+//     Column.ForKey(dbType).
+//
+// Falling through to the name pass instead of declaring ID-unmatched columns
+// dropped/added immediately avoids a spurious drop+add when an ID is missing on
+// one side. The name pass guards the inverse case: two same-named columns with
```
makalaaneesh: I would think either ID or name is simpler. What is this scenario where we expect ID to be missing on one side.

yb-voyager/src/schemadiff/diff.go
```diff
@@ -0,0 +1,136 @@
...
+// WHY: every column of a wholly added/dropped table also appears as COLUMN_ADDED/
+// COLUMN_DROPPED — pure noise once TABLE_ADDED/TABLE_DROPPED already conveys the
+// full change.
+//
+// A renamed table is matched (emits TABLE_NAME_CHANGED, not TABLE_ADDED/DROPPED),
+// so its real column changes are actual mutations and are preserved.
+//
+// Only COLUMN_ADDED under TABLE_ADDED and COLUMN_DROPPED under TABLE_DROPPED are
+// suppressed; every other finding type passes through unchanged.
+func suppressLifecycleTableColumns(diffs []Difference) []Difference {
```
makalaaneesh: Can we push this into diffColumns itself? 

We probably should have discussed this in the earlier PR, but it seems to me that `columns` field should be inside `Table` rather than alongside it. 
let's discuss this on call. If you model it that way, then you will never need to have this function because you will match tables first before you diff columns.

yb-voyager/src/schemadiff/filter.go
```diff
@@ -0,0 +1,241 @@
...
+//
+// Scope is permissive and total — it never errors:
+//   - If an include and its exclude are both set, exclude wins (includes apply
+//     first, excludes last).
+//   - An entry that matches nothing is a silent no-op.
+//
+// Flag-level policy — e.g. --table-list and --exclude-table-list being mutually
+// exclusive — is the command's to enforce, so FilterByScope stays pure for any caller.
+type Scope struct {
+	Tables             []schemasnapshot.ObjectRef // empty = all; matched against AnchorTable
```
makalaaneesh: Let's have "include" in the name so that it's explicit. `IncludeTables`, `IncludeObjects`

makalaaneesh (commented):
high level structure looks good! 
added some comments. 
Yet to look at some files (filter, tables)

yb-voyager/src/schemadiff/tables.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	if !objectRefSetEqual(tA.InheritedBy, tB.InheritedBy) {
+		diffs = append(diffs, newDifference(TableInheritedByChanged, tA.ObjectRef, &tA.ObjectRef, "", cloneObjectRefs(tA.InheritedBy), cloneObjectRefs(tB.InheritedBy)))
+	}
+
+	return diffs
+}
+
+// partitionParentEqual compares two *ObjectRef values for equality.
```
makalaaneesh, line 99: every function below this is a helper? (or maybe better defined inside schemasnapshot? because they are all operations on ObjectRef)

yb-voyager/src/schemadiff/tables.go
```diff
@@ -0,0 +1,262 @@
...
+// limitations under the License.
+
+package schemadiff
+
+import "github.com/yugabyte/yb-voyager/yb-voyager/src/schemasnapshot"
+
+// diffTables computes table-level differences between snapshots a and b using a
+// hybrid two-pass match:
+//
+//  1. ID pass — tables with a usable stable ID (OID) are matched by ID. Enabled
```
makalaaneesh: again, not clear why we need two passes. Let's discuss on call

yb-voyager/src/schemadiff/tables.go
ShivanshGahlot: Fixed the comment

yb-voyager/src/schemadiff/columns.go
ShivanshGahlot: Yeah correct. They dont have as such OIDs. What they have is attnum. And we are using {ParentTableOID}:{attnum} as the unique identifier for the column. I have verified locally that renaming a column does not change its attnum and dropping a column still reserves its attnum and any new column added is assigned a new attnum regardless. pg docs also somewhat say this:

> attnum int2
> 
> The number of the column. Ordinary columns are numbered from 1 up. System columns, such as ctid, have (arbitrary) negative numbers.

> attisdropped bool
> 
> This column has been dropped and is no longer valid. A dropped column is still physically present in the table, but is ignored by the parser and so cannot be accessed via SQL.

According to this the dropped column is still physically present but ignored and consequently the attnum will be retained.

ShivanshGahlot: This article also explain that attnum is permanent: https://www.bytebase.com/blog/why-its-hard-to-add-a-column-in-the-middle-of-postgres-table/#the-holes-that-never-fill

ShivanshGahlot: Yep. I think thats correct. Since we are adding amp too, this becomes more important.

ShivanshGahlot: Issue is that Side is in the SnapshotHeader and we only pass the SnapshotContent to the Diff. It will require a lot of refactoring at the backend to surface it in the Content. Or we pass the whole SchemaSnapshot with both Header and Content which will change the signatures of all the functions. This side issue is inconsequential till we have source vs target though.
So what do you think I should do here? The second option of passing the entire SchemaSnapshot seems easier. Should I do that?

yb-voyager/src/schemadiff/columns.go
ShivanshGahlot: So residueA/B basically hold the columns(/tables in the tables.go) that go to the name matching function. 
1. When matchId==false i.e. cross database type, all the columns/tables are routed here.
2. In case of same database type match, columns on side A whose ID match nothing on side B or vice versa (basically a new column added/dropped) falls into residue. Or if lets say column `purchases` was dropped and then same name column `purchases` was added again to the same table, the ids wont match on side A and B and these columns will fall into residue A/B and propagated as genuine add and drops.
3. In case of PG the IDs will technically never be null. But Claude has defensively added the c.ID != "" check below for a future same database type provider that lacks a stable per-column ID. I am not fully conviced with this check either but since its defensive, I did not remove it.

yb-voyager/src/schemadiff/columns.go
ShivanshGahlot: Yeah the idea of the name pass is more or less for the cross engine case. But I see your point here. I feel it will be easier to read and to maintain if we just have a switch for same db/side vs other cases and in that we can launch either id based pass or name based pass. This was designed keeping in mind that StableIdentity field we have removed. But I guess that also should be solvable by just doing name pass directly.

ShivanshGahlot: I am redoing this pass. Isnt a big change but it makes the structure easier to understand and read too.

ShivanshGahlot: So I have extracted out the common code from tables and columns both into match.go and using generic functions over there. Both columns.go and tables.go now call that and we select the keys for the matchByKey function: either IDs or the names.

yb-voyager/src/schemadiff/columns.go
ShivanshGahlot: Not using this residue logic now. You can ignore. Simplified approach is in match.go

yb-voyager/src/schemadiff/match.go
```diff
@@ -0,0 +1,88 @@
...
+//
+// The predicate is "IDs usable", NOT merely "same engine": a same-engine source
+// without IDs must still match by name, else every object collides on "".
+func chooseMatchKeys[T any](
+	dbTypeA, dbTypeB string,
+	as, bs []T,
+	idOf func(T) string,
+	nameKey func(obj T, dbType string) string,
+) (keyA, keyB func(T) string) {
+	if dbTypeA == dbTypeB && allHaveID(as, idOf) && allHaveID(bs, idOf) {
```
ShivanshGahlot: The question for comparing Sides here still remains.

yb-voyager/src/schemadiff/diff.go
ShivanshGahlot: Yeah valid. I had deliberately kept it flat as I did not want the diffing logic to become too nested. We would have to consider wether Indexes, Contraints etc also should be nested in Tables. Consequence of keeping this flas is this one function. But this will only happen for columns anyways. Lets discuss on call.

yb-voyager/src/schemadiff/filter.go
ShivanshGahlot: Done

yb-voyager/src/schemadiff/columns.go
ShivanshGahlot: The restructure does not need this

yb-voyager/src/schemadiff/columns.go
```diff
@@ -0,0 +1,153 @@
...
+func diffColumnsByName(residueA, residueB []schemasnapshot.Column, matchByID bool, dbTypeA, dbTypeB string) []Difference {
+	var diffs []Difference
+
+	// identity seam — the single point where a column's name-match key is
+	// derived. Today: ForKey (case-sensitive, collision-safe, per-part-quoted
+	// composite key). When cross-engine diffing lands, this becomes a
+	// NameRegistry handle — change only here; the match loop stays generic
+	// over the key string.
+	keyA := func(c schemasnapshot.Column) string { return c.ForKey(dbTypeA) }
+	keyB := func(c schemasnapshot.Column) string { return c.ForKey(dbTypeB) }
```
priyanshi-yb: nit: only one function for a columnKey should be enough.

yb-voyager/src/schemadiff/filter.go
```diff
@@ -0,0 +1,241 @@
...
+
+// buildTableRenameAliases maps each table identity to the identities it aliases,
+// from TABLE_NAME_CHANGED (new name) and TABLE_SCHEMA_CHANGED (new schema)
+// findings, so either side of a rename/move can stand in during table matching.
+//
+// A table may change both name and schema in one interval; those arrive as two
+// findings sharing the same side-A anchor, so we group by the old ref and rebuild
+// the full new ref from whichever parts changed. Aliases accumulate ([]ObjectRef)
+// so chains like a→b, b→c don't clobber each other.
+func buildTableRenameAliases(diffs []Difference) map[schemasnapshot.ObjectRef][]schemasnapshot.ObjectRef {
```
priyanshi-yb, line 122: I don't understand the use of creating aliases

This is what I understood about how are we handling the rename case, for ex-
Let's say we have two – Snapshot 1 and Snapshot 2
Snapshot 1 has table 1  for x oid and  
Snapshot 2 has table 2 for x oid
So we know table1 is renamed to tabl2
Now you have a diff of -` Table name changed - table1->table2`
and there is one more diff on the table: the column's default value is changed, so another diff - Column default changed for anchor table 1 (As I can see from diffTables we are calling diffColumns which is always putting snapshot 1's table name in Anchor when two columns are compared) 
So now we have these two diffs
```
diff 1 - table name changed: table1->table2 Anchor: table1
diff2 - column default value changes: table1.col1, val1->val2 Anchor: table1
```

Now in the diffs, the anchor always has the old name in a table rename scenario, so if table1 is being excluded in the list, then we will filter out the diff. 
Where are we using this alias handling, or in which cases its useful?

yb-voyager/src/schemadiff/differ.go
```diff
@@ -0,0 +1,43 @@
...
+// callers who want the raw, unfiltered diff.
+type Differ struct {
+	cfg Config
+}
+
+// NewDiffer constructs a Differ with the given Config. A zero Config yields a
+// pass-through Differ whose Diff output matches the package-level Diff function.
+func NewDiffer(cfg Config) *Differ {
+	return &Differ{cfg: cfg}
+}
```
priyanshi-yb, line 37: I think we had discussed this in the initial review of the design, but I can't recall - what is the use of a struct differ other than just having a config and doing diff? I feel this isn't really required, as we can also just have a function that can take scope and snapshots and do diff first, and then filter by scope if required.
Or have two functions where the caller can decide which one to call based on whether we have a scope defined or not. (Diff, DiffAndFilterByScope).

yb-voyager/src/schemadiff/differ_test.go
```diff
@@ -0,0 +1,155 @@
...
+	b := snapWithTables(ordersB, legacyB)
+
+	scope := Scope{IncludeTables: []schemasnapshot.ObjectRef{ref("public", "orders")}}
+	d := NewDiffer(Config{Scope: scope})
+	got := d.Diff(a, b)
+
+	// The Differ result must equal FilterByScope(Diff(a,b), scope).
+	want := FilterByScope(Diff(a, b), scope)
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("scoped Differ diverged from FilterByScope(Diff(...))\ngot:  %v\nwant: %v", got, want)
```
priyanshi-yb: I am not sure what is being tested with these, as got and want are essentially calling the same function, so if anything regress in the filter with the same scope, it won't be caught here, right?

yb-voyager/src/schemadiff/diff.go
```diff
@@ -0,0 +1,86 @@
...
+type QualifiedObject struct {
+	Key     string
+	Display string
+}
+
+// Difference describes a single detected schema change between two snapshots,
+// where A is the old (side-A) snapshot and B is the new (side-B) snapshot. It is
+// the unit element of the slice Diff returns. One shape covers added, dropped, and
+// changed findings; which fields are populated depends on Type — see each field.
+type Difference struct {
```
makalaaneesh, line 37:
```
Fields: 
- DiffType(TABLE_ADD, COLUMN_TYPE_CHANGED)
- op,(add/drop/modify) 
- object_type, object_name (col, <col_name>) (table, <table_name>) (index, <index_name>)
- table_name (depends if we want table-wise grouping)
-  field/attribute = (when op=modify) (NOTNULL)
- before
- after
```

yb-voyager/src/schemadiff/filter.go
```diff
@@ -0,0 +1,245 @@
...
+
+// ObjectType is the user-facing object-type selector for Scope filtering, matching
+// the values accepted by --object-type-list. v1 emits and filters TABLE and COLUMN
+// findings, so both are declared; the remaining selectors are re-enabled as the
+// engine starts emitting their findings.
+//
+// COLUMN is its own selector: a column change is filtered directly by
+// --object-type-list=COLUMN (it is NOT swept in under TABLE). Table-scoping is
+// orthogonal — a column finding still anchors to its host table for --table-list.
+type ObjectType string
```
makalaaneesh: I wonder if this is also an enum that should be defined in the `ObjectRef` / `TableScopedRef` ? That will make it easier to filter.

yb-voyager/src/schemasnapshot/snapshot.go
```diff
@@ -135,14 +134,40 @@ type Table struct {
...
 	InheritsFrom      []ObjectRef `json:"inherits_from,omitempty"`      // legacy table inheritance (INHERITS): parent table(s) this table inherits from; can be multiple. Distinct from declarative partitioning.
 	InheritedBy       []ObjectRef `json:"inherited_by,omitempty"`       // legacy table inheritance (INHERITS): tables that inherit from this one.
+	Columns           []Column    `json:"columns,omitempty"`            // columns of this table, in attnum order.
+}
+
+// TableScopedRef is the identity of an object that lives under a table — a column
+// today, an index or constraint later. Table is the parent table's (schema, name);
+// Name is the object's own name within that table. A table-scoped object has no
+// independent schema — its schema is its table's schema.
+type TableScopedRef struct {
```
makalaaneesh: If we're calling it `ObjectRef` then this should be `TableScopedObjectRef` for consistency?

yb-voyager/src/schemadiff/match.go
makalaaneesh: I think i may have added a comment before as well, but reiterating: why would they not have ID? If they don't have IDs, we should fail rather than just silently fallback to name

yb-voyager/src/schemadiff/columns.go
makalaaneesh: pls add comments if not already added

makalaaneesh: >this side issue is inconsequential till we have source vs target though.


My recommendation would be to do it sooner rather than later so that we wouldn't have to do a lot of refactors. 
We can do it in a separate PR , but pls do add a TODO in code as well as create a task

yb-voyager/src/schemadiff/filter.go
makalaaneesh: I also don't quite understand the alias requirement.

yb-voyager/src/schemadiff/columns.go
ShivanshGahlot: Not using this anymore

yb-voyager/src/schemadiff/match.go
ShivanshGahlot: I just completely removed this logic. Claude had unnecessarily kept it when its not even possible to happen.

yb-voyager/src/schemadiff/columns.go
ShivanshGahlot: Added a TODO for it.

yb-voyager/src/schemadiff/differ_test.go
ShivanshGahlot: Yeah fixed this in the test. Now we are checking against a fixed expected value.

yb-voyager/src/schemadiff/differ.go
ShivanshGahlot: So we were avoiding this single function with multiple parameters or multiple different function approach only. We wanted to standardise the differ so that it is easily extensible with other knobs in the future. Like Ignore Rules will also come in the future.

ShivanshGahlot: We can discuss this on call once

yb-voyager/src/schemadiff/filter.go
ShivanshGahlot: Hmm I feel its better to keep it separate. ObjectRef/TableScopedRef start from the capture part. Over there they are already embedded in the Column/Table struct. Added ObjectType too to the ObjectRef is kind of redundant. And this ObjectType is basically created when creating a diff, a logic that has the context of what the object actually is. It just acts as a knob for filtering and it might be cleaner to keep it as a top level field in the Difference struct rather than a nested field.

yb-voyager/src/schemasnapshot/snapshot.go
ShivanshGahlot: Yeah makes sense. Done.

yb-voyager/src/schemadiff/tables.go
ShivanshGahlot: Keeping them in place since they are only used here

yb-voyager/src/schemadiff/filter.go
ShivanshGahlot: Disabled the alias logic

makalaaneesh (approved):
LGTM! pls make sure pr description is up to date with all the changes.

PR #3652, opened 2026-07-07
Fetch Unique index information from target db and use that for PG->YB migrations

priyanshi-yb:

Live-migration unique-key conflict detection previously sourced its unique-index metadata from the **source** database: during `export data` the indexes were queried from the source and persisted to metaDB (`table_to_unique_indexes_key`), then read back by the importer. That is the wrong source of truth — the **import target** is the DB that actually enforces the unique constraints, and the source/target index sets can differ. This PR removes the source-side path entirely and instead fetches unique indexes **live from the import target** when the conflict-detection cache is initialized during the streaming phase.

Key changes:
- Adds `GetTableToUniqueIndexesMap` to the `TargetDB` interface, implemented for PostgreSQL/YugabyteDB via a catalog query (unique constraints + unique indexes), with leaf-partition indexes merged into their root table (import events only reference the root). The Oracle target returns an empty map.
- `initializeConflictDetectionCache` now calls `tdb.GetTableToUniqueIndexesMap(importTableList)`  instead of reading metaDB; `importTableList` is threaded through  `streamChanges` → `streamChangesFromSegment` `initializeConflictDetectionCache`.
- Deletes all source-side unique-index code: `srcdb` PostgreSQL/YugabyteDB/Oracle/MySQL `GetTableToUniqueIndexesMap`/`MergeUniqueIndexes` (and their tests), `saveTableToUniqueIndexesMapInMetaDB` in `exportData.go`, `getTableToUniqueIndexesMapFromMetaDB` in `live_migration.go`, and the `TABLE_TO_UNIQUE_INDEXES_KEY` metaDB constant.
- For Oracle sources, live migration now forces `PARTITION_BY_TABLE` so all events of a table run sequentially on one channel, making unique-key conflicts impossible (we do not fetch unique indexes for Oracle). Correspondingly, `--cdc-partitioning-strategy` is now restricted to PostgreSQL sources.

- `--cdc-partitioning-strategy` is now rejected for non-PostgreSQL sources with the error: `--cdc-partitioning-strategy is only supported for PostgreSQL source` (previously it was accepted regardless of source type).
- No new flags, no installation changes, no report changes.

- New target-side integration tests: `TestPostgresTargetGetTableToUniqueIndexesMap` (`tgtdb/postgres_test.go`) and the YugabyteDB equivalent in `tgtdb/yugabytedb_test.go`, covering unique constraints, unique indexes, and partitioned-table merging.
- New end-to-end live-migration test `live_migration_uk_conflict_test.go` exercising unique-key conflict handling with the target-fetched indexes.
- Removed now-obsolete source-side tests (`srcdb/postgres_test.go`, `srcdb/oracle_test.go`,`srcdb/yugabytedb_test.go` unique-index cases).
- Verified `go build ./...` and `go vet ./...` with no import cycles.

No.

Yes — **MetaDB**. The `table_to_unique_indexes_key` entry is no longer written or read. Within a single migration, this is safe because the new importer fetches indexes live from the target instead of relying on the exported value. The only mixed-version caveat is a new-binary `export data` (which no longer writes the key) followed by an old-binary `import data` (which would look for the key) — but the recommended flow uses the same binary version across phases, so no on-disk migration is required.


yb-voyager/cmd/live_migration.go
```diff
@@ -187,6 +187,17 @@ func getCdcPartitioningStrategyPerTable(tableNames []sqlname.NameTuple) (*utils.
 		return tableToPartitioningStrategyMap, nil
 	}
 
+	if sourceDBType == ORACLE {
```
makalaaneesh, line 190: [AI] **Oracle→YB partitioning is forced to `PARTITION_BY_TABLE` before the metaDB-persisted strategy is read — breaks resume across a mid-migration upgrade.**

This new `if sourceDBType == ORACLE` branch (lines 190–199) is in the target-importer path and returns *before* the `metaDB.GetImportDataStatusRecord()` lookup (line 201) that honors a previously persisted `TableToCDCPartitioningStrategyMap`.

Before this PR, an Oracle→YB target importer took the same `auto` path as PG and persisted `PARTITION_BY_PK` for normal tables. Failure case: an Oracle→YB live migration started on the prior binary persists `pk` and streams events routed across channels by PK hash (`hashEvent`). If voyager is upgraded mid-stream — an explicitly supported scenario (root BUGBOT, "Upgrade and Backward Compatibility") — the new code ignores the persisted `pk` strategy and forces `table`, so `hashEvent` now routes every event of a table to a single channel keyed on table name. Because resume dedup relies on per-channel `LastAppliedVsn`, the changed channel assignment can cause events to be silently skipped or re-applied. This is exactly the case flagged by the pre-existing `TODO: handle upgrade scenario for PG/Oracle pk->table change` (line 175), now actualized for Oracle.

Suggested fix: for the Oracle target-importer path, still read `importDataStatus.TableToCDCPartitioningStrategyMap` and honor a persisted strategy when present (default to `PARTITION_BY_TABLE` only on first run) — or explicitly guard/document that an in-flight Oracle live migration cannot be resumed across this upgrade.

---

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -379,6 +379,241 @@ func (pg *TargetPostgreSQL) GetPrimaryKeyConstraintNames(table sqlname.NameTuple
...
+		schemaList = append(schemaList, parts[0])
+		tableList = append(tableList, parts[1])
+	}
+	return lo.Uniq(schemaList), lo.Uniq(tableList)
+}
+
+// queryPGUniqueIndexesMap runs the PG/YB unique-index discovery query for the given
+// tables using the provided query function and returns a map from table to its list
+// of unique indexes (each represented as an ordered column list).
+func queryPGUniqueIndexesMap(queryFn func(query string) (*sql.Rows, error), tableList []sqlname.NameTuple) (*utils.StructMap[sqlname.NameTuple, [][]string], error) {
```
makalaaneesh: [AI] **Dead code: `queryPGUniqueIndexesMap` is defined but never called.**

`queryPGUniqueIndexesMap` (lines 539–566) is a fully-implemented helper with zero callers — a repo-wide search finds only this definition. The two real callers, `TargetPostgreSQL.GetTableToUniqueIndexesMap` and `TargetYugabyteDB.GetTableToUniqueIndexesMap`, call `queryPGUniqueIndexesByCatalog` directly (`postgres.go:399`, `yugabytedb.go:654`), not this wrapper. Go does not flag unused package-level functions, so it compiles silently.

Per root BUGBOT: "Remove dead code, unused functions, and leftover debugging artifacts before merging."

Suggested fix: delete the function (lines 539–566).

---

yb-voyager/src/tgtdb/target_db_interface.go
```diff
@@ -45,6 +45,11 @@ type TargetDB interface {
 	QuoteAttributeNames(tableNameTup sqlname.NameTuple, columns []string) ([]string, error)
 	GetPrimaryKeyColumns(table sqlname.NameTuple) ([]string, error)
 	GetPrimaryKeyConstraintNames(tableNameTup sqlname.NameTuple) ([]string, error)
+	// GetTableToUniqueIndexesMap returns, for each table, the list of unique
+	// indexes/constraints (each an ordered column list). Used by live-migration
+	// conflict detection to fetch unique-key metadata directly from the import
+	// target. Not supported for Oracle targets (those use the metaDB path).
```
makalaaneesh: [AI] **Misleading interface doc: says Oracle targets "use the metaDB path", which this PR deletes.**

The doc on `GetTableToUniqueIndexesMap` states "Not supported for Oracle targets (those use the metaDB path)." This PR removes the metaDB path entirely — `TABLE_TO_UNIQUE_INDEXES_KEY`, `saveTableToUniqueIndexesMapInMetaDB`, and `getTableToUniqueIndexesMapFromMetaDB` are all deleted, and `initializeConflictDetectionCache` now unconditionally calls `tdb.GetTableToUniqueIndexesMap`. The actual Oracle behavior (`oracle.go:214`) is to return an empty map; what makes that safe is the `PARTITION_BY_TABLE` guard in `getCdcPartitioningStrategyPerTable`, not any metaDB lookup. The comment points a future maintainer at code that no longer exists.

Suggested fix: replace with e.g. "Oracle targets return an empty map; Oracle live migration always uses PARTITION_BY_TABLE, so conflict detection never runs for them."

---

makalaaneesh (commented):
[AI] Automated review — 3 findings (0 critical, 3 warning). See inline comments. This change moves unique-index metadata from the source/metaDB to a live query against the import target DB, and forces PARTITION_BY_TABLE for Oracle→YB targets. Findings cover a mid-upgrade resume hazard on the Oracle path, one dead helper, and a stale interface doc. Verified the new PG/YB unique-index query compiles and runs against a live PostgreSQL. One non-blocking question was also raised (not posted): whether target-side unique indexes are guaranteed to exist before the streaming/conflict-detection phase, since unique-index metadata now comes from the target at cache-build time rather than the source at export time.

---

yb-voyager/cmd/live_migration.go
makalaaneesh: @priyanshi-yb it caught an interesting upgrade scenario :) (not that we need to address it)

yb-voyager/cmd/live_migration.go
```diff
@@ -214,6 +187,17 @@ func getCdcPartitioningStrategyPerTable(tableNames []sqlname.NameTuple) (*utils.
...
+		//Oracle sources do not support unique-key conflict detection during live migration
+		//(we do not fetch unique indexes for Oracle). Force PARTITION_BY_TABLE so that all
+		//events of a table run sequentially on a single channel, which makes unique-key
+		//conflicts impossible and hence conflict detection unnecessary.
+		for _, t := range tableNames {
```
makalaaneesh, line 195: can we validate: if your source db is oracle, you cannot pass --cdc-partitioning-strategy config?

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -379,6 +379,211 @@ func (pg *TargetPostgreSQL) GetPrimaryKeyConstraintNames(table sqlname.NameTuple
 	return nil, nil
 }
 
+func (pg *TargetPostgreSQL) GetTableToUniqueIndexesMap(tableList []sqlname.NameTuple) (*utils.StructMap[sqlname.NameTuple, [][]string], error) {
```
makalaaneesh, line 382: I'm assuming all of this is just copy paste from source, so not going through

yb-voyager/cmd/live_migration.go
```diff
@@ -650,40 +634,22 @@ func processEvents(chanNo int, evChan chan *tgtdb.Event, lastAppliedVsn int64, d
...
-func initializeConflictDetectionCache(evChans []chan *tgtdb.Event, exporterRole string, sourceDBTypeForConflictCache string) error {
-	tableToUniqueIndexes, err := getTableToUniqueIndexesMapFromMetaDB(exporterRole)
+// initializeConflictDetectionCache builds the conflict detection cache used during
+// the streaming phase. The per-table unique indexes are fetched live from the import
+// target DB (the DB that actually enforces unique constraints). Oracle import targets
+// always use PARTITION_BY_TABLE (see getCdcPartitioningStrategyPerTable), so conflict
+// detection never runs for them and their target driver returns an empty map.
+func initializeConflictDetectionCache(evChans []chan *tgtdb.Event, sourceDBTypeForConflictCache string, importTableList []sqlname.NameTuple) error {
+	log.Infof("fetching table to unique indexes map from import target (%s)", tconf.TargetDBType)
+	tableToUniqueIndexes, err := tdb.GetTableToUniqueIndexesMap(importTableList)
```
makalaaneesh, line 647: Let's add a comment here with our offline discussion: 

* "attribute name registry is not really required for PG->YB scenario since everything is same in both sides and we deal with catalog column names in both the places - event's fields and the unique index column map
It was only required for Oracle scenario which. we are not doing" 

* we are making the assumption that column names between PG <> YB do not change.

makalaaneesh, line 647: as discussed let's add a integration test with case sensitive column names and ensure that conflicts are detected

yb-voyager/cmd/live_migration.go
priyanshi-yb: yeah.

yb-voyager/src/tgtdb/postgres.go
priyanshi-yb: yes it mostly, just that some changes as per how things are available in a different way target side than export.

yb-voyager/cmd/live_migration.go
priyanshi-yb: I just checked the validations we have for that flag - it doens't have a source type check, adding it now

yb-voyager/cmd/import.go
```diff
@@ -225,12 +225,17 @@ func validateCdcPartitioningStrategyFlag(cmd *cobra.Command) error {
 	if importerRole != TARGET_DB_IMPORTER_ROLE {
 		return nil
 	}
+	parameterPassed := cmd.Flags().Changed("cdc-partitioning-strategy")
 	if !changeStreamingIsEnabled(importType) {
-		if cmd.Flags().Changed("cdc-partitioning-strategy") {
+		if parameterPassed {
```
makalaaneesh: nit: `parameterPassed` is too generic a name. `cdcPartitioningStrategyChanged` is a better name IMO

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1096,3 +1096,192 @@ FROM generate_series(1, 20) as i;`,
...
+		},
+	}, 200, 5)
+	testutils.FatalIfError(t, err, "failed to wait for forward streaming complete")
+
+	require.False(t, liveMigrationTest.GetImportRunner().IsStopped(),
+		"import should keep running during count failpoint mode")
+
+	conflictStats, err := testutils.ReadUniqueKeyConflictStats(uniqueKeyConflictStatsPath)
+	testutils.FatalIfError(t, err, "failed to read unique key conflict stats")
+	require.GreaterOrEqual(t, conflictStats.Total, 500,
```
makalaaneesh, line 1232: nit: this has potential to become flaky because this is timing dependent. Keep an eye out for this later

makalaaneesh (approved):
LGTM with minor nits

PR #3655, opened 2026-07-08
Capture source schema snapshots during export (for schema-drift detection)

ShivanshGahlot:

Records a snapshot of the source schema at each migration milestone, so a later command can report what changed on the source while the migration was running. Nothing consumes the snapshots yet — `schema detect-drift` does, and it is stacked on this PR.

- **Capture points** — `export schema`, and `export data` at start, periodically, and on exit. Covers offline and live, snapshot and streaming phases.
- **Reasons** — the start snapshot records why the run started (`initial` / `resume` / `clean_restart`), derived from the data directory rather than the flag, so `--start-clean` on an empty dir is still `initial`. The exit snapshot records how it ended (`complete` / `cutover` / `interrupt` / `error`).
- **Best-effort by construction** — capture never fails or blocks a migration. Every error is logged and swallowed, and one 10s budget bounds the catalog read and the MetaDB write together.
- **Placeholders** — when the catalog cannot be read (source already gone at exit), a metadata-only marker still records the milestone, on its own fresh context so it survives whatever killed the capture.
- **Exit paths** — the error return captures inline, while the source connection is still open; an `atexit` handler covers signals and `utils.ErrExit`, whose `os.Exit` skips deferred cleanup. A signal-driven shutdown is classified `interrupt` (or `complete` for the end-migration teardown), not `error`.
- **Scope** — PostgreSQL only, source exporter only; a no-op for `export data from target`.
- **Schema names** — capture compares raw catalog identifiers. The min-quoted form (`"Odd Schema"`) matched no `nspname`, silently capturing zero tables, so `Source.GetSchemaListUnquoted` was added for this use.

Two new flags on the export commands, both hidden and off by default until `schema detect-drift` ships:
```
--suppress-schema-snapshot-capture   (default true)
--schema-snapshot-capture-interval   (default 60, minutes)
```

New unit tests for the start/exit reason classification and the schema-list accessors; integration tests for the capture hooks (in-process and command-driven) and a live-migration E2E for periodic capture during streaming. Manually verified the exit paths that tests cannot easily drive: SIGINT/SIGTERM/SIGUSR2 mid-export, and a source killed mid-export producing a placeholder.

No.

Yes — MetaDB. Adds a `schema_snapshots` table, created on demand. Additive only: an older binary against the same export dir ignores it.


yb-voyager/cmd/exportData.go
```diff
@@ -1846,6 +1894,31 @@ func validateAndExtractTableNamesFromFile(filePath string, flagName string) (str
...
+//
+// Admissibility is a SEPARATE concern owned by clearMigrationStateIfRequired (the
+// guard), which runs immediately after: it ErrExits a non-empty, non-start-clean
+// offline or mid-snapshot rerun (pg_dump can't resume), so the only non-empty run that
+// actually reaches this capture is the streaming-continue resume — for which "resume"
+// is the correct label.
+func snapshotStartReasonFor(startClean, dataDirEmpty bool) string {
+	switch {
+	case startClean:
+		return schemasnapshot.ReasonCleanRestart
```
ShivanshGahlot: Start clean can be passed in the first run too

yb-voyager/cmd/exportData.go
```diff
@@ -762,6 +769,31 @@ func exportData() bool {
...
+			reason := schemasnapshot.ReasonError
+			if ProcessShutdownRequested {
+				// SIGINT/SIGTERM is a user interrupt; SIGUSR2 is the end-migration
+				// command's controlled teardown, treated as a clean exit.
+				reason = schemasnapshot.ReasonInterrupt
+				if EndMigrationStopRequested {
+					reason = schemasnapshot.ReasonComplete
+				}
+			}
+			exitCtx, cancel := context.WithTimeout(context.Background(), schemaSnapshotExitCaptureTimeout)
```
ShivanshGahlot: Timeout can be lower: bad user experience

yb-voyager/cmd/exportData.go
```diff
@@ -762,6 +769,31 @@ func exportData() bool {
 
 	//finalTableList is with leaf partitions and root tables after this in the whole export flow to make all the catalog queries work fine
 
+	if exporterRole == SOURCE_DB_EXPORTER_ROLE {
+		captureSourceSchemaSnapshot(ctx, schemasnapshot.LabelExportDataFromSourceStart, snapshotStartReason, true)
+		atexit.Register(func() {
+			if exportDataExitSnapshotCaptured {
```
makalaaneesh: [AI] **Data race on `exportDataExitSnapshotCaptured` between the export goroutine and the signal-handler goroutine.**

`exportDataExitSnapshotCaptured` is a plain `bool` (declared at `cmd/schemaSnapshotCapture.go:34`). It is written by the main export goroutine in `exportData()` (lines 855, 862, 878) and read here inside the `atexit`-registered handler.

On a normal exit this is safe: `root.go`'s `PersistentPostRun` calls `atexit.Exit(0)` on the main goroutine, sequentially after `exportData()` returns. But on SIGINT/SIGTERM/SIGUSR2 the signal handler in `main.go` (`registerSignalHandlers`, a separate goroutine) calls `atexit.Exit(1)` — which runs this handler on the signal goroutine while the main goroutine is still executing `exportData()`. The two goroutines then access `exportDataExitSnapshotCaptured` (write vs. read) with no synchronization. That is a data race under the Go memory model and would be flagged by `go test -race`; it also violates the `.cursor/BUGBOT.md` rule that shared mutable state must be protected by a mutex or channel.

The signal path is exactly when this handler is meant to fire, so the race is on the live path, not a corner case. Practical worst case is benign (a duplicate/spurious best-effort exit snapshot, which collides on the second-granularity UNIQUE name and is swallowed), and the feature is off by default — hence Warning rather than Critical.

Suggested fix: make the flag an atomic to establish a happens-before edge and silence `-race`:

```go
// schemaSnapshotCapture.go
var exportDataExitSnapshotCaptured atomic.Bool

// atexit handler
if exportDataExitSnapshotCaptured.Load() { return }

// success/cutover/complete paths
exportDataExitSnapshotCaptured.Store(true)
```

---

makalaaneesh (commented):
[AI] Automated review — 1 finding (0 critical, 1 warning). See inline comment. Well-structured change that wires the `schemasnapshot` library into the export commands; correctly gated (PG-only, source-exporter-only, suppression-honored, off by default behind hidden flags) and honors the best-effort/swallow-error contract so the data path is never blocked. The one warning is an unsynchronized data race on `exportDataExitSnapshotCaptured` between the export goroutine and the signal-handler goroutine that runs the `atexit` exit-capture handler — benign in outcome but a genuine race `-race` would flag. A few non-blocking questions were also raised but not posted: unbounded contexts on the synchronous start/export-schema captures (vs. the 10s-bounded exit path), `schema-snapshot-capture-interval` being exposed on `export data from target` where capture never runs, config-template omission of the new flags, and the periodic ticker signaling stop without joining the goroutine.

---

---

yb-voyager/cmd/schemaSnapshotCapture.go
```diff
@@ -0,0 +1,154 @@
...
+// context timeout is a sufficient bound here.
+const schemaSnapshotExitCaptureTimeout = 10 * time.Second
+
+// captureSourceSchemaSnapshot captures the source schema and persists it as a
+// snapshot for the given label/reason. It is BEST-EFFORT and off the data path:
+// it never returns an error and never blocks the migration — every failure is
+// logged and swallowed.
+// Honors --suppress-schema-snapshot-capture. PostgreSQL only (no-op with a log for
+// other engines). Callers are responsible for exporter-role gating.
+func captureSourceSchemaSnapshot(ctx context.Context, label, reason string, placeholderOnFailure bool) {
```
makalaaneesh: let's return proper errors here. (not just log). It's the caller's responsibility to decide what to do with the error.

yb-voyager/cmd/exportData.go
```diff
@@ -762,6 +769,31 @@ func exportData() bool {
 
 	//finalTableList is with leaf partitions and root tables after this in the whole export flow to make all the catalog queries work fine
 
+	if exporterRole == SOURCE_DB_EXPORTER_ROLE {
+		captureSourceSchemaSnapshot(ctx, schemasnapshot.LabelExportDataFromSourceStart, snapshotStartReason, true)
+		atexit.Register(func() {
```
makalaaneesh: pls define a separate function here; don't want to pollute the main exportData func

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -396,6 +396,7 @@ func isCutoverInitiatedAndCutoverDetected(exporterRole string) (bool, error) {
 func debeziumExportData(config *dbzm.Config, tableNameToApproxRowCountMap map[string]int64) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
+	defer startPeriodicSourceSchemaSnapshotCapture(ctx)()
```
makalaaneesh: why are these calls `defer`s ? Why can't we just call the function when we need to?

yb-voyager/cmd/exportSchema.go
```diff
@@ -237,6 +239,8 @@ func exportSchema(cmd *cobra.Command) error {
 	saveSourceDBConfInMSR()
 	setSchemaIsExported()
 
+	captureSourceSchemaSnapshot(context.Background(), schemasnapshot.LabelExportSchema, "", true)
```
makalaaneesh: do we need a timeout based context everywhere else as well ? (not just in the exit handlers)?

yb-voyager/cmd/exportData.go
makalaaneesh: do we have a test case that actually tests that this timeout is honored by downstream layers (schema snapshot/pgx eventually)? 
If not , we should add

yb-voyager/cmd/schemaSnapshotCapture.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// startPeriodicSourceSchemaSnapshotCapture launches a background ticker that
+// captures a source schema snapshot every --schema-snapshot-capture-interval
+// minutes for the full duration of the export — both the snapshot and streaming
+// phases. It returns a stop function to be invoked via defer.
+// Best-effort and off the data path: a no-op when suppressed, when the interval is
+// <= 0, or when this is not the source exporter; periodic capture failures are
+// logged and never affect export.
+func startPeriodicSourceSchemaSnapshotCapture(ctx context.Context) func() {
```
makalaaneesh: IMO, better to take `schemaSnapshotCaptureInterval` as an input to this func rather than relying on globals

yb-voyager/cmd/schemaSnapshotCapture.go
```diff
@@ -0,0 +1,154 @@
...
+				return
+			case <-stop:
+				return
+			case <-ticker.C:
+				captureSourceSchemaSnapshot(ctx, schemasnapshot.LabelExportDataFromSourcePeriodic, "", false)
+			}
+		}
+	}()
+	var once sync.Once
+	return func() { once.Do(func() { close(stop) }) }
```
makalaaneesh: do we really need to return a stop function? Can the caller not re-use the context `cancel` ?

yb-voyager/cmd/exportSchema.go
makalaaneesh: won't exportSchema also require a atexit handler?

yb-voyager/cmd/exportData.go
makalaaneesh: [AI] [WARNING] The 10s exit budget does not bound the metaDB write — signal-path exit can hang up to ~30s

  This is the most important new finding. The handler has two I/O operations, but schemaSnapshotExitCaptureTimeout (10s) only covers one:

  - Source catalog read (pgx): takes exitCtx → bounded at 10s ✓
  - metaDB insert: InsertSchemaSnapshot → m.db.Exec(...) (metadb/schemaSnapshot.go:110) — plain Exec, no context. metaDB is opened with _txlock=exclusive&_timeout=30000 (metadataDB.go:52), i.e. a 30-second SQLite busy timeout.

  On the signal path, the handler runs in the signal goroutine concurrently with the still-live main goroutine, which during live streaming writes metaDB frequently (event stats, MSR). If the main goroutine holds the write lock at Ctrl-C, the
  handler's INSERT (and the CREATE TABLE IF NOT EXISTS that precedes it) blocks up to 30s on the busy timeout — the exitCtx can't interrupt it. The comment says the timeout "caps how long a wedged … source DB delays the exit," but the true
  worst-case exit delay is ~10s (read) + up to 30s (metaDB write), not 10s.

yb-voyager/cmd/exportData.go
```diff
@@ -762,6 +769,31 @@ func exportData() bool {
...
+			if exportDataExitSnapshotCaptured {
+				return // normal complete/cutover path already recorded the exit
+			}
+			// Abnormal exit (error or signal). Attempt a full capture so a drift-caused
```
makalaaneesh: [AI] [WARNING] yb-voyager/cmd/exportData.go:778 — Error-path exit capture silently degrades to a placeholder because the DB connection is already closed

- Problem/Suggestion: The atexit exit handler's comment states it attempts a full capture "so a drift-caused failure records the drifted end-state schema," falling back to a placeholder only if the schema "can't be read." But the two clean-error paths (return false) can never deliver a full capture: exportData has defer source.DB().Disconnect() at line 617, and both error returns (offline export failure ~`exportData.go:872-873, debezium-start failure ~exportData.go:800-803) run that defer *before* exportDataCommandFnreachesatexit.Exit(1) (exportData.go:217). By the time the handler runs, GetDB() returns a *closed* pool (Disconnectcallspg.db.Close()without nilling the handle —srcdb/postgres.go:178), so Capture's BeginTxfails and only a metadata-only placeholder is written. This defeats the stated goal precisely for the most likely drift-caused failure (offlinepg_dumpfailing because schema changed mid-export). Note: the SIGINT/SIGTERM interrupt path and the mid-functionutils.ErrExitpaths are *not* affected — thereatexit.Exit→os.Exitbypasses theDisconnectdefer, so the connection is still open. Only thereturn false` error paths regress. (Feature is off by default, so this is latent — but the code should meet its documented contract when enabled.)
- Code line: yb-voyager/cmd/exportData.go:778-793 (handler comment/logic), :617 (defer), :872-873 (offline error return)
- Failure Code Case: Enable capture, run offline export data, drop/alter a table so pg_dump fails → exportDataOffline returns error → return false → Disconnect() → atexit.Exit(1) → exit handler hits closed pool → records a placeholder instead of the drifted schema.
- Suggested Code Change: Capture the exit snapshot inline on the error paths, while the connection is still open, mirroring the success paths (and set the flag so the atexit handler no-ops):
err = exportDataOffline(ctx, cancel, finalTableList, tablesColumnList, "")
if err != nil {
    log.Errorf("Export Data failed: %v", err)
    if exporterRole == SOURCE_DB_EXPORTER_ROLE {
        // Capture here (connection still open) so a drift-caused failure records the
        // drifted end-state schema; the atexit handler only sees closed pools.
        captureSourceSchemaSnapshot(ctx, schemasnapshot.LabelExportDataFromSourceExit, schemasnapshot.ReasonError, true)
        exportDataExitSnapshotCaptured = true
    }
    return false
}
Apply the same to the debezium-start return false branch. The atexit handler then remains the fallback purely for interrupt/ErrExit exits (where it genuinely still has an open connection).

yb-voyager/cmd/exportData.go
```diff
@@ -762,6 +769,31 @@ func exportData() bool {
...
+			// failure records the drifted end-state schema; captureSourceSchemaSnapshot
+			// falls back to a placeholder if the schema can't be read. The run's own ctx
+			// is already cancelled by now, so use a fresh, bounded context.
+			reason := schemasnapshot.ReasonError
```
makalaaneesh: i'm a little iffy about this assumption that it's going to be an error by default that landed us here. 
I know you're handling other cases below, but let's discuss this once on call.

yb-voyager/cmd/exportData.go
ShivanshGahlot: Done

yb-voyager/cmd/exportData.go
ShivanshGahlot: Fixed with atomic.Bool

yb-voyager/cmd/exportData.go
ShivanshGahlot: Yep fixed this. Capturing the error snapshots before the function returns/ defer source.DB().Disconnect() runs

yb-voyager/cmd/exportData.go
ShivanshGahlot: This now only handles utils.ErrExit exits. In the above comment, the two error paths explicity capture the snapshots now.

yb-voyager/cmd/exportData.go
ShivanshGahlot: Lets discuss this on call once.

ShivanshGahlot: Threading the ctx into the metadb functions too to fix this

ShivanshGahlot: Doing this only for the write functions now in metdb.

yb-voyager/cmd/exportDataDebezium.go
ShivanshGahlot: It is a new idiom claude used. It basically is defer f()(), it immediately calls the startPeriodicSourceSchemaSnapshotCapture and returns a stops function, which is deferred.

ShivanshGahlot: But yeah I removed this due to another one of your comments. Basically not using the stop channel this function was creating. Directly using the ctx and the goroutine stops when the ctx is cancelled by the caller.

yb-voyager/cmd/schemaSnapshotCapture.go
ShivanshGahlot: Yep good point. Its redundant. Fixed it.

yb-voyager/cmd/exportSchema.go
ShivanshGahlot: Yeah we can do that. I am wondering how much the timer should be.

ShivanshGahlot: We can discuss this once on call

ShivanshGahlot: I am overall worried about latency in connections. What if the customer's connection has a high latency which results in the context being cancelled?

ShivanshGahlot: image

ShivanshGahlot: image

ShivanshGahlot: The whole capture flow chart and a few points of discussion I had ^^

yb-voyager/cmd/exportData.go
ShivanshGahlot: Threading the context into the write functions of metadb

yb-voyager/cmd/exportSchema.go
ShivanshGahlot: We dont right? The atexit handles abnormal exits. If export schema is exiting abnormally, it means it was not completed properly, we dont need to capture the snapshot at that point.
We only need to capture when it completes normally.

yb-voyager/cmd/schemaSnapshotCapture.go
ShivanshGahlot: I have done it intentionally.
So the idea here is that its a best effort capture. The capture library does return errors and all and they are handled here. This function basically makes capture off the data path. The caller does not need to see the errors as it is off the data path.
If in the future some caller does want to see the errors, they can call CaptureAndSaveSnapshot directly instead of this wrapper.
Moreover, the skip snapshots path is also handled here. Returning an error/returning something means also returning the context why the snapshot was not captured i.e. was it skipped or it errored out. Not a very strong point but just pointing this out.

ShivanshGahlot: We can discuss this on call.

yb-voyager/cmd/schemaSnapshotCapture.go
ShivanshGahlot: Yeah done

yb-voyager/cmd/config.go
```diff
@@ -108,6 +108,7 @@ var allowedExportSchemaConfigKeys = mapset.NewThreadUnsafeSet[string](
 	"skip-colocation-recommendations", "assessment-report-path",
 	"skip-performance-recommendations",
 	"assess-schema-before-export",
+	"suppress-schema-snapshot-capture",
```
makalaaneesh, line 111: can we use "disable-" instead of "suppress" just for consistency of naming?

yb-voyager/cmd/exportData.go
```diff
@@ -781,6 +791,33 @@ func exportData() bool {
...
+		// for the capture. Panics are covered too.
+		//
+		// utils.ErrExit and signals do not unwind, so they never reach this defer --
+		// registerExportDataExitSnapshotHook above covers them.
+		defer func() {
+			if ok {
+				captureExportDataExitSnapshot(ctx, successReason)
+				return
+			}
+			captureExportDataExitSnapshotFresh(exportDataExitReason())
```
makalaaneesh, line 817: naming is a bit confusing; what's the difference between captureExportDataExitSnapshotFresh and captureExportDataExitSnapshot?

yb-voyager/cmd/schemaSnapshotCapture.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// registerExportDataExitSnapshotHook covers the exit paths that never unwind, so
+// exportData's exit defer cannot run: signals, and utils.ErrExit (whose os.Exit skips
+// defers, leaving the connection open). Whichever path gets there first wins the claim.
+func registerExportDataExitSnapshotHook() {
+	atexit.Register(func() {
+		captureExportDataExitSnapshotFresh(exportDataExitReason())
+	})
+}
```
makalaaneesh, line 93: these two funcs should be in exportData.go?

yb-voyager/cmd/exportData.go
makalaaneesh, line 817: here `if !ok`, we already know the reason, no? It's got to be an error.  Is there any other possibility?

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -396,6 +396,8 @@ func isCutoverInitiatedAndCutoverDetected(exporterRole string) (bool, error) {
 func debeziumExportData(config *dbzm.Config, tableNameToApproxRowCountMap map[string]int64) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
+	// Periodic capture runs until ctx is cancelled; the defer cancel() above stops it.
+	startPeriodicSourceSchemaSnapshotCapture(ctx, time.Duration(schemaSnapshotCaptureInterval)*time.Minute)
```
makalaaneesh, line 400: why only here and not in exportDAta.go (when you take the first snapshot and register the hook).

yb-voyager/cmd/schemaSnapshotCapture.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	if err := captureSourceSchemaSnapshot(ctx, schemasnapshot.LabelExportDataFromSourceExit, reason, true); err != nil {
+		log.Warnf("schema-snapshot exit capture (%s) failed, migration unaffected: %v", reason, err)
+	}
+}
+
+// captureExportDataExitSnapshotFresh is captureExportDataExitSnapshot on a fresh
+// context, for exit sites where the run's own context may already be cancelled: the
+// error `return false` paths (which must capture inline, before exportData's deferred
+// Disconnect closes the connection) and the atexit fallback.
```
makalaaneesh, line 60: don't understand why we need a separate func. Why can't the caller just pass `context.Background()` to `captureExportDataExitSnapshot` ?

yb-voyager/cmd/exportData.go
```diff
@@ -1773,6 +1814,9 @@ func handleGetInitialTableListError(err error) {
 }
 
 func exportDataOffline(ctx context.Context, cancel context.CancelFunc, finalTableList []sqlname.NameTuple, tablesColumnList *utils.StructMap[sqlname.NameTuple, []string], snapshotName string) error {
+	// Periodic capture runs until ctx is cancelled; the caller (exportData) owns ctx
+	// and cancels it via its defer cancel() when the export finishes.
+	startPeriodicSourceSchemaSnapshotCapture(ctx, time.Duration(schemaSnapshotCaptureInterval)*time.Minute)
```
makalaaneesh, line 1819: [AI] **CRITICAL: two periodic-capture goroutines run concurrently for the whole PG live-migration streaming phase.**

`startPeriodicSourceSchemaSnapshotCapture` is called from *both* `exportDataOffline` (this line) and `debeziumExportData` (`yb-voyager/cmd/exportDataDebezium.go:400`).

For PostgreSQL snapshot-and-changes, `exportDataOffline` is reached via `createReplicationSlotAndExportSnapshotIfRequiredForPG` (`exportData.go:1195`) with `exportData`'s own `ctx` — and that ctx is not cancelled until `exportData` returns, so this ticker keeps firing through the entire streaming phase. `debeziumExportData` then starts a second, independent ticker.

Result: two periodic snapshots per configured interval for the rest of the run — doubled metaDB writes and a duplicated point at every interval on the drift timeline that the stacked `detect-drift` PR will consume. When the two tickers happen to align within the same wall-clock second, the second insert also collides on the second-granularity `PRIMARY KEY (side, label, captured_at)` and logs a spurious warning (that collision behaviour is acknowledged in `schemaSnapshotCapture_integration_test.go:392`).

Neither test catches this: the live E2E asserts `>= 1` periodic snapshot, and the function-level test drives `startPeriodic` directly.

**Repro:** `yb-voyager export data from source --source-db-type postgresql --export-type snapshot-and-changes --suppress-schema-snapshot-capture false --schema-snapshot-capture-interval 1`. After the snapshot phase completes, the `export_data_from_source_periodic` row count grows at ~2/min instead of 1/min.

**Suggested fix** — start the ticker exactly once, at the single owner of the whole export lifetime, and drop both inner call sites:

```go
// exportData.go, alongside the start capture at :799
if exporterRole == SOURCE_DB_EXPORTER_ROLE {
	if err := captureSourceSchemaSnapshot(ctx, schemasnapshot.LabelExportDataFromSourceStart, snapshotStartReason, true); err != nil {
		log.Warnf("schema-snapshot start capture failed, export unaffected: %v", err)
	}
	// One ticker for the whole export -- snapshot AND streaming phases, offline and live.
	// ctx is this function's, cancelled by its defer cancel() when the export finishes.
	startPeriodicSourceSchemaSnapshotCapture(ctx, time.Duration(schemaSnapshotCaptureInterval)*time.Minute)
	registerExportDataExitSnapshotHook()
	...
}
```

That also covers the `changes-only` flow (which skips `exportDataOffline`) and the PG live snapshot phase, so no coverage is lost. If you'd rather keep two call sites, make the start idempotent with a `sync.Once` / atomic CAS, the way `exportDataExitSnapshotCaptured` already does for the exit capture.

yb-voyager/cmd/export.go
```diff
@@ -56,6 +58,14 @@ func registerCommonExportFlags(cmd *cobra.Command) {
...
 
 	BoolVar(cmd.Flags(), &source.RunGuardrailsChecks, "run-guardrails-checks", true, "run guardrails checks before export. (only valid for PostgreSQL) Setting this to false is unsafe: it skips critical pre-migration validations (such as source/target database permissions, binary dependencies, and version compatibility) and may lead to migration failures or data issues. Leave the default (true) unless you have a specific reason to disable checks.")
+
+	// Defaults to true (capture off) for now: nothing consumes the snapshots until the
+	// detect-drift command ships. Flip to false to enable capture by default once it does.
+	BoolVar(cmd.Flags(), &suppressSchemaSnapshotCapture, "suppress-schema-snapshot-capture", true,
+		"disable best-effort schema-snapshot capture during export. (only valid for PostgreSQL)")
+	// Hidden for now: capture is off by default and nothing consumes the snapshots until
+	// the detect-drift command ships. Still settable via CLI/config for internal use.
+	cmd.Flags().MarkHidden("suppress-schema-snapshot-capture")
```
makalaaneesh, line 68: [AI] **WARNING: `MarkHidden` errors are dropped; the repo's `mustMarkFlagHidden` helper is bypassed.**

`cmd.Flags().MarkHidden(...)` returns an error that is discarded here and again at `export.go:278`. `cmd/root.go:546` already provides `mustMarkFlagHidden(cmd, name)` for exactly this, added by the recent errcheck pass.

Per `yb-voyager/.cursor/BUGBOT.md` ("Never silently swallow errors. If a function returns an error, either handle it, return it, or log it with sufficient context."), and for consistency with the rest of the tree, use the helper.

**Failure case:** rename either flag and forget to update the `MarkHidden` string — the flag silently becomes visible in `--help` with no build-time or runtime signal.

```go
mustMarkFlagHidden(cmd, "suppress-schema-snapshot-capture")
...
mustMarkFlagHidden(cmd, "schema-snapshot-capture-interval")
```

makalaaneesh (commented):
[AI] Automated review — 7 findings (1 critical, 3 warnings, 3 suggestions). See inline comments. Suggestion-severity items are included at the author's request and are non-blocking polish.

yb-voyager/cmd/root.go
```diff
@@ -54,6 +54,7 @@ var (
 	migrationUUID                      uuid.UUID
 	perfProfile                        utils.BoolStr
 	ProcessShutdownRequested           bool
+	EndMigrationStopRequested          bool
```
makalaaneesh, line 57: [AI] **WARNING: new `EndMigrationStopRequested` is a plain `bool` written from the signal goroutine and read from the export goroutine — a data race.**

`main.go:64` sets this on the signal goroutine. `exportDataExitReason()` (`cmd/schemaSnapshotCapture.go:76`) reads it from the export goroutine's exit defer, and the `atexit` hook reads it from the signal goroutine. That is an unsynchronised race that `-race` will report.

This PR already recognised this exact class of problem and made `exportDataExitSnapshotCaptured` an `atomic.Bool` (commit `85f856e8f`, "make exit-capture flag atomic to fix data race"); the same treatment is cheap here. `yb-voyager/.cursor/BUGBOT.md` also requires shared mutable state to be protected by a mutex or channel.

The comment at `schemaSnapshotCapture.go:73-75` argues the race is benign, but a "benign" data race is still undefined behaviour in Go and will fail any `-race` integration run.

**Failure case:** `go test -race -tags integration_live_migration ./src/testlivemigration/...` with a SIGUSR2 or SIGINT mid-export → `WARNING: DATA RACE` on `cmd.EndMigrationStopRequested`.

```go
// root.go -- set only by the SIGUSR2 handler (main.go), read by the export
// exit-snapshot classifier on a different goroutine; atomic for that reason.
EndMigrationStopRequested atomic.Bool
```
```go
// main.go
cmd.EndMigrationStopRequested.Store(true)

// schemaSnapshotCapture.go
if EndMigrationStopRequested.Load() {
	return schemasnapshot.ReasonComplete
}
```

Since this PR is what makes `ProcessShutdownRequested` cross-goroutine-read on this path, consider converting it too (or reading both through one small accessor).

yb-voyager/cmd/schemaSnapshotCapture_integration_test.go
```diff
@@ -0,0 +1,415 @@
...
+// asserts a REAL snapshot (not a placeholder) lands with every table: because the capture
+// is bounded by CaptureTimeout, a real snapshot is itself proof the budget sufficed for
+// this size — had it been exceeded, the capture would have degraded to a placeholder and
+// this test would fail. The elapsed time is logged so the headroom below the budget is
+// visible.
+//
+// Note: on a healthy testcontainer this proves "the budget is ample for size", not
+// "the budget survives a slow/loaded/high-latency source" — that adverse case isn't
+// deterministically reproducible here and is a manual/load-test concern.
+func TestSchemaSnapshotCaptureLargeSchema(t *testing.T) {
```
makalaaneesh, line 245: [AI] **WARNING: `TestSchemaSnapshotCaptureLargeSchema` is a wall-clock-budget test that will flake on a loaded CI runner.**

The test seeds 1000 tables / ~10k columns and asserts (`:300`) that the result is a *real* snapshot rather than a placeholder. That only holds if the catalog read + JSON marshal + SQLite insert all finish inside the hard-coded `CaptureTimeout = 10s`. On a contended CI box (Docker + testcontainers + parallel packages) that budget is not comfortably out of reach, and exceeding it produces a confusing "got a placeholder" failure rather than "the machine was slow". The test's own doc comment already concedes it proves nothing about a slow source.

**Failure case:** loaded runner; the 1000-table capture takes 10.4s → `CaptureAndSaveSnapshot` returns a deadline error → placeholder is written → `require.False(t, headers[0].IsPlaceholder, ...)` fails, unrelated to the change under test.

**Suggested fix:** either drop the size to something safely inside the budget, or make the test *measure* rather than *gate* — capture with an explicitly generous context and report elapsed time against the production budget:

```go
// Measure with generous headroom; report against the production budget instead of
// letting a loaded CI box turn "slow machine" into "wrong behaviour".
ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
defer cancel()
...
if elapsed > schemasnapshot.CaptureTimeout {
	t.Logf("WARNING: %d-table capture took %s, exceeding the %s production budget",
		numTables, elapsed, schemasnapshot.CaptureTimeout)
}
```

yb-voyager/cmd/schemaSnapshotCapture.go
```diff
@@ -0,0 +1,210 @@
...
+// whole export, snapshot and streaming phases alike. The interval is a parameter, not
+// the global, so tests can use a small one. The goroutine stops when ctx is cancelled,
+// so no separate stop function is needed.
+//
+// Best-effort: a no-op when suppressed, when interval <= 0, or off the source exporter.
+func startPeriodicSourceSchemaSnapshotCapture(ctx context.Context, interval time.Duration) {
+	if exporterRole != SOURCE_DB_EXPORTER_ROLE || source.DBType != POSTGRESQL || bool(suppressSchemaSnapshotCapture) {
+		return
+	}
+	if interval <= 0 {
```
makalaaneesh, line 166: [AI] **SUGGESTION: a non-positive capture interval silently disables periodic capture.**

This early return fires with no log line, so `--schema-snapshot-capture-interval 0` (or a negative value from a config file) looks accepted but quietly turns periodic capture off.

**Failure case:** a user sets `schema-snapshot-capture-interval: 0` in the config file expecting "capture every tick" and gets no periodic snapshots and no diagnostic anywhere in the log.

Either validate the flag at registration/PreRun time, or at minimum say so:

```go
if interval <= 0 {
	log.Infof("periodic schema-snapshot capture disabled: interval is %s (non-positive)", interval)
	return
}
```

yb-voyager/cmd/schemaSnapshotCapture.go
```diff
@@ -0,0 +1,210 @@
...
+// for the given label/reason, returning why it could not.
+//
+// A skip is not an error: a non-PostgreSQL source or --suppress-schema-snapshot-capture
+// returns nil, since nothing went wrong. Exporter-role gating is the caller's.
+//
+// The caller decides what a failure means. Today every caller is an export hook that
+// logs and carries on, because schema capture is off the data path and must never fail
+// or stall an export.
+func captureSourceSchemaSnapshot(ctx context.Context, label, reason string, placeholderOnFailure bool) error {
+	if source.DBType != POSTGRESQL {
```
makalaaneesh, line 105: [AI] **SUGGESTION: the PostgreSQL / suppression gate is duplicated across three functions.**

`source.DBType != POSTGRESQL` and `bool(suppressSchemaSnapshotCapture)` are re-checked in `captureSourceSchemaSnapshot` (here), `startPeriodicSourceSchemaSnapshotCapture` (`:163`), and `saveSourceSchemaSnapshotPlaceholder` (`:191`).

One predicate keeps the "when is capture live at all?" rule in a single place, and makes it obvious at a glance that the placeholder path honours the same gate as the real capture path:

```go
// schemaSnapshotCaptureEnabled reports whether capture is live at all: PostgreSQL
// source, not suppressed. Role gating stays at the call sites that know the role.
func schemaSnapshotCaptureEnabled() bool {
	return source.DBType == POSTGRESQL && !bool(suppressSchemaSnapshotCapture)
}
```

yb-voyager/cmd/export.go
```diff
@@ -252,6 +262,22 @@ func registerExportDataFlags(cmd *cobra.Command) {
...
+// registerSchemaSnapshotIntervalFlag registers the periodic-capture interval.
+//
+// Deliberately NOT in registerExportDataFlags, which `export data from target`
+// shares: capture is source-side only, so the flag would be accepted there and do
+// nothing. It also can't join --suppress-schema-snapshot-capture in
+// registerCommonExportFlags, which export schema calls — that has no interval.
+//
+// Config-file key: "export-data.schema-snapshot-capture-interval".
+func registerSchemaSnapshotIntervalFlag(cmd *cobra.Command) {
+	cmd.Flags().IntVar(&schemaSnapshotCaptureInterval, "schema-snapshot-capture-interval", 60,
```
makalaaneesh, line 274: [AI] **SUGGESTION: new flags are not in the config templates.**

`yb-voyager/.cursor/BUGBOT.md` ("Flag and Config Handling") asks that every new flag be added to `yb-voyager/config-templates`. Neither `--schema-snapshot-capture-interval` nor `--suppress-schema-snapshot-capture` appears there.

Both are deliberately hidden until `detect-drift` ships, and `--metrics-port` set the same precedent, so this is fine to defer — but please add them (commented out) in the follow-up PR that un-hides them, so it does not get forgotten.

yb-voyager/cmd/config.go
ShivanshGahlot: Renamed in 20be1ed0 to `--disable-schema-snapshot-capture`, along with the config key and the Go variable. It was the only `suppress-` flag in the tree against three existing `disable-` ones. The flag is hidden and has never shipped, so the rename costs nothing now; after a release it would need an alias and a deprecation cycle.

Semantics unchanged. Verified the scoping is identical: accepted on `export data` and `export schema`, rejected on `export data from target`, both flags still hidden.

yb-voyager/cmd/exportData.go
ShivanshGahlot: Agreed the name carried no information. `captureExportDataExitSnapshotFresh` is removed in 188d0d62 — it was a one-line alias, so both callers now pass `context.Background()` directly.

The reason a fresh context is needed there is kept as a comment at the call site: on a failing export path the run's ctx may already be cancelled (`exportDataOffline` cancels on its quit path), so the exit capture must not reuse it, or it aborts exactly when the drifted end-state matters most.

yb-voyager/cmd/schemaSnapshotCapture.go
ShivanshGahlot: Moved in be30d7ef. I moved all four export-data-exit items rather than two, since splitting that cluster across files would be worse than either extreme: `exportDataExitSnapshotCaptured`, `captureExportDataExitSnapshot`, `exportDataExitReason` and `registerExportDataExitSnapshotHook`.

`schemaSnapshotCapture.go` is now only the reusable capture primitives, and no longer needs the `atexit` or `sync/atomic` imports at all.

yb-voyager/cmd/exportData.go
ShivanshGahlot: Yes — a signal can reach `return false`. The signal kills the in-flight child, the export reports failure, and the defer runs with `ProcessShutdownRequested` already set, so the correct reason on that path is `interrupt`, not `error`. Hardcoding `ReasonError` here is what previously recorded every Ctrl-C as an error.

On the signal path the atexit hook usually claims the capture first, because the export goroutine stays blocked in `pg_dump` until an atexit handler kills it — so `!ok` combined with a signal is a race that this branch loses more often than it wins. It is still a real interleaving, and hardcoding the reason would also split the classification across two sites with different logic.

Behaviour confirmed against a PostgreSQL source: killing the `pg_dump` child with no signal present, this defer does claim the capture first and records `error`.

yb-voyager/cmd/exportDataDebezium.go
ShivanshGahlot: It should have been in exportData.go. Starting it in both places meant PostgreSQL snapshot-and-changes ran two tickers concurrently through the streaming phase, recording two periodic snapshots per interval.

Fixed in 4fd7bb58: started once in `exportData`, the single owner of the export lifetime, with both inner call sites removed.

yb-voyager/cmd/schemaSnapshotCapture.go
ShivanshGahlot: It can, and now does — the helper is removed in 188d0d62 and both callers pass `context.Background()` directly.

The one thing kept is a comment recording why a fresh context is required rather than the run's ctx: on a failing export path that ctx may already be cancelled (`exportDataOffline` cancels on its quit path), which would abort the exit capture.

yb-voyager/cmd/exportData.go
ShivanshGahlot: Confirmed and fixed in 4fd7bb58, using the suggested fix: start the ticker once in `exportData` next to the start capture, and drop both inner call sites. Coverage is unchanged for offline, PG live, debezium snapshot-only and changes-only.

Reproduced with `--schema-snapshot-capture-interval 1` before the fix: periodic snapshots arrive in pairs 60s apart, with roughly 20s between the two members of each pair — two 1-minute tickers offset by however long the snapshot phase took. Six snapshots over an interval where a single ticker produces three.

Two clarifications on the scope:

- It only doubles on a first run. `createReplicationSlotAndExportSnapshotIfRequiredForPG` is behind `if !dataIsExported()`, so a resume already gets a single ticker.
- `changes-only` was not previously uncovered: it skips `exportDataOffline` but still reaches `debeziumExportData`, so it had exactly one ticker before this change as well.

On the test gap: the ticker now logs when it starts, and the live E2E asserts exactly one start, since the previous `>= 1 periodic snapshot` assertion passed at either rate. Reintroducing the second call site fails that test with `2 starts means concurrent tickers`.

yb-voyager/cmd/export.go
ShivanshGahlot: Fixed in 798aa3a9.

Note that `mustMarkFlagHidden` did not exist on this branch — it landed on main in 6ab7594f after this PR was opened, so the branch is now rebased onto main. After that rebase these two were the only bare `Flags().MarkHidden` calls left in the tree.

yb-voyager/cmd/root.go
ShivanshGahlot: Confirmed and fixed in 09aeae5f. The race reproduces under `-race` with a goroutine setting the flags alongside `exportDataExitReason()`:

```
WARNING: DATA RACE
Read at 0x000105b90bf5 by goroutine 23:
  github.com/yugabyte/yb-voyager/yb-voyager/cmd.exportDataExitReason()
```

`ProcessShutdownRequested` is converted too, as suggested — fixing one and leaving the other racy on the same code path is hard to justify, and it was only about nine sites. Both are `atomic.Bool` now, and the comment claiming the race was benign is removed.

Separately: `go test -race` over the whole `cmd` package also reports a race in `TestRandomBatchProducer_Resumption_AllBatchesProduced_AllConsumed`, inside `importDataRandomFileBatchProducer.go` / `importDataSequentialFileBatchProducer.go`. Both files are byte-identical to main and untouched here, so it is pre-existing and out of scope for this PR, but it does mean a `-race` run on that package will not be clean even with this fix. Worth a separate issue.

yb-voyager/cmd/schemaSnapshotCapture_integration_test.go
ShivanshGahlot: The failure mode described here no longer applies. Since `captureSourceSchemaSnapshot` returns its error (e77602c5, bcdff834) and the test does `require.NoError` on it, a budget overrun fails at the capture call rather than on the placeholder assertion:

```
Error: Received unexpected error:
       schemasnapshot: opening snapshot transaction: context deadline exceeded
```

Confirmed by forcing `CaptureTimeout` to 1ns: `require.False(..., IsPlaceholder)` is unreachable on the timeout path, so a slow machine now reports a deadline, not a placeholder.

On the flake risk: the 1000-table / ~10k-column capture measures 35.8ms and 48.5ms against the 10s budget, roughly 200-280x headroom. The 10.4s failure case needs about a 215x slowdown.

Left as-is on that basis. Happy to switch it to `CaptureAndSaveSnapshot` with a generous context if you would rather it not be wall-clock bound at all — the trade-off is that it stops covering `captureSourceSchemaSnapshot`'s wiring.

yb-voyager/cmd/schemaSnapshotCapture.go
ShivanshGahlot: Fixed in 52175596, taking the stricter option: `0` or negative now fails at startup rather than being logged, and the flag description states the minimum.

```
failed to validate export flags: --schema-snapshot-capture-interval must be at least 1 (minute), got 0
```

The validation is scoped by flag presence rather than by exporter role, which is worth noting: `export data from target` calls export data's PreRun but never registers this flag, so the global there is still its `0` zero value (`IntVar`'s default of 60 only applies where the flag is registered) and a value-only check would break that command. There is a unit test case covering it.

yb-voyager/cmd/schemaSnapshotCapture.go
ShivanshGahlot: Done in e2e727e0, with one deviation from the snippet: `schemaSnapshotCaptureEnabled` returns `(bool, string)` rather than just `bool`.

A plain bool would collapse `captureSourceSchemaSnapshot`'s two distinct log lines — "only PostgreSQL sources are supported" and "disabled via --disable-schema-snapshot-capture" — into one generic message, and that distinction is the answer to "why am I getting no snapshots?". The two silent call sites discard the reason.

yb-voyager/cmd/export.go
ShivanshGahlot: Agreed, and deferring as suggested.

One gap: the follow-up PR this defers to does not currently do it. The stacked detect-drift PR (#3769) does not un-hide these flags, even though the comment here says they are hidden until the detect-drift command ships. So either #3769 should un-hide them and add the commented-out template entries, or this comment needs rewording. Planning to handle it on #3769 unless you would prefer it elsewhere.

yb-voyager/cmd/exportData.go
makalaaneesh: pls add a comment here in code to explain that reason could be interrupt OR error

yb-voyager/cmd/schemaSnapshotCapture_integration_test.go
```diff
@@ -0,0 +1,415 @@
+//go:build integration
```
makalaaneesh: can we move these four test files into schemasnapshot? Don't want to create new files in cmd. 
Also there's scope for merging some of these unit tests into a single file

yb-voyager/cmd/exportData.go
```diff
@@ -1829,6 +1883,63 @@ func handleGetInitialTableListError(err error) {
...
+// first, and the atexit hook then no-ops. Hardcoding ReasonError there recorded
+// every Ctrl-C as an error.
+func exportDataExitReason() string {
+	if !ProcessShutdownRequested.Load() {
+		return schemasnapshot.ReasonError
+	}
+	if EndMigrationStopRequested.Load() {
+		return schemasnapshot.ReasonComplete
+	}
+	return schemasnapshot.ReasonInterrupt
```
makalaaneesh: Would suggest this as it's more easy to understand:

if EndMigrationStopRequested: 
    complete
if ProcessShutdownRequested: 
    interrupt 
error

makalaaneesh: just take a look at tests and verify that we have test for these different cases asserting the expected reason

yb-voyager/cmd/schemaSnapshotCapture_integration_test.go
makalaaneesh: actually integration tests make more sense in a new pkg testschemadrift

yb-voyager/cmd/export.go
```diff
@@ -252,6 +262,39 @@ func registerExportDataFlags(cmd *cobra.Command) {
...
+		"interval (in minutes, must be at least 1) at which voyager periodically captures a source schema snapshot throughout export data (both snapshot and streaming phases; offline and live). (only valid for PostgreSQL)")
+	// Hidden for now: capture is off by default and nothing consumes the snapshots until
+	// the detect-drift command ships. Still settable via CLI/config for internal use.
+	mustMarkFlagHidden(cmd, "schema-snapshot-capture-interval")
+}
+
+// validateSchemaSnapshotCaptureInterval rejects an interval that would silently turn
+// periodic capture off, so a bad value fails at startup instead of looking accepted.
+//
+// Scoped by flag presence, not by role: `export data from target` runs export data's
```
makalaaneesh: [AI][Suggestion] **Docstring justification for `validateSchemaSnapshotCaptureInterval` is inaccurate.**

The comment claims "the global sits at its 0 zero value there (IntVar's default of 60 only applies where the flag is registered)". That's not how `IntVar` works: at registration time it writes the default (60) into the pointer. Because both `exportDataCmd` and `exportDataFromSrcCmd` call `registerSchemaSnapshotIntervalFlag(...)` in `init()`, the global is `60` at runtime even during `export data from target` (which shares those globals via package init). The guard itself is still correct — the `Lookup(...) == nil` check reliably scopes it to commands that registered the flag — but the reasoning in the docstring is wrong and future readers may over-trust it.

Suggested fix: reword to reflect that the guard is defensive by *flag-presence* on the command, not a compensation for a still-0 global. Something like:

```go
// Scoped by flag presence on `cmd`, not by role: `export data from target` shares
// the export-data package globals but does not register this flag, so its Lookup
// returns nil and the guard is a no-op there. (The global itself is 60 at runtime,
// written by IntVar in the commands that do register the flag.)
```

---

yb-voyager/cmd/exportData.go
```diff
@@ -834,6 +848,42 @@ func exportData() bool {
...
+	if exporterRole == SOURCE_DB_EXPORTER_ROLE {
+		if err := captureSourceSchemaSnapshot(ctx, schemasnapshot.LabelExportDataFromSourceStart, snapshotStartReason, true); err != nil {
+			log.Warnf("schema-snapshot start capture failed, export unaffected: %v", err)
+		}
+		// One ticker for the whole export -- snapshot AND streaming phases, offline and
+		// live. ctx is this function's, cancelled by its defer cancel() when the export
+		// ends. Started here, at the single owner of the export lifetime: starting it in
+		// both exportDataOffline and debeziumExportData ran two tickers at once for PG
+		// snapshot-and-changes, doubling the periodic snapshots.
+		startPeriodicSourceSchemaSnapshotCapture(ctx, time.Duration(schemaSnapshotCaptureInterval)*time.Minute)
```
makalaaneesh: [AI][Suggestion] **Consider stopping the periodic ticker deterministically before the exit-capture defer runs.**

The periodic ticker goroutine (started on the previous line via `startPeriodicSourceSchemaSnapshotCapture(ctx, ...)`) stops on `ctx.Done()`, but `defer cancel()` runs *after* the exit-capture defer registered here (LIFO). During the exit-capture window (bounded by `CaptureTimeout = 10s`), the ticker is still active. Production intervals are >= 1 minute so a tick landing in that window is unlikely, but the outcome — a periodic snapshot persisted *after* the exit snapshot — is a mislabelled timeline entry that drift-detection consumers would have to filter out.

Failure case: in an offline PG export that finishes at second N, a tick queued at N+e lands within the `ok=true` exit-capture defer; the periodic snapshot ends up with `captured_at` strictly after the exit snapshot, and drift-detection treats it as post-exit noise.

Suggested fix: give the periodic ticker its own child context that the exit-capture defer cancels first:

```go
periodicCtx, stopPeriodic := context.WithCancel(ctx)
startPeriodicSourceSchemaSnapshotCapture(periodicCtx, time.Duration(schemaSnapshotCaptureInterval)*time.Minute)
registerExportDataExitSnapshotHook()

defer func() {
    stopPeriodic() // ensure no periodic tick races the exit capture
    if ok {
        captureExportDataExitSnapshot(ctx, successReason)
        return
    }
    captureExportDataExitSnapshot(context.Background(), exportDataExitReason())
}()
```

---

makalaaneesh (commented):
[AI] Automated review (delta since 0b31f09) — 2 findings (0 critical, 0 warning, 2 suggestion). All previously-posted [AI] findings verified as addressed by the new commits. Two additional suggestion findings about `yb-voyager/src/srcdb/postgres.go` (`GetGeneratedStoredColumns` — SQL identifier interpolation at line 838 and English-error-text fallback at line 859) were dropped because that function was added on `main` (PR #3741), not by this PR's diff — please treat them as follow-up cleanups in that file, not review comments here.

---

yb-voyager/cmd/exportData.go
ShivanshGahlot: Added in 5660c31a.

It also reconciles with the comment a few lines above, which says signals never reach this defer: that is true of unwinding (`os.Exit` skips defers), but not of the side effect — the signal kills the in-flight child, so the export reports failure and can still reach this path, racing `os.Exit`, with `ProcessShutdownRequested` already set.

yb-voyager/cmd/exportData.go
ShivanshGahlot: Reordered in 03393e51 — it does read better without the leading negation.

One behavioural note, since the two orderings are not quite equivalent. They agree on every reachable input and differ only when the end-migration flag is set *without* a shutdown request: previously `error`, now `complete`. `main.go` stores `ProcessShutdownRequested` before `EndMigrationStopRequested`, so that combination cannot be observed in practice. The test case pinning it is updated and annotated as unreachable, so the classification stays defined for all four inputs.

On test coverage, split by level:

- The classifier itself is fully covered — `TestExportDataExitReason` pins all four flag combinations.
- The *persisted* reason is only asserted for `complete`, in `schemaSnapshotCaptureCommand_integration_test.go`. There is no committed test asserting a persisted `interrupt`, `error`, or `cutover`. (The `ReasonError` assertion in `schemaSnapshotCapture_integration_test.go` is on a placeholder whose reason was passed in directly, so it is not exercising the classification.)

Those three were verified by hand — a signal matrix over SIGINT/SIGTERM/SIGUSR2/mid-export failure asserting the recorded reason, and a live migration through cutover — but that is not in the repo.

The cheapest gap to close is `interrupt`: the live E2E already stops the export with SIGTERM via `StopExportData`, so asserting the exit snapshot's reason there is a handful of lines in a test that already runs. `error` and `cutover` each need their own scenario. Happy to add any or all of them — say which and I will.

yb-voyager/cmd/schemaSnapshotCapture_integration_test.go
ShivanshGahlot: Had a look at how the repo places integration tests, and the convention cuts the other way for these files.

`integration`-tagged tests are co-located with the package under test in every case — `src/tgtdb`, `src/srcdb`, `cmd`, `src/schemasnapshot`, `src/schemadiff`, `src/pgss`, `src/cp/yugabyted`. Likewise all 7 `integration_voyager_command` tests live in `cmd`. The only test-only package is `src/testlivemigration`, and it earns that two ways: those tests are black-box (a separate package, so they shell out to the binary and cannot reach package internals) and they share a 1,688-line harness in `live_migration_testing_framework.go`.

`schemaSnapshotCapture_integration_test.go` is the opposite shape — white-box, with about 100 references to 9 unexported package-`cmd` identifiers: `captureSourceSchemaSnapshot`, `startPeriodicSourceSchemaSnapshotCapture`, `saveSourceSchemaSnapshotPlaceholder`, `initMetaDB`, the mutable globals `source`, `metaDB`, `exporterRole`, `disableSchemaSnapshotCapture`, plus `testPostgresSource` from this package's `TestMain`. Moving it to another package means exporting all of that purely to relocate a test, which seems like a bad trade.

`schemaSnapshotCaptureCommand_integration_test.go` could move with modest work — it already shells out to the binary, and reading the snapshots could go through the exported `metadb.NewMetaDB` the way `testlivemigration`'s `InitMetaDB()` does. But since every other `integration_voyager_command` test sits in `cmd`, moving just this one would make the tree less consistent rather than more.

Where a `testschemadrift` package would genuinely fit is as a new black-box tier mirroring `testlivemigration`: run a migration, change the source schema mid-flight, run `schema detect-drift`, assert the report. That is new coverage rather than a relocation, and it belongs on the detect-drift PR. Happy to scope it there if you want it.

yb-voyager/cmd/exportData.go
ShivanshGahlot: Done

yb-voyager/cmd/export.go
```diff
@@ -295,6 +300,31 @@ func validateSchemaSnapshotCaptureInterval(cmd *cobra.Command) error {
...
 
+// sourceCapture resolves the command globals into a schemasnapshot.SourceCapture, which
+// holds the capture policy itself. This adapter is the only place that reads them, so
+// the policy stays testable without command state.
+//
+// A source that is not a *srcdb.PostgreSQL yields a nil DB rather than an error: the
+// capture's own gate rejects non-PostgreSQL sources first, so the handle is never used.
+func sourceCapture() export.SchemaSnapshotCapture {
+	var db *sql.DB
+	if pg, ok := source.DB().(*srcdb.PostgreSQL); ok {
```
makalaaneesh: if !pg, let's just return?

makalaaneesh (approved):
LGTM with a minor comment

makalaaneesh:
let's revisit
- the export.SourceSchemaCapture 
- tests and where the tests should live (which pkg?)


notes: 
1. everything other than the periodic function can go into schemasnapshot package directly. 
2. no need to test the periodic function; periodic function can go into utils perhaps? 
3. all unit + integration (db) tests go into schemasnapshot. 
4. all cmd-level integration tests (that run export-data command and assert that snapshots were captured) go into a new testschemadrift package

PR #3658, opened 2026-07-08
tgtdb: skip amp target validation in Init

AmeyaKasture:

---

| Component        

### DIFF
--- a/yb-voyager/src/tgtdb/yugabytedb_amp.go
+++ b/yb-voyager/src/tgtdb/yugabytedb_amp.go
@@ -49,7 +49,7 @@ func (amp *TargetYugabyteDBAmp) Init() error {
 	if err := amp.TargetPostgreSQL.Init(); err != nil {
 		return err
 	}
-	return amp.validateAmpTarget()
+	return nil
 }
 
 // validateAmpTarget confirms the connected compute is a yb-amp endpoint


yb-voyager/src/tgtdb/yugabytedb_amp.go
```diff
@@ -49,7 +49,7 @@ func (amp *TargetYugabyteDBAmp) Init() error {
 	if err := amp.TargetPostgreSQL.Init(); err != nil {
 		return err
 	}
-	return amp.validateAmpTarget()
+	return nil
```
makalaaneesh, line 52: [AI] **Silently disabling `validateAmpTarget()` removes a guardrail and leaves dead code.**

`Init()` now returns `nil` instead of calling `amp.validateAmpTarget()`. Two concerns:

1. **Guardrail removed with no explanation.** `validateAmpTarget()` "confirms the connected compute is a yb-amp endpoint." After this change, an `Init()` against a non-amp target succeeds silently and any failure is deferred to a later, more confusing point in the migration. There is no comment or `TODO` explaining *why* the check is being skipped, so a future reader cannot tell whether this is intentional/temporary or an accidental regression.

2. **Dead code.** `validateAmpTarget()` (and any helpers it exclusively uses) now has no callers. Per the repo-root `.cursor/BUGBOT.md` — *"Remove dead code, unused functions, and leftover debugging artifacts before merging"* — the now-unreachable validation should either be removed or explicitly gated.

Failure case: a user points `import data` at a compute endpoint that is not a yb-amp endpoint; previously `Init()` failed fast with a clear message, now it proceeds and fails later (or behaves incorrectly) with no early signal.

Suggested fix: if the skip is temporary, keep the call behind a documented flag/condition and add a `// TODO(<owner>): re-enable amp validation once ...` comment; if it is permanent, delete `validateAmpTarget()` and its now-unused helpers so no dead code remains.

---

---

makalaaneesh (commented):
[AI] Automated review — 1 finding (0 critical, 1 warning). See inline comment. Single-line change disabling amp target validation in `Init()`; the warning covers the removed guardrail plus the resulting dead code.

---

PR #3661, opened 2026-07-09
Handle conflict detection for null values based on the index property…

priyanshi-yb:

Live-migration unique-key conflict detection previously handled NULLs via a blunt, hidden `--disable-null-conflicts` import flag: when set, `NULL == NULL` was never treated as a conflict; otherwise it always was. That is incorrect for schemas mixing default (`NULLS DISTINCT`) and `NULLS NOT DISTINCT` unique indexes, and forced an unsafe all-or-nothing choice on the user.

This PR removes the flag and instead derives NULL-conflict behavior **per unique index** from the index’s actual `NULLS [NOT] DISTINCT` property, read live from the import-target catalog. A new `tgtdb.UniqueIndex` type (`Columns` + `NullsNotDistinct`) replaces the previous `[][]string` representation across the target drivers and the conflict-detection cache. With `NULLS NOT DISTINCT`, two NULLs are treated as equal (conflict); with the default `NULLS DISTINCT`, NULLs never conflict with each other.

The target unique-index discovery query was simplified to a single `pg_index`-based path that reads `indnullsnotdistinct` defensively (`COALESCE(to_jsonb(ix)->>'indnullsnotdistinct', false)`), so it stays correct on PostgreSQL < 15 and YugabyteDB, and OR-merges the flag when leaf-partition indexes are folded into their root table. The cdcbench mock/artifact path was aligned to the same `UniqueIndex` shape (including capturing `NullsNotDistinct` in the artifact manifest), and NULL-distinct workloads that previously expected false-positive conflicts (`oltp-insert-null-then-fill`, `schema-composite-uk-null-component`) now correctly expect no conflicts.

- Removes the hidden `--disable-null-conflicts` import-data flag and its config key `disable-null-conflicts`. NULL-conflict handling is now automatic and per-index. Users who previously set it (CLI or config file) should remove it; leaving the key in a config file will fail config validation after upgrade.
- No changes to installation or reports.

- Updated unit tests in `conflictDetectionCache_test.go` exercising both `NULLS DISTINCT` and `NULLS NOT DISTINCT` conflict scenarios.
- Updated target integration tests `TestPostgresTargetGetTableToUniqueIndexesMap` and `TestYugabyteGetTableToUniqueIndexesMap` to assert the `NullsNotDistinct` property is captured for unique indexes/constraints (including partition merge).
- Updated `live_migration_uk_conflict_test.go` and `live_migration_test.go` for the new `UniqueIndex` type.
- Updated cdcbench artifact/mock to serve `tgtdb.UniqueIndex`, and flipped `ExpectConflicts` for the now-correct NULL DISTINCT workloads; `conflict-nulls-not-distinct` still expects conflicts.
- Verified with `go build -tags cdc_benchmark ./test/cdcbench/` and `./cmd/`.

No.

No. There are no metaDB/on-disk format changes — the removed flag was a CLI/config option only, and unique-index metadata (including the NULLS-NOT-DISTINCT property) is fetched live from the target at runtime.


yb-voyager/cmd/config.go
```diff
@@ -159,7 +159,7 @@ var allowedImportDataConfigKeys = mapset.NewThreadUnsafeSet[string](
 	"adaptive-parallelism-frequency-seconds", "min-available-memory-threshold", "max-batch-size-bytes",
 	"ybvoyager-use-task-picker-for-import",
 
-	"cdc-partitioning-strategy", "disable-null-conflicts",
+	"cdc-partitioning-strategy",
```
makalaaneesh, line 162: [AI] **Removing `disable-null-conflicts` from the allowed config keys is a hard-breaking upgrade for existing config files.**

`validateConfigFile` (config.go:605) returns a `ConfigValidationError` whenever any key in the config file is not in the section's allow-set. After this PR, a user who upgrades mid-migration with `import-data.disable-null-conflicts` still present in their config file will get a hard validation failure — not a warning — and the error ("invalid key") won't hint that the option was intentionally removed.

Failure case: existing `config.yaml` containing `import-data:\n  disable-null-conflicts: true` → next `import data` invocation aborts on `ConfigValidationError`.

Per the root `.cursor/BUGBOT.md` "Upgrade and Backward Compatibility" rule (serialized/user state must stay compatible across mid-migration upgrades, or the release must be flagged breaking). The flag was hidden and documented UNSAFE, so impact is limited, but consider one of:
- Gracefully accept-and-ignore the deprecated key with a one-line deprecation warning (keep it in the allow-set for one release), or
- Explicitly call this out in the release notes and mark the release as breaking.

---

---

.github/workflows/go.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches: ['main', '*.*-dev', '*.*.*-dev']
   pull_request:
-    branches: [main]
+    branches: [priyanshi/tech-debt-target-queries-unique-indexes]
```
makalaaneesh: [AI] **CI `pull_request` triggers are retargeted to the stacked base branch — must be reverted to `main` before merge.**

This PR points the `pull_request:` trigger of 7 workflows (`go.yml`, `integration-tests.yml`, `issue-tests.yml`, `failpoint-tests.yml`, `misc-migtests.yml`, `mysql-migtests.yml`, `pg-13-migtests.yml`) from `[main]` to `[priyanshi/tech-debt-target-queries-unique-indexes]`.

Failure case: if this diff reaches `main` unreverted, the `pull_request` triggers on `main` will only fire for PRs whose base is `priyanshi/tech-debt-target-queries-unique-indexes` — i.e. all normal PRs targeting `main` would stop running these checks.

The PR description already notes this is temporary and should be reverted before the base merges; flagging it here as a pre-merge blocker so it isn't lost. Restore `branches: [main]` in all 7 files (or rebase onto `main`) before this merges.

---

---

makalaaneesh (commented):
[AI] Automated review — 2 findings (0 critical, 2 warning). See inline comments. Core change is clean: `[][]string` → `tgtdb.UniqueIndex{Columns, NullsNotDistinct}` across the target drivers and conflict-detection cache, deriving NULL-conflict behavior per-index from the target catalog's `NULLS [NOT] DISTINCT` property instead of the removed global `--disable-null-conflicts` flag. Verified the PG/YB query is defensive on PG<15/YB (`COALESCE(to_jsonb(ix)->>'indnullsnotdistinct', false)`), that the hardcoded `false` in the `unique_constraints` CTE is harmless (the `unique_indexes` CTE captures constraint-backed indexes with the real flag and `bool_or` merges them by shared index name), and that unit + integration tests cover both DISTINCT and NOT DISTINCT for before-after, before-before, and composite keys. The two warnings are a backward-compat break on config removal and a temporary CI-trigger retarget that must be reverted before merge.

Non-blocking (not posted as findings): the `unique_constraints` CTE is arguably redundant now that `unique_indexes` captures all non-PK unique indexes; and the NULLS NOT DISTINCT *constraint* (as opposed to index) path isn't directly exercised by the integration test, though the logic covers it.

---

---

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -437,7 +437,8 @@ WITH unique_constraints AS (
         tc.table_name,
         tc.constraint_name AS index_key,
         kcu.column_name,
-        kcu.ordinal_position
+        kcu.ordinal_position,
+        false AS nulls_not_distinct
```
makalaaneesh: not possible to specify this for unique constraints?

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -456,7 +457,8 @@ unique_indexes AS (
         t.relname AS table_name,
         i.relname AS index_key,
         a.attname AS column_name,
-        array_position(ix.indkey, a.attnum) + 1 AS ordinal_position
+        array_position(ix.indkey, a.attnum) + 1 AS ordinal_position,
+        COALESCE((to_jsonb(ix) ->> 'indnullsnotdistinct')::boolean, false) AS nulls_not_distinct
```
makalaaneesh: is this a version compatibility thing? earlier versions did not have this field? pls add comments if so

makalaaneesh (approved):
LGTM with minor nits/questions

.github/workflows/go.yml
makalaaneesh: pls revert before merging

yb-voyager/src/tgtdb/postgres.go
priyanshi-yb: It is possible for unique constraints to also have the NULLS NOT DISTINCT property; it's a miss. Fixing it.
Thanks for catching that.

yb-voyager/src/tgtdb/postgres.go
priyanshi-yb: yeah this property is added in PG15+ versions

PR #3668, opened 2026-07-14
Fix flakiness in the couple of unit tests -  TestRandomBatchProducer_StashAndContinue_LastBatchHasAllErrors and TestExportSchemaSchemaOptimizationReportPerfOptimizationsAutofix

priyanshi-yb:

`TestRandomBatchProducer_StashAndContinue_LastBatchHasAllErrors` was flaky because `RandomBatchProducer` returns batches in non-deterministic order and `NextBatch()` does not block. The test assumed a fixed order (1-record batch first, then empty) and called `NextBatch()` a second time without waiting, which could fail with `RecordCount` mismatches or "no batches available".

This PR rewrites the test to use the existing `consumeAllBatches` helper to drain both batches, then asserts the expected shape in an order-independent way: one empty non-last batch (number 1) and one last batch (number 0) carrying the good row, with the oversized row still stashed against the last batch. Comments document why the small `maxBatchSizeBytes=15` setup produces that empty first batch (header + first row exceeds the limit).

`TestExportSchemaSchemaOptimizationReportPerfOptimizationsAutofix` was flaky, as the import schema was being run asynchronously in the test and then we were doing some validation on the schema created in the target, but in cases where import schema can take time, the validation can fail - import schema should run synchronously in the test, as it takes a bit of time to complete, and then anything else can be run after that

No user-facing changes.

- Updated unit test `TestRandomBatchProducer_StashAndContinue_LastBatchHasAllErrors` in `importDataRandomFileBatchProducer_test.go`.
- Updated unit test `TestExportSchemaSchemaOptimizationReportPerfOptimizationsAutofix` in `exportSchema_test.go`
- No production code changes; no additional integration tests needed.

No.

No.

### DIFF
--- a/yb-voyager/cmd/importDataRandomFileBatchProducer_test.go
+++ b/yb-voyager/cmd/importDataRandomFileBatchProducer_test.go
@@ -726,9 +726,17 @@ func TestRandomBatchProducer_StashAndContinue(t *testing.T) {
 	assert.True(t, producer.Done(), "Done() should be true after producer finishes")
 }
 
+// Issue: this test was flaky because RandomBatchProducer returns batches in non-deterministic
+// order and NextBatch() does not block. The test assumed a fixed order (batch with 1 record
+// first, then empty) and called NextBatch() a second time without waiting, so it could fail
+// with RecordCount mismatches or "no batches available".
+//
+// Fix: use consumeAllBatches to wait for both batches, then assert the expected shape
+// order-independently. With maxBatchSizeBytes=15, header (7) + first data row (11) = 18
+// exceeds the limit, so the first batch is finalized empty and the first row is carried to
+// the last batch; the oversized second row is stashed against that last batch.
 func TestRandomBatchProducer_StashAndContinue_LastBatchHasAllErrors(t *testing.T) {
-	// Set max batch size in bytes to a small value to trigger the row-too-large error
-	maxBatchSizeBytes := int64(15) // deliberately small to trigger error
+	maxBatchSizeBytes := int64(15)
 	ldataDir, lexportDir, state, _, progressReporter, err := setupExportDirAndImportDependencies(1000, maxBatchSizeBytes)
 	require.NoError(t, err)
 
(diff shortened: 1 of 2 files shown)


yb-voyager/cmd/importDataRandomFileBatchProducer_test.go
```diff
@@ -726,9 +726,17 @@ func TestRandomBatchProducer_StashAndContinue(t *testing.T) {
 	assert.True(t, producer.Done(), "Done() should be true after producer finishes")
 }
 
+// Issue: this test was flaky because RandomBatchProducer returns batches in non-deterministic
+// order and NextBatch() does not block. The test assumed a fixed order (batch with 1 record
+// first, then empty) and called NextBatch() a second time without waiting, so it could fail
+// with RecordCount mismatches or "no batches available".
+//
+// Fix: use consumeAllBatches to wait for both batches, then assert the expected shape
```
makalaaneesh: this seems more like a PR comment, not a code comment?

makalaaneesh (approved):
LGTM !

makalaaneesh:

Reviewed both test-only changes against the branch-review lenses (test-correctness, determinism, whether assertions match intended behavior) and the applicable `.cursor/BUGBOT.md` hierarchy (repo root → `yb-voyager/` → `yb-voyager/cmd/`).

- `exportSchema_test.go`: switching the `import schema` invocation from async (`true`) to sync (`false`) makes `RunVoyagerCommand` call `cmd.Wait()` before the test queries YugabyteDB for index sharding strategies, fixing the race where indexes were checked before `import schema` finished. Consistent with every other synchronous `RunVoyagerCommand` call in the file.
- `importDataRandomFileBatchProducer_test.go`: the rework drops the order-dependent `NextBatch()` assertions in favor of `consumeAllBatches` + classification by `RecordCount`. This matches the producer's actual behavior — an empty first batch (Number 1, header carried forward) and a last batch with the good row (Number `LAST_SPLIT_NUM`=0); the oversized-row error file is renamed to batch 0 on `isLastBatch` (`errorHandlers.go`), so anchoring the assertion at `lastBatch.Number` is correct. Ran the test 5× with `-tags unit`; passes deterministically.

Only Suggestion-level observations (not posted): the `doDuringCmd` `sleep(10s)` in the export test is now redundant under sync mode, and the trailing `assert.True(producer.Done())` is tautological since `consumeAllBatches` only returns once `Done()` is true.

---

---

PR #3671, opened 2026-07-15
Add configuration in import data for users to be able to override the partition key per table as `cdc-partition-key-overrides` and renaming `cdc-partition-strategy` to `cdc-partition-key`

priyanshi-yb:

This PR reworks how the CDC partition strategy is configured during live migration (`import data to target`, PG→YB). It renames the existing global flag `--cdc-partitioning-strategy` to `--cdc-partition-key` (values `auto`, `pk`, `table`) and adds a new optional `--cdc-partition-key-overrides` flag that lets users override the strategy for specific tables using `schema.table:pk|table` pairs separated by `;` (e.g. `public.orders:table;sales.events:pk`). Overrides only support `pk`/`table` (not `auto` or custom columns); tables not listed keep the global partition key.

The resolution logic was extracted into a new file (`cmd/live_migration_cdc_partition_strategy.go`) and split into pure, unit-testable helpers: `parseCdcPartitionKeyOverrides` (syntax), `resolveCdcPartitionKeyOverrides` (name-registry lookup + import-table-list membership + duplicate detection on the resolved `NameTuple`), and `resolveEffectiveCdcPartitionKeys` (applies the global strategy, overlays per-table overrides, then rejects an effective `pk` for tables with expression-based unique indexes, which must run sequentially via `table`). Parsing, validation, and expr-UK checks now run in a new `prepareCdcPartitionKey` step that executes **before snapshot import** (after start-clean, before marking import started) so bad configs fail fast instead of erroring at `streamChanges`. The per-table map is persisted to metaDB, and a change-guard rejects modifying the partition key or overrides on resume unless `--start-clean` is used.

Both config values are persisted in `ImportDataStatusRecord` (the overrides string is a new additive field), and start-clean now clears the persisted per-table map so it is recomputed for the new run. Flag validation was also hardened to return errors (rather than calling `utils.ErrExit` directly) and start-clean state clearing now checks its metaDB update errors.

- **Command line:** `--cdc-partitioning-strategy` is renamed to `--cdc-partition-key`; new `--cdc-partition-key-overrides` flag added. Both are hidden flags for `import data to target`. These flags are only valid for live migration with a PostgreSQL source and a YugabyteDB target (rejected for offline/non-PG/non-YB/non-target with clear errors).
- **Configuration:** `cdc-partition-key` and `cdc-partition-key-overrides` are now allowed keys in the import-data config (replacing `cdc-partitioning-strategy`).
- **Behavior on resume:** changing the partition key or overrides after import has started is rejected unless `--start-clean` is passed; resuming an import started by an older voyager (no partition key persisted) requires `--start-clean`.
- **Guardrail:** `pk` (global or via override) is rejected for tables with an expression-based unique index; the fix is to set that table to `table`.
- No changes to installation or reports.

- **New unit tests** in `yb-voyager/cmd/live_migration_test.go`:
  - `TestParseCdcPartitionKeyOverrides` — override string parsing (whitespace, trailing `;`, rejecting `auto`/custom columns/missing colon/empty strategy/duplicates).
  - `TestResolveEffectiveCdcPartitionKeys` — global-strategy resolution, per-table override overlay, AMP forcing `table`, and rejection of `pk` on expression-UK tables.
  - `TestResolveCdcPartitionKeyOverrides` — name-registry resolution, import-table-list membership, and duplicate/conflict detection across different spellings of the same table.
  - `TestValidateCdcPartitionKeyFlags` — context guardrails (target/YB/PG/streaming), value validation, and resume change-guards (same-config pass, changed key/overrides rejection, older-version resume, and start-clean bypass).
- **New integration tests** in `src/testlivemigration/live_migration_uk_conflict_test.go`:
  - `TestLiveMigrationCdcPartitionKeyConfigs` — invalid override fails before snapshot; global `pk` succeeds and a resume rejecting an overrides change; a `--start-clean` run completing live migration through cutover with a mix of `pk`/`table` strategies via overrides, asserting the persisted per-table map.
  - `TestLiveMigrationCdcPartitionKeyRejectsPkOnExpressionUniqueIndex` — end-to-end verification of the expression-UK guardrail (real `getExpressionUniqueIndexTables` query) and the documented `table` remedy, through cutover.
- Existing live-migration integration tests (`live_migration_integration_test.go`) and the CDC ingest benchmark shim were updated for the renamed flag/variable.

No.

Yes — `ImportDataStatusRecord` (MetaDB / Import Data State) gains a new `cdcPartitionKeyOverridesConfig` field. The change is additive and the existing `cdcPartitioningStrategyConfig` JSON tag is preserved for backward compatibility. Resuming an import started by an older version (which lacks a persisted partition key) is explicitly guarded and requires `--start-clean`.

---

| Component |
| MetaDB |
| Name registry json |
| Data File Descriptor Json |
| Callhome Json |
| Export Status Json |
| YugabyteD Tables |
| TargetDB Metadata Tables |
| Schema Dump |
| AssessmentDB |
| Import Data State |
| Data .sql files of tables |


yb-voyager/cmd/importData.go
```diff
@@ -924,13 +925,14 @@ func setupImportDataObservability() error {
...
-func updateImportDataStartedInMetaDB() error {
+func updateImportDataStartedAndSomeConfigsInMetaDB() error {
 	switch importerRole {
 	case TARGET_DB_IMPORTER_ROLE:
-		log.Infof("updating import data started in meta db with cdc partitioning strategy: %s", cdcPartitioningStrategy)
+		log.Infof("updating import data started in meta db with cdc-partition-key: %s, overrides: %q", cdcPartitionKey, cdcPartitionKeyOverrides)
 		err := metaDB.UpdateImportDataStatusRecord(func(record *metadb.ImportDataStatusRecord) {
 			record.ImportDataStarted = true
-			record.CdcPartitioningStrategyConfig = cdcPartitioningStrategy
+			record.CdcPartitioningStrategyConfig = cdcPartitionKey
```
makalaaneesh, line 934: [AI] **Flagless resume overwrites the persisted cdc-partition-key config, which can later falsely block a legitimate resume.**

`updateImportDataStartedAndSomeConfigsInMetaDB` runs unconditionally on every `import data` invocation (called at `importData.go:1246`) and always writes `record.CdcPartitioningStrategyConfig = cdcPartitionKey` and `record.CdcPartitionKeyOverridesConfig = cdcPartitionKeyOverrides` using the *current* flag values, which fall back to their defaults (`"auto"` / `""`) when the flags are not passed.

Failure sequence (PG->YB live import to target):
1. First run: `import data ... --cdc-partition-key pk`. Stored config becomes `pk`; per-table `TableToCDCPartitioningStrategyMap` is computed for `pk`.
2. Process is killed and resumed with a bare `import data ...` (no cdc flags). `validateCdcPartitionKeyFlags` returns early at the `!anyPassed` guard, so no change-guard runs; then this function silently rewrites the stored `CdcPartitioningStrategyConfig` from `pk` to the default `auto` (and overrides to `""`). Streaming still works because the per-table map is preserved.
3. A later resume that re-passes `--cdc-partition-key pk` now hits the change-guard at `import.go:345` and compares `pk != "auto"`, so it `ErrExit`s with "changing cdc-partition-key is not allowed. Current: auto, new: pk" — a false positive that blocks a resume where nothing actually changed. The stored config is now permanently out of sync with the map that is actually driving streaming.

The persisted config should only be written on the first start (when it has not been set yet / the per-table map is nil), mirroring how `prepareCdcPartitionKey` no-ops on resume. Suggested fix: guard these two assignments so a resume does not clobber an already-set value, e.g. only set them when `record.CdcPartitioningStrategyConfig == "" && record.TableToCDCPartitioningStrategyMap == nil`, or move the config persistence into `prepareCdcPartitionKey`/`computeAndPersistCdcPartitioningStrategyPerTable` where the first-run vs resume distinction is already made.

---

yb-voyager/cmd/config.go
```diff
@@ -159,7 +159,7 @@ var allowedImportDataConfigKeys = mapset.NewThreadUnsafeSet[string](
 	"adaptive-parallelism-frequency-seconds", "min-available-memory-threshold", "max-batch-size-bytes",
 	"ybvoyager-use-task-picker-for-import",
 
-	"cdc-partitioning-strategy",
+	"cdc-partition-key", "cdc-partition-key-overrides", "disable-null-conflicts",
```
makalaaneesh: [AI] **`disable-null-conflicts` is added as an allowed config key but no such flag exists anywhere — dead/misleading config surface, and out of scope for this PR.**

This line adds three keys to `allowedImportDataConfigKeys`. `cdc-partition-key` and `cdc-partition-key-overrides` are backed by real flags registered in `import.go`, but `disable-null-conflicts` has no corresponding flag or reader anywhere in the codebase (a repo-wide search finds it only on this line).

Because config values are applied by iterating over the command's registered flags (`cmd.Flags().VisitAll` in `applyConfigToFlags`), a `disable-null-conflicts` key in a user's config file is silently ignored rather than rejected. That is worse than rejecting it: a user who sets `disable-null-conflicts: true` gets no error and no effect, and the allow-list implies the option is supported.

This also appears unrelated to the stated goal of the PR (partition-key override + rename). Suggested fix: drop `disable-null-conflicts` from this set until the backing flag is actually introduced (in the PR that adds that feature), keeping the allow-list to `"cdc-partition-key", "cdc-partition-key-overrides"`.

---

makalaaneesh (commented):

---

---

yb-voyager/cmd/import.go
```diff
@@ -249,19 +337,19 @@ func validateCdcPartitioningStrategyFlag(cmd *cobra.Command) error {
...
 	if importDataStatus == nil || !importDataStatus.ImportDataStarted || bool(startClean) {
-		//if import data has not started or start-clean flag is used, allow the change in cdc partitioning strategy
 		return nil
 	}
 	if importDataStatus.CdcPartitioningStrategyConfig == "" {
-		//if not a first run and the cdc partitioning strategy is not set
-		//this can be the case when the import data is resumed from an earlier version of yb-voyager
-		//So we should use the cdc partitioning strategy as pk to be upgrade safe
-		utils.ErrExit("Resuming from an earlier version of yb-voyager is not supported as cdc partition strategy was not set. Use --start-clean to start a fresh import with the new yb-voyager version.")
+		utils.ErrExit("Resuming from an earlier version of yb-voyager is not supported as cdc partition key was not set. Use --start-clean to start a fresh import with the new yb-voyager version.")
```
makalaaneesh: this msg is not really clear. What is this case?

yb-voyager/cmd/import.go
```diff
@@ -249,19 +331,19 @@ func validateCdcPartitioningStrategyFlag(cmd *cobra.Command) error {
...
 	}
-	if cdcPartitioningStrategy != importDataStatus.CdcPartitioningStrategyConfig {
-		utils.ErrExit("changing the cdc partitioning strategy is not allowed after the import data has started. Current strategy: %s, new strategy: %s\n Use --start-clean to start a fresh import with the new strategy.", importDataStatus.CdcPartitioningStrategyConfig, cdcPartitioningStrategy)
+	storedOverrides := importDataStatus.CdcPartitionKeyOverridesConfig //TODO: just checking the string equality is not enough, we might need to compare the overrides by properly
+	if cdcPartitionKeyOverrides != storedOverrides {
+		utils.ErrExit("changing cdc-partition-key-overrides is not allowed after the import data has started. Current: %q, new: %q\n Use --start-clean to start a fresh import with the new overrides.", storedOverrides, cdcPartitionKeyOverrides)
 	}
-	log.Infof("cdc partitioning strategy: %s", cdcPartitioningStrategy)
+	log.Infof("cdc-partition-key: %s, cdc-partition-key-overrides: %q", cdcPartitionKey, cdcPartitionKeyOverrides)
```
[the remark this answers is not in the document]
makalaaneesh: this is part of what i was mentioning in our discussion @priyanshi-yb

yb-voyager/cmd/import.go
```diff
@@ -218,29 +220,115 @@ func validateImportUsePartitionRootFlag() error {
...
+		}
+		tableName := strings.TrimSpace(parts[0])
+		strategy := strings.TrimSpace(parts[1])
+		if tableName == "" || strategy == "" {
+			return nil, goerrors.Errorf("invalid cdc-partition-key-overrides entry %q: table and strategy must both be non-empty", entry)
+		}
+		if strategy != PARTITION_BY_PK && strategy != PARTITION_BY_TABLE {
+			return nil, goerrors.Errorf("invalid cdc-partition-key-overrides strategy %q for table %q: supported values are %s, %s",
+				strategy, tableName, PARTITION_BY_PK, PARTITION_BY_TABLE)
+		}
```
makalaaneesh, line 249: nit: use validCdcPartitionKeys?

yb-voyager/cmd/import.go
```diff
@@ -218,29 +220,115 @@ func validateImportUsePartitionRootFlag() error {
...
+			return nil, goerrors.Errorf("duplicate table %q in cdc-partition-key-overrides", tableName)
+		}
+		result[tableName] = strategy
+	}
+	return result, nil
+}
+
+// resolveCdcPartitionKeyOverrides looks up override table names in namereg and
+// validates each is present in importTableList. Returns a map keyed by NameTuple.
+func resolveCdcPartitionKeyOverrides(rawOverrides map[string]string, importTableList []sqlname.NameTuple) (*utils.StructMap[sqlname.NameTuple, string], error) {
```
makalaaneesh: this func is called by live_migration.go. Can we put it somewhere other than import? maybe in live_migration.go itself?

yb-voyager/cmd/import.go
```diff
@@ -218,29 +220,115 @@ func validateImportUsePartitionRootFlag() error {
...
+		if existing, ok := resolved.Get(tuple); ok && !strings.EqualFold(existing, strategy) {
+			return nil, goerrors.Errorf("cdc-partition-key-overrides: table %q (resolved to %s) specified multiple times with conflicting strategies %q and %q",
+				tableSpec, tuple.ForOutput(), existing, strategy)
+		}
+		resolved.Put(tuple, strategy)
+	}
+	return resolved, nil
+}
+
+func validateCdcPartitionKeyFlags(cmd *cobra.Command) error {
```
makalaaneesh, line 261: no ErrExit. return proper errors. Pls take care of this in this function

yb-voyager/cmd/live_migration.go
```diff
@@ -190,14 +190,40 @@ func streamChanges(state *ImportDataState, tableNames []sqlname.NameTuple) error
...
-PK for normal tables and table for tables having expression based unique indexes
+prepareCdcPartitionKey validates overrides against the import table list, resolves
+per-table strategies (global --cdc-partition-key + --cdc-partition-key-overrides +
+auto/expr-UK rules), and persists TableToCDCPartitioningStrategyMap.
 
-If its not the first run, it will fetch the strategy from the metadb key IMPORT_DATA_STATUS_KEY
+It is intentionally called before snapshot import so bad configs fail fast.
+On resume (map already in metaDB) this is a no-op.
+*/
+func prepareCdcPartitionKey(tableNames []sqlname.NameTuple) error {
```
makalaaneesh: I would recommend all these new functions related to cdc partitioning to go into a new file. Don't want to clutter live_migration.go

yb-voyager/cmd/live_migration.go
```diff
@@ -236,67 +261,139 @@ func getCdcPartitioningStrategyPerTable(tableNames []sqlname.NameTuple) (*utils.
...
+			if strategy == PARTITION_BY_PK {
+				needsExprUKCheck = true
+				return false, nil
+			}
+			return true, nil
+		})
+		if err != nil {
+			return nil, fmt.Errorf("error iterating cdc-partition-key-overrides: %w", err)
+		}
+	}
```
makalaaneesh: nit: separate function

yb-voyager/cmd/live_migration.go
```diff
@@ -236,67 +261,139 @@ func getCdcPartitioningStrategyPerTable(tableNames []sqlname.NameTuple) (*utils.
...
+
+	var expressionUniqueIndexTables []sqlname.NameTuple
+	if needsExprUKCheck && tconf.TargetDBType != YUGABYTEDB_AMP {
```
makalaaneesh: AMP check here seems out of place. 
What is the plan for AMP? While ingesting, are we using partition-by-table? If so, then that check should be made much earlier, no?

yb-voyager/cmd/live_migration.go
makalaaneesh: log the value also

yb-voyager/cmd/live_migration.go
```diff
@@ -236,67 +261,139 @@ func getCdcPartitioningStrategyPerTable(tableNames []sqlname.NameTuple) (*utils.
...
+		overrides = utils.NewStructMap[sqlname.NameTuple, string]()
+	}
+	if exprUKSet == nil {
+		exprUKSet = utils.NewStructMap[sqlname.NameTuple, bool]()
+	}
+
+	switch globalKey {
 	case "auto":
-		if tconf.TargetDBType == YUGABYTEDB_AMP {
+		if isAMP {
```
makalaaneesh: i saw some logic above but just checking again: do we have guardrails to prevent user from setting cdc-partition-* settings if target is AMP?

yb-voyager/cmd/import.go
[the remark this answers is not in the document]
priyanshi-yb: So the thing is- for parsing and preparing a list of partition strategies for the tables, we need to know the tables that have expression-based unique indexes (at the start, as fetching new ones can result in a change in config in case some schema drift and expression-based unique indexes are added/removed) so we may have to store and have a guardrail properly on the maps (stored vs parsed from config) to error out with change in config

yb-voyager/cmd/import.go
priyanshi-yb: This is in case the first run was with an old Voyager version, which doesn't have support for this config, and then the user re-runs the import data; then it's not allowed, and hence the user is supposed to pass --start-clean.

yb-voyager/cmd/import.go
priyanshi-yb: validCdcPartitionKeys - has auto also, do we want to allow passing auto in the config for a table overrides?

yb-voyager/cmd/live_migration.go
priyanshi-yb: >What is the plan for AMP? While ingesting, are we using partition-by-table? 

Yes, it's partition_by_table for AMP

priyanshi-yb: >If so, then that check should be made much earlier, no?

fixed it

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+// resolveEffectiveCdcPartitionKeys applies global strategy, then per-table overlays,
+// then rejects effective pk for expression-UK tables. Pure helper for unit tests.
+func resolveEffectiveCdcPartitionKeys(
+	tableNames []sqlname.NameTuple,
+	globalKey string,
+	overrides *utils.StructMap[sqlname.NameTuple, string],
+	exprUKSet *utils.StructMap[sqlname.NameTuple, bool],
+	isAMP bool,
```
makalaaneesh: nit: i'd suggest just take targetDBType as input, and then you can decide isAMP or not inside.

yb-voyager/cmd/import.go
makalaaneesh: isn't this just a breaking change then? can we just mark it as a breaking release?

yb-voyager/cmd/config.go
makalaaneesh: pls remove this

yb-voyager/cmd/importData.go
makalaaneesh: @priyanshi-yb this seems like a valid problem?

yb-voyager/cmd/import.go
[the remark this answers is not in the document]
makalaaneesh: let's create a task and take this up separately

yb-voyager/cmd/import.go
```diff
@@ -218,29 +218,79 @@ func validateImportUsePartitionRootFlag() error {
...
+	}
+	return result, nil
+}
+
+func validateCdcPartitionKeyFlags(cmd *cobra.Command) error {
+	globalPassed := cmd.Flags().Changed("cdc-partition-key")
+	overridesPassed := cmd.Flags().Changed("cdc-partition-key-overrides")
+	anyPassed := globalPassed || overridesPassed
+
+	if importerRole != TARGET_DB_IMPORTER_ROLE && tconf.TargetDBType != YUGABYTEDB_AMP {
```
makalaaneesh: if importerRole != TARGET_DB_IMPORTER_ROLE || tconf.TargetDBType != YUGABYTEDB

yb-voyager/src/testlivemigration/live_migration_integration_test.go
```diff
@@ -1095,6 +1095,164 @@ FROM generate_series(1, 10);`,
 
 }
 
+// TestLiveMigrationCdcPartitionKeyConfigs covers three import-data config cases:
+//  1. invalid overrides fail before snapshot
+//  2. global pk succeeds; changing overrides on resume is rejected
+//  3. start-clean with mixed pk/table overrides completes live migration through cutover
+func TestLiveMigrationCdcPartitionKeyConfigs(t *testing.T) {
```
makalaaneesh: let's put it in the _conflict_test.go file?

makalaaneesh (approved):
LGTM with nits (one important bug fix required)
See if we need more tests around guardrails

yb-voyager/cmd/import.go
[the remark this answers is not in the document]
priyanshi-yb: https://yugabyte.atlassian.net/browse/DB-22597

yb-voyager/cmd/import.go
priyanshi-yb: As discussed, this change was there from the start of this config cdc-partition-strategy, and it wasn't a breaking change, as it just requires import-data with start-clean

PR #3673, opened 2026-07-16
Add a config in import data to disable sequential scan on update/deletes to avoid retriable errors during cdc streaming with repeatable read isolation level

priyanshi-yb:

During live migration CDC import into YugabyteDB under repeatable-read isolation, UPDATE/DELETE statements that fall back to sequential scans can hit retryable conflicts under concurrent writes. This PR adds a hidden import-data setting, `--disable-sequential-scan-on-update-deletes` (default `true`), that prefixes UPDATE and DELETE SQL with the YB planner hint `/*+ Set(enable_seqscan off) */` so those statements prefer index access.

Updates (which are queued as plain SQL) and deletes (prepared statements) both honor the flag. The key is allowed in import-data config files and stored on `TargetConf` so it can be toggled off if needed.

1. **Command line:** Adds hidden flag `--disable-sequential-scan-on-update-deletes` (default `true`) on import-data.
2. **Configuration:** Same key is allowed in import-data config files.
3. **Installation:** No.
4. **Reports:** No.

Default behavior changes for CDC update/delete ingest (seqscan disabled via hint). The flag is hidden, so typical CLI usage is unchanged unless someone opts out.

- Exercised via live-migration / partition CDC import paths that run `ExecuteBatch` against YugabyteDB (including partition tests that cover update/delete streaming).
- No dedicated unit test for the hint string itself; behavior is covered by CDC import integration/live-migration tests against a real YB target.
- Manual verification: with the flag on (default), UPDATE/DELETE statements in the import batch include the `enable_seqscan off` hint; setting the flag to `false` leaves statements unchanged.

No.

No. `TargetConf` gains an additive JSON field (`disable_sequential_scan_on_update_deletes`); older Voyager versions ignore unknown fields, and the new default is applied from the CLI flag when unset.

---

| Component        

### DIFF
--- a/migtests/tests/resumption/live-migration/partition-iteration-test/scenario.yaml.template
+++ b/migtests/tests/resumption/live-migration/partition-iteration-test/scenario.yaml.template
@@ -1,6 +1,6 @@
 name: partition-iteration-test
 
-num_iterations: 5
+num_iterations: 20
 
 env:
   LOG_LEVEL: info
--- a/migtests/tests/resumption/live-migration/partition-iteration-test/scenario.yaml.template
+++ b/migtests/tests/resumption/live-migration/partition-iteration-test/scenario.yaml.template
@@ -1,6 +1,6 @@
 name: partition-iteration-test
 
-num_iterations: 20
+num_iterations: 5
 
 env:
   LOG_LEVEL: info
--- a/migtests/tests/resumption/live-migration/partition-iteration-test/scenario.yaml.template
+++ b/migtests/tests/resumption/live-migration/partition-iteration-test/scenario.yaml.template
@@ -134,7 +134,6 @@ voyager:
   import_data:
     flags:
       use-partition-root: "false"
-      disable-sequential-scan-on-update-deletes: "true"
   import_to_source:
     flags:
       use-partition-root: "false"
--- a/migtests/tests/resumption/live-migration/partition-iteration-test/scenario.yaml.template
+++ b/migtests/tests/resumption/live-migration/partition-iteration-test/scenario.yaml.template
@@ -134,6 +134,7 @@ voyager:
   import_data:
     flags:
       use-partition-root: "false"
+      disable-sequential-scan-on-update-deletes: "true"
   import_to_source:
     flags:
       use-partition-root: "false"
--- a/yb-voyager/src/tgtdb/event.go
+++ b/yb-voyager/src/tgtdb/event.go
@@ -404,7 +404,7 @@ func (event *Event) getUpdateStmt(tdb TargetDB, usePartitionRoot bool, disableSe
 	}
 	stmt = fmt.Sprintf(updateTemplate, tableName.ForUserQuery(), setClause, whereClause)
 	if disableSequentialScanOnUpdateDeletes {
-		stmt = fmt.Sprintf("%s %s", stmt, DISABLE_SEQUENTIAL_SCAN_ON_UPDATE_DELETES_HINT)
+		stmt = fmt.Sprintf("%s %s", DISABLE_SEQUENTIAL_SCAN_ON_UPDATE_DELETES_HINT, stmt)
 	}
 	return stmt, nil
 }
@@ -429,7 +429,7 @@ func (event *Event) getDeleteStmt(tdb TargetDB, usePartitionRoot bool, disableSe
 	}
 	stmt = fmt.Sprintf(deleteTemplate, tableName.ForUserQuery(), whereClause)
 	if disableSequentialScanOnUpdateDeletes {
-		stmt = fmt.Sprintf("%s %s", stmt, DISABLE_SEQUENTIAL_SCAN_ON_UPDATE_DELETES_HINT)
+		stmt = fmt.Sprintf("%s %s", DISABLE_SEQUENTIAL_SCAN_ON_UPDATE_DELETES_HINT, stmt)
 	}
 	return stmt, nil
 }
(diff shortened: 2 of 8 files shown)


migtests/tests/resumption/live-migration/partition-iteration-test/scenario.yaml.template
```diff
@@ -1,6 +1,6 @@
 name: partition-iteration-test
 
-num_iterations: 20
+num_iterations: 5
```
makalaaneesh: [AI] **Unrelated `num_iterations` reduction cuts resumption-test coverage by 75%.**

This PR lowers `num_iterations` from 20 to 5 in the `partition-iteration-test` Jenkins scenario template. Tracing the branch history, this drop was introduced in commit ce33340 ("miss for the prepare statement") — a commit whose message is about the prepared-statement code path, not the iteration count — and the `disable-sequential-scan-on-update-deletes` flag line that the PR originally added to this scenario was later removed (redundant once the default flipped to true). That leaves the 20→5 change looking like an accidental/leftover debugging edit with no connection to the PR's stated goal.

Running only 5 iterations instead of 20 materially weakens the live-migration + fall-back resumption soak this scenario is meant to provide. Per `migtests/.cursor/BUGBOT.md` ("Remove unnecessary artifacts ... no leftover debugging comments, no commented-out code") and general scope hygiene, incidental changes unrelated to the PR should not ride along.

Suggested fix: restore `num_iterations: 20`. If the reduction is deliberate (e.g., to cut Jenkins runtime), call it out explicitly in the PR description and keep it in a separate commit rather than folding it into this feature.

---

yb-voyager/src/tgtdb/event.go
```diff
@@ -403,10 +403,13 @@ func (event *Event) getUpdateStmt(tdb TargetDB, usePartitionRoot bool) (string,
 		return "", err
 	}
 	stmt = fmt.Sprintf(updateTemplate, tableName.ForUserQuery(), setClause, whereClause)
+	if disableSequentialScanOnUpdateDeletes {
+		stmt = fmt.Sprintf("%s %s", DISABLE_SEQUENTIAL_SCAN_ON_UPDATE_DELETES_HINT, stmt)
```
makalaaneesh: [AI] **No test asserts the seqscan hint is injected (or correctly omitted) in the generated SQL.**

This PR adds conditional hint injection to `getUpdateStmt`, `getDeleteStmt`, `getPreparedUpdateStmt`, and `getPreparedDeleteStmt`, and the flag defaults to `true`, so every YugabyteDB live-migration UPDATE/DELETE now emits `/*+ Set(enable_seqscan off) */ ...` by default. This is new behavior on the per-event CDC apply path, yet there is no unit test in `yb-voyager/src/tgtdb/` exercising these statement builders. A regression — hint dropped, wrong placement/spacing, or the hint leaking onto the Oracle/PostgreSQL callers that intentionally pass `false` (`oracle.go:524/531`, `postgres.go:763/770`) — would go completely unnoticed.

The delete path is worth asserting explicitly: deletes are built via `getPreparedDeleteStmt` and the hint is prepended to the *prepared* statement text (`event.go:525`), so a test should confirm the hint actually appears in the prepared SQL, not only the inline UPDATE case.

Suggested fix: add a table-driven unit test (per `yb-voyager/src/tgtdb/.cursor/BUGBOT.md`, include a case-sensitive table-name case) that calls all four builders with `disableSequentialScanOnUpdateDeletes` set true and false and asserts the exact generated SQL — hint present when true, absent when false. This both prevents regressions and documents the intended contract for the hardcoded-`false` Oracle/PG callers.

---

makalaaneesh (commented):

---

---

yb-voyager/src/tgtdb/event.go
```diff
@@ -463,7 +469,7 @@ func (event *Event) getPreparedInsertStmt(tdb TargetDB, targetDBType string, use
 }
 
 // NOTE: PS for each event of same table can be different as it depends on columns being updated
-func (event *Event) getPreparedUpdateStmt(tdb TargetDB, usePartitionRoot bool) (string, error) {
+func (event *Event) getPreparedUpdateStmt(tdb TargetDB, usePartitionRoot bool, disableSequentialScanOnUpdateDeletes bool) (string, error) {
```
makalaaneesh: did we test this? 
In case of prepared statements, does the hint need to be given in the PREPARE or when we actually execute?

AFAIK, planning happens when you execute, not when you prepare. Worth testing this manually

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -760,7 +760,7 @@ func (pg *TargetPostgreSQL) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
 	for i := 0; i < len(batch.Events); i++ {
 		event := batch.Events[i]
 		if event.Op == "u" {
-			stmt, err := event.GetSQLStmt(pg, pg.tconf.UsePartitionRoot)
+			stmt, err := event.GetSQLStmt(pg, pg.tconf.UsePartitionRoot, false)
```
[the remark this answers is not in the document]
makalaaneesh: NA

yb-voyager/src/tgtdb/event.go
priyanshi-yb: I believe it does work, as this issue is for both UPDATES/DELETEs and for DELETES we do use the prepared statement. When I was testing the change at the starting i only made changes for the UPDATE statement, and the test was failing even after that. Then I realised deletes also need to be handled, so I made this change for DELETEs also. And it worked after that. So prepared statements seem to be working with this hint, but I will double check.

yb-voyager/src/tgtdb/event.go
```diff
@@ -403,10 +403,13 @@ func (event *Event) getUpdateStmt(tdb TargetDB, usePartitionRoot bool) (string,
 		return "", err
 	}
 	stmt = fmt.Sprintf(updateTemplate, tableName.ForUserQuery(), setClause, whereClause)
+	if disableSequentialScanOnUpdateDeletes {
```
makalaaneesh: Can we do this only in the targetYugabyteDB layer? since it's only really releavant for YB? After calling event.GetStatement(), we just prepend the hint to it.

That way, 
change is localized to target YB. much less code changes overall.

makalaaneesh (approved):
LGTM

PR #3675, opened 2026-07-16
Add FORCE_CONFLICT mode to event-generator for testing CDC conflict detection

yaronp68:
## Summary
- Adds a `force_conflicts` mode to the event-generator (`migtests/scripts/event-generator/`) that deliberately induces unique-key value collisions between CDC events during live migration: frees a unique (non-PK) column's value on one row via `DELETE`/`UPDATE`, then immediately reuses it on a different row via `INSERT`/`UPDATE`. Because the two rows have different primary keys, the resulting CDC events can land on different parallel apply channels and race — this is the exact scenario `ConflictDetectionCache` (`yb-voyager/cmd/conflictDetectionCache.go`) is built to detect and serialize.
- Adds `faker_for_key()` for deterministic, reproducible payloads on the reused row (same key → same generated row, useful for correlating logs/reruns).
- Adds discovery of plain single-column `UNIQUE` indexes (`_find_unique_columns()`) so the new operation only targets qualifying tables.
- Opt-in via a new `force_conflicts` config block (`enabled`, `weight`, `free_via`, `reuse_via`), disabled by default — fully backward compatible with existing configs.

## Test plan
- [x] Verified end-to-end against a real live migration: Postgres source → yb-voyager (`--parallel-jobs 4`) → YugabyteDB target, with `force_conflicts.enabled: true`.
- [x] Confirmed dozens of `"conflict detected for table ..."` log lines from `ConflictDetectionCache`, each followed by the expected `"waiting for event(vsn=X) to be complete..."` serialization.
- [x] Confirmed source and target data stayed fully consistent afterward (no duplicate unique values, row-for-row identical between source and target).
- [x] Config validates against the existing schema; new keys are optional and additive.

### DIFF
--- a/migtests/scripts/event-generator/utils.py
+++ b/migtests/scripts/event-generator/utils.py
@@ -1106,7 +1106,11 @@ def force_conflict_operation(
                 f"UPDATE {table_name} SET {unique_col} = %s WHERE {pk_where}",
                 [new_value] + source_pk_values,
             )
-        conn.commit()
+        # Deliberately not committed here: the free and reuse statements must
+        # land in the same source transaction so they're exported at the same
+        # time and only then raced against each other in parallel import - two
+        # separately-committed transactions could straddle an export batch
+        # boundary and never actually race. See PR #3675 review discussion.
 
         fake = faker_for_key(conflict_value)
         if reuse_via == "INSERT":
@@ -1129,7 +1133,9 @@ def force_conflict_operation(
             )
             target_row = cursor.fetchone()
             if not target_row:
-                conn.commit()
+                # No reuse target: roll back the free too, so an incomplete
+                # round doesn't leave a stray free-only mutation committed.
+                conn.rollback()
                 return None
             set_clause, params = build_update_values(
                 table_schemas,
--- a/migtests/scripts/event-generator/utils.py
+++ b/migtests/scripts/event-generator/utils.py
@@ -180,8 +180,10 @@ def faker_for_key(key: Any) -> Faker:
     regenerate the same fake payload associated with a given unique-key
     value, so induced conflicts are reproducible across runs/logs.
     """
+    import zlib
     f = Faker()
-    f.seed_instance(hash(str(key)))
+    seed = zlib.crc32(str(key).encode("utf-8"))
+    f.seed_instance(seed)
     return f
 
 
--- a/migtests/scripts/event-generator/utils.py
+++ b/migtests/scripts/event-generator/utils.py
@@ -487,13 +487,13 @@ def _find_unique_columns(
         SELECT a.attname
         FROM pg_index i
         JOIN pg_class c ON c.oid = i.indrelid
-        JOIN pg_attribute a ON a.attrelid = c.oid AND a.attnum = i.indkey[0]
+        JOIN pg_attribute a ON a.attrelid = c.oid AND a.attnum = (i.indkey::smallint[])[1]
         WHERE c.oid = %s::regclass
           AND i.indisunique
           AND NOT i.indisprimary
           AND i.indpred IS NULL
-          AND array_length(i.indkey, 1) = 1
-          AND i.indkey[0] > 0
+          AND array_length(i.indkey::smallint[], 1) = 1
+          AND (i.indkey::smallint[])[1] > 0
     """
     cursor.execute(query, (regclass,))
     return [r[0] for r in cursor.fetchall()]
--- a/migtests/scripts/event-generator/utils.py
+++ b/migtests/scripts/event-generator/utils.py
@@ -474,31 +474,38 @@ def _find_primary_key(
     return None
 
 
-def _find_unique_columns(
+def _find_unique_indexes(
     cursor: Any,
     table_name: str,
     schema_name: Optional[str],
-) -> List[str]:
-    """Return columns covered by a plain single-column UNIQUE index (excluding the PK).
+) -> List[Dict[str, Any]]:
+    """Return UNIQUE indexes on plain columns (excluding the PK), one dict per index:
+    {"columns": [col, ...], "predicate": <partial-index WHERE text, or None>}.
 
-    Restricted to non-partial, non-expression, single-column unique indexes so
-    that forcing a value collision on one of them has an unambiguous meaning.
+    Supports single- and multi-column (composite) unique indexes, and partial
+    unique indexes (predicate is returned so callers can scope row selection
+    to it - a value freed/reused outside the predicate's scope wouldn't
+    actually violate the constraint). Expression indexes are excluded: there's
+    no general way to pick an underlying-column value that reproduces a given
+    expression result on a different row, so a collision can't be forced on
+    one reliably.
     """
     regclass = _qualify_regclass(table_name, schema_name)
     query = """
-        SELECT a.attname
+        SELECT array_agg(a.attname ORDER BY k.ord) AS columns,
+               pg_get_expr(i.indpred, i.indrelid) AS predicate
         FROM pg_index i
         JOIN pg_class c ON c.oid = i.indrelid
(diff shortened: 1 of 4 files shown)


migtests/scripts/event-generator/utils.py
```diff
@@ -972,6 +1048,114 @@ def execute_with_retry(
...
+                1,
+                column_overrides=column_overrides,
+                forced_values={unique_col: conflict_value},
+                faker_instance=fake,
+            )
+            cursor.execute(f"INSERT INTO {table_name} ({columns}) VALUES {values_list}")
+        else:
+            cursor.execute(
+                f"SELECT {pk_select} FROM {table_name} "
+                f"WHERE {unique_col} IS NULL OR {unique_col} != %s ORDER BY random() LIMIT 1",
```
makalaaneesh: [AI] **`free_via=UPDATE` + `reuse_via=UPDATE` can re-select the just-freed source row, producing no cross-PK conflict.**

When `free_via == "UPDATE"` the source row is not deleted — its unique column is set to a new random value (or `NULL` for enum/bit types via `generate_random_data`). The reuse-UPDATE target is then chosen with `WHERE {unique_col} IS NULL OR {unique_col} != %s ... ORDER BY random() LIMIT 1`, which does **not** exclude the source row's primary key. So the freed source row still qualifies and can be picked as the target.

When that happens, the value is freed and re-applied on the *same* primary key, both CDC events key to the same apply channel, and no cross-PK race is generated — silently defeating FORCE_CONFLICT for that round. Probability is ~1/N, which is non-negligible on the small tables typical in migtests. (`free_via=DELETE` is unaffected, since the source row is gone.)

Suggested fix: exclude the source PK from the target selection, e.g. append `AND NOT ({pk_where})` and pass `source_pk_values` into the query params (or, for the common single-column PK, `AND {pk_col} != %s`).

---

---

migtests/scripts/event-generator/utils.py
```diff
@@ -171,6 +172,21 @@ def set_faker_seed(seed: int) -> None:
     _fake.seed_instance(seed)
 
 
+def faker_for_key(key: Any) -> Faker:
```
makalaaneesh, line 175: [AI] **`faker_for_key` does not actually make the reused row reproducible — `generate_random_data` still draws from the module-global `random` (and `_fake` for uuid).**

The README and the docstrings here state the reused row is "regenerated deterministically from the value ... so it's reproducible across runs/logs." But passing `faker_instance` into `generate_random_data` only redirects its Faker-based branches. Integer, smallint, boolean, numeric, money, array-element, and `date`-via-`random` values are drawn from the module-level `random` module, whose state depends on how many prior operations ran — not on the per-key seed. Worse, the `uuid` branch calls `_fake.uuid4()` (the module-global Faker), ignoring the passed `faker_instance` entirely — and uuid is a very common unique-key/payload type.

As a result the non-forced columns of the reused row differ run-to-run, so the documented reproducibility guarantee does not hold. This will cost debugging time for anyone who relies on it to correlate logs.

Suggested fix: either (a) route every random draw in `generate_random_data` through the passed `faker_instance` (e.g. `_fake.uuid4()` → `fake.uuid4()`, and replace bare `random.*` calls with a per-key-seeded RNG), or (b) soften the README/docstring claim to state that only the Faker-generated columns are reproducible from the key.

---

---

makalaaneesh (commented):
[AI] Automated review — 2 findings (0 critical, 2 warning). See inline comments. Scope is limited to the migtests event-generator FORCE_CONFLICT mode; no product or hot-path code is touched. Both findings concern the test helper's own correctness: a self-targeting case that can silently skip the intended cross-PK conflict, and a reproducibility claim the implementation only partially delivers.

---

---

migtests/scripts/event-generator/utils.py
yaronp68: fixed

migtests/scripts/event-generator/utils.py
yaronp68: Fixed via option (a): every value generate_random_data produces now routes through the resolved fake instance — either the module-level _fake when no faker_instance is passed, or a per-key seeded one from faker_for_key. Verified live that same-key calls now match exactly across the board (int, uuid, inet, money, numeric, text) and differ across different keys, so the "reproducible from a seed" claim in the docstrings now actually holds.

migtests/scripts/event-generator/utils.py
```diff
@@ -972,6 +1049,115 @@ def execute_with_retry(
...
+        else:
+            cursor.execute(
+                f"SELECT {pk_select} FROM {table_name} "
+                f"WHERE ({unique_col} IS NULL OR {unique_col} != %s) AND NOT ({pk_where}) "
+                f"ORDER BY random() LIMIT 1",
+                [conflict_value] + source_pk_values,
+            )
+            target_row = cursor.fetchone()
+            if not target_row:
+                conn.commit()
```
priyanshi-yb: To deterministically produce the conflict, it is required to perform these two statements in a single transaction so that they can be exported at a similar time, and then only import will try to import them in parallel. 
I would suggest we commit all these statements in a single transaction

migtests/scripts/event-generator/utils.py
```diff
@@ -458,6 +474,33 @@ def _find_primary_key(
...
+        SELECT a.attname
+        FROM pg_index i
+        JOIN pg_class c ON c.oid = i.indrelid
+        JOIN pg_attribute a ON a.attrelid = c.oid AND a.attnum = (i.indkey::smallint[])[1]
+        WHERE c.oid = %s::regclass
+          AND i.indisunique
+          AND NOT i.indisprimary
+          AND i.indpred IS NULL
+          AND array_length(i.indkey::smallint[], 1) = 1
+          AND (i.indkey::smallint[])[1] > 0
```
priyanshi-yb: `i.indpred IS NULL` -> why are we skipping a partial index? We can generate conflicts on that also 
And the other two filters also, not able to understand why

migtests/scripts/event-generator/utils.py
yaronp68: Good catch, thanks. Fixed in d93bc8c8 - removed the intermediate `conn.commit()` after the free statement so free+reuse now land in one transaction, with a single commit at the end. Also switched the "no reuse target found" path from committing to rolling back, so an incomplete round doesn't leave a stray free-only mutation applied.

---
_automated · Claude Code (Sonnet 5)_

migtests/scripts/event-generator/utils.py
yaronp68: Fair point on the first two - fixed in 09013425:

- **Composite (multi-column) unique indexes**: now supported. `_find_unique_indexes` returns each index's full column list, and `force_conflict_operation` threads a list through (source/target selection, free/reuse writes, `forced_values`) instead of a single column.
- **Partial unique indexes**: also now supported. The predicate is captured via `pg_get_expr(i.indpred, i.indrelid)` and used to scope every row-selection query - a value freed/reused outside the predicate's scope wouldn't actually violate the constraint, so we only ever pick rows within it.
- **Expression indexes** (`(i.indkey::smallint[])[1] > 0`): left excluded, on purpose. Unlike the other two, there's no general way to force a collision here - to reuse a value on a different row we'd need to pick underlying-column inputs that reproduce a given expression's *output*, which isn't solvable for an arbitrary expression (only for specific invertible ones like `lower(x)`). Since the schema is arbitrary, I don't think this is safely automatable, so I left it out rather than build something that only works for a subset of expressions.

Verified all three combinations (single-column, composite, partial) against a live Postgres instance across DELETE/UPDATE-free and INSERT/UPDATE-reuse.

---
_automated · Claude Code (Sonnet 5)_

priyanshi-yb (approved):
LGTM, thanks for this idea!

PR #3676, opened 2026-07-16
Add native Prometheus metrics for import and export

JoeSouthan:

Adds a native Prometheus metrics surface, replacing the hidden, import-snapshot-only
Prometheus server with a first-class, documented one covering import, export, CDC,
errors, and throughput.

- New `src/metrics/` package: a `Recorder` interface with a no-op default (zero
  allocation, active when metrics are disabled) and a `PrometheusRecorder`
  implementation backed by its own `prometheus.Registry` (not the global default
  registry), plus a dedicated `/metrics` HTTP server on its own `http.ServeMux`.
- A new `--metrics-port` flag, decoupled from `--profile` (which now only starts
  pprof), available on both import and export commands.
- Preserves the 5 existing snapshot counter names and label set exactly
  (`yb_voyager_import_snapshot_{rows,bytes}_total`,
  `yb_voyager_import_snapshot_batch_{created,submitted,ingested}_total`).
- Adds new metrics: batch-size histograms, a last-batch-ingested timestamp gauge
  (stall detection), import error counters (by `error_kind`), CDC events
  imported/exported + import rate, exported snapshot row counts, and
  parallelism/parallel-connections/node-CPU gauges for adaptive parallelism.
- `--prometheus-metrics-port` and its old per-role default ports (9101-9104) are
  kept as a deprecated fallback (used only when `--metrics-port` is unset and
  `--profile` is set), with a deprecation warning.

**Live-migration cutover observability expansion** (this update):
- CDC catch-up/cutover-gate gauges: `yb_voyager_cdc_events_pending`,
  `yb_voyager_cdc_estimated_seconds_to_catch_up`,
  `yb_voyager_cdc_last_event_applied_timestamp_seconds`, wired into the
  streaming-import stats reporter.
- Per-table import progress and lifecycle gauges:
  `yb_voyager_import_snapshot_table_total_rows`,
  `yb_voyager_import_table_start_timestamp_seconds`,
  `yb_voyager_import_table_completed_timestamp_seconds`.
- `yb_voyager_export_errors_total`, a counter of export errors by operation.
- `yb_voyager_build_info`, a gauge always set to 1, labelled with version and
  commit, for correlating scrape data with the running binary.
- `yb_voyager_source_replication_slot_retained_wal_bytes`, polled every 30s
  during PostgreSQL streaming export, so operators can be alerted before a
  stalled slot causes the source to run out of disk.
- Added an `exporter_role` label to the existing
  `yb_voyager_export_cdc_events_total` counter (this metric has not yet
  shipped in a released build, so this is not a compatibility break).
- Test-quality cleanups from review feedback: `TestResolveMetricsPort` and
  the `src/metrics/` package tests are now table-driven / grouped into single
  top-level tests with sub-tests, rather than one assertion per top-level
  test function.

**Earlier metrics-server startup + export progress/health metrics** (this update):
- `--metrics-port` now starts serving as soon as `migration_uuid` is resolved,
  rather than after later blocking steps: import starts it before the
  "waiting for snapshot data export to complete" poll loop; export starts it
  before the target-version compatibility check and cutover checks. No new
  metric series, just earlier availability of the existing ones.
- `yb_voyager_export_snapshot_table_total_rows`: expected total rows per table
  during export, mirroring the existing import-side
  `yb_voyager_import_snapshot_table_total_rows`. Enables percent-complete and
  ETA calculation on the export side (previously only available for import).
- `yb_voyager_import_snapshot_batches_in_flight`: a gauge of batches submitted
  to the worker pool but not yet ingested (submitted minus ingested), per
  table. A stall detector independent of the existing last-ingest-timestamp
  gauge — a healthy pipeline keeps this bounded, a stuck one lets it grow.
- `yb_voyager_import_pool_pending_close_connections`: connections the target
  pool has been asked to close but hasn't yet, sampled from the existing
  adaptive-parallelism 10s polling loop. Explains transient throughput dips
  during parallelism scale-down that would otherwise look like a stall.
- `yb_voyager_cdc_debezium_up`: liveness (1/0) of the Debezium CDC process,
  set from the existing export-side Debezium poll loop. Export-only — the
  import process has no direct handle on Debezium, which always runs in the
  exporter process.
- `yb_voyager_export_parallelism`: the configured `--parallel-jobs` value,
  static for the process lifetime, for parity with the import-side
  parallelism gauge.

- New `--metrics-port <port>` flag on import and export commands (default `0`,
  disabled). Exposes `GET http://<host>:<port>/metrics` in Prometheus text format.
- `--profile` no longer implicitly starts a metrics server; it only starts pprof
  now. `--prometheus-metrics-port` still works as a deprecated alias and logs a
  warning recommending `--metrics-port`.
- `metrics-port` is a config-file key on the relevant import/export command
  sections, following the same config-binding convention as other flags.
- No changes to reports or the installation process. No new flags added by the
  cutover-observability update — it only adds metric series to the existing
  `/metrics` endpoint.

- New unit tests (`//go:build unit`) for the `src/metrics/` package: the no-op
  recorder never panics and is the default, `SetRecorder`/`Get` install and
  return the active recorder, all `PrometheusRecorder` metrics register on their
  own registry with the correct names/labels/values (verified with
  `prometheus/client_golang/prometheus/testutil`), and the HTTP server actually
  serves `/metrics` end-to-end on a real listener (port 0, OS-assigned).
- A unit test for the new `--metrics-port`/`--prometheus-metrics-port`/`--profile`
  resolution logic (`resolveMetricsPort`), table-driven.
- Unit tests for the new cutover-observability wiring: the streaming-import
  reporter's last-event-applied hook, per-table total-rows wiring, and the
  export-error counter, each verified against a `//go:build unit` recording
  test-double (`metrics.RecordingRecorder`) so call sites can be tested without
  a live Prometheus registry.
- Full existing unit suite (`go test -tags unit ./...`) passes unchanged
  (`go build ./...`, `go vet ./...`, and `staticcheck ./...` clean on all
  touched packages).
- New unit tests for the batches-in-flight lifecycle (submit increments,
  ingest decrements), pending-close-connections and export-parallelism
  gauges, and Debezium liveness (up/down transition).
- Manually verified every new metric name/label/value renders correctly on
  `/metrics` via a standalone harness against the Prometheus recorder.
- Not covered: end-to-end validation against a live migration with a real
  Prometheus scraper attached to the new endpoint; happy to add
  `integration_voyager_command`-tagged coverage if maintainers want it before
  merge.

No. Callhome, control-plane, and YugabyteD payload structs are untouched.

No. `MigrationStatusRecord`/metaDB, the assessment DB schema, and all other
on-disk/serialized state are untouched. This PR only adds an in-memory metrics
package and a new opt-in HTTP endpoint.

---

| Component        


yb-voyager/cmd/importData_observability_test.go
```diff
@@ -0,0 +1,26 @@
...
+	if p, on := resolveMetricsPort("target_db_importer", 0, true); !on || p != 9101 {
+		t.Fatalf("legacy --profile fallback must use role default port; got (%d,%t)", p, on)
+	}
+	if p, on := resolveMetricsPort("import_file", 0, true); !on || p != 9102 {
+		t.Fatalf("legacy --profile fallback must use role default port; got (%d,%t)", p, on)
+	}
+	if p, on := resolveMetricsPort("source_db_exporter", 0, true); on || p != 0 {
+		t.Fatalf("legacy --profile fallback has no default for export roles; got (%d,%t)", p, on)
+	}
+}
```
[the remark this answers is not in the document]
JoeSouthan: Addressed — `TestResolveMetricsPort` is now table-driven with the same cases (disabled, explicit port enables, explicit port wins over legacy, legacy fallback per role, no legacy fallback for export roles).

yb-voyager/src/metrics/prometheus_test.go
```diff
@@ -0,0 +1,119 @@
...
+	if got := testutil.ToFloat64(r.parallelism.WithLabelValues("uuid-1", "sess-1", "target_db_importer")); got != 8 {
+		t.Fatalf("parallelism = %v", got)
+	}
+	if got := testutil.ToFloat64(r.parallelConns.WithLabelValues("uuid-1", "sess-1", "target_db_importer")); got != 4 {
+		t.Fatalf("parallel_connections = %v", got)
+	}
+	if got := testutil.ToFloat64(r.nodeCPU.WithLabelValues("uuid-1", "sess-1", "node-1")); got != 55.5 {
+		t.Fatalf("node cpu = %v", got)
+	}
+}
```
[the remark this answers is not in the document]
JoeSouthan: Addressed — all `PrometheusRecorder` tests are now grouped as sub-tests under a single `TestPrometheusRecorder`.

yb-voyager/src/metrics/recorder_test.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+}
+
+// countingRecorder is a test double embedding noopRecorder and counting one method.
+type countingRecorder struct {
+	noopRecorder
+	setParallelismCalls int
+}
+
+func (c *countingRecorder) SetParallelism(role string, level int) { c.setParallelismCalls++ }
```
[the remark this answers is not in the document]
JoeSouthan: Addressed — the recorder tests are now grouped as sub-tests under a single `TestRecorder` ("Get returns no-op by default", "No-op recorder does not panic", "SetRecorder installs active recorder").

yb-voyager/src/errs/export.go
```diff
@@ -125,6 +127,7 @@ func NewExportDataError(flow string, failedStep string, err error) *ExportDataEr
 
 // NewExportDataErrorWithCompletedFlows creates a new error with flow context and completed flows
 func NewExportDataErrorWithCompletedCalls(currentFlow string, callExecutionHistory *stack.Stack, failedStep string, err error) *ExportDataError {
+	metrics.Get().RecordExportError(currentFlow)
```
makalaaneesh: [AI] **`yb_voyager_export_errors_total` will over-count; recording lives in the wrong layer.**

`RecordExportError` is invoked from inside the error *constructors* (`NewExportDataError` at line 119 and `NewExportDataErrorWithCompletedCalls` here at line 130). But `propagateIfExportDataError` re-wraps an already-constructed `*errs.ExportDataError` by calling `NewExportDataErrorWithCompletedCalls` again at every propagation hop:

```go
func propagateIfExportDataError(err error, currentFlow string) error {
    var exportDataErr *errs.ExportDataError
    if errors.As(err, &exportDataErr) {
        exportDataErr.AddCall(exportDataErr.CurrentFlow())
        return errs.NewExportDataErrorWithCompletedCalls(currentFlow, ...) // records AGAIN
    }
    ...
}
```

So a single underlying export failure is counted once at creation (`NewExportDataError`, flow = X) and then once more per propagation level (e.g. `GET_INITIAL_TABLE_LIST_OPERATION` at `exportData.go:1542/1554/1579/1584`, `APPLY_TABLE_LIST_FLAGS_ON_SUBSEQUENT_RUN` at 1608, etc.). The counter is inflated and the `operation` label is smeared across every intermediate flow rather than attributing the one real failure.

This also conflicts with `yb-voyager/src/errs/.cursor/BUGBOT.md`: *"This package should remain lightweight — it defines error types, not error-handling logic."* Emitting a metric as a side effect of `new`-ing an error object is behavior/logic, and it adds an `errs -> metrics` dependency to a package that is meant to only define types.

Suggested fix: remove `metrics.Get().RecordExportError(...)` from both constructors and record the metric once at the point where an export error is actually surfaced/handled (e.g. in the `export data` command flow where the terminal error is reported), so each logical failure increments the counter exactly once with the correct `operation` label.

---

---

makalaaneesh (commented):
[AI] Automated review — 1 finding (0 critical, 1 warning). See the inline comment.

Overall the metrics package is cleanly designed: the `Recorder` interface with an atomic no-op default keeps the disabled/hot-path case allocation-free, each `PrometheusRecorder` owns its own registry (safe for tests), and the deprecated `--prometheus-metrics-port` / `--profile` legacy-port fallback is resolved by a well-tested pure function (`resolveMetricsPort`). The one concern worth fixing before merge is metric over-counting caused by recording an export-error metric inside the error constructors, which are also used on the propagation path. Everything else (per-batch/per-poll recording sites, PG-guarded type assertion for the WAL monitor, signature propagation of the new `importerRole` arg) checked out and the touched packages build clean.

---

---

yb-voyager/src/errs/export.go
JoeSouthan: Fixed — moved `RecordExportError` out of the `errs` constructors (which was double/multi-counting on every propagation hop) and into the single terminal handling site in `exportData()` where the error is finally surfaced (`handleGetInitialTableListError`), keyed off the failed flow. This also drops the `errs -> metrics` dependency, keeping the `errs` package free of side effects per the package's review rules. Added a unit test verifying the counter increments exactly once even when the error is propagated through multiple flows.

yb-voyager/src/metrics/recorder.go
```diff
@@ -0,0 +1,91 @@
...
+type ErrorKind string
+
+const (
+	ErrorKindRowProcessing  ErrorKind = "row_processing"
+	ErrorKindBatchIngestion ErrorKind = "batch_ingestion"
+)
+
+// Recorder is the only type metrics call sites depend on. A no-op implementation
+// is the default so call sites never need to check whether metrics are enabled.
+type Recorder interface {
```
makalaaneesh, line 19: Can we organize these by these high-level groups so it's easy to follow? Also names of the methods can indicate the category (like RecordImportSnapshot* )

- export snapshot
- export CDC
- import snapshot 
- import CDC
- misc if any

yb-voyager/src/metrics/prometheus.go
```diff
@@ -0,0 +1,380 @@
...
+func SessionID() string { return sessionID }
+
+// PrometheusRecorder implements Recorder backed by a dedicated registry, so it
+// holds no global state and is safe to construct in tests.
+type PrometheusRecorder struct {
+	reg           *prometheus.Registry
+	migrationUUID string
+	sessionID     string
+
+	importRowsTotal         *prometheus.CounterVec
```
makalaaneesh: same comment as in the recorder: 
Can we organize these by these high-level groups so it's easy to follow:

export snapshot
export CDC
import snapshot
import CDC
misc if any

yb-voyager/src/metrics/prometheus.go
```diff
@@ -0,0 +1,380 @@
...
+	exportParallelism   *prometheus.GaugeVec
+	snapshotTablesTotal *prometheus.GaugeVec
+
+	debeziumUp *prometheus.GaugeVec
+}
+
+func NewPrometheusRecorder(migrationUUID, sessionID string) *PrometheusRecorder {
+	reg := prometheus.NewRegistry()
+	f := promauto.With(reg)
+	rec := &PrometheusRecorder{
```
makalaaneesh, line 84: can we follow the naming pattern for metrics for easy identification: 
export_data_snapshot_*
export_data_cdc_*
import_data_snapshot_*
import_data_cdc_*

yb-voyager/src/metrics/prometheus.go
```diff
@@ -0,0 +1,380 @@
...
+var errorLabels = append(append([]string{}, snapshotLabels...), "error_kind")
+var cdcRateLabels = []string{"migration_uuid", "session_id", "importer_role"}
+var cdcEventLabels = append(append([]string{}, cdcRateLabels...), "event_type")
+var exportRowLabels = []string{"migration_uuid", "session_id", "table_name", "schema_name"}
+var exportCDCLabels = []string{"migration_uuid", "session_id", "exporter_role"}
+var nodeCPULabels = []string{"migration_uuid", "session_id", "node"}
+var exportErrorLabels = []string{"migration_uuid", "session_id", "operation"}
+var replicationSlotLabels = []string{"migration_uuid", "session_id", "slot_name"}
+var buildInfoLabels = []string{"migration_uuid", "session_id", "version", "commit"}
+var roleLabels = []string{"migration_uuid", "session_id", "role"}
```
makalaaneesh: some metrics use "role", and some use the more specific "importer_role/exporter_role". Can we make this consistent?

yb-voyager/src/metrics/recorder.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// throughput / parallelism gauges
+	SetParallelism(role string, level int)
+	SetParallelConnections(role string, n int)
+	SetPendingConnsToClose(role string, n int)
+	SetNodeCPUPercent(node string, pct float64)
+	SetExportParallelism(role string, level int)
+
+	// snapshot scope
+	SetSnapshotTablesTotal(role string, count int)
```
makalaaneesh: We seem to be using "role" to indicate export/import, which i think can be confusing because we don't use that elsewhere in the codebase. 

Can we have one metric each for Import/Export?

yb-voyager/src/metrics/prometheus.go
```diff
@@ -0,0 +1,380 @@
...
+		// PromQL: sum by (event_type) (rate(yb_voyager_cdc_events_imported_total[5m]))
+		cdcEventsImported: f.NewCounterVec(prometheus.CounterOpts{
+			Name: "yb_voyager_cdc_events_imported_total",
+			Help: "Total CDC events imported during streaming, by event_type",
+		}, cdcEventLabels),
+		cdcImportRate: f.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "yb_voyager_cdc_import_rate_events_per_second",
+			Help: "CDC events imported per second (3-minute average)",
+		}, cdcRateLabels),
+		exportSnapshotRows: f.NewGaugeVec(prometheus.GaugeOpts{
```
makalaaneesh: this sounds like it should be a counter, not a gauge?

yb-voyager/src/metrics/recorder.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// process liveness
+	SetDebeziumUp(role string, up bool)
```
makalaaneesh: there are a lot of unaccounted paths in the code where we detect an error in debezium running, wherein we'd actually have to set debeziumup=false.

I'd suggest we skip this metric for now; and then in a later PR, we can expose all of debezium's prometheus metrics directly

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -494,11 +497,16 @@ func reportStreamingProgress(ctx context.Context) {
 
 func calculateStreamingProgress(ctx context.Context) {
 	var err error
+	var lastRecordedEventCount int64
 	for {
 		totalEventCount, totalEventCountRun, err = metaDB.GetTotalExportedEventsByExporterRole(exporterRole, runId)
 		if err != nil {
 			utils.ErrExit("failed to get total exported count from metadb: %w", err)
 		}
+		if delta := totalEventCountRun - lastRecordedEventCount; delta > 0 {
```
makalaaneesh, line 503: this function is not always invoked (for instance, if progress bar is disabled, this won't run, and we won't get metrics. 

Need to find a better place; will check and get back.

yb-voyager/src/metrics/recorder.go
```diff
@@ -0,0 +1,91 @@
...
+	SetCDCEstimatedSecondsToCatchUp(role string, seconds float64)
+	SetCDCLastEventApplied(role string)
+
+	// export
+	SetExportedSnapshotRowCount(t sqlname.NameTuple, rows int64)
+	SetExportSnapshotTableTotalRows(t sqlname.NameTuple, rows int64)
+	SetExportTableStarted(t sqlname.NameTuple)
+	SetExportTableCompleted(t sqlname.NameTuple)
+	RecordExportedCDCEvents(role string, events int64)
+	RecordExportError(operation string)
```
makalaaneesh: I think we can skip this metric. there is only error that is recorded; it's not really useful IMO.

makalaaneesh (commented):
Hi @JoeSouthan thanks for your contribution! 
Sharing some comments (only reviewed the export-side metrics; yet to go through the import-side metrics)

yb-voyager/src/metrics/recorder.go
makalaaneesh: For import, this actually takes into consideration `pendingTasks` not the total no. of tables. On resume, it will report the tables-remaining, not total tables.

yb-voyager/src/metrics/prometheus.go
```diff
@@ -0,0 +1,380 @@
...
+		}, cdcRateLabels),
+		cdcEstimatedSecondsToCatchUp: f.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "yb_voyager_cdc_estimated_seconds_to_catch_up",
+			Help: "Estimated seconds for the importer to apply all pending CDC events at the current rate",
+		}, cdcRateLabels),
+		cdcLastEventApplied: f.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "yb_voyager_cdc_last_event_applied_timestamp_seconds",
+			Help: "Unix timestamp of the most recent successfully applied CDC event batch",
+		}, cdcRateLabels),
+		importTableTotalRows: f.NewGaugeVec(prometheus.GaugeOpts{
```
makalaaneesh: this sounds like it should be a counter?

makalaaneesh: Also, this sounds too similar to `yb_voyager_import_snapshot_rows_total`. we should find a better way to distinguish between the two.

yb-voyager/src/metrics/prometheus.go
```diff
@@ -0,0 +1,380 @@
...
+			Help: "Total number of batches created for import",
+		}, snapshotLabels),
+		snapshotBatchSubmitted: f.NewCounterVec(prometheus.CounterOpts{
+			Name: "yb_voyager_import_snapshot_batch_submitted_total",
+			Help: "Total number of batches submitted to worker pool",
+		}, snapshotLabels),
+		snapshotBatchIngested: f.NewCounterVec(prometheus.CounterOpts{
+			Name: "yb_voyager_import_snapshot_batch_ingested_total",
+			Help: "Total number of batches successfully ingested",
+		}, snapshotLabels),
```
makalaaneesh, line 109: I want to highlight that these metrics (which were created previously) are counters that do not get reset to 0 on resume. This can be misleading. 

This is worth discussing.

yb-voyager/src/metrics/prometheus.go
```diff
@@ -0,0 +1,380 @@
...
+		// PromQL: yb_voyager_import_snapshot_batches_in_flight > 0 for an extended period indicates a stall.
+		snapshotBatchesInFlight: f.NewGaugeVec(prometheus.GaugeOpts{
```
makalaaneesh: this seems redundant when the submitted, ingested metrics are already presented.

yb-voyager/cmd/importData.go
```diff
@@ -2356,9 +2402,12 @@ func createSnapshotImportCompletedEvent() cp.SnapshotImportCompletedEvent {
 
 func createInitialImportDataTableMetrics(tasks []*ImportFileTask) []*cp.UpdateImportedRowCountEvent {
 	result := []*cp.UpdateImportedRowCountEvent{}
+	metrics.Get().SetSnapshotTablesTotal(importerRole, len(tasks))
```
makalaaneesh: One thing to note here is: This will only be set for pending tasks (i.e. tables which are yet to be imported). On import-data resumes, this can be lower than the total no. of. tables.

yb-voyager/cmd/importData.go
```diff
@@ -2356,9 +2402,12 @@ func createSnapshotImportCompletedEvent() cp.SnapshotImportCompletedEvent {
...
 	for _, task := range tasks {
 		var schemaName, tableName string
 		schemaName, tableName = task.TableNameTup.ForKeyTableSchema()
+		metrics.Get().SetImportSnapshotTableTotalRows(importerRole, task.TableNameTup, getTotalProgressAmount(task))
```
makalaaneesh: same as comment above: this will only be set for pending tasks

yb-voyager/cmd/importDataFileTaskImporter.go
```diff
@@ -111,6 +112,8 @@ func NewFileTaskImporter(task *ImportFileTask, state *ImportDataState, batchProd
 	currentProgressAmount := getImportedProgressAmount(task, state)
 	progressReporter.AddProgressAmount(task, currentProgressAmount)
 
+	metrics.Get().SetImportTableStarted(importerRole, task.TableNameTup)
```
makalaaneesh: Note: On resume, this will be set again.

yb-voyager/src/adaptiveparallelism/adaptive_parallelism.go
```diff
@@ -120,6 +123,25 @@ func fetchClusterMetricsAndUpdateParallelism(yb TargetYugabyteDBWithConnectionPo
 	if callhomeMetricsCollector != nil {
 		callhomeMetricsCollector.SetCurrentParallelConnections(currentNumConnections)
 	}
+	metrics.Get().SetParallelConnections(constants.TARGET_DB_IMPORTER_ROLE, currentNumConnections)
+	metrics.Get().SetParallelism(constants.TARGET_DB_IMPORTER_ROLE, currentNumConnections)
```
makalaaneesh: these are duplicate. Let's pick one?

yb-voyager/src/metrics/recorder.go
```diff
@@ -0,0 +1,91 @@
...
+	InitImportSnapshotTable(role string, t sqlname.NameTuple)
+	SetImportTableStarted(role string, t sqlname.NameTuple)
+	SetImportTableCompleted(role string, t sqlname.NameTuple)
+
+	// source health
+	SetSourceReplicationSlotRetainedWALBytes(slotName string, bytes int64)
+
+	// throughput / parallelism gauges
+	SetParallelism(role string, level int)
+	SetParallelConnections(role string, n int)
```
makalaaneesh: These two seem to only be called in case of adaptive parallelism. It's possible for the user to turn off adaptive parallelism and specify a fixed `--parallel-jobs`. We should consider that as well.

yb-voyager/src/metrics/recorder.go
```diff
@@ -0,0 +1,91 @@
...
+	SetPendingConnsToClose(role string, n int)
```
makalaaneesh: Do we really need this? Are you folks using this? Seems like an internal detail that might not be too relevant for users.

yb-voyager/src/adaptiveparallelism/adaptive_parallelism.go
```diff
@@ -120,6 +123,25 @@ func fetchClusterMetricsAndUpdateParallelism(yb TargetYugabyteDBWithConnectionPo
...
+			log.Warnf("adaptive: parsing cpu usage user as float for node %s: %v", node, err)
+			continue
+		}
+		sys, err := strconv.ParseFloat(nodeMetrics.Metrics[CPU_USAGE_SYSTEM_METRIC], 64)
+		if err != nil {
+			log.Warnf("adaptive: parsing cpu usage system as float for node %s: %v", node, err)
+			continue
+		}
+		metrics.Get().SetNodeCPUPercent(node, (user+sys)*100)
+	}
```
makalaaneesh: nit: separate function for all metrics related logic

yb-voyager/src/metrics/recorder.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// batch-size distribution
+	ObserveSnapshotBatchSize(role string, t sqlname.NameTuple, rows, bytes int64)
+
+	// import errors
+	RecordImportError(role string, t sqlname.NameTuple, kind ErrorKind, rows, bytes int64)
+
+	// cdc / streaming
+	RecordCDCEventsImported(role string, inserts, updates, deletes int64)
+	SetCDCImportRate(role string, eventsPerSec float64)
```
makalaaneesh: I think we can skip this "rate" metric. These are usually calculated on prometheus via rate(counter)

yb-voyager/src/metrics/recorder.go
```diff
@@ -0,0 +1,91 @@
...
+	RecordSnapshotBatchIngested(role string, t sqlname.NameTuple, rows, bytes int64)
+
+	// batch-size distribution
+	ObserveSnapshotBatchSize(role string, t sqlname.NameTuple, rows, bytes int64)
+
+	// import errors
+	RecordImportError(role string, t sqlname.NameTuple, kind ErrorKind, rows, bytes int64)
+
+	// cdc / streaming
+	RecordCDCEventsImported(role string, inserts, updates, deletes int64)
```
makalaaneesh: Note: on resume, this will reset to 0. We can potentially re-initialize it on resume to the non-zero value 

(We can consider doing this in a subsequent PR)

yb-voyager/src/reporter/stats/streamImportReporter.go
```diff
@@ -182,4 +186,6 @@ func (s *StreamImportStatsReporter) UpdateRemainingEvents() {
 	if lastMinIngestionRate > 0 {
 		s.estimatedTimeToCatchUp = time.Duration(s.remainingEvents/lastMinIngestionRate) * time.Minute
 	}
+	metrics.Get().SetCDCEventsPending(s.importerRole, s.remainingEvents)
+	metrics.Get().SetCDCEstimatedSecondsToCatchUp(s.importerRole, s.estimatedTimeToCatchUp.Seconds())
```
makalaaneesh: This function seems to be gated on `!disablePb` (i.e. this will not run if progress-bar is disabled)

makalaaneesh (commented):
Some more comments on import-side of metrics

makalaaneesh:
At a high-level, structure looks good! 
A few discussion points: 
- What happens on resume? Are counters reset to 0, or re-initialized to the latest value? Can metrics be aggregated across runs?
- What happens on start-clean? How will prometheus users be able to distinguish between a start-clean run vs a non-start-clean run? 
- Should we have a config to enable/disable prometheus server? 
- Server lifetime is limited to the process lifetime. If voyager sets some metrics and exits within the scrape interval, those metrics will be list. ( I think this is ok for now ) 
- Consistency in names of metrics/methods/labels. 
- node_exporter text collector approach

Not everything needs to be addressed in this PR; we can decide how to take it up incrementally.

yb-voyager/src/metrics/recorder.go
JoeSouthan: Done. The Recorder interface is now grouped into // import snapshot, // import CDC, // export snapshot, // export CDC, and // misc, and methods are named by category (RecordImportSnapshot*, RecordExportSnapshot*, RecordImportCDC*, RecordExportCDC*).

yb-voyager/src/metrics/prometheus.go
JoeSouthan: Done. PrometheusRecorder fields and the constructor are grouped by the same import/export x snapshot/CDC sections.

yb-voyager/src/metrics/prometheus.go
JoeSouthan: Done. All metrics now follow yb_voyager_<import|export>_data_<snapshot|cdc>_* (for example yb_voyager_import_data_snapshot_rows_total, yb_voyager_export_data_cdc_events_total).

yb-voyager/src/metrics/prometheus.go
JoeSouthan: Done. The generic role label is removed. Import metrics carry importer_role and export metrics carry exporter_role, consistently across every metric including the per-table export snapshot metrics.

yb-voyager/src/metrics/recorder.go
JoeSouthan: Done. Direction is now encoded in the metric name (import_data_* vs export_data_*), so there is one metric per direction and no role-as-direction label. The remaining importer_role/exporter_role labels only distinguish the specific importer/exporter (target_db_importer, source_db_exporter, and so on), which are existing codebase concepts.

yb-voyager/src/metrics/prometheus.go
JoeSouthan: Done. export_data_snapshot_rows_total is now a monotonic counter (delta-tracked from the reporter cumulative counts) rather than a gauge.

yb-voyager/src/metrics/recorder.go
JoeSouthan: Done. Dropped the debezium_up metric. Agreed the liveness paths are not fully covered; we plan to expose Debezium own Prometheus metrics directly in a later PR.

yb-voyager/cmd/exportDataDebezium.go
JoeSouthan: Done. calculateStreamingProgress (which records export CDC events) is now started unconditionally; only the console progress rendering stays gated on !disablePb. Export CDC metrics now record even with the progress bar and diagnostics off.

yb-voyager/src/metrics/recorder.go
JoeSouthan: Done. Dropped export_errors_total entirely, per your point that only a single error kind is ever recorded.

yb-voyager/src/metrics/recorder.go
JoeSouthan: Done. Table-scope metrics are now driven from all tasks rather than only pending tasks, so import_data_snapshot_tables_total and per-table expected_rows stay accurate across resume.

yb-voyager/src/metrics/prometheus.go
JoeSouthan: Renamed to yb_voyager_import_data_snapshot_table_expected_rows to distinguish it from ..._rows_total. Kept it a gauge deliberately: it is the expected total (a target/denominator), not a cumulative count, so counter semantics do not fit.

yb-voyager/src/metrics/prometheus.go
JoeSouthan: Addressed. On resume, the import snapshot row/byte counters are now seeded from the persisted per-table counts in the metaDB (InitImportSnapshotTable), so they reflect true cumulative progress across restarts instead of restarting at 0. resumption.py now asserts this continuity across restarts and start-clean.

yb-voyager/src/metrics/prometheus.go
JoeSouthan: Done. Dropped the in-flight gauge; it is derivable in PromQL as submitted_total - ingested_total.

yb-voyager/cmd/importData.go
JoeSouthan: Done. createInitialImportDataTableMetrics now sets totals and expected-rows over all tasks, so this is no longer limited to pending tasks on resume.

yb-voyager/cmd/importData.go
JoeSouthan: Done. Same fix; driven from all tasks now.

yb-voyager/src/metrics/recorder.go
JoeSouthan: Done. Parallelism and parallel-connections are now also emitted once for a fixed --parallel-jobs run (when adaptive parallelism is disabled), not only from the adaptive poller.

yb-voyager/src/metrics/recorder.go
JoeSouthan: Done. Dropped pending_close_connections; agreed it is an internal detail that is not useful to users.

yb-voyager/src/adaptiveparallelism/adaptive_parallelism.go
JoeSouthan: Done. Metric emission is extracted into a dedicated emitParallelismMetrics helper.

yb-voyager/src/metrics/recorder.go
JoeSouthan: Done. Dropped the rate metric; it is computed in Prometheus via rate(yb_voyager_import_data_cdc_events_total[...]).

yb-voyager/src/reporter/stats/streamImportReporter.go
JoeSouthan: Done. UpdateRemainingEvents (the pending and ETA gauges) now runs on the periodic tick regardless of disablePb; only the console table render is gated. ReportStats takes a displayEnabled flag and keeps emitting metrics when the progress bar is off.

yb-voyager/src/metrics/recorder.go
JoeSouthan: Snapshot row/byte counters are now seeded from the metaDB on resume. The CDC-events counter (import_data_cdc_events_total) still restarts at 0 per process; leaving that re-init as a follow-up PR as you suggested, since it needs the applied-event count persisted per run.

yb-voyager/src/adaptiveparallelism/adaptive_parallelism.go
JoeSouthan: These two are currently set to the same value (currentNumConnections), so they are effectively duplicated today. They are conceptually distinct (adaptive parallelism level vs actual parallel connection count), but since nothing sets them differently yet, I am inclined to keep just import_data_parallelism and drop import_data_parallel_connections. Does that work?

yb-voyager/cmd/importDataFileTaskImporter.go
JoeSouthan: Correct. On resume the table start-timestamp is set again to the resumed run start. That is intentional for now (the gauge reflects when the current run began working the table); the completed-timestamp and the cumulative counters are the resume-safe signals. Happy to guard it to set-once if you would prefer the original start preserved.

yb-voyager/src/errs/export.go
JoeSouthan: For the record: export_errors_total was subsequently dropped entirely (see the recorder.go discussion), so the over-counting concern is fully moot now.

yb-voyager/src/adaptiveparallelism/adaptive_parallelism.go
JoeSouthan: Done. Dropped import_data_parallel_connections and kept import_data_parallelism (emitted from the adaptive poller, and once for a fixed --parallel-jobs run). No more duplicate.

yb-voyager/cmd/exportDataStatus.go
```diff
@@ -212,6 +228,7 @@ func startExportPB(progressContainer *mpb.Progress, mapKey string, quitChan chan
 		log.Infof("Replacing actualRowCount=%d inplace of expectedRowCount=%d for table=%s",
 			actualRowCount, tableMetadata.CountTotalRows, tableMetadata.TableName.ForUserQuery())
 		pbr.SetTotalRowCount(actualRowCount, false)
+		metrics.Get().SetExportSnapshotTableExpectedRows(exporterRole, tableMetadata.TableName, actualRowCount)
```
makalaaneesh: I just realised: This will work for offline migration (where the source database is not receiving traffic), but for live migration, this will not be accurate because this parallel goroutine that is calculating the actual row count might run the count(*) after a consistent snapshot is set by pg_dump; so it's likely that this number will be higher than what actually gets exported by pg_dump in the snapshot. 

Can we skip updating this metric here? 

cc: @priyanshi-yb correct me if i'm wrong here.

yb-voyager/cmd/exportDataStatus.go
```diff
@@ -285,6 +303,11 @@ func startExportPB(progressContainer *mpb.Progress, mapKey string, quitChan chan
 	*/
 	readLines()
 
+	// Land the exported gauge exactly on the table total so the "% complete"
+	// panel reaches 100% (the polling goroutine above can stop a few rows short).
+	metrics.Get().RecordExportSnapshotRowCount(exporterRole, tableMetadata.TableName, tableMetadata.CountTotalRows)
```
makalaaneesh: it will be more accurate to call both `RecordExportSnapshotRowCount` and `SetExportSnapshotTableExpectedRows` here and set them both to `CountLiveRows`. (`CountTotalRows` is just an estimate)

cc: @priyanshi-yb

yb-voyager/cmd/exportDataDebezium.go
makalaaneesh: this might become problematic in that the metric will get updated only every 12 min in this case. 
Since we will always have metrics exposed now, perhaps we can remove this if-else and always just default to 10s. 

@priyanshi-yb any concerns?

yb-voyager/cmd/importData.go
```diff
@@ -1146,8 +1195,13 @@ func importSnapshotData(msr *metadb.MigrationStatusRecord, errorHandler importda
 	if err != nil {
 		utils.ErrExit("Failed to get max parallel connections: %s", err)
 	}
+	if !tconf.AdaptiveParallelismMode.IsEnabled() {
+		// Adaptive parallelism emits this gauge itself once it starts polling;
+		// for a fixed --parallel-jobs run there's no such poller, so emit once here.
+		metrics.Get().SetImportParallelism(importerRole, maxParallelConns)
+	}
+	importDataAllTableMetrics := createInitialImportDataTableMetrics(state, importFileTasks, pendingTasks)
```
makalaaneesh: was it intentional to take this out of the `if importerRole == TARGET_DB_IMPORTER_ROLE {` block? If so, why?

yb-voyager/cmd/importData.go
```diff
@@ -2366,11 +2405,32 @@ func createSnapshotImportCompletedEvent() cp.SnapshotImportCompletedEvent {
...
+// createInitialImportDataTableMetrics sets table-scope metrics (totals, expected
+// rows, seeded counters) from allTasks so they stay accurate on resume, when
+// pendingTasks alone would under-report tables already completed in a prior run.
+// The control-plane event list is still built from pendingTasks only (unchanged
+// behaviour: the control plane only expects updates for tables being worked on
+// in this run).
+func createInitialImportDataTableMetrics(state *ImportDataState, allTasks, pendingTasks []*ImportFileTask) []*cp.UpdateImportedRowCountEvent {
+	metrics.Get().SetImportSnapshotTablesTotal(importerRole, len(allTasks))
+	for _, task := range allTasks {
+		metrics.Get().SetImportSnapshotTableExpectedRows(importerRole, task.TableNameTup, getTotalProgressAmount(task))
```
makalaaneesh: can we use task.RowCount directly here? `getTotalProgressAmount` could also return filesize from the looks of it, but the metric is about rows.

yb-voyager/cmd/importData.go
```diff
@@ -2366,11 +2405,32 @@ func createSnapshotImportCompletedEvent() cp.SnapshotImportCompletedEvent {
...
+		if err != nil { // best-effort; do not abort migration for a metric seed
+			log.Warnf("failed to read imported row count for %s, seeding metric at 0: %v", task.TableNameTup, err)
+			seedRows = 0
+		}
+		seedBytes, err := state.GetImportedByteCount(task.FilePath, task.TableNameTup)
+		if err != nil {
+			log.Warnf("failed to read imported byte count for %s, seeding metric at 0: %v", task.TableNameTup, err)
+			seedBytes = 0
+		}
+		metrics.Get().InitImportSnapshotTable(importerRole, task.TableNameTup, seedRows, seedBytes)
```
makalaaneesh: apologies for the back-and-forth here but can we drop the seeding of this counter? 

Rationale: 

1. Apparently, this is not common-place in prometheus to seed counters; rate() functions are expected to work with counters resetting to 0
2. After a little bit of research, i think the right way to go about it is to have both counters and gauges. counters give you info per run/session (i.e. if you stop/start, then your counters get reset to 0). gauges are absolutes, which will have to be seeded and then keep getting updated. (gauges will help get metrics like the total rows imported/ total cdc events imported/etc across resumes). 

so, my recommendation would be : 
1. for this PR: let's remove seeding. 
2. in a subsequent PR, we can introduce guage metrics which will be seeded. 

let me know if this makes sense, or if you have a better alternative solution.

yb-voyager/cmd/importDataFileTaskImporter.go
```diff
@@ -111,6 +112,8 @@ func NewFileTaskImporter(task *ImportFileTask, state *ImportDataState, batchProd
 	currentProgressAmount := getImportedProgressAmount(task, state)
 	progressReporter.AddProgressAmount(task, currentProgressAmount)
 
+	metrics.Get().SetImportSnapshotTableStarted(importerRole, task.TableNameTup)
```
makalaaneesh: this might be confusing since this gets set on every resume. Can we only set this if it's NOT a resume (we can use `currentProgressAmount > 0` to deduce if it's a resume)

yb-voyager/cmd/importData.go
```diff
@@ -912,27 +917,71 @@ func applyTableListFilter(importFileTasks []*ImportFileTask) []*ImportFileTask {
...
+	}
+	if override == 0 {
+		log.Warnf("starting metrics on default port %d because --profile is set; this fallback is deprecated, use --metrics-port %d explicitly instead", port, port)
+	}
+	rec := metrics.NewPrometheusRecorder(migrationUUID.String(), metrics.SessionID())
+	metrics.SetRecorder(rec)
+	srv := metrics.NewServer(port, rec.Registry())
+	return srv.Start()
+}
+
```
makalaaneesh: can we move this to a separate file (since it's being used by both export/import

yb-voyager/cmd/exportData.go
```diff
@@ -934,9 +935,35 @@ func initPGLiveMigrationAndExportSnapshotIfRequired(ctx context.Context, cancel
 	config.ReplicationSlotName = msr.PGReplicationSlotName
 	config.PublicationName = msr.PGPublicationName
 	config.InitSequenceMaxMapping = sequenceInitValues.String()
+
+	pgDB := source.DB().(*srcdb.PostgreSQL)
```
makalaaneesh: this could panic in the worst case; 
Can we do a `pgDB, ok := ...` and return an err if `!ok`

yb-voyager/cmd/live_migration.go
```diff
@@ -153,10 +153,10 @@ func streamChanges(state *ImportDataState, tableNames []sqlname.NameTuple) error
 		return fmt.Errorf("error handling cdc partitioning strategy: %w", err)
 	}
 
-	if !disablePb {
+	{
```
makalaaneesh: nit: remove the parentheses for the block since there's no if condition anymore.

yb-voyager/src/adaptiveparallelism/adaptive_parallelism.go
```diff
@@ -77,6 +79,7 @@ type TargetYugabyteDBWithConnectionPool interface {
 	GetClusterMetrics() (map[string]tgtdb.NodeMetrics, error) // node_uuid:metric_name:metric_value
 	GetNumConnectionsInPool() int
 	GetNumMaxConnectionsInPool() int
+	GetPendingConnsToCloseInPool() int
```
makalaaneesh: don't see this being used anywhere; so can we remove it?

makalaaneesh (commented):
thanks for the changes @JoeSouthan ! Few more comments (should be the last round mostly). 
We can take up further enhancements in subsequent PRs

migtests/scripts/resumption.py
makalaaneesh, line 1: We might have to write a different test specifically for metrics; wouldn't want to meddle with resumption.py as it is used in our long running test pipelines. 

We can take up tests in a subsequent PR as well. Can you pls revert the changes in this file?

makalaaneesh:
pls also resolve the conflicts; after that, i will trigger the test pipelines

yb-voyager/cmd/exportDataStatus.go
priyanshi-yb: Agree, I think we shouldn't use actual row count for the export snapshot rows metrics and just adding this should be good enough for the metrics
```
	metrics.Get().SetExportSnapshotTableExpectedRows(exporterRole, tableMetadata.TableName, tableMetadata.CountTotalRows)
```

yb-voyager/cmd/importData.go
```diff
@@ -1196,9 +1250,9 @@ func importSnapshotData(msr *metadb.MigrationStatusRecord, errorHandler importda
 }
 
 func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorPolicy) {
-	err := setupImportDataObservability()
+	err := updateImportDataStartedInMetaDB()
```
priyanshi-yb: Can you please rebase from main? There are some changes that have been made around this, where this is moved below around line 1237

yb-voyager/cmd/exportDataStatus.go
priyanshi-yb: Agree, CountLiveRows is the count of rows getting exported during the run

yb-voyager/cmd/exportDataDebezium.go
priyanshi-yb: Should be okay 👍

yb-voyager/cmd/importData.go
JoeSouthan: Intentional. Import parallelism applies to every importer role (target / source / source-replica / import-file), not just the target DB, so the gauge is emitted for all roles rather than only inside the `if importerRole == TARGET_DB_IMPORTER_ROLE` block. It lives in the `if !AdaptiveParallelismMode.IsEnabled()` branch because adaptive parallelism emits the same gauge itself once its poller starts; this branch only covers the fixed `--parallel-jobs` path where there is no poller. Happy to move it back inside the target-only block if you'd rather report parallelism for the target import alone.

JoeSouthan:
Rebased onto latest `main` and resolved the conflicts, and addressed the second review round (export metrics now use `CountLiveRows`, streaming progress always polls at 10s, counter seeding dropped in favour of pre-registering series at 0 with seeded gauges to follow in a later PR, expected-rows from `task.RowCount`, table-started timestamp only on fresh start, shared metrics-server helpers moved to `cmd/metricsServer.go`, guarded the PG type assertion, removed the dead pending-close getter chain, flattened the conditionless block, and reverted `resumption.py`). Ready for the pipelines whenever you are.

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -512,12 +517,7 @@ func calculateStreamingProgress(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		default:
-			if disablePb && callhome.SendDiagnostics {
-				// to not do unneccessary frequent calls to metadb in case we only require this info for callhome
```
makalaaneesh, line 516: @priyanshi-yb what was the context for this selective change before?

priyanshi-yb: It was just this callhome use case where we needed to calculate stats in case of disable-pb, since there was no other use case to calculate them. So we had different timing for normal vs only callhome paths: 12 mins, because the callhome payload sending frequency is 15 mins, so we were just calculating the stats every 12 mins.

makalaaneesh: thanks! i don't see a problem with removing it then. Just wanted to double-check in case there was some other reason.

yb-voyager/cmd/importData.go
```diff
@@ -1196,9 +1202,9 @@ func importSnapshotData(msr *metadb.MigrationStatusRecord, errorHandler importda
 }
 
 func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorPolicy) {
-	err := setupImportDataObservability()
+	err := updateImportDataStartedAndSomeConfigsInMetaDB()
 	if err != nil {
-		utils.ErrExit("Failed to setup import data observability: %s", err)
+		utils.ErrExit("Failed to update import data started in meta DB: %s", err)
```
priyanshi-yb: In main, there is a recent change where this function call is moved to line 1252 after the `prepareCdcPartitionKey` call. Can you please check and remove this call

priyanshi-yb: This might be the reason for the failure for the couple of tests in the failpoints action

JoeSouthan: Fixed in 2436fe6a — removed the duplicate early call, kept the one after prepareCdcPartitionKey.

makalaaneesh:
jenkins PG + import-data-file tests - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline-test/1514/

makalaaneesh (approved):
LGTM! 
I will soon come up with a list of items that we had previously discussed and deferred to subsequent PRs

PR #3690, opened 2026-07-21
Optimize conflict detection logic to maintain a lookup of events on same table and same unique index so that it is faster to check for conflicts

priyanshi-yb:

Live-migration CDC import detects unique-key conflicts by holding in-flight UPDATE/DELETE events in `ConflictDetectionCache` and blocking conflicting later events until those clear. Previously, every incoming event scanned the entire cache, which becomes expensive as the cache grows under heavy UK-conflict load.

This PR adds a value-keyed secondary index (`ukLookup`) over cached source events, keyed by table + unique-index identity + BeforeFields value tuple, plus a reverse map (`vsnToBuckets`) for O(#buckets) removal. On the normal source path, conflict checks look up only matching buckets (before-after and before-before) instead of scanning all cached events. Fall-back/fall-forward (target-DB-exporter) keeps the existing table-level full scan, since those conflicts are value-agnostic. Logging is tightened so a blocked incoming event emits one concise INFO, with per-index detail at DEBUG. `UniqueIndex` also carries `IndexName` for clearer conflict diagnostics.

Improvement in benchmark-

main
```
pth-max	         0 conflicts/op	      1344 events/s

--- cdcbench summary ---
WORKLOAD                               RUNS  EVENTS/S  CONFLICTS/RUN  BATCHES/RUN  CACHE DEPTH AVG/MAX  TIME/RUN
oltp-checkout-multi-table              1     8,585     0              200          1,138 / 1,791        2.329s
oltp-insert-null-then-fill             1     1,585     0              696          1,544 / 2,304        12.618s
oltp-job-queue                         1     1,920     0              600          1,845 / 2,767        10.414s
oltp-ledger                            1     360,103   0              105          0 / 0                56ms
oltp-mixed-crud                        1     2,343     0              500          1,181 / 1,776        8.534s
oltp-mixed-uk-and-plain-tables         1     7,906     0              200          1,658 / 2,648        2.529s
oltp-order-lifecycle                   1     1,608     0              666          1,566 / 2,328        12.434s
oltp-session-churn                     1     2,926     0              400          1,797 / 2,734        6.8s
oltp-skewed-hot-rows                   1     1,361     0              798          1,766 / 2,640        14.685s
oltp-state-machine-transitions         1     65,287    10,000         11,986       1 / 1                306ms
schema-composite-uk-null-component     1     1,404     0              800          1,850 / 2,741        14.235s
schema-composite-uk-shared-tenants     1     1,378     0              800          1,796 / 2,680        14.514s
schema-composite-uk-unique-tenants     1     1,387     0              800          1,809 / 2,694        14.412s
schema-no-uk-control                   1     213,673   0              100          0 / 0                94ms
schema-partial-index-inactive-rows     1     51,030    12,820         20,000       1 / 2                392ms
schema-two-unique-indexes              1     1,301     0              800          1,718 / 2,564        15.364s
schema-wide-rows                       1     1,344     0              800          1,757 / 2,619        14.877s
edge-all-deletes                       1     213,024   0              100          0 / 0                94ms
edge-all-inserts                       1     300,369   0              100          0 / 0                67ms
edge-all-updates                       1     1,361     0              800          1,773 / 2,644        14.692s
edge-updates-and-deletes               1     1,936     0              600          2,502 / 3,724        10.33s
conflict-delete-reinsert               1     53,857    10,000         19,828       1 / 1                371ms
conflict-documented-types              1     57,740    11,755         19,964       1 / 2                346ms
conflict-nulls-not-distinct            1     65,116    10,000         10,001       2 / 2                307ms
conflict-second-index                  1     51,528    12,969         20,000       1 / 2                388ms
conflict-update-pairs                  1     51,207    13,184         19,828       2 / 2                391ms
conflict-value-swap                    1     43,164    16,201         19,994       1 / 2                463ms
conflict-versioned-rows-partial-index  1     57,536    12,190         19,888       1 / 2                348ms

PASS
ok  	github.com/yugabyte/yb-voyager/yb-voyager/cmd	174.389s
```

this PR

```

--- cdcbench summary ---
WORKLOAD                               RUNS  EVENTS/S  CONFLICTS/RUN  BATCHES/RUN  CACHE DEPTH AVG/MAX  TIME/RUN
oltp-checkout-multi-table              1     328,471   0              100          817 / 1,691          61ms
oltp-insert-null-then-fill             1     195,075   0              100          5,358 / 8,252        103ms
oltp-job-queue                         1     257,607   0              100          8,524 / 12,978       78ms
oltp-ledger                            1     352,389   0              105          0 / 0                57ms
oltp-mixed-crud                        1     176,732   0              100          4,435 / 7,376        113ms
oltp-mixed-uk-and-plain-tables         1     326,786   0              100          1,332 / 2,739        61ms
oltp-order-lifecycle                   1     188,942   0              100          9,182 / 15,000       106ms
oltp-session-churn                     1     304,363   0              100          6,534 / 9,900        65ms
oltp-skewed-hot-rows                   1     49,807    0              107          11,066 / 16,168      402ms
oltp-state-machine-transitions         1     61,827    10,000         11,986       1 / 2                323ms
schema-composite-uk-null-component     1     242,435   0              100          0 / 0                82ms
schema-composite-uk-shared-tenants     1     236,493   0              100          7,815 / 16,163       85ms
schema-composite-uk-unique-tenants     1     239,628   0              100          8,191 / 16,583       83ms
schema-no-uk-control                   1     198,819   0              100          0 / 0                101ms
schema-partial-index-inactive-rows     1     50,401    10,000         20,000       1 / 2                397ms
schema-two-unique-indexes              1     201,934   0              100          23,707 / 38,040      99ms
schema-wide-rows                       1     74,784    0              100          9,111 / 18,474       267ms
edge-all-deletes                       1     229,419   0              100          7,541 / 15,514       87ms
edge-all-inserts                       1     293,507   0              100          0 / 0                68ms
edge-all-updates                       1     165,252   0              100          8,959 / 17,939       121ms
edge-updates-and-deletes               1     206,210   0              100          11,801 / 19,213      97ms
conflict-delete-reinsert               1     62,594    10,000         19,828       0 / 1                320ms
conflict-documented-types              1     61,497    10,000         19,964       1 / 2                325ms
conflict-nulls-not-distinct            1     70,985    10,000         10,001       1 / 2                282ms
conflict-second-index                  1     56,607    10,000         20,000       3 / 6                353ms
conflict-update-pairs                  1     54,029    10,000         19,828       1 / 2                370ms
conflict-value-swap                    1     34,671    13,332         19,994       1 / 2                577ms
conflict-versioned-rows-partial-index  1     59,271    10,000         19,888       1 / 2                337ms

PASS
ok  	github.com/yugabyte/yb-voyager/yb-voyager/cmd	7.322s
```

No user-facing changes. Behavior of conflict ordering is unchanged; this is an internal performance/logging improvement for CDC import. No CLI, config, installation, or report changes.

- New/extended unit tests in `conflictDetectionCache_test.go` (build tag `unit`), including:
  - `TestConflictLookup_FindsBeforeAfterConflict`
  - `TestConflictLookup_FindsCompositeConflict`
  - `TestConflictLookup_FindsBeforeBeforeConflict_NullsNotDistinct`
  - `TestConflictLookup_NullsDistinctNotIndexed`
  - `TestConflictLookup_SamePKNoConflict`
  - `TestConflictLookup_NoConflictDoesNotBlock`
  - `TestConflictLookup_RemoveDeindexes`
  - `TestComputeConflictBucketKey_NoAmbiguity`
  - plus existing `eventsConfict` / NULLS DISTINCT coverage
- Run: `go test -tags unit ./cmd/ -run 'TestConflictLookup_|TestComputeConflictBucketKey_|TestEventsConfict_|TestEventsConflict_'`
- Existing live-migration UK conflict / partition failpoint tests remain the end-to-end coverage for this path (`live_migration_uk_conflict_test.go`).
- CDC ingest bench shim updated so cache-depth sampling includes `ukLookup` and seeds CDC partitioning metadata needed by the streaming path.

fallback unique conflict resumption test - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/990/

three basic live all workflow resumption test - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/989

PG/Oracle live tests - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/8567/

iteration test -https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/992/parameters/

partition iteration test - https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/991

No.

No.

---

| Component        


yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,271 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+	c.indexEventLocked(cachedEvent)
 	log.Infof("adding event vsn(%d) to conflict cache", event.Vsn)
 }
 
+// indexEventLocked adds the cached event to the value-keyed lookup index for every
+// unique index of its table whose columns are all present (and indexable) in the
+// event's BeforeFields. It is a no-op for events with nil/empty BeforeFields (e.g.
+// target-DB-exporter events). Caller must hold the lock.
+func (c *ConflictDetectionCache) indexEventLocked(event *tgtdb.Event) {
+	if event.BeforeFields == nil {
```
makalaaneesh: when is this expected? should we error out?

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,271 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+// computeConflictBucketKey builds an unambiguous lookup key for the given table,
+// unique-index position and the index-column values found in fields. It returns
+// ok=false when the tuple is not indexable, mirroring the conflict semantics in
+// uniqueIndexColumnsExistInBothFields and uniqueKeyColumnValuesEqual:
+//   - a column is missing from fields, OR
+//   - a column value is NULL under the default NULLS DISTINCT (NULLs never conflict).
+//
+// Values are length-prefixed so distinct tuples never collide (e.g. {"ab",""} vs
+// {"a","b"}); NULLs under NULLS NOT DISTINCT use a dedicated sentinel.
+func computeConflictBucketKey(table string, indexPos int, fields map[string]*string, index tgtdb.UniqueIndex) (string, bool) {
```
makalaaneesh: `indexPos` is a bit error-prone. Can we just use the name? does it really matter for performance?

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,271 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+	if fields == nil {
```
makalaaneesh, line 470: when is this possible? should we error out instead?

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,271 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+		return "", false
+	}
+	var b strings.Builder
+	b.WriteString(table)
+	b.WriteByte(0)
+	b.WriteString(strconv.Itoa(indexPos))
+	for _, column := range index.Columns {
+		val, exists := fields[column]
+		if !exists {
```
makalaaneesh, line 479: when is this possible? should we error out instead?

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,271 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+//
+// Values are length-prefixed so distinct tuples never collide (e.g. {"ab",""} vs
+// {"a","b"}); NULLs under NULLS NOT DISTINCT use a dedicated sentinel.
+func computeConflictBucketKey(table string, indexPos int, fields map[string]*string, index tgtdb.UniqueIndex) (string, bool) {
+	if fields == nil {
+		return "", false
+	}
+	var b strings.Builder
+	b.WriteString(table)
+	b.WriteByte(0)
```
makalaaneesh: why the `0`s everywhere?

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,271 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+		if !exists {
+			return "", false
+		}
+		b.WriteByte(0)
+		if val == nil {
+			if !index.NullsNotDistinct {
+				// default NULLS DISTINCT: NULLs never conflict, so this tuple is not indexable.
+				return "", false
+			}
+			b.WriteString(conflictBucketNull)
```
makalaaneesh: what if there's an actual value `N`? this seems error-prone

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,271 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
 }
 
 func (c *ConflictDetectionCache) Put(event *tgtdb.Event) {
 	c.Lock()
 	defer c.Unlock()
-	c.m[event.Vsn] = event.Copy()
+	// caches a separate copy of the event, not a pointer, otherwise it will be modified
+	// by ConvertEvent() causing issues in events comparison for conflict detection.
+	cachedEvent := event.Copy()
+	c.m[cachedEvent.Vsn] = cachedEvent
```
makalaaneesh: do we really need to ALSO put it in the map now that we have this index?

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,271 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+}
+
+// findConflictLocked returns the cached events that conflict with the incoming event,
+// or an empty slice if there is none, along with detail describing the value-path match
+// (empty conflictInfo on the target-DB-exporter path). Caller must hold the lock.
+//
+// For the target-DB-exporter path (fall-back/fall-forward) conflicts are table-level
+// and value-agnostic (see eventsConfict), so we retain the full scan over the cache.
+// For the normal source path we use the value-keyed lookup index (findValueConflictLocked).
+func (c *ConflictDetectionCache) findConflictLocked(incomingEvent *tgtdb.Event) ([]*tgtdb.Event, conflictInfo) {
```
makalaaneesh: `conflictInfo` could be different for different events, no? wouldn't we need one per event that we conflict with?

yb-voyager/cmd/conflictDetectionCache.go
makalaaneesh: table type should be NameTuple?

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,271 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+			cachedEvents := c.getNonSamePKEventsWithSameBucketKey(key, incomingEvent)
+			for _, cachedEvent := range cachedEvents {
+				log.Debugf("conflict detected for table %s, index columns %v, between before value of cached-event1(vsn=%d, colVal=%s) and after value of incoming-event2(vsn=%d, colVal=%s)",
+					table, index.Columns,
+					cachedEvent.Vsn, formatUniqueIndexColumnValuesForLog(cachedEvent.BeforeFields, index.Columns),
+					incomingEvent.Vsn, formatUniqueIndexColumnValuesForLog(incomingEvent.Fields, index.Columns))
+			}
+			if len(cachedEvents) > 0 {
+				//If before-after conflict is detected, then return the cached events
+				return cachedEvents, conflictInfo{
```
makalaaneesh: Oh so we're not returning all possible conflicts? We return early as soon as we find any conflict.

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: In target db exporter case, before fields are empty

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: hmm will see, I think it's just that the key length will be more with name

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: In case of delete, the after fields are empty so we only need to calculate the key for before

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: So the field here is any field before/after so in case of incoming UPDATE/INSERT event doesn't have the unique column in the after fields, so we don't need to check further for the conflict of before-after case for such event so returning false

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: for the source exporter we don't need this as we will use the index lookup but for the target exporter we need this map, will change this to only populate it in case it's coming from target exporter

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: we have two types of conflict checks: before-before/before-after, so right now if we find any conflicting events for the before-after case, then we are returning those events, but if there are none, then we check for before-before. I know we were thinking of returning all the possible conflicts and then have the caller decide what to do with that, and in the future we need to improve the logic to just flush the event on which those events are present. But I think it's okay for now to have these two separate checks and then return if one has found conflicts for the incoming one; it's easier for logging also.
Let me know what do you think

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,271 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
 	c.tableToUniqueIndexes = tableToUniqueIndexes
 	c.sourceDBType = sourceDBType
 	c.evChans = evChans
+	c.ukLookup = make(map[string]map[int64]*tgtdb.Event)
+	c.vsnToBuckets = make(map[int64][]string)
 	return c
 }
 
 func (c *ConflictDetectionCache) Put(event *tgtdb.Event) {
```
makalaaneesh: let's have guardrails and error out if: 
- it's NOT an UPDATE/DELETE
- BeforeFields is nil

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,271 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
-				select {
-				case c.evChans[i] <- FLUSH_BATCH_EVENT:
-				default:
-					// channel is full, so it's okay not to send FLUSH_BATCH_EVENT
-					// because MAX_EVENTS_PER_BATCH would likely be reached in the next batch.
-					log.Infof("channel %d is full with size %d, not sending FLUSH_BATCH_EVENT", i, len(c.evChans[i]))
-				}
+	conflictLogged := false
+	for {
+		cachedEvents, info := c.findConflictLocked(incomingEvent)
```
makalaaneesh: I'd recommend: 

```
type conflictInfo struct {
	indexName string 
	eventsConflicting []tgtdb.Event
	indexColumns []string // matched unique-index columns
	matchType    string   // "before-after" or "before-before"
	matchedValue string   // formatted incoming-side index-column values (shared by all cached events)
}
```
return an array of conflictInfos

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: As discussed it won't because the format will be something like 
```
<table>\0<index>\0V<len>:<val>\0N\0V<len>:<val>
```

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: its a separator for table/index and every field data

yb-voyager/cmd/conflictDetectionCache_test.go
```diff
@@ -456,3 +457,234 @@ func TestUniqueKeyConflictPairKey_OrdersVsns(t *testing.T) {
...
+		ExporterRole: SOURCE_DB_EXPORTER_ROLE,
+	}
+	assert.Empty(t, findConflictForTest(cache, incoming))
+}
+
+// computeConflictBucketKey must not collide for tuples that differ only in the
+// split of characters between adjacent columns.
+func TestComputeConflictBucketKey_NoAmbiguity(t *testing.T) {
+	idx := uidx("a", "b")
+	k1, ok1 := computeConflictBucketKey("public.users", 0, map[string]*string{"a": strPtr("ab"), "b": strPtr("")}, idx)
```
makalaaneesh: [AI] **New unit test does not compile — breaks the entire `cmd` unit-test build.**

`TestComputeConflictBucketKey_NoAmbiguity` calls `computeConflictBucketKey` with the wrong argument types. The signature is:

```go
func computeConflictBucketKey(table sqlname.NameTuple, indexName string, fields map[string]*string, index tgtdb.UniqueIndex) (string, bool)
```

but the test passes a string literal for `table` and `0` for `indexName` (lines 681, 682, 688). `go vet -tags unit ./cmd/` fails to compile the whole `cmd` test package:

```
cmd/conflictDetectionCache_test.go:681:38: cannot use "public.users" (untyped string constant) as sqlname.NameTuple value in argument to computeConflictBucketKey
```

Because the test binary won't build, CI's `go test -tags unit ./...` (`go.yml`) fails and **none** of the `cmd` unit tests run — including the new lookup-index tests this PR adds. (Verified locally: `go build ./cmd/` passes; only this test file is stale.)

Suggested fix — pass a real `NameTuple` and the index name (empty string, matching how the cache builds keys with `uidx`'s empty `IndexName`):

```go
tbl := testTableTuple()
k1, ok1 := computeConflictBucketKey(tbl, "", map[string]*string{"a": strPtr("ab"), "b": strPtr("")}, idx)
k2, ok2 := computeConflictBucketKey(tbl, "", map[string]*string{"a": strPtr("a"), "b": strPtr("b")}, idx)
_, ok3 := computeConflictBucketKey(tbl, "", map[string]*string{"a": strPtr("a")}, idx)
```

---

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,58 +155,305 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
 	c.sourceDBType = sourceDBType
 	c.evChans = evChans
+	c.ukLookup = make(map[string]map[int64]*tgtdb.Event)
+	c.vsnToBuckets = make(map[int64][]string)
 	return c
 }
 
-func (c *ConflictDetectionCache) Put(event *tgtdb.Event) {
+func (c *ConflictDetectionCache) Put(event *tgtdb.Event) error {
+	if event.BeforeFields == nil || event.Op == "i" {
```
makalaaneesh: [AI] **`Put` now hard-errors on nil `BeforeFields`, aborting the whole migration where the old code tolerated it.**

`Put` was previously infallible and unconditionally cached every event. It now returns an error when `event.BeforeFields == nil` (or `Op == "i"`), and `handleEvent` (`live_migration.go`) propagates that as a fatal error that aborts the import:

```go
err = conflictDetectionCache.Put(event)
if err != nil {
    return goerrors.Errorf("error putting event into conflict detection cache: %v", err)
}
```

Any `d`/`u` event that reaches `Put` with a nil `BeforeFields` now **kills the migration** instead of being cached. This is most concerning on the target-DB-exporter (fall-back / fall-forward) path — the surrounding code itself notes YB-CDC before-images are unreliable. Previously such an event just sat in `c.m` and was still handled by the table-level target-path scan (which treats any cached `d` as a conflict regardless of before-values); now it aborts. The message `"before fields are nil"` is also misleading for the `Op == "i"` half of the condition (inserts legitimately have no before fields).

Suggested fix: confirm every `d`/`u` event across all flows (forward, fall-back, fall-forward, changes-only, all partition strategies) is guaranteed a non-nil `BeforeFields`. If it is not guaranteed, don't fail the migration — keep tolerating such events on the legacy `c.m` scan path (or log-and-skip). Split the conditions so the message is accurate:

```go
if event.Op == "i" {
    return goerrors.Errorf("insert event vsn(%d) cannot be cached", event.Vsn)
}
if event.BeforeFields == nil {
    // decide: tolerate (store on scan path) vs. error — see above
}
```

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -546,8 +542,8 @@ func TestLiveMigrationWithUniqueKeyValuesWithPartialPredicateConflictDetectionCa
 	conflictStats, err := testutils.ReadUniqueKeyConflictStats(uniqueKeyConflictStatsPath)
 	testutils.FatalIfError(t, err, "failed to read unique key conflict stats")
 
-	require.LessOrEqual(t, conflictStats.Total, 2500, "true-positive delta should produce at most 2500 UK conflicts")
-	require.LessOrEqual(t, conflictStats.ByTable[`"test_schema"."test_live"`], 2500, "test_live should have at most 2500 UK conflicts")
+	require.GreaterOrEqual(t, conflictStats.Total, 0, "true-positive delta should produce at least 0 UK conflicts")
```
makalaaneesh: [AI] **UK-conflict assertions gutted to `>= 0`, so live-migration tests no longer detect conflict-detection regressions.**

This PR removes the upper-bound (`require.LessOrEqual`) assertions and relaxes the lower bounds to `>= 0` across the UK-conflict live-migration tests (lines 347, 545–548, 765, 1083–1084, 1229). `require.GreaterOrEqual(t, count, 0)` on a count is vacuously true and asserts nothing.

These tests exist specifically to bound the number of UK conflicts produced during streaming. With every bound now `>= 0`, a regression in the new lookup index that **misses a real conflict** — a false negative, which in production means a unique-constraint violation / data corruption during apply — would no longer fail the test. The messages are self-contradictory too (line 545 asserts `>= 0` but says "at least 3000 UK conflicts"), and lines 547–548 are exact duplicates of 545–546.

Conflict detection is a data-correctness hot path (root `.cursor/BUGBOT.md`, "Performance-Critical (Hot) Paths"), and this PR's whole purpose is to change that logic — the tests guarding it should be **tightened, not neutralized**. If the optimization genuinely changes the observed conflict counts, re-derive and assert the new lower/upper bounds and explain why the count changed.

Suggested fix: restore meaningful lower/upper bounds (recomputed if necessary) and delete the duplicated lines 547–548.

---

yb-voyager/src/tgtdb/main_test.go
```diff
@@ -126,8 +126,8 @@ func TestMain(m *testing.M) {
 	// 4. Create a yugabytedb 3 node cluster container
 	yugabytedbClusterContainer := testcontainers.NewYugabyteDBCluster(&testcontainers.ContainerConfig{
 		DBType:            testcontainers.YUGABYTEDB,
-		NodeCount:         3,
-		ReplicationFactor: 3,
+		NodeCount:         1,
```
makalaaneesh: [AI] **Unrelated test-infra downgrades ride along: Oracle target disabled and YB cluster reduced from 3 nodes/RF3 to 1/1.**

This PR (about CDC conflict detection) also comments out the entire Oracle target-DB setup/teardown (lines 115–118, 167) and reduces the `tgtdb` integration cluster from `NodeCount: 3, ReplicationFactor: 3` to `NodeCount: 1, ReplicationFactor: 1` (lines 129–130). These look like leftover local-iteration changes to speed up local runs, unrelated to the stated goal.

Merging this silently drops Oracle-target integration coverage in the `tgtdb` package and stops exercising a real 3-node RF3 cluster, weakening the integration suite for everyone. Per root `.cursor/BUGBOT.md` ("Remove dead code, unused functions, and leftover debugging artifacts before merging") and the scope/necessity lens, incidental test-infra changes shouldn't ride in a conflict-detection PR.

Suggested fix: revert `main_test.go` to restore the Oracle target setup/teardown and the 3-node/RF3 cluster, unless there is a deliberate, separately-justified reason to change them (in which case it belongs in its own PR).

---

makalaaneesh (commented):
[AI] Automated review — 4 findings (1 critical, 3 warning). See inline comments.

The lookup-index optimization to `ConflictDetectionCache` (secondary `ukLookup` + `vsnToBuckets` reverse index) is a reasonable design for avoiding the O(N) cache scan on the source-exporter path. Two of the findings are the highest priority: the new `cmd` unit test does not compile (which fails CI's `go test -tags unit ./...` and prevents *all* `cmd` unit tests, including the new ones, from running), and `Put` now aborts the whole migration on a nil `BeforeFields` where the old code tolerated it. The other two warnings concern test assertions being relaxed to `>= 0` (they can no longer catch a missed-conflict regression on this data-correctness hot path) and unrelated `tgtdb` test-infra downgrades (Oracle target disabled, cluster reduced to 1 node/RF1) riding along in this PR. A couple of non-blocking observations (eager log formatting on the conflict path; the flush loop bounded by the `NUM_EVENT_CHANNELS` global rather than `len(c.evChans)`) were noted but not posted.

---

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,58 +155,305 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+// computeConflictBucketKey builds an unambiguous lookup key for the given table,
+// unique-index position and the index-column values found in fields. It returns
+// ok=false when the tuple is not indexable, mirroring the conflict semantics in
+// uniqueIndexColumnsExistInBothFields and uniqueKeyColumnValuesEqual:
+//   - a column is missing from fields, OR
+//   - a column value is NULL under the default NULLS DISTINCT (NULLs never conflict).
+//
+// Values are length-prefixed so distinct tuples never collide (e.g. {"ab",""} vs
+// {"a","b"}); NULLs under NULLS NOT DISTINCT use a dedicated sentinel.
+func computeConflictBucketKey(table sqlname.NameTuple, indexName string, fields map[string]*string, index tgtdb.UniqueIndex) (string, bool) {
```
makalaaneesh: tgtdb.UniqueIndex.IndexName is already there?  why pass indexName again?

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,58 +155,305 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+	var b strings.Builder
+	b.WriteString(table.ForKey())
+	b.WriteByte(0) //separator for the table name and index name
+	b.WriteString(indexName)
```
makalaaneesh: pls change comments everywhere. I still see indexpos in some places

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,58 +155,305 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+	uniqueIndexes, _ := c.tableToUniqueIndexes.Get(incomingEvent.TableNameTup)
+	if len(uniqueIndexes) == 0 {
+		return []Conflict{}
+	}
+	totalConflictInfo := make([]Conflict, 0)
+	for _, index := range uniqueIndexes {
+		// before-after conflict: cachedEvent.BeforeFields == incomingEvent.Fields
+		if key, ok := computeConflictBucketKey(incomingEvent.TableNameTup, index.IndexName, incomingEvent.Fields, index); ok {
+			cachedEvents := c.getNonSamePKEventsWithSameBucketKey(key, incomingEvent)
+			for _, cachedEvent := range cachedEvents {
```
makalaaneesh: nits: 

1. Put this logic inside the if block `if len(cachedEvents) > 0`. 
2. write a String() on `Conflict` struct with all the conflict information; then you can just log the conflict struct

yb-voyager/cmd/conflictDetectionCache.go
makalaaneesh: let's error out? right now it just access fields[col] which can panic

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
makalaaneesh: pls review these test changes @priyanshi-yb

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,58 +155,305 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
-	c.m[event.Vsn] = event.Copy()
+	// caches a separate copy of the event, not a pointer, otherwise it will be modified
+	// by ConvertEvent() causing issues in events comparison for conflict detection.
+	cachedEvent := event.Copy()
+	if isTargetDBExporter(event.ExporterRole) {
+		c.m[cachedEvent.Vsn] = cachedEvent
+	} else {
+		c.indexEventLocked(cachedEvent)
+	}
 	log.Infof("adding event vsn(%d) to conflict cache", event.Vsn)
```
makalaaneesh: could we downgrade to debug?

yb-voyager/cmd/conflictDetectionCache.go
makalaaneesh:
```
Put()
	if ! U/D 
		return 
	if U: 
		if changedColumns <intersection> (UK columns <union> predictate columns) == EMPTY: 
			return 
		add-to-cache (uklookup) <---- computeKey (beforeFields) (all columns should be present)

FindConflicts()
	for each unique index:
		if D:
			skip-check 
		if I: 
			if uniqueINdex.NullsDistinct AND ANY unique key column value is NULL: 
				skip-check
			check before-after conflicts; <---- computeKey (afterFields) (all columns should be present)
			 no before-before
		if U: 
			
			if changedColumns <intersection> (UK columns <union> predictate columns) == EMPTY: 
				skip-check 
			before-after:
				if uniqueINdex.NullsDistinct AND ANY unique key column value (after fields) is NULL: 
					skip-check 
				check before-after conflicts <---- computeKey (afterFields) (possible that only a subset of columns are present)
			
			if uniqueIndex has partial predicate: (before-before)
				if uniqueINdex.NullsDistinct AND ANY unique key column value (before fields) is NULL: 
					skip-check 
				check before-before conflicts.  <---- computeKey (beforeFields) (all columns should be present)
```

makalaaneesh: ok to fix this flow later

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +154,200 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+			b.WriteString(conflictBucketNull)
+			continue
 		}
+		b.WriteString(conflictBucketValue)
+		b.WriteString(strconv.Itoa(len(*val)))
+		b.WriteByte(':')
+		b.WriteString(*val)
 	}
+	return b.String(), true
 }
```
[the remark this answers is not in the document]
makalaaneesh: minor optimization; can ignore

yb-voyager/cmd/conflictDetectionCache_test.go
```diff
@@ -21,12 +21,14 @@ import (
 	"os"
```
makalaaneesh, line 21: the richer family (16 NULL/composite cases) exercises the production-dead eventsConfict value path, while the production findConflictLocked path gets only 7 comparatively basic tests, and the encoding function
  gets exactly one adversarial case.

makalaaneesh: All tests should be at the level of `.Put()` and `.findConflictLocked()`

makalaaneesh, line 21: Multi-index lookup tests — currently zero lookup tests with two indexes

makalaaneesh, line 21: don't set IndexName, so every test index shares IndexName: ""

makalaaneesh, line 21: 4. Behavior tests for this commit's new semantics (all untested)

  - Put returns error for Op == "i" and for nil BeforeFields (and: every existing cache.Put(cached) call ignores the new error — add require.NoError).
  - Cached event is indexed by BeforeFields only: cached update before{email:"old"} after{email:"new"}; incoming insert email:"new" → no conflict; incoming insert email:"old" → conflict. Nothing asserts the after-image isn't indexed.
  - Multiple cached events in one bucket: two deletes with the same UK value, different PKs; incoming insert → both VSNs returned.
  - Table isolation: two tables with same index name and same values → no cross-table match.
  - Target-exporter branch of findConflictLocked (the c.m full scan): one cached + one incoming with target-exporter roles → table-level conflict found; also that a target-exporter cached event never enters ukLookup.
  - Duplicate before-after/before-before entries for unchanged-UK update — assert current behavior (2 identical entries) or the deduped behavior if you take my earlier suggestion; either way pin it.

yb-voyager/cmd/conflictDetectionCache.go
makalaaneesh: add a TODO

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -137,47 +155,310 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
...
+			// detail is logged at DEBUG in findValueConflictLocked. The target-DB-exporter
+			// path already emits its own "conflict detected" INFO in eventsConfict, so here
+			// we only log at DEBUG for that path to avoid double counting.
+			cachedVsns := lo.Map(info.eventsConflicting, func(e *tgtdb.Event, _ int) int64 { return e.Vsn })
+			if isTargetDBExporter(incomingEvent.ExporterRole) {
+				log.Debugf("event(vsn=%d) waiting for in-flight event(s) %v to be applied", incomingEvent.Vsn, cachedVsns)
+			} else if !conflictLogged {
+				log.Infof("conflict detected: event(vsn=%d) on table %s, unique index columns %v with matched value %s (%s conflict) waiting for in-flight event(s) %v to be applied",
+					incomingEvent.Vsn, incomingEvent.TableNameTup.ForKey(), info.indexColumns, info.matchedValue, info.matchType, cachedVsns)
+				conflictLogged = true
```
makalaaneesh: should be after the for loop

makalaaneesh (approved):
LGTM

PR #3691, opened 2026-07-21
Fix concurrent CREATE/DROP DATABASE failures in migtests on YB 2026.1+

Abhi-yb:

Serialize target `CREATE`/`DROP DATABASE` in the migtest harness so parallel tests no longer fail with concurrent-DDL errors on **YB 2026.1+**.

**Problem.** YB 2026.1+ no longer supports concurrent `CREATE DATABASE`. When migtests run in parallel against the same target and create/drop databases at the same time, we see:
- `Restart read required / Catalog Version Mismatch` — a concurrent DDL bumped the catalog version mid-statement.
- `Keyspace '<name>' already exists` — a concurrent `CREATE` partially applied (the DocDB keyspace was created) and then aborted, leaving an **orphaned keyspace with no `pg_database` row**. A later `DROP DATABASE IF EXISTS` is then a no-op (nothing in the catalog to drop), so the next `CREATE` keeps failing. Retries cannot recover this case.

This was clean on 2025.2.3.0-b149 and started failing on 2026.1.0.0-b118. Confirmed with the YB team that concurrent `CREATE DATABASE` is unsupported from 2026.1.

**Fix.** Guard target create/drop with a host-level `flock` (`YB_DB_LOCK_FILE`) so only one test does a create-or-drop `DATABASE` at a time — the collision never happens. This does **not** serialize whole tests, only the few seconds of create/drop. A small retry loop remains as a fallback for any residual transient failure.

- `functions.sh`: add `YB_DB_LOCK_FILE` + `create_target_database` helper; the standalone drop (`ysql_terminate_and_drop_database`) now takes the same lock, so create and drop are serialized together. The DB-14314 terminate-stale-sessions workaround is preserved.
- `run-test.sh`, `live-migration-run-test.sh`, `live-migration-fallb-run-test.sh`, `live-migration-fallf-run-test.sh`, `run-schema-migration.sh`, `callhome/run-callhome-test.sh`: create the target DB via `create_target_database`.

`flock` only serializes processes that share the lockfile's filesystem (i.e. run on the same agent), which the parallel migtests do. `YB_DB_LOCK_FILE` is env-overridable so the path can be pointed at a shared volume if any tests ever run in isolated containers.

Out of scope: the import-file and resumption test frameworks create their target DB directly and are not part of the failing set.

None. Test-harness only (`migtests/scripts/`); no changes to `yb-voyager` itself.

- Existing migtests are the coverage; this only changes how they create/drop the target DB.
- Verified the lock logic in isolation (stubbed `run_ysql`): concurrent creates + a standalone drop are fully serialized with no overlap, no self-deadlock (create's internal drop runs on the unlocked core while holding the lock), and return codes propagate.
- To be validated by running both pipelines: passes on the older YB version (no regression) and fixes the failures on 2026.1.

No.

No.

### DIFF
--- a/migtests/scripts/callhome/run-callhome-test.sh
+++ b/migtests/scripts/callhome/run-callhome-test.sh
@@ -143,8 +143,7 @@ main() {
     fi
 
     step "Create target database."
-    ysql_terminate_and_drop_database "${TARGET_DB_NAME}"
-    run_ysql yugabyte "CREATE DATABASE  \"${TARGET_DB_NAME}\" with COLOCATION=TRUE;"
+    create_target_database "${TARGET_DB_NAME}" true
 
     step "Export schema"
     export_schema --send-diagnostics=true
--- a/migtests/scripts/functions.sh
+++ b/migtests/scripts/functions.sh
@@ -93,17 +93,73 @@ run_ysql() {
 	PGPASSWORD="${TARGET_DB_ADMIN_PASSWORD}" psql -P pager=off -h ${TARGET_DB_HOST} -p ${TARGET_DB_PORT} -U ${TARGET_DB_ADMIN_USER} -d ${db_name} -c "${sql}"
 }
 
-# TODO: Remove this helper (and all its call-sites) once the underlying
+# Newer YB versions (2026.1+) no longer support concurrent CREATE DATABASE.
+# When parallel migtests create/drop databases on the same target at once, a
+# concurrent CREATE can partially apply and then abort with "Catalog Version
+# Mismatch", leaving an orphaned keyspace with no pg_database row. A later
+# DROP DATABASE IF EXISTS is then a no-op (nothing in the catalog to drop) and
+# the next CREATE fails with "Keyspace '<name>' already exists" — which retries
+# cannot recover from.
+#
+# So we serialize create/drop with a host-level file lock: only one test does a
+# create-or-drop DATABASE at a time, and the collision never happens. flock only
+# serializes processes that share this lockfile's filesystem (i.e. run on the
+# same agent), which our parallel migtests do. It does NOT serialize whole tests
+# — only the few seconds of create/drop.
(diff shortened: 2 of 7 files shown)


makalaaneesh:

Reviewed the 7 changed files (+72/-22, test-harness only under `migtests/scripts/`) against base `main`. The change serializes target `CREATE`/`DROP DATABASE` behind a host-level `flock` (`YB_DB_LOCK_FILE`) to avoid the concurrent-`CREATE DATABASE` failures on YB 2026.1+, adding a `create_target_database` helper and folding the drop into it. The implementation is correct and careful:

- **No self-deadlock**: `create_target_database` and the public `ysql_terminate_and_drop_database` hold the lock and call the `_ysql_terminate_and_drop_database_unlocked` core, so the lock is never re-acquired while held.
- **Return codes propagate**: the `( … ) 200>lockfile` subshell `exit 0/1` becomes the function's status, so a failed create fails-fast under the callers' `set -e`.
- **Call sites consistent**: every `ysql_terminate_and_drop_database` + `CREATE DATABASE` pair in the run scripts was replaced by `create_target_database <name> <colocated>`, preserving the per-source colocation branch; standalone drops still use the locked public wrapper. The DB-14314 terminate-stale-sessions workaround is preserved.
- Follows the `migtests/.cursor/BUGBOT.md` rules: reuses the shared helper instead of reimplementing, and all callers of the changed shared function still work.

Non-blocking observations (not defects): (1) the retry loop holds the lock across the `sleep 10` back-offs, so a genuinely-failing create can block other parallel create/drop for up to ~55s — acceptable since serialization should make retries rare, and keeping the lock during retry intentionally prevents any concurrent `CREATE`; (2) the default `/tmp/yb_create_drop_database.lock` could hit cross-user permission issues on a shared agent, already mitigated by the env-overridable path.

---

---

Abhi-yb:
Superseded: concurrency is handled by enabling the concurrent-DDL gflags on the test cluster (jenkins-helpers#1545) — no harness-side create/drop serialization needed. Remaining test fixes live in #3700.

PR #3703, opened 2026-08-03
Disable sequential scan on live-migration metadata UPDATE queries

makalaaneesh:

Applies the existing `DISABLE_SEQUENTIAL_SCAN_ON_UPDATE_DELETES_HINT` (`/*+ Set(enable_seqscan off) */`) to the two live-migration metadata `UPDATE` queries in `TargetYugabyteDB.ExecuteBatch` (`yb-voyager/src/tgtdb/yugabytedb.go`):

- the channel-metadata update (`ybvoyager_import_data_event_channels_metainfo`)
- the per-table event-stats update (`ybvoyager_imported_event_count_by_table`)

Both are gated on the existing `tconf.DisableSequentialScanOnUpdateDeletes` flag (CLI `--disable-sequential-scan-on-update-deletes`, default `true`). Previously the hint was applied only to data-table `UPDATE`/`DELETE` events, not to these metadata tables.

**Root cause:** These metadata tables are very small (row count = channels x tables), so YugabyteDB's cost-based optimizer sometimes picks a sequential scan for the `UPDATE`s. Under `REPEATABLE READ` isolation, a sequential scan concurrent with other writes triggers read-restart errors.

**Fix:** Applying the hint forces an index/PK scan for these metadata `UPDATE`s, which avoids the read restarts.

No user-facing changes. The behavior is controlled by the pre-existing `--disable-sequential-scan-on-update-deletes` flag (default `true`); no new commands, configuration, installation, or report changes.

The change reuses the existing hint constant and the existing flag gating already covered for data-table events, extending it to the two metadata `UPDATE` queries.

No. There are no changes to callhome or yugabyted payloads.

No. This only prepends a query hint to two existing `UPDATE` statements against the TargetDB metadata tables; it does not change any on-disk structure or schema.

---

### Context

Slack thread: https://yugabyte.slack.com/archives/C0BLZN7BWTF/p1785747616392559?thread_ts=1785747284.077929&cid=C0BLZN7BWTF

---
_Generated by [Claude Code](https://claude.ai/code/session_01HWrxpbXuw4PDdY3b6gS1e2)_

### DIFF
--- a/yb-voyager/src/tgtdb/yugabytedb.go
+++ b/yb-voyager/src/tgtdb/yugabytedb.go
@@ -1290,6 +1290,9 @@ func (yb *TargetYugabyteDB) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
 		}
 
 		updateVsnQuery := batch.GetChannelMetadataUpdateQuery(migrationUUID)
+		if yb.tconf.DisableSequentialScanOnUpdateDeletes {
+			updateVsnQuery = DISABLE_SEQUENTIAL_SCAN_ON_UPDATE_DELETES_HINT + updateVsnQuery
+		}
 		res, err = tx.Exec(context.Background(), updateVsnQuery)
 		if err != nil || res.RowsAffected() == 0 {
 			log.Errorf("error executing stmt for batch(%s): %v, rowsAffected: %v", batch.ID(), err, res.RowsAffected())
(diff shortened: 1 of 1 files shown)


makalaaneesh:
Two CI checks are red, but both are pre-existing/environmental failures unrelated to this change (a 6-line Go string-concat in `yb-voyager/src/tgtdb/yugabytedb.go`):

- **`Build on macos-latest`** — the Go build itself passed; both darwin binaries compiled and were archived. The job failed only at the release-publishing step (`softprops/action-gh-release`: "GitHub Releases requires a tag"). That "Create yb-voyager binary" workflow ran on the branch-`create` event, which has no tag, so the release step errors regardless of the diff.
- **`test-issues-against-all-yb-versions (2025.2.3.0-b149)`** — the `issues_integration` schema-detection tests all passed. The job failed on stale version-metadata assertions: `TestLatestStable` (`src/ybversion`) because the matrix's YB version `2025.2.3.0` is older than versions already committed to the repo, and `TestConnectorLatestStable` (`versions/`) because the bundled Debezium connector is behind the latest release. Both are "bump the version JSON" staleness checks, and the `tgtdb` package isn't compiled into those tests.

Neither failure touches the code path this PR changes. Will re-check as the remaining build and migration-test jobs finish.

---

priyanshi-yb (approved):
LGTM!

makalaaneesh:
The `Failpoint: Cutover` check failed at the `gotestfmt-action@v2` setup step with `HttpError: API rate limit exceeded` — before any Go test ran (no `/tmp/gotest.log` was produced). This is a transient CI-infra flake unrelated to the change; the failpoint tests never reached the modified `ExecuteBatch` code path. Re-running the job should clear it.

---

PR #3705, opened 2026-08-03
Guardrail for cdc-partition-key-overrides to error out properly in case it changes in between the runs

priyanshi-yb:

This PR hardens the resume change-guard for `--cdc-partition-key-overrides` during live migration (`import data to target`, PG→YB). Previously the guard compared the raw overrides string on resume, which was fragile: two strings that differ only in ordering, quoting/casing, or whitespace resolve to the same effective per-table strategy but were wrongly rejected (and a genuinely different config could slip through). This addresses the leftover `TODO` from the original feature.

The raw-string comparison for overrides is removed from `validateCdcPartitionKeyFlags` and replaced with a semantic comparison in `prepareCdcPartitionKey` (via the new `validateCdcPartitioningStrategyUnchanged`): on resume the current flags are re-resolved into an effective per-table strategy map and diffed against the map persisted on the first run (`diffCdcPartitioningStrategy`), producing an actionable error that lists exactly which tables changed. To avoid re-querying the target DB for expression-based unique-index tables on every resume, the expression-UK table set is now captured on the first prepare and persisted in a new `ImportDataStatusRecord` field (`CdcExpressionUniqueIndexTables`); resume reuses it. For older records that predate this capture (nil), the code falls back to the previous raw-string check for overrides. The global `--cdc-partition-key` still uses the existing raw-string guard.

- **Command line:** No new or renamed flags. Behavior improvement: on resume (without `--start-clean`), a `--cdc-partition-key-overrides` string that is written differently but resolves to the same per-table strategy (reordered entries, quoted identifiers, extra whitespace) is now accepted instead of being rejected.
- **Error messages:** The resume change-guard error is reworded to `changing cdc-partition-key / cdc-partition-key-overrides is not allowed after the import data has started; effective strategy changed for: <tables>...` and now names the specific tables whose effective strategy changed. Switching an expression-UK table to `pk` on resume is rejected with the expression-UK error.
- **Configuration / installation / reports:** No changes.

- **New unit test** `TestValidateCdcPartitioningStrategyUnchanged` in `yb-voyager/cmd/live_migration_test.go`: same config passes; semantically-equivalent overrides (quoting / whitespace) pass; a changed override target table and a removed override are rejected. The old raw-string `resume rejects changed overrides` subtest in `TestValidateCdcPartitionKeyFlags` was removed (superseded by the semantic test).
- **New integration test** `TestLiveMigrationCdcPartitionKeyOverridesEquivalentOnResume` in `src/testlivemigration/live_migration_uk_conflict_test.go`: first run with global `pk` + `orders:table` override; a resume that flips the effective strategies is rejected, while a resume with a differently-written but semantically-equivalent overrides string is accepted and runs through streaming and cutover, asserting the persisted per-table map is preserved and that `CdcExpressionUniqueIndexTables` is captured on the first run.
- **Updated integration tests**: `TestLiveMigrationCdcPartitionKeyConfigs` was updated for the new error wording, and `TestLiveMigrationCdcPartitionKeyRejectsPkOnExpressionUniqueIndex` now also verifies that switching the expression-UK table to `pk` on resume is rejected (reusing the captured expression-UK set, no target-DB re-query).

No.

Yes — `ImportDataStatusRecord` (MetaDB / Import Data State) gains a new additive field `CdcExpressionUniqueIndexTables` (`json:"cdcExpressionUniqueIndexTables"`). It is backward compatible: a nil value (records written by an older voyager, or a first run that did not need the expression-UK check) is handled explicitly by falling back to the previous raw-string overrides check. No existing JSON tags are changed.

---

| Component |
| MetaDB |
| Name registry json |
| Data File Descriptor Json |
| Callhome Json |
| Export Status Json |
| YugabyteD Tables |
| TargetDB Metadata Tables |
| Schema Dump |
| AssessmentDB |
| Import Data State |
| Data .sql files of tables |

### DIFF
--- a/yb-voyager/cmd/live_migration_cdc_partition_strategy.go
+++ b/yb-voyager/cmd/live_migration_cdc_partition_strategy.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	goerrors "github.com/go-errors/errors"
+	"github.com/samber/lo"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/metadb"
@@ -320,6 +321,9 @@ func getExpressionUniqueIndexTablesIfRequired(tableNames []sqlname.NameTuple, ov
 		if importDataStatus == nil {
 			return nil, goerrors.Errorf("import data status record not found")
 		}
+		if importDataStatus.CdcExpressionUniqueIndexTables == nil {
+			return nil, goerrors.Errorf("cdc expression unique index tables not found in import data status record")
+		}
 		for _, key := range importDataStatus.CdcExpressionUniqueIndexTables {
 			tuple, err := namereg.NameReg.LookupTableName(key)
 			if err != nil {
@@ -361,11 +365,11 @@ func validateCdcPartitioningStrategyUnchanged(tableNames []sqlname.NameTuple, im
 // whose effective strategy changed.
 func diffCdcPartitioningStrategy(resolved *utils.StructMap[sqlname.NameTuple, string], stored map[string]string) error {
 	var changed []string
-	var missing []string
+	var missingInStored []string
 	err := resolved.IterKV(func(t sqlname.NameTuple, strategy string) (bool, error) {
 		storedStrategy, ok := stored[t.ForKey()]
 		if !ok {
-			missing = append(missing, t.ForKey())
+			missingInStored = append(missingInStored, t.ForKey())
 		} else if storedStrategy != strategy {
 			changed = append(changed, fmt.Sprintf("%s (persisted: %q, new: %q)", t.ForOutput(), storedStrategy, strategy))
 		}
@@ -374,13 +378,32 @@ func diffCdcPartitioningStrategy(resolved *utils.StructMap[sqlname.NameTuple, st
 	if err != nil {
 		return err
 	}
-	if len(missing) > 0 {
-		sort.Strings(missing)
-		return goerrors.Errorf("cdc-partition-key-overrides: table %q is not in the import table list", strings.Join(missing, ", "))
+	var missingInResolved []string
+	storedTables := lo.Keys(stored)
+	for _, storedTable := range storedTables {
+		tuple, err := namereg.NameReg.LookupTableName(storedTable)
+		if err != nil {
+			return fmt.Errorf("error looking up table name: %w", err)
+		}
+		if _, ok := resolved.Get(tuple); !ok {
+			missingInResolved = append(missingInResolved, storedTable)
+		}
+	}
+	errorMsg := ""
+	if len(missingInStored) > 0 {
+		sort.Strings(missingInStored)
+		errorMsg += fmt.Sprintf("cdc-partition-key-overrides: table %q is in the current import table list but was not part of the original import; use --start-clean to start a fresh import with the new configuration\n", strings.Join(missingInStored, ", "))
+	}
+	if len(missingInResolved) > 0 {
+		sort.Strings(missingInResolved)
+		errorMsg += fmt.Sprintf("cdc-partition-key-overrides: table %q was part of the original import but is missing from the current import table list; use --start-clean to start a fresh import with the new configuration\n", strings.Join(missingInResolved, ", "))
 	}
 	if len(changed) > 0 {
 		sort.Strings(changed)
-		return goerrors.Errorf("changing cdc-partition-key / cdc-partition-key-overrides is not allowed after the import data has started; effective strategy changed for: %s.\nUse --start-clean to start a fresh import with the new configuration.", strings.Join(changed, "; "))
+		errorMsg += fmt.Sprintf("changing cdc-partition-key / cdc-partition-key-overrides is not allowed after the import data has started; effective strategy changed for: %s.\nUse --start-clean to start a fresh import with the new configuration.", strings.Join(changed, "; "))
+	}
+	if errorMsg != "" {
+		return goerrors.Errorf(errorMsg)
 	}
 	return nil
 }
--- a/yb-voyager/cmd/live_migration_cdc_partition_strategy.go
+++ b/yb-voyager/cmd/live_migration_cdc_partition_strategy.go
@@ -403,7 +403,7 @@ func diffCdcPartitioningStrategy(resolved *utils.StructMap[sqlname.NameTuple, st
 		errorMsg += fmt.Sprintf("changing cdc-partition-key / cdc-partition-key-overrides is not allowed after the import data has started; effective strategy changed for: %s.\nUse --start-clean to start a fresh import with the new configuration.", strings.Join(changed, "; "))
 	}
 	if errorMsg != "" {
-		return goerrors.Errorf(errorMsg)
+		return goerrors.Errorf("change in cdc-partition-key-overrides in between runs detected: %s", errorMsg)
 	}
 	return nil
 }
--- a/yb-voyager/cmd/live_migration_cdc_partition_strategy.go
+++ b/yb-voyager/cmd/live_migration_cdc_partition_strategy.go
@@ -65,7 +65,7 @@ func prepareCdcPartitionKey(tableNames []sqlname.NameTuple) error {
 		return nil
 	}
 
-	_, err = computeAndPersistCdcPartitioningStrategyPerTable(tableNames)
+	_, err = computeAndPersistCdcPartitioningStrategyPerTable(tableNames, importDataStatus)
 	return err
 }
 
@@ -243,44 +243,10 @@ func checkIfNeedsExprUKCheck(cdcPartitionKey string, overrides *utils.StructMap[
 
 // computeAndPersistCdcPartitioningStrategyPerTable resolves global + overrides + auto/expr-UK
 // rules and writes TableToCDCPartitioningStrategyMap to metaDB.
-func computeAndPersistCdcPartitioningStrategyPerTable(tableNames []sqlname.NameTuple) (*utils.StructMap[sqlname.NameTuple, string], error) {
-	rawOverrides, err := parseCdcPartitionKeyOverrides(cdcPartitionKeyOverrides)
+func computeAndPersistCdcPartitioningStrategyPerTable(tableNames []sqlname.NameTuple, importDataStatus *metadb.ImportDataStatusRecord) (*utils.StructMap[sqlname.NameTuple, string], error) {
+	tableToPartitioningStrategyMap, exprUKKeysForStorage, err := computeCdcPartitioningStrategyPerTable(tableNames, true, importDataStatus)
 	if err != nil {
-		return nil, err
-	}
-	overrides, err := resolveCdcPartitionKeyOverrides(rawOverrides, tableNames)
-	if err != nil {
-		return nil, err
-	}
-
-	needsExprUKCheck, err := checkIfNeedsExprUKCheck(cdcPartitionKey, overrides)
-	if err != nil {
-		return nil, err
-	}
-
-	var expressionUniqueIndexTables []sqlname.NameTuple
-	// nil unless the expression-UK check ran; a non-nil (possibly empty) slice records
-	// that we resolved the set so resume can reuse it without re-querying the target DB.
-	var exprUKKeysForStorage []string
-	if needsExprUKCheck {
-		expressionUniqueIndexTables, err = getExpressionUniqueIndexTables(tableNames)
-		if err != nil {
-			return nil, fmt.Errorf("error getting expression unique index tables: %w", err)
-		}
-		exprUKKeysForStorage = make([]string, 0, len(expressionUniqueIndexTables))
-		for _, t := range expressionUniqueIndexTables {
-			exprUKKeysForStorage = append(exprUKKeysForStorage, t.ForKey())
-		}
-	}
-	exprUKSet := utils.NewStructMap[sqlname.NameTuple, bool]()
-	for _, t := range expressionUniqueIndexTables {
-		exprUKSet.Put(t, true)
-	}
-
-	tableToPartitioningStrategyMap, err := resolveEffectiveCdcPartitionKeys(
-		tableNames, cdcPartitionKey, overrides, exprUKSet, tconf.TargetDBType)
-	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error computing cdc partitioning strategy per table: %w", err)
 	}
 
 	metadbMap := make(map[string]string)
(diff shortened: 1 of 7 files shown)


makalaaneesh:

Reviewed the 5 changed files (+370/-23) against base `main`, applying the root and `yb-voyager/cmd` / `yb-voyager/src/metadb` BUGBOT rules. This hardens the resume change-guard for `--cdc-partition-key-overrides` (PG→YB `import data to target`): the fragile raw-string comparison in `validateCdcPartitionKeyFlags` is replaced with a semantic, per-table effective-strategy diff in `prepareCdcPartitionKey` (`validateCdcPartitioningStrategyUnchanged` / `diffCdcPartitioningStrategy`), and the expression-UK table set is persisted in a new additive `ImportDataStatusRecord.CdcExpressionUniqueIndexTables` field so resume avoids re-querying the target.

Verified during review:
- **No false-accept of a real config change.** Any override change that alters an effective per-table strategy is reflected in the re-resolved map and caught by the diff; equal override strings early-return; differing strings with a nil expr-UK set conservatively fall back to the old raw-string reject; the global `--cdc-partition-key` is still string-compared in `validateCdcPartitionKeyFlags`, which runs before `prepareCdcPartitionKey`.
- **Ordering is safe** — `prepareCdcPartitionKey` runs before the config strings are persisted, so a rejected resume never overwrites `CdcPartitionKeyOverridesConfig`.
- **Backward-compat** — the new JSON field is additive; the nil-vs-empty distinction survives the JSON round-trip, and old records take the nil fallback. No existing tags changed.
- **Expr-UK reconstruction round-trips correctly** (leaf partitions map back to root tuples drawn from the import-table list), and `start-clean` nils the map then unconditionally rewrites the field, so there is no stale-cache issue. The resolution happens at prepare/resume, not on the per-event CDC hot path. `go build`, `go vet -tags unit ./cmd/`, and the relevant unit tests pass.

Non-blocking observations raised but not posted as findings: the nil-fallback conflates a genuine upgrade with a new-binary first run where the expr-UK check wasn't needed (still conservative, never false-accepts); a test-coverage gap on the equivalent-overrides path with a *non-empty* reconstructed expr-UK set and on the nil upgrade branch; and a couple of clarity/robustness nits (flatten `validateCdcPartitioningStrategyUnchanged`, prefer a positive streaming-progressed signal over the `time.Sleep` + absence assertion in the integration test).

---

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -284,6 +302,73 @@ func computeAndPersistCdcPartitioningStrategyPerTable(tableNames []sqlname.NameT
...
+// cdc-partition-key-overrides.
+func validateCdcPartitioningStrategyUnchanged(tableNames []sqlname.NameTuple, importDataStatus *metadb.ImportDataStatusRecord) error {
+
+	exprUKSet := utils.NewStructMap[sqlname.NameTuple, bool]()
+	storedOverrides := importDataStatus.CdcPartitionKeyOverridesConfig
+	if importDataStatus.CdcExpressionUniqueIndexTables == nil {
+		//In case the expression-UK tables were not captured on the first run, then it is only possible that its a resume with upgrade scenario so lets keep it simple
+		//so keeping the old string check for the overrides
+		if storedOverrides != cdcPartitionKeyOverrides {
+			return goerrors.Errorf("changing cdc-partition-key-overrides is not allowed after the import data has started. Current: %q, new: %q\n Use --start-clean to start a fresh import with the new overrides.", storedOverrides, cdcPartitionKeyOverrides)
```
makalaaneesh: if we fail here, then we need to make this a breaking release, right?

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -284,6 +302,73 @@ func computeAndPersistCdcPartitioningStrategyPerTable(tableNames []sqlname.NameT
...
+		}
+	}
+	if storedOverrides == cdcPartitionKeyOverrides {
```
makalaaneesh: do we now really need to store the config as well in metadb?

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -284,6 +302,73 @@ func computeAndPersistCdcPartitioningStrategyPerTable(tableNames []sqlname.NameT
...
+	overrides, err := resolveCdcPartitionKeyOverrides(rawOverrides, tableNames)
+	if err != nil {
+		return err
+	}
+
+	resolved, err := resolveEffectiveCdcPartitionKeys(tableNames, cdcPartitionKey, overrides, exprUKSet, tconf.TargetDBType)
+	if err != nil {
+		return err
+	}
+
```
makalaaneesh: this is essentially the compute part of the computeAndPersist, right? 
Can we restructure to reuse the same code?

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -284,6 +302,73 @@ func computeAndPersistCdcPartitioningStrategyPerTable(tableNames []sqlname.NameT
...
+// partitioning strategy from the current flags and fails if it differs from the map
+// persisted on the first run. It reuses the expression-UK tables captured on the first
+// run (falling back to a target-DB lookup only for older records that predate that
+// capture) so resume never silently applies a changed cdc-partition-key /
+// cdc-partition-key-overrides.
+func validateCdcPartitioningStrategyUnchanged(tableNames []sqlname.NameTuple, importDataStatus *metadb.ImportDataStatusRecord) error {
+
+	exprUKSet := utils.NewStructMap[sqlname.NameTuple, bool]()
+	storedOverrides := importDataStatus.CdcPartitionKeyOverridesConfig
+	if importDataStatus.CdcExpressionUniqueIndexTables == nil {
```
makalaaneesh: [AI] : 

1. The nil vs empty-slice ambiguity defeats the fix for --cdc-partition-key table. exprUKKeysForStorage is only assigned when needsExprUKCheck is true, so a first run with global table and no pk override persists nil — indistinguishable from
  a pre-upgrade record. Resume then falls back to exactly the raw string compare the PR removes. Confirmed:

  first run:  --cdc-partition-key table --cdc-partition-key-overrides "test_schema.orders:table;test_schema.events:table"
  resume:     ...same two entries, reordered
  result:     "changing cdc-partition-key-overrides is not allowed after the import data has started."

  Fix is one line — var exprUKKeysForStorage = []string{} — since [] unmarshals to a non-nil empty slice, preserving nil as the genuine "older voyager" marker. The comment in importDataStatus.go:40-46 currently documents the ambiguity as if it
  were intended; it should be tightened along with the fix.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
makalaaneesh: [AI] 

2. The storedOverrides == cdcPartitionKeyOverrides → return nil fast path skips the diff, so a changed global key slips through when ImportDataStarted is false. validateCdcPartitionKeyFlags returns early when !ImportDataStarted, and
  prepareCdcPartitionKey (importData.go:1241) persists the map five lines before updateImportDataStartedAndSomeConfigsInMetaDB (:1246). If the process dies in that window, a rerun with a different --cdc-partition-key passes both guards: the
  stale map drives streaming while CdcPartitioningStrategyConfig gets overwritten with the new key, leaving metaDB self-inconsistent. Confirmed — validateCdcPartitioningStrategyUnchanged returns nil for a persisted all-pk map re-run with
  --cdc-partition-key table. Dropping the early return (it saves only a few namereg lookups) makes the guard independent of ImportDataStarted:

  if importDataStatus.CdcExpressionUniqueIndexTables == nil {
      if storedOverrides != cdcPartitionKeyOverrides { return goerrors.Errorf(...) }
      return nil   // legacy record: nothing better available
  }
  // always diff, regardless of whether the overrides string changed

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -284,6 +302,73 @@ func computeAndPersistCdcPartitioningStrategyPerTable(tableNames []sqlname.NameT
...
+
+	return diffCdcPartitioningStrategy(resolved, importDataStatus.TableToCDCPartitioningStrategyMap)
+}
+
+// diffCdcPartitioningStrategy compares a freshly-resolved per-table strategy map against
+// the one persisted on the first run and returns an actionable error listing the tables
+// whose effective strategy changed.
+func diffCdcPartitioningStrategy(resolved *utils.StructMap[sqlname.NameTuple, string], stored map[string]string) error {
+	var changed []string
+	err := resolved.IterKV(func(t sqlname.NameTuple, strategy string) (bool, error) {
```
makalaaneesh, line 369: maybe this isn't a real problem, but for completeness of this function, shouldn't we also be checking if there are extra in stored as compared to resolved?

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
priyanshi-yb: No, so two things, right - breaking release where the upgrade happens in between commands export/import, and the user has to start the migration from scratch, and the other one is the command itself is not resumable so it requires start-clean. This is the second case where overrides are changing and require start-clean, as suggested in the error msg

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -284,6 +302,73 @@ func computeAndPersistCdcPartitioningStrategyPerTable(tableNames []sqlname.NameT
...
+// validateCdcPartitioningStrategyUnchanged re-resolves the effective per-table CDC
+// partitioning strategy from the current flags and fails if it differs from the map
+// persisted on the first run. It reuses the expression-UK tables captured on the first
+// run (falling back to a target-DB lookup only for older records that predate that
+// capture) so resume never silently applies a changed cdc-partition-key /
+// cdc-partition-key-overrides.
+func validateCdcPartitioningStrategyUnchanged(tableNames []sqlname.NameTuple, importDataStatus *metadb.ImportDataStatusRecord) error {
+
+	exprUKSet := utils.NewStructMap[sqlname.NameTuple, bool]()
+	storedOverrides := importDataStatus.CdcPartitionKeyOverridesConfig
```
makalaaneesh: nit: cdcPartitionKeyOverridesStored

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -284,6 +302,73 @@ func computeAndPersistCdcPartitioningStrategyPerTable(tableNames []sqlname.NameT
 	return tableToPartitioningStrategyMap, nil
 }
 
+// validateCdcPartitioningStrategyUnchanged re-resolves the effective per-table CDC
+// partitioning strategy from the current flags and fails if it differs from the map
+// persisted on the first run. It reuses the expression-UK tables captured on the first
+// run (falling back to a target-DB lookup only for older records that predate that
+// capture) so resume never silently applies a changed cdc-partition-key /
+// cdc-partition-key-overrides.
+func validateCdcPartitioningStrategyUnchanged(tableNames []sqlname.NameTuple, importDataStatus *metadb.ImportDataStatusRecord) error {
```
makalaaneesh, line 348: 1. this func expects that certain fields in MSR will be populated. If that is not the case, we should error out. 
2. let's mark it a breaking release.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
priyanshi-yb: Marking next release as breaking as discussed and removing this check.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
priyanshi-yb: Not really, then, as now we are doing parsing and populating the map per table but just for some cases where we don't need to do any further checks if its as is, keeping it.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
makalaaneesh: the case "extra in stored" still does not seem to be handled.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -233,55 +243,146 @@ func checkIfNeedsExprUKCheck(cdcPartitionKey string, overrides *utils.StructMap[
...
 		if err != nil {
 			return nil, fmt.Errorf("error getting expression unique index tables: %w", err)
 		}
+	} else {
+		// Resume: reuse the set captured on the first run (persisted in metaDB) so we do not
+		// re-query the target DB.
+		if importDataStatus == nil {
+			return nil, goerrors.Errorf("import data status record not found")
+		}
+		for _, key := range importDataStatus.CdcExpressionUniqueIndexTables {
```
makalaaneesh, line 327: we should also check if `importDataStatus.CdcExpressionUniqueIndexTables` is not nil

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -233,55 +243,146 @@ func checkIfNeedsExprUKCheck(cdcPartitionKey string, overrides *utils.StructMap[
...
 	var expressionUniqueIndexTables []sqlname.NameTuple
-	if needsExprUKCheck {
+	if isFirstRun {
+		// First run: query the target DB for the authoritative set of expression-UK tables.
 		expressionUniqueIndexTables, err = getExpressionUniqueIndexTables(tableNames)
 		if err != nil {
 			return nil, fmt.Errorf("error getting expression unique index tables: %w", err)
 		}
+	} else {
+		// Resume: reuse the set captured on the first run (persisted in metaDB) so we do not
```
makalaaneesh, line 319: can you elaborate here -- as to why it's a bad idea to re-query? (it's not just an optimization if i remember correctly from our previous discussions)

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
priyanshi-yb: Yeah, I only fixed the return error for missing in stored from resolved

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
priyanshi-yb: schema drift is one case because of which we can't re-query db to fetch new expression indexes in resume case as its will change the partition strategy resolution and in resume we can't allow that.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
priyanshi-yb: will add but I have marked this release as breaking so this is anyways not possible

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
makalaaneesh: right, yeah. I was just asking to add it into the comment here, so that we don't forget about this case later.

makalaaneesh (approved):
LGTM with one nit about a comment

PR #3707, opened 2026-08-04
Add a POC script for PostgreSQL to YugabyteDB AMP migration assessment

makalaaneesh:

AMP's compute is PostgreSQL, so voyager's YugabyteDB compatibility analysis — unsupported datatypes, PL/pgSQL objects, query constructs, colocation and sharding — does not apply to a PG → AMP migration. This adds a small standalone script that reports only the two things that do matter for that path: complexity `LOW`, and a vCPU recommendation matched to the source database's own capacity.

The script writes the same control-plane rows `assess-migration` writes (phase 1, `IN PROGRESS` at invocation_sequence *n* then `COMPLETED` at *n+1*), carrying a payload shaped like `AssessMigrationPayloadYugabyteD`, so the AMP controller can read it without changes.

Source vCPU is detected in two tiers. `exact` reads the host's real limits through `pg_read_file`, taking the smaller of `Cpus_allowed_list` and the cgroup `cpu.max` quota — both are needed, because `/proc` is host-scoped inside a container and a cpuset restriction leaves the quota reading `max`. `high` inverts a vCPU-derived GUC: PostgreSQL ships `max_worker_processes` and `max_parallel_workers` at exactly 8, so a larger value means the platform sized them from the machine via `GREATEST(vCPU*2,8)` / `GREATEST(vCPU/2,8)`. Neither tier yielding a number reports `-1` and lets the caller decide, rather than inventing one. `MemoryPerInstance` is a fixed 16 — AMP provisioning takes vCPU only.

This is deliberately throwaway scaffolding for the AMP POC. It lives in `yb-voyager/scripts/`, outside the installed tree, and is not wired into any command or the build. The real implementation belongs inside `assess-migration --target-db-type yugabytedb-amp`.

No user-facing changes. No new commands or flags, no configuration changes, no change to the installation process, and no change to any report voyager itself produces. Nothing in `cmd/` or `src/` is touched — the PR is purely additive, five new files under a new `yb-voyager/scripts/amp/` directory.

Manual testing only; there are no automated tests, and no existing test tier covers standalone shell scripts. Verified end to end against a local PostgreSQL 17 instance, with a second local database standing in for the control plane (the bootstrap DDL is plain PostgreSQL):

- **No signal → `-1`**: reports `VCPUsPerInstance: -1`, sets `Sizing.FailureReasoning`, still writes the report.
- **`high` tier**: forced `max_parallel_workers=16` via `PGOPTIONS` → 48 vCPU at `max_parallel_workers*2` with `Confidence: high`.
- **Control-plane writes**: bootstrap, then `IN PROGRESS` at seq 1 with an empty payload and `COMPLETED` at seq 2 with the report JSON. A re-run correctly lands at seq 3 and 4 rather than colliding on the primary key. The resulting table is a 15-column match for voyager's effective shape, and `host_ip` / `voyager_info` carry the same JSON-in-a-column format voyager writes.

**Not verified:** the `exact` tier never executed — the dev machine is macOS with no `/proc`, and no Linux container was available. Its two halves were validated separately instead: the SQL regexes against synthetic `/proc/self/status`, `/proc/meminfo`, `/proc/cpuinfo` and `/proc/self/cgroup` content, and the bash cpuset/quota arithmetic against synthetic values (`0-3,8` → 5, `0-3,8,10-11` → 7, `150000 100000` → 2). **This path still needs one run against a containerised PostgreSQL on Linux before the POC is demoed.**

No changes to voyager's payloads. `ASSESS_MIGRATION_YBD_PAYLOAD_VERSION` and the `AssessMigrationPayloadYugabyteD` struct are untouched.

The script does *write* a yugabyted-shaped payload from outside voyager. It stamps `PayloadVersion` as `1.8-amp.1` — deliberately distinguishable from a genuine `1.8` — and adds one field voyager does not emit, `SourceEnvironment` (`VCPUs`, `MemoryGiB`, `DetectionMethod`, `Confidence`), so a reader can tell how the number was arrived at. `SchemaSummary`, `ParsedSchemaSummary` and `AssessmentIssues` are present but empty; `SourceSizeDetails` carries `TotalDBSize` only.

No schema changes. The script touches **YugabyteD Tables** and **Migration Assessment Report Json** from the reference table, but in both cases it conforms to the existing shape rather than altering it:

- `amp-controlplane-bootstrap.psql` is copied verbatim from `setupDatabase()` in `src/cp/yugabyted/yugabyted.go`, including the original 11-column `CREATE TABLE` followed by the 7 `ALTER`s, so a database bootstrapped by this script and one bootstrapped by voyager end up identical in either order. All statements are `IF NOT EXISTS`.
- The rows written match voyager's column set and value conventions exactly.

### DIFF
--- a/yb-voyager/scripts/amp/yb-voyager-amp-assess-migration.sh
+++ b/yb-voyager/scripts/amp/yb-voyager-amp-assess-migration.sh
@@ -1,0 +1,307 @@
+#!/bin/bash
+#   Copyright (c) YugabyteDB, Inc.
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#	    http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+# POC: a minimal migration assessment for a PostgreSQL -> YugabyteDB AMP
+# migration. AMP's compute is PostgreSQL, so none of voyager's YugabyteDB
+# compatibility analysis applies; the assessment reports LOW complexity and a
+# vCPU recommendation matched to the source's own capacity.
+#
+# Writes the same control-plane rows voyager's yugabyted control plane writes
+# for `assess-migration`, with a payload shaped like AssessMigrationPayloadYugabyteD,
+# so the AMP controller can read it without changes.
+#
+# This is throwaway scaffolding. The real implementation belongs inside
+# `yb-voyager assess-migration --target-db-type yugabytedb-amp`.
+
+set -e
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE:-$0}" )" &> /dev/null && pwd )
+SCRIPT_NAME=$(basename $0)
+
+PAYLOAD_VERSION="1.8-amp.1"
+VOYAGER_VERSION="amp-assess-poc"
+RECOMMENDED_MEMORY_GB=16   # fixed for now; AMP provisioning takes vCPU only
+
+HELP_TEXT="
+Usage: $SCRIPT_NAME <pg_connection_string> <schema_list> <output_dir>
+
+Arguments:
+  pg_connection_string   Source PostgreSQL connection string, e.g.
+                         'postgresql://username@hostname:port/dbname'
+                         Quote it to avoid shell interpretation.
+
+  schema_list            Pipe-separated schema list, e.g. 'public|sales'.
+                         Recorded on the control-plane row; not otherwise used.
+
+  output_dir             Directory for the local copy of the assessment report.
+
+Environment:
+  PGPASSWORD                  Source database password.
+  YB_VOYAGER_MIGRATION_UUID   Required. The migration UUID the control-plane
+                              rows are keyed by (AMP's migration.id).
+  CONTROL_PLANE_TYPE          'yugabyted' to write control-plane rows. Anything
+                              else (or unset) writes only the local report.
+  YUGABYTED_DB_CONN_STRING    Required when CONTROL_PLANE_TYPE=yugabyted.
+
+Example:
+  PGPASSWORD=secret \\
+  YB_VOYAGER_MIGRATION_UUID=90b8b0b8-1111-2222-3333-444455556666 \\
+  CONTROL_PLANE_TYPE=yugabyted \\
+  YUGABYTED_DB_CONN_STRING='postgresql://yugabyte:yugabyte@127.0.0.1:5433' \\
+  $SCRIPT_NAME 'postgresql://user@dbhost:5432/prod' 'public' /tmp/amp-assessment
+"
+
+if [ "$1" = "-h" ] || [ "$1" = "--help" ] || [ $# -lt 3 ]; then
+    echo "$HELP_TEXT"
+    [ $# -lt 3 ] && exit 1
+    exit 0
+fi
+
+SOURCE_CONN="$1"
+SCHEMA_LIST="$2"
+OUTPUT_DIR="$3"
+
+if [ -z "$YB_VOYAGER_MIGRATION_UUID" ]; then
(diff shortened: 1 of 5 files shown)


yb-voyager/scripts/amp/yb-voyager-amp-assess-migration.sh
```diff
@@ -0,0 +1,307 @@
...
+# Mirrors voyager: the started row is written before any assessment work, and
+# the completed row gets the next invocation_sequence rather than updating it.
+if [ "$WRITE_CONTROL_PLANE" = true ]; then
+    echo "Bootstrapping control-plane schema..."
+    cp_ -v ON_ERROR_STOP=on -f "$SCRIPT_DIR/amp-controlplane-bootstrap.psql" > /dev/null
+
+    SEQ=$(cp_ -v ON_ERROR_STOP=on -c "
+        SELECT coalesce(MAX(invocation_sequence), 0) + 1
+          FROM ybvoyager_visualizer.ybvoyager_visualizer_metadata
+         WHERE migration_uuid = '$YB_VOYAGER_MIGRATION_UUID' AND migration_phase = 1;")
```
makalaaneesh, line 118: [AI] **SEQ query interpolates `YB_VOYAGER_MIGRATION_UUID` directly into SQL (injection / quote-breakage).**

The invocation-sequence lookup builds its SQL by shell-interpolating the environment value directly into a `psql -c` string:

```sh
SEQ=$(cp_ -v ON_ERROR_STOP=on -c "
    SELECT coalesce(MAX(invocation_sequence), 0) + 1
      FROM ybvoyager_visualizer.ybvoyager_visualizer_metadata
     WHERE migration_uuid = '$YB_VOYAGER_MIGRATION_UUID' AND migration_phase = 1;")
```

Every other statement in this script deliberately uses psql's `-v name=... :'name'` binding (the file even documents *"psql interpolates `:'var'` only for script input"*) precisely so untrusted values are safely quoted. This one query is the exception.

**Why it matters:** `YB_VOYAGER_MIGRATION_UUID` is caller-supplied. A value containing a single quote breaks the query, and because `psql -c` executes multiple semicolon-separated statements, a crafted value such as `'; DROP TABLE ybvoyager_visualizer.ybvoyager_visualizer_metadata; --` yields arbitrary SQL execution against the control-plane database. Even absent a malicious actor, a malformed UUID silently aborts the run. In the intended AMP deployment the UUID is a trusted `migration.id`, which is why this is a Warning rather than Critical, but it is a real, trivially avoidable defect.

**Suggested fix** — use the same binding pattern as the rest of the script, feeding the statement on stdin so `:'uuid'` interpolation applies:

```sh
SEQ=$(cp_ -v ON_ERROR_STOP=on -v uuid="$YB_VOYAGER_MIGRATION_UUID" <<'SQL'
SELECT coalesce(MAX(invocation_sequence), 0) + 1
  FROM ybvoyager_visualizer.ybvoyager_visualizer_metadata
 WHERE migration_uuid = :'uuid' AND migration_phase = 1;
SQL
)
```

---

makalaaneesh (commented):
[AI] Automated review — 1 finding (0 critical, 1 warning). See the inline comment.

This PR adds a self-contained, throwaway POC under `yb-voyager/scripts/amp/` (five new files, not wired into `cmd/`/`src/` or the build) that produces a PostgreSQL → YugabyteDB AMP migration assessment: `LOW` complexity plus a source-matched vCPU sizing recommendation, written as the same yugabyted control-plane rows voyager writes. Cross-checks passed: the payload key names match the `AssessMigrationPayloadYugabyteD` / `SizingRecommendation` Go structs, `migration_phase=1` matches the phase map, the bootstrap DDL is byte-for-byte with `setupDatabase()` in `src/cp/yugabyted/yugabyted.go`, and the vCPU arithmetic (`max_worker_processes/2`, `max_parallel_workers*2`, CFS `ceil(quota/period)`, cpuset counting) and the `/proc`+cgroup regexes are correct.

The one warning is a SQL-injection / quote-breakage risk in the invocation-sequence lookup, which shell-interpolates the caller-supplied `YB_VOYAGER_MIGRATION_UUID` directly into a `psql -c` string instead of using the `:'var'` binding the rest of the script uses. A few non-blocking observations were raised but not posted: an empty-`df` case that can emit invalid JSON in the STARTED row's `voyager_info`, a CFS-quota tier that's dropped when the cpu-list count is 0, the `high`-tier GUC heuristic misfiring on hand-tuned GUCs, and confirming the AMP controller tolerates the empty-string payload on the IN PROGRESS row.

---

makalaaneesh:
CLosing this as this has been implemented on AMP side by @AmeyaKasture

PR #3709, opened 2026-08-05
Support custom partition key for cdc streaming of import data in cdc-partition-key-overrides

priyanshi-yb:

This PR adds configurable CDC event partitioning for `import data to target` in live migration, so users can control how change events are hashed across the parallel import channels. Two flags are introduced: a global `--cdc-partition-key` (`auto` | `pk` | `table`) and a per-table `--cdc-partition-key-overrides` that accepts `pk`, `table`, or a **custom key column list** `(col1,col2)` per table. Previously partitioning was effectively fixed; now a table can be routed by its primary key, serialized onto a single channel, or hashed by an arbitrary set of columns.

The effective per-table strategy is resolved and validated up-front (before the snapshot phase) rather than mid-stream: `auto` picks `pk` for normal tables and forces `table` for tables with expression-based unique indexes; `pk` is rejected for expression-UK tables; overrides are validated against the import table list and de-duplicated on the resolved `NameTuple`. The resolved map is persisted in the MetaDB `ImportDataStatusRecord`, and a resume change-guard re-resolves the current flags and fails (unless `--start-clean`) if the effective strategy or custom key columns for any table changed between runs.

Routing is unified behind `GetEventPartitionKey`/`hashEvent`: the partition key is the PK values under `pk`, the table name under `table`, and the ordered custom column values under `custom` (custom key columns are required to be immutable). Conflict detection is generalized with a single routing-key exclusion in `ConflictDetectionCache`: two events that share the same partition key are guaranteed to land on the same channel and apply in commit order, so they can never be a cross-channel unique-key conflict and are excluded from conflict detection. This replaces the earlier special-case same-PK handling and lets custom-key tables skip conflict detection for same-key events.

The in-memory (`cdcPartitionKeyOverride`) and on-disk (`metadb.CDCPartitionKey`) representations were consolidated into a single struct (strategy + columns), replacing the previous separate strategy/columns maps. Custom key support is shipped as **Tech Preview**.

Yes — command-line only:
- New flag `--cdc-partition-key` on `import data` / `import data to target` (default `auto`; values `auto`, `pk`, `table`).
- New flag `--cdc-partition-key-overrides` for per-table overrides, e.g.`public.orders:table;sales.events:pk;public.ledger:(account_id)`. Unlisted tables keep the global value. Both flags are also recognized by the config-file layer.
- Guardrails: the flags are only valid for `import data to target`, a PostgreSQL source, and live (snapshot-and-changes / changes-only) migrations; changing them on resume is rejected unless `--start-clean` is used.
- Using a custom key column list prints a `[Tech Preview]` warning.

No changes to installation or reports.

- **Unit tests** (`yb-voyager/cmd/live_migration_test.go`): override parsing including the parenthesized custom-key syntax (`TestParseCdcPartitionKeyOverrides`), effective-strateg resolution (`TestResolveEffectiveCdcPartitionKeys`), override resolution/validation (`TestResolveCdcPartitionKeyOverrides`), resume change-guards (`TestValidateCdcPartitioningStrategyUnchanged`, `TestValidateCdcPartitionKeyFlags`), and event routing (`TestHashEventCustomKey`, `TestGetEventPartitionKey`). Conflict-cache behavior updated in `conflictDetectionCache_test.go`.
- **Integration (live migration) tests** (`yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go`): an end-to-end migration mixing strategies via `--cdc-partition-key-overrides`
  (`TestLiveMigrationWithCustomCdcPartitionKey`), and a dedicated test that drives a partial
  unique index on the custom-key column and asserts, via the `uniqueKeyConflictDetected`
  count failpoint, that zero conflicts are detected because same-custom-key events co-locate
  on one channel (`TestLiveMigrationCustomCdcPartitionKeyNoConflict`).
- Ran `go test -tags unit ./cmd/...` (custom-key suites pass) and verified the live-migration
  test package builds under `-tags 'integration_live_migration failpoint_import'`. The live
  E2E tests require PG + YugabyteDB containers and a failpoint-enabled binary.

No.

Yes — the MetaDB `ImportDataStatusRecord` (MetaDB / Import Data State). The per-table map
`TableToCDCPartitioningStrategyMap` (`map[string]string`, JSON tag
`tableToCDCPartitioningStrategyMap`) is replaced by `TableToCDCPartitionKey`
(`map[string]metadb.CDCPartitionKey`, JSON tag `tableToCDCPartitionKey`). This is a
non-additive JSON-tag change; the ongoing release is marked breaking, so a mid-migration
downgrade is not supported for import-data state. New config fields
(`CdcPartitioningStrategyConfig`, `CdcPartitionKeyOverridesConfig`,
`CdcExpressionUniqueIndexTables`) are additive.

---

| Component        

### DIFF
--- a/yb-voyager/cmd/conflictDetectionCache.go
+++ b/yb-voyager/cmd/conflictDetectionCache.go
@@ -131,6 +131,13 @@ type ConflictDetectionCache struct {
 	evChans              []chan *tgtdb.Event
 	sourceDBType         string
 
+	// Partitioning strategy + custom key columns per table, used to compute an event's
+	// partition key (see GetEventPartitionKey). Two events with the same partition key are
+	// routed to the same channel and applied in commit order, so they are excluded from
+	// conflict detection (they can never race).
+	tableToPartitioningStrategyMap *utils.StructMap[sqlname.NameTuple, string]
+	tableToCustomKeyColumnsMap     *utils.StructMap[sqlname.NameTuple, []string]
+
 	/*
 		ukLookup is a value-keyed secondary index over the cached events used to avoid
 		scanning the entire cache on every incoming event (see WaitUntilNoConflict).
@@ -148,7 +155,7 @@ type ConflictDetectionCache struct {
 	vsnToBuckets map[int64][]string
 }
 
-func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.NameTuple, []tgtdb.UniqueIndex], evChans []chan *tgtdb.Event, sourceDBType string) *ConflictDetectionCache {
+func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.NameTuple, []tgtdb.UniqueIndex], evChans []chan *tgtdb.Event, sourceDBType string, tableToPartitioningStrategyMap *utils.StructMap[sqlname.NameTuple, string], tableToCustomKeyColumnsMap *utils.StructMap[sqlname.NameTuple, []string]) *ConflictDetectionCache {
 	c := &ConflictDetectionCache{}
 	c.m = make(map[int64]*tgtdb.Event)
 	c.cond = sync.NewCond(&c.Mutex)
@@ -157,6 +164,8 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
 	c.evChans = evChans
 	c.ukLookup = make(map[string]map[int64]*tgtdb.Event)
 	c.vsnToBuckets = make(map[int64][]string)
+	c.tableToPartitioningStrategyMap = tableToPartitioningStrategyMap
+	c.tableToCustomKeyColumnsMap = tableToCustomKeyColumnsMap
 	return c
 }
 
@@ -328,12 +337,14 @@ func (c *ConflictDetectionCache) findConflictLocked(incomingEvent *tgtdb.Event) 
 //
 // A cached event conflicts when, for some unique index, its BeforeFields equal the
 // incoming event's Fields (before-after) or BeforeFields (before-before) AND the two
-// events have different PKs. The exact value equality (including column existence and
-// NULLS [NOT] DISTINCT semantics) is already guaranteed by the bucket-key match built
-// in computeConflictBucketKey, so - unlike the full-scan target path - this does not
-// re-run eventsConfict. The only residual filter is the same-PK exclusion: same-PK
-// events are routed to the same channel and applied in order, so they are never
-// cross-channel conflicts.
+// events have different partition keys. The exact value equality (including column
+// existence and NULLS [NOT] DISTINCT semantics) is already guaranteed by the bucket-key
+// match built in computeConflictBucketKey, so - unlike the full-scan target path - this
+// does not re-run eventsConfict. The only residual filter is the partition-key exclusion
+// (getEventsWithDifferentPartitionKey): events that share the incoming event's partition
+// key are routed to the same channel and applied in order, so they are never
+// cross-channel conflicts. The partition key is the PK under PARTITION_BY_PK and the
+// custom key columns under PARTITION_BY_CUSTOM.
 //
 // All unique indexes and both checks (before-after, before-before) are evaluated so the
 // incoming event can wait on the complete set of conflicting cached events at once. The
@@ -351,7 +362,10 @@ func (c *ConflictDetectionCache) findValueConflictLocked(incomingEvent *tgtdb.Ev
 			return []Conflict{}, err
 		}
 		if ok {
-			cachedEvents := c.getNonSamePKEventsWithSameBucketKey(key, incomingEvent)
+			cachedEvents, err := c.getEventsWithDifferentPartitionKey(key, incomingEvent)
+			if err != nil {
+				return []Conflict{}, err
+			}
 			if len(cachedEvents) > 0 {
 				for _, cachedEvent := range cachedEvents {
 					log.Debugf("conflict detected for table %s, index columns %v, between before value of cached-event1(vsn=%d, colVal=%s) and after value of incoming-event2(vsn=%d, colVal=%s)",
@@ -378,7 +392,10 @@ func (c *ConflictDetectionCache) findValueConflictLocked(incomingEvent *tgtdb.Ev
 			return []Conflict{}, err
 		}
 		if ok {
-			cachedEvents := c.getNonSamePKEventsWithSameBucketKey(key, incomingEvent)
+			cachedEvents, err := c.getEventsWithDifferentPartitionKey(key, incomingEvent)
+			if err != nil {
+				return []Conflict{}, err
+			}
 			if len(cachedEvents) > 0 {
 				for _, cachedEvent := range cachedEvents {
 					log.Debugf("conflict detected for table %s, index columns %v, between before value of cached-event1(vsn=%d, colVal=%s) and before value of incoming-event2(vsn=%d, colVal=%s)",
@@ -399,19 +416,30 @@ func (c *ConflictDetectionCache) findValueConflictLocked(incomingEvent *tgtdb.Ev
 	return totalConflictInfo, nil
 }
 
-// getNonSamePKEventsWithSameBucketKey returns all cached events in the given lookup
-// bucket whose PK differs from the incoming event's PK (nil if none). Same-PK events
-// are excluded because they are routed to the same channel and applied in order.
-// Caller must hold the lock.
-func (c *ConflictDetectionCache) getNonSamePKEventsWithSameBucketKey(key string, incomingEvent *tgtdb.Event) []*tgtdb.Event {
+// getEventsWithDifferentPartitionKey returns all cached events in the given lookup bucket
+// whose partition key differs from the incoming event's (nil if none). Events sharing the
+// incoming event's partition key are excluded because they route to the same channel (see
+// GetEventPartitionKey / hashEvent) and are therefore applied in commit order - they can
+// never be a cross-channel conflict. This generalizes the same-PK exclusion: the partition
+// key is the primary key under PARTITION_BY_PK and the custom key columns under
+// PARTITION_BY_CUSTOM. Caller must hold the lock.
+func (c *ConflictDetectionCache) getEventsWithDifferentPartitionKey(key string, incomingEvent *tgtdb.Event) ([]*tgtdb.Event, error) {
+	incomingPartitionKey, err := GetEventPartitionKey(incomingEvent, c.tableToPartitioningStrategyMap, c.tableToCustomKeyColumnsMap)
+	if err != nil {
+		return nil, goerrors.Errorf("error computing partition key for incoming event(vsn=%d): %v", incomingEvent.Vsn, err)
+	}
 	var cachedEventsMatchingKey []*tgtdb.Event
 	for _, cachedEvent := range c.ukLookup[key] {
-		if c.sameTableEventsHaveSamePK(cachedEvent, incomingEvent) {
+		cachedPartitionKey, err := GetEventPartitionKey(cachedEvent, c.tableToPartitioningStrategyMap, c.tableToCustomKeyColumnsMap)
+		if err != nil {
+			return nil, goerrors.Errorf("error computing partition key for cached event(vsn=%d): %v", cachedEvent.Vsn, err)
+		}
+		if cachedPartitionKey == incomingPartitionKey {
 			continue
 		}
 		cachedEventsMatchingKey = append(cachedEventsMatchingKey, cachedEvent)
 	}
-	return cachedEventsMatchingKey
+	return cachedEventsMatchingKey, nil
 }
 
 const (
--- a/yb-voyager/cmd/conflictDetectionCache.go
+++ b/yb-voyager/cmd/conflictDetectionCache.go
@@ -131,12 +131,11 @@ type ConflictDetectionCache struct {
 	evChans              []chan *tgtdb.Event
 	sourceDBType         string
 
-	// Partitioning strategy + custom key columns per table, used to compute an event's
-	// partition key (see GetEventPartitionKey). Two events with the same partition key are
-	// routed to the same channel and applied in commit order, so they are excluded from
-	// conflict detection (they can never race).
-	tableToPartitioningStrategyMap *utils.StructMap[sqlname.NameTuple, string]
-	tableToCustomKeyColumnsMap     *utils.StructMap[sqlname.NameTuple, []string]
+	// Per-table CDC partition key (strategy + custom key columns), used to compute an
+	// event's partition key (see GetEventPartitionKey). Two events with the same partition
+	// key are routed to the same channel and applied in commit order, so they are excluded
+	// from conflict detection (they can never race).
+	tablePartitionKeyMap *utils.StructMap[sqlname.NameTuple, cdcPartitionKeyOverride]
 
 	/*
 		ukLookup is a value-keyed secondary index over the cached events used to avoid
@@ -155,7 +154,7 @@ type ConflictDetectionCache struct {
 	vsnToBuckets map[int64][]string
 }
 
-func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.NameTuple, []tgtdb.UniqueIndex], evChans []chan *tgtdb.Event, sourceDBType string, tableToPartitioningStrategyMap *utils.StructMap[sqlname.NameTuple, string], tableToCustomKeyColumnsMap *utils.StructMap[sqlname.NameTuple, []string]) *ConflictDetectionCache {
+func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.NameTuple, []tgtdb.UniqueIndex], evChans []chan *tgtdb.Event, sourceDBType string, tablePartitionKeyMap *utils.StructMap[sqlname.NameTuple, cdcPartitionKeyOverride]) *ConflictDetectionCache {
 	c := &ConflictDetectionCache{}
 	c.m = make(map[int64]*tgtdb.Event)
 	c.cond = sync.NewCond(&c.Mutex)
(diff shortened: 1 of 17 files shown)


yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -131,6 +131,13 @@ type ConflictDetectionCache struct {
 	evChans              []chan *tgtdb.Event
 	sourceDBType         string
 
+	// Partitioning strategy + custom key columns per table, used to compute an event's
+	// partition key (see GetEventPartitionKey). Two events with the same partition key are
+	// routed to the same channel and applied in commit order, so they are excluded from
+	// conflict detection (they can never race).
+	tableToPartitioningStrategyMap *utils.StructMap[sqlname.NameTuple, string]
+	tableToCustomKeyColumnsMap     *utils.StructMap[sqlname.NameTuple, []string]
```
makalaaneesh: can we create a struct to represent 
- strategy
- columns 
?

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -399,19 +416,30 @@ func (c *ConflictDetectionCache) findValueConflictLocked(incomingEvent *tgtdb.Ev
...
-// Caller must hold the lock.
-func (c *ConflictDetectionCache) getNonSamePKEventsWithSameBucketKey(key string, incomingEvent *tgtdb.Event) []*tgtdb.Event {
+// getEventsWithDifferentPartitionKey returns all cached events in the given lookup bucket
+// whose partition key differs from the incoming event's (nil if none). Events sharing the
+// incoming event's partition key are excluded because they route to the same channel (see
+// GetEventPartitionKey / hashEvent) and are therefore applied in commit order - they can
+// never be a cross-channel conflict. This generalizes the same-PK exclusion: the partition
+// key is the primary key under PARTITION_BY_PK and the custom key columns under
+// PARTITION_BY_CUSTOM. Caller must hold the lock.
+func (c *ConflictDetectionCache) getEventsWithDifferentPartitionKey(key string, incomingEvent *tgtdb.Event) ([]*tgtdb.Event, error) {
```
makalaaneesh: `getConflictingEventsWithDifferentPartitionKey` ?

yb-voyager/cmd/live_migration.go
```diff
@@ -403,41 +407,111 @@ func handleEvent(event *tgtdb.Event,
...
+	hash := fnv.New64a()
+	hash.Write([]byte(partitionKey))
 	return int(hash.Sum64() % (uint64(NUM_EVENT_CHANNELS))), nil
 }
 
+// customKeyColumnValue returns the value of a custom partition key column for the event,
+// preferring BeforeFields (full row image for update/delete under REPLICA IDENTITY FULL)
+// and falling back to Fields (inserts). The second return is false if the column is
+// present in neither map.
+func customKeyColumnValue(e *tgtdb.Event, col string) (*string, bool) {
```
makalaaneesh: let's name this func more explicitly: `customPartitionKeyColumnValue`

yb-voyager/cmd/live_migration.go
```diff
@@ -403,41 +407,111 @@ func handleEvent(event *tgtdb.Event,
...
+		return 0, err
+	}
+	hash := fnv.New64a()
+	hash.Write([]byte(partitionKey))
 	return int(hash.Sum64() % (uint64(NUM_EVENT_CHANNELS))), nil
 }
 
+// customKeyColumnValue returns the value of a custom partition key column for the event,
+// preferring BeforeFields (full row image for update/delete under REPLICA IDENTITY FULL)
+// and falling back to Fields (inserts). The second return is false if the column is
```
makalaaneesh: rather than falling back, we should check event type (insert/update/delete) and then choose the fields accordingly

yb-voyager/cmd/live_migration.go
```diff
@@ -403,41 +407,111 @@ func handleEvent(event *tgtdb.Event,
...
+}
+
+// Returns a hash value between 0..NUM_EVENT_CHANNELS
+func hashEvent(e *tgtdb.Event, tableToPartitioningStrategyMap *utils.StructMap[sqlname.NameTuple, string], tableToCustomKeyColumnsMap *utils.StructMap[sqlname.NameTuple, []string]) (int, error) {
+	partitionKey, err := GetEventPartitionKey(e, tableToPartitioningStrategyMap, tableToCustomKeyColumnsMap)
+	if err != nil {
+		return 0, err
+	}
+	hash := fnv.New64a()
+	hash.Write([]byte(partitionKey))
```
makalaaneesh: could this be a perf regression? because `GetEventPartitionKey` would have to return a string first before it's written to the hash struct. (before, we were directly writing to the hash struct). 
Can you check by running the benchmark test?

yb-voyager/cmd/live_migration.go
```diff
@@ -148,7 +150,7 @@ func streamChanges(state *ImportDataState, tableNames []sqlname.NameTuple) error
 		return fmt.Errorf("failed to initialize stats reporter: %w", err)
 	}
 
-	tableToPartitioningStrategyMap, err := getCdcPartitioningStrategyPerTable(tableNames)
+	tableToPartitioningStrategyMap, tableToCustomKeyColumnsMap, err := getCdcPartitioningStrategyPerTable(tableNames)
```
makalaaneesh: nit: `tableToCustomPartitionKeyColumnsMap`

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -257,41 +292,57 @@ func computeAndPersistCdcPartitioningStrategyPerTable(tableNames []sqlname.NameT
...
+		return nil, nil, nil, err
 	}
 	overrides, err := resolveCdcPartitionKeyOverrides(rawOverrides, tableNames)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
-	exprUKSet, err := getExpressionUniqueIndexTablesIfRequired(tableNames, overrides, cdcPartitionKey, isFirstRun, importDataStatus)
+	strategyOverrides, customColumns, err := splitResolvedOverrides(overrides)
```
makalaaneesh: i didn't really understand why we need to split. WHy can't we continue operating/working with `cdcPartitionKeyOverride`?

makalaaneesh: where do we check if the columns passed are actual valid columns of the table?

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -99,33 +101,47 @@ func getCdcPartitioningStrategyPerTable(tableNames []sqlname.NameTuple) (*utils.
...
+			}
+			if strategy == PARTITION_BY_CUSTOM {
+				columns, ok := tableToCustomKeyColumnsMap.Get(t)
+				if !ok || len(columns) == 0 {
+					return nil, nil, goerrors.Errorf("cdc custom partition key columns not found for table: %s", t.ForKey())
+				}
 			}
 		}
-		return tableToPartitioningStrategyMap, nil
+		return tableToPartitioningStrategyMap, tableToCustomKeyColumnsMap, nil
```
makalaaneesh: I don't quite follow this logic. We are seeming to iterate through all these three separately. 
- TableToCDCPartitioningStrategyMap
- TableToCustomPartitionKeyColumns
- tableNames

Won't the last for-loop suffice?

yb-voyager/cmd/live_migration.go
```diff
@@ -403,41 +407,111 @@ func handleEvent(event *tgtdb.Event,
...
+		}
+		columns, ok := tableToCustomKeyColumnsMap.Get(e.TableNameTup)
+		if !ok || len(columns) == 0 {
+			return "", goerrors.Errorf("custom partition key columns not found for table %v", e.TableNameTup)
+		}
+		// Custom key columns are immutable (guardrail enforced separately), so a row's
+		// key value is the same in BeforeFields (update/delete) and Fields (insert).
+		// Prefer BeforeFields (present for update/delete under REPLICA IDENTITY FULL);
+		// fall back to Fields for inserts.
+		for _, col := range columns {
```
makalaaneesh, line 467: don't we need a sort(columns) here as well? for deterministic hash

yb-voyager/cmd/import.go
```diff
@@ -237,28 +246,75 @@ func parseCdcPartitionKeyOverrides(overrides string) (map[string]string, error)
 		}
 		parts := strings.SplitN(entry, ":", 2)
 		if len(parts) != 2 {
-			return nil, goerrors.Errorf("invalid cdc-partition-key-overrides entry %q: expected format schema.table:pk|table", entry)
+			return nil, goerrors.Errorf("invalid cdc-partition-key-overrides entry %q: expected format schema.table:pk|table|<col1,col2,...>", entry)
```
makalaaneesh: as discussed, we should probably mandatae a "(col1,col2)" or some other special character to be able to distinguish between `table`/`pk`/`col`

yb-voyager/cmd/import.go
```diff
@@ -237,28 +246,75 @@ func parseCdcPartitionKeyOverrides(overrides string) (map[string]string, error)
...
+
+		override, err := parseCdcPartitionKeyOverrideValue(tableName, value)
+		if err != nil {
+			return nil, err
 		}
-		if _, exists := result[tableName]; exists {
-			if strings.EqualFold(strategy, result[tableName]) {
+
+		if existing, exists := result[tableName]; exists {
+			if cdcPartitionKeyOverridesEqual(existing, override) {
```
makalaaneesh: do we really need this? Let's just throw an error if specified twice, asking the user to fix this.

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: I have that structure in the starting place where we are trying to parse/resovle the config; will see if it should also be here

yb-voyager/cmd/live_migration.go
priyanshi-yb: No regression; ran the CDC benchmark on main vs this branch; it is similar only

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: Made this change, now everywhere in memory/metadb we are usign the struct with strategy and columns(in case of custom).

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
priyanshi-yb: In this PR, we have only one check: if the custom partition key is not found in the event at all, we error out. There is a planned guardrail in future PRs to check beforehand if the passed custom partition is part of that table or not, and error out in the validate step itself.

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -399,19 +416,30 @@ func (c *ConflictDetectionCache) findValueConflictLocked(incomingEvent *tgtdb.Ev
...
+	incomingPartitionKey, err := GetEventPartitionKey(incomingEvent, c.tableToPartitioningStrategyMap, c.tableToCustomKeyColumnsMap)
+	if err != nil {
+		return nil, goerrors.Errorf("error computing partition key for incoming event(vsn=%d): %v", incomingEvent.Vsn, err)
+	}
 	var cachedEventsMatchingKey []*tgtdb.Event
 	for _, cachedEvent := range c.ukLookup[key] {
-		if c.sameTableEventsHaveSamePK(cachedEvent, incomingEvent) {
+		cachedPartitionKey, err := GetEventPartitionKey(cachedEvent, c.tableToPartitioningStrategyMap, c.tableToCustomKeyColumnsMap)
```
makalaaneesh: [AI] **Hot path: `GetEventPartitionKey` is recomputed for every cached event on every conflict check.**

This loop calls `GetEventPartitionKey(cachedEvent, ...)` for each event in the bucket, and `getEventsWithDifferentPartitionKey` is itself invoked up to `2 * len(uniqueIndexes)` times per incoming event in `findValueConflictLocked`, and re-run on every `c.cond.Wait()` wake-up inside `WaitUntilNoConflict`'s loop. `GetEventPartitionKey` allocates a `strings.Builder` and, under `PARTITION_BY_PK`, also allocates a key-column slice and `sort.Strings` on every call.

The code this replaced (`sameTableEventsHaveSamePK`) did a cheap column-wise map comparison with no allocation or sort. Per `.cursor/BUGBOT.md` ("Performance-Critical (Hot) Paths"): "conflict detection (`conflictDetectionCache`) ... run for every streamed change. This path has regressed before; a small per-event allocation multiplies by millions of events ... Avoid re-doing map/JSON/marshal work per event when it can be cached or done once."

A cached event's partition key is invariant once it is in the cache. Compute it once when the event is indexed (e.g. store `vsn -> partitionKey`, or add a memoized field on the cached `*tgtdb.Event`) and compare the stored value here, so the sort+build cost is paid once per cached event rather than on every conflict re-check.

---

yb-voyager/cmd/live_migration.go
```diff
@@ -403,41 +407,111 @@ func handleEvent(event *tgtdb.Event,
...
+		// table prefix alone is the partition key.
+	case PARTITION_BY_CUSTOM:
+		if tableToCustomKeyColumnsMap == nil {
+			return "", goerrors.Errorf("table to custom partition key columns map is not initialized")
+		}
+		columns, ok := tableToCustomKeyColumnsMap.Get(e.TableNameTup)
+		if !ok || len(columns) == 0 {
+			return "", goerrors.Errorf("custom partition key columns not found for table %v", e.TableNameTup)
+		}
+		// Custom key columns are immutable (guardrail enforced separately), so a row's
```
makalaaneesh, line 463: [AI] **The custom-key immutability invariant this code relies on is not enforced — a mutable custom key column mis-routes a row's events across channels.**

The comment states "Custom key columns are immutable (guardrail enforced separately)", but that guardrail does not appear to be implemented (the final commit message is "warning for custom key as its not implemented properly" and the only safeguard is the `[Tech Preview]` log line in `parseCdcPartitionKeyOverrideValue`). Correctness of routing depends entirely on this invariant.

Concrete failure: for a custom key column `k` that is later UPDATEd from `V1` to `V2`:
- INSERT of the row → `BeforeFields` nil → routes by `Fields[k] = V1`.
- A later UPDATE of the same row (k now stably `V2`) → `customKeyColumnValue` prefers `BeforeFields[k] = V2` → routes to a *different* channel than the earlier events.

The row's events are now split across channels and can be applied out of commit order, which is exactly what the partition key is supposed to prevent — leading to a unique/PK violation or a stale final row. Preferring `BeforeFields` over `Fields` does not close this gap; any change to a custom-key value across the row's lifetime splits it.

Suggested: actually enforce the guardrail before streaming (reject `--cdc-partition-key-overrides` custom columns that are not part of an immutable/PK-like set, or at least fail fast in `prepareCdcPartitionKey`) rather than documenting an invariant the code assumes but never checks.

---

yb-voyager/cmd/import.go
```diff
@@ -237,28 +246,76 @@ func parseCdcPartitionKeyOverrides(overrides string) (map[string]string, error)
...
+// comma-separated custom column list (PARTITION_BY_CUSTOM).
+func parseCdcPartitionKeyOverrideValue(tableName, value string) (cdcPartitionKeyOverride, error) {
+	switch value {
+	case PARTITION_BY_PK:
+		return cdcPartitionKeyOverride{Strategy: PARTITION_BY_PK}, nil
+	case PARTITION_BY_TABLE:
+		return cdcPartitionKeyOverride{Strategy: PARTITION_BY_TABLE}, nil
+	}
+
+	utils.PrintAndLogfWarning("[Tech Preview] Using custom cdc partition key for table %q: %s", tableName, value)
```
makalaaneesh: [AI] **Any unrecognized override value silently becomes a custom key, and custom columns are never validated to exist — typos fail late, mid-stream.**

`parseCdcPartitionKeyOverrideValue` treats everything that is not literally `pk` or `table` as a custom column list. Two consequences:

1. A mistyped strategy is silently reinterpreted as a custom key. `schema.t:tabel` becomes a custom key on a column named `tabel`; `schema.t:auto` (a valid value for the global `--cdc-partition-key`) becomes a custom key on a column named `auto`. Before this change these produced an immediate, clear validation error.
2. The custom column names are never checked against the actual table schema. A bogus/typo'd column passes parse and `resolveCdcPartitionKeyOverrides`, is persisted, and only fails much later during streaming when `GetEventPartitionKey` errors with "custom partition key column %q not found in event" — aborting the migration after snapshot import rather than at flag-validation time.

Per the product review rules, features like this "must have explicit validation or guardrails". Suggested: validate custom key columns against the target table's columns in `prepareCdcPartitionKey` (which runs before snapshot import and has the name registry / import table list), so misconfiguration fails fast with an actionable error instead of mid-stream.

---

makalaaneesh (commented):
[AI] Automated review — 3 findings (0 critical, 3 warning). See inline comments.

Scope reviewed: generalization of CDC event routing from PK-only to a per-table partition key (`pk` / `table` / `custom`), the conflict-detection cache's switch from same-PK exclusion to partition-key exclusion, the new `TableToCustomPartitionKeyColumns` field on `ImportDataStatusRecord` (additive JSON — backward-compatible), override parsing/persistence, and resume/diff validation. The pk/table byte layout is preserved so channel assignment is stable across upgrades, and partition-key collisions in the conflict path are safe (same key → same channel → in-order). The warnings concern hot-path recomputation, the unenforced custom-key immutability invariant, and missing custom-column validation.

---

yb-voyager/cmd/import.go
```diff
@@ -510,8 +574,9 @@ Note that for the cases where a table doesn't have a primary key, this may lead
 		table: Partition CDC events by table (all events for a table share one channel).`)
 
 	cmd.Flags().StringVar(&cdcPartitionKeyOverrides, "cdc-partition-key-overrides", "",
-		`Optional per-table CDC partition-key overrides as schema.table:pk|table pairs, separated by ';'.
-		Example: public.orders:table;sales.events:pk. Unlisted tables keep the global --cdc-partition-key.`)
+		`Optional per-table CDC partition-key overrides as schema.table:pk|table|(col1,col2,...) pairs, separated by ';'.
```
makalaaneesh: do we want to update this now? considering we're not officially releasing support for this,  i think we should defer any help msg changes

yb-voyager/cmd/conflictDetectionCache.go
makalaaneesh: NA. verified with benchmark

yb-voyager/cmd/live_migration.go
makalaaneesh: NA

yb-voyager/cmd/import.go
makalaaneesh: later PR

makalaaneesh (approved):
LGTM

PR #3711, opened 2026-08-12
Fix NPE on hstore values that are SQL NULL

ShivanshGahlot:

Stacked on #3739 (issue #3743), which must merge first.

An hstore value can be SQL NULL — `'k=>NULL'`. That is a different value from `'k=>""'` and from a NULL column.

- **Crash** — voyager sets `hstore.handling.mode=map`, so debezium delivers that entry as a null inside a Kafka Connect MAP. `DebeziumRecordTransformer` escaped map values with no null check and threw an NPE.
- **Impact** — streaming stops permanently. The exporter throws before the offset is committed, so every restart replays the same WAL record. Deleting the key on the source does not help.
- **Fix** — `PostgresToYbValueConverter` now registers a string pass-through for hstore. The column arrives as `STRING` holding postgres' own text, so the MAP branch is never entered.
- **Backstop** — the MAP branch also writes a null entry as unquoted `NULL`. A domain over hstore still reaches that path, so it stays.
- **Why it went unnoticed** — the snapshot phase uses `pg_dump` and is unaffected. A stored value stays dormant until an unrelated UPDATE writes the full row image to the WAL.

No user-facing changes.

Extended the Go live migration datatype tests with a null-valued hstore entry, a domain over hstore, and literal `"NULL"` / `'@'` / path values in both the INSERT and UPDATE DMLs. `TestLiveMigrationWithDatatypeEdgeCases` and `TestLiveMigrationWithDatatypeEdgeCasesAndFallback` both pass, covering the forward and fall-back paths. `migtests/scripts/live-migration-run-test.sh pg/datatypes` passes through cutover. I also reproduced the original failure against a real PG source, then confirmed a stalled export dir resumes once the fix is in.

No.

No.


debezium-server-voyager/debezium-server-voyagerexporter/src/test/java/io/debezium/server/ybexporter/DebeziumRecordTransformerTest.java
```diff
@@ -0,0 +1,175 @@
+/*
```
makalaaneesh: let's add/modify the golang live migration datatype tests (including fall-back)

makalaaneesh:

---

ShivanshGahlot:
### Why the MAP branch is kept (and not deleted)

We considered deleting the `case MAP:` block in `DebeziumRecordTransformer`, since the new pass-through in `PostgresToYbValueConverter` means hstore no longer goes through it. Tested it, and it turns out the MAP branch is still live code. Keeping it.

**Why it is still reached.** The two paths pick their route using different keys:

- the pass-through registers by type **name**: `case "hstore"`
- debezium picks the MAP schema by type **OID**, and it resolves a domain to its base type

So a domain over hstore slips between them:

```sql
CREATE DOMAIN meta_map AS hstore;
CREATE TABLE t (id text PRIMARY KEY, plain_h hstore, domain_h meta_map);
INSERT INTO t VALUES ('D1', 'k=>NULL, s=>"v"', 'k=>NULL, s=>"v"');
```

Both columns hold the same hstore value, but debezium reports:

```
Processing converter for column: plain_h,  type: hstore
Processing converter for column: domain_h, type: meta_map     <- name is the domain, not "hstore"

Configuring stringify converter for column: plain_h, type: hstore
                                            <- domain_h not registered, so it goes to the MAP branch
```

**What each variant does with `domain_h`:**

| | export | value written for `domain_h` |
| --- | --- | --- |
| this PR (pass-through + MAP null guard) | succeeds | `"s" => "v","k" => NULL` — valid hstore |
| MAP branch deleted | succeeds | `{s=v, k=null}` — Java's map format |
| MAP null guard deleted | **NPE**, streaming stalls | — |

**Why deleting it is the worst of the three.** With the branch gone the value falls through to `fieldValue.toString()`, which on a Map gives Java's format. Export reports success, and the failure surfaces later in `import data`:

```
ERROR yugabytedb.go:1160 error closing batch(1:1): ERROR: Syntax error near "v" at position 3 (SQLSTATE XX000)
WARN  live_migration.go:551 retriable error executing batch(1:1) ...
```

It is classified **retriable**, so import retries the same poisoned event indefinitely (14 attempts before I stopped it) and the row never advances. Same permanent-stall shape as the original bug, but moved to the import side and reported as an hstore syntax error, with nothing pointing back at the exporter that produced it.

So the guard is not dead code: today it is what makes domain-over-hstore columns export correctly.

**Known gap, not addressed here:** `domain_h` does not get the pass-through at all, so domains still rely on the hand-rolled escaping. Fixing that properly means matching on the resolved base type rather than the type name, which is a larger change than this PR. Happy to file it separately.

ShivanshGahlot:
@makalaaneesh haven't deleted the MAP path. Please take a look at the comment claude posted above.

ShivanshGahlot:
### Re: is this a "domain" problem, and does it affect more than hstore?

Yes to both. Tested it — a plain column and a domain column side by side, same data.

The hook has two switches, and they behave differently under domains:

```java
switch (JDBCType.valueOf(column.jdbcType()))  // BIT, STRUCT, ARRAY  -> safe
switch (column.typeName())                    // varbit, tsvector,
                                              //  tsquery, hstore   -> broken
```

Debezium resolves a domain's **jdbcType** to the base type, but reports **typeName** as the domain's own name:

```
p_intarr,    type: _int4,      JDBC type: 2003
dm_intarr,   type: d_intarr,   JDBC type: 2003   <- still matches
p_tsvector,  type: tsvector,   JDBC type: 1111
dm_tsvector, type: d_tsvector, JDBC type: 1111   <- no longer matches
```

So all four typeName-matched types miss the hook under a domain.

### Is a domain the only UDT affected?

Yes. Checked every `pg_type` kind that can wrap one of these types:

| column | typname | typtype | jdbcType | registered? |
| --- | --- | --- | --- | --- |
| `hstore` | `hstore` | b | 1111 | yes (name) |
| `hstore[]` | `_hstore` | b | 2003 | yes (jdbcType ARRAY) |
| `d_hstore[]` (array of DOMAIN) | `_d_hstore` | b | 2003 | yes (jdbcType ARRAY) |
| composite with hstore/tsvector | `c_comp` | c | 2002 | yes (jdbcType STRUCT) |
| DOMAIN over DOMAIN | `d2_hstore` | d | 1111 | **no** |

Arrays and composites are safe because their jdbcType is always ARRAY/STRUCT regardless of what is inside, and the first switch matches on jdbcType. A domain is the only kind that keeps a scalar jdbcType while changing the type name. (Multi-dimensional arrays are a non-issue — `int[][]` is still `_int4` in PG, dimensionality is not part of the type.)

### End-to-end result: export -> import into YugabyteDB

Segment file (one event):

```json
{"vsn":1,"op":"u",
 "before_fields":{
   "p_hstore":"\"k\"=>NULL, \"s\"=>\"v\"",
   "dm_hstore":"\"s\" => \"v\",\"k\" => NULL",
   "p_tsvector":"'cat':3 'fat':2",
   "dm_tsvector":"J2NhdCc6MyAnZmF0Jzoy",
   "p_tsquery":"'fat' & 'cat'",
   "dm_tsquery":"J2ZhdCcgJiAnY2F0Jw==",
   "p_varbit":"1010101010",
   "dm_varbit":"qgI="},
 "table_name":"alltypes","key":{"id":"A1"}}
```

Source vs target after import (the CDC update touched only the `dm_*` columns, which is why the `p_*` source values stay at their originals):

| column | source | target | |
| --- | --- | --- | --- |
| `p_hstore` | `"k"=>NULL, "s"=>"v"` | `"k"=>NULL, "s"=>"v"` | ok |
| `dm_hstore` | `"k2"=>NULL, "s2"=>"v2"` | `"k2"=>NULL, "s2"=>"v2"` | ok |
| `p_tsvector` | `'cat':3 'fat':2` | `'cat':3 'fat':2` | ok |
| `dm_tsvector` | `'brown':3 'fox':4 'quick':2` | `'x2762726f776e273a33…'` | **corrupt** |
| `p_tsquery` | `'fat' & 'cat'` | `'fat' & 'cat'` | ok |
| `dm_tsquery` | `'quick' & 'fox'` | `'x27717569636b272026…'` | **corrupt** |
| `p_varbit` | `1010101010` | `1010101010` | ok |
| `dm_varbit` | `1111000011` | `0000001111000011` | **corrupt** |

**Import reported 0 errors and caught up cleanly.** The base64 is decoded to bytes, rendered as hex, and stored as a single lexeme — valid tsvector/tsquery syntax, so nothing complains:

```
2762726f776e273a33…  ->  'brown':3 'fox':4 'quick':2
27717569636b272026…  ->  'quick' & 'fox'
```

`dm_varbit` is subtler: 10 bits became 16, because the base64 decoded to 2 whole bytes and got zero-padded to the column width. It looks like an ordinary bit string.

Note `dm_hstore` round-trips correctly — it is the only one of the four that survives a domain, because it falls to the MAP branch. That is why this PR keeps that branch.

### Suggested fix (tried it, it works)

`nativeType` is already the resolved base OID — debezium reports the same value for a plain column and a domain over it:

```
PROBE col=p_hstore typeName=hstore     jdbc=1111 nativeType=25744
PROBE col=dom2     typeName=d2_hstore  jdbc=1111 nativeType=25744   <- identical
```

So match the base OID as a fallback. varbit/tsvector/tsquery are built-in types with OIDs fixed by PG, so they can be constants:

```java
private static final Set<Integer> PASSTHROUGH_BASE_OIDS = Set.of(
        1562,  // varbit
        3614,  // tsvector
        3615); // tsquery

// in the typeName switch:
default:
    if (PASSTHROUGH_BASE_OIDS.contains(column.nativeType())) {
        registration.register(SchemaBuilder.string(), this::stringify);
    }
    break;
```

hstore is deliberately not listed: it is an extension, so its OID is per-database (verified — `hstore` was 26038 in one DB and 26463 in another, depending on what was installed first), and hstore-over-domain is already correct via the MAP branch.

Applied this and re-ran the same scenario:

```
Configuring stringify converter for column: dm_tsvector, type: d_tsvector (base oid 3614)
Configuring stringify converter for column: dm_tsquery,  type: d_tsquery  (base oid 3615)
Configuring stringify converter for column: dm_varbit,   type: d_varbit   (base oid 1562)
```

All 8 columns now match between source and target, import errors 0:

```
dm_tsvector | 'brown':3 'fox':4 'quick':2   (was 'x2762726f776e273a33…')
dm_tsquery  | 'quick' & 'fox'               (was 'x27717569636b272026…')
dm_varbit   | 1111000011                    (was 0000001111000011)
```

Keeping it out of this PR since it is a separate bug with a different blast radius — happy to raise it as its own PR, patch is ready.

debezium-server-voyager/debezium-server-voyagerexporter/src/test/java/io/debezium/server/ybexporter/DebeziumRecordTransformerTest.java
```diff
@@ -0,0 +1,136 @@
...
```
makalaaneesh: can we get rid of the debezium side tests, since we're never really running them

yb-voyager/src/testlivemigration/live_migration_datatype_edge_cases_test.go
```diff
@@ -3471,34 +3526,47 @@ func getDatatypeEdgeCasesTestConfig(dbName string) *TestConfig {
...
+				map_domain = '"dnk" => "", "dek" => NULL'::test_schema.meta_map
 			WHERE id = 2;`,
 
 			`UPDATE test_schema.map_edge_cases
 			SET map_empty_values = '"fallback" => "value"',
-				map_multiple_pairs = '"a" => "1", "b" => "2"'
+				map_multiple_pairs = '"a" => "1", "b" => "2"',
+				map_null_values = '"back" => NULL',
+				map_domain = '"dback" => NULL'::test_schema.meta_map
 			WHERE id = 5 AND map_empty_values IS NULL;`,
```
priyanshi-yb, line 3573: One case of each NULL literal / special char like `@` in emails / path  in the value in UPDATE DML also similar to INSERT?
`"key" => "NULL"` 
`"Key" => "sda@sdf.com"`
`"Key"=> "/U://dasa/dasda/`

priyanshi-yb (approved):
LGTM, just add few more cases in UPDATE statements

yb-voyager/src/testlivemigration/live_migration_datatype_edge_cases_test.go
ShivanshGahlot: Done

PR #3715, opened 2026-08-17
[cherry-pick] Propagate WaitUntilNoConflict error in handleEvent instead of ignorin…

makalaaneesh:
…g it (#3714)

WaitUntilNoConflict returns an error when a conflict probe cannot be computed (e.g. an incoming UPDATE with nil BeforeFields when the source table lacks REPLICA IDENTITY FULL, or a failed partition-key computation). The call site in handleEvent discarded that error, so the event proceeded without any conflict serialization - silently disabling unique-key conflict detection for that event. Inside findValueConflictLocked the same error path also discards before-after conflicts that were already found.

Propagate the error so the import fails loudly, consistent with how the DELETE path already hard-fails when Put cannot cache an event with nil BeforeFields.

---

| Component        

### DIFF
--- a/yb-voyager/cmd/conflictDetectionCache_test.go
+++ b/yb-voyager/cmd/conflictDetectionCache_test.go
@@ -793,3 +793,22 @@ func TestConflictWithMultipleIndexes(t *testing.T) {
 	assert.Equal(t, "idx_a_b", actualConflicts[0].indexName)
 	assert.Equal(t, "idx_nnd_c_d", actualConflicts[1].indexName)
 }
+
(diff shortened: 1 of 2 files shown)


priyanshi-yb (approved):

priyanshi-yb:
nit: remove `[Cherry-pick]` from the title

makalaaneesh:
> nit: remove `[Cherry-pick]` from the title

This was intentional, just to have a differentiator from the original PR. Any reasons why you think it's a bad idea?

priyanshi-yb:
>This was intentional, just to have a differentiator from the original PR. Any reasons why you think it's a bad idea?

The reason is to have the same commit msg in both branches for better readability

PR #3716, opened 2026-08-17
Fix: false negative case where subset of unique index columns being changed in update

priyanshi-yb:
The write-up covers the false-negative: Debezium’s after-image only has changed columns, so a composite/partial unique-index UPDATE that touches a subset of columns never reconstructed the full index tuple. Import now merges BeforeFields + Fields into AfterFields and uses that in the before-after check. Testing called out is TestLiveMigrationWithSubsetOFPartialUNiqueIndexColumnsBeingChangedInUpdate.

---

| Component        

### DIFF
--- a/yb-voyager/src/tgtdb/event.go
+++ b/yb-voyager/src/tgtdb/event.go
@@ -88,7 +88,7 @@ func (e *Event) UnmarshalJSON(data []byte) error {
 	e.Fields = rawEvent.Fields
 	e.BeforeFields = rawEvent.BeforeFields
 	e.ExporterRole = rawEvent.ExporterRole
-	e.AfterFields = mergeBeforeAndAfterFields(e.BeforeFields, e.Fields)
+	e.AfterFields = mergeBeforeAndChangedFields(e.Op, e.BeforeFields, e.Fields)
 	e.partitionSchemaName = rawEvent.PartitionSchemaName
 	e.partitionTableName = rawEvent.PartitionTableName
 	if !e.IsCutoverEvent() {
@@ -101,15 +101,23 @@ func (e *Event) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-func mergeBeforeAndAfterFields(beforeFields, afterFields map[string]*string) map[string]*string {
-	fullFields := make(map[string]*string)
-	for column, value := range beforeFields {
-		fullFields[column] = value
+func mergeBeforeAndChangedFields(op string, beforeFields, changeFields map[string]*string) map[string]*string {
+	switch op {
+	case "c":
+		return changeFields
+	case "u":
+		allAfterFields := make(map[string]*string)
+		for column, value := range beforeFields {
+			allAfterFields[column] = value
+		}
+		for column, value := range changeFields {
+			allAfterFields[column] = value
+		}
+		return allAfterFields
+	case "d":
+		return nil
 	}
-	for column, value := range afterFields {
-		fullFields[column] = value
-	}
-	return fullFields
+	return nil
 }
 
 var cachePreparedStmt = sync.Map{}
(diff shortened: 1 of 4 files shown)


yb-voyager/src/tgtdb/event.go
```diff
@@ -99,6 +101,17 @@ func (e *Event) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func mergeBeforeAndAfterFields(beforeFields, afterFields map[string]*string) map[string]*string {
```
makalaaneesh: nit: we are merging `BeforeAndChangedFields`

makalaaneesh: we should NOT be merging for `DELETE` or `INSERT` because it does not make sense. 
For insert, `AfterFields` is the full set of fields anyway. before is empty.
For delete, `AfterFields` should be empty 

(trying to be consistent with debezium terminology)

yb-voyager/src/tgtdb/event.go
```diff
@@ -87,6 +88,7 @@ func (e *Event) UnmarshalJSON(data []byte) error {
 	e.Fields = rawEvent.Fields
 	e.BeforeFields = rawEvent.BeforeFields
 	e.ExporterRole = rawEvent.ExporterRole
+	e.AfterFields = mergeBeforeAndChangedFields(e.Op, e.BeforeFields, e.Fields)
```
makalaaneesh, line 91: [AI] **`AfterFields` is eagerly allocated for every UPDATE event on the CDC hot path.**

`UnmarshalJSON` runs for every CDC event — a documented hot path (root `.cursor/BUGBOT.md`, "Performance-Critical (Hot) Paths"). For every `op=="u"` event, `mergeBeforeAndChangedFields` allocates a brand-new map and copies **all** of `BeforeFields` into it. Under `REPLICA IDENTITY FULL` (which is exactly what makes this fix work), `BeforeFields` holds every column, so this is an O(#columns) map allocation + copy per update event — and the cache comments elsewhere note rows of 30–50 columns / ~7 kB.

This cost is paid unconditionally, including for tables with **no** unique indexes and migrations where conflict detection never reads `AfterFields` (only source-exporter events on tables with unique indexes ever consult it via `checkUniqueIndexBeforeAfterConflict`).

Failure case: high-throughput CDC of wide tables with RI FULL incurs a new per-event heap allocation with zero benefit in the common no-unique-index case.

Suggested fix: compute the merged after-image lazily, on demand inside conflict detection (a helper called from `checkUniqueIndexBeforeAfterConflict` only for tables that actually have unique indexes), or gate the eager computation on "table has unique indexes", rather than doing it for every update in `UnmarshalJSON`.

---

yb-voyager/src/tgtdb/event.go
```diff
@@ -44,6 +44,7 @@ type Event struct {
 	Key                 map[string]*string
 	Fields              map[string]*string //all the column values of the row - worst
 	BeforeFields        map[string]*string //all the column values of the row - worst
+	AfterFields         map[string]*string //all the column values of the row - worst //all fields of the row adn columns that are chnges in fields are updated to fields value
```
makalaaneesh, line 47: [AI] **`Event.Copy()` does not copy the new `AfterFields` field.**

The new `AfterFields` field is not copied by `Event.Copy()` (which uses `lo.MapEntries` for `Key`/`Fields`/`BeforeFields` but omits `AfterFields`), so every event cloned via `ConflictDetectionCache.Put` → `event.Copy()` has `AfterFields == nil`.

This is non-fatal only by luck today: `checkUniqueIndexBeforeAfterConflict` reads `AfterFields` from the *incoming* (pre-conversion, un-copied) event and only `BeforeFields` from the cached copy. But `Copy()` is the canonical clone used for the whole cache, and this omission is a latent correctness trap — any future check that reads `cachedEvent.AfterFields` will silently get `nil` and miss conflicts, i.e. re-introduce exactly the false-negative class this PR fixes.

Suggested fix: add `AfterFields: lo.MapEntries(e.AfterFields, idFn)` to `Copy()`. (Alternatively, if you adopt lazy computation per the other comment and drop the stored field, there's nothing to forget here.)

---

makalaaneesh (commented):

---

PR #3717, opened 2026-08-18
migtests: wait for importer to drain queue before cutover in fallback-unique-conflict-test

Abhi-yb:
Nightly hit `TimeoutError: cutover_to_target_status_completed` — `remaining_events_eq_0` only proves the exporter wrote the cutover marker into the queue segments; the importer can still be far behind, so the whole drain was happening inside the 900s cutover wait, which isn't enough under conflict load.

Adds `forward_importer_caught_up` / `fallback_importer_caught_up` wait conditions (checks `queue_segment_meta` per-segment imported flags in metaDB, mirroring `MetaDB.GetPendingSegments`) and a drain-wait stage **after cutover is initiated** on both legs (per review — initiating cutover closes the final open segment, so all segments can reach imported and the condition is `pending == 0`). The cutover status waits stay at 900s and now only cover the status flip, not the drain.

Verified: a pre-cutover `== 0` variant deadlocks by construction (run 1112 — the open tail segment can never be marked imported before cutover); post-initiate placement avoids that.

### DIFF
--- a/migtests/scripts/live-migration-resumption/helpers.py
+++ b/migtests/scripts/live-migration-resumption/helpers.py
@@ -309,9 +309,7 @@ def importer_caught_up(export_dir: str, leg: str) -> bool:
     backlog_marker_present() only proves the EXPORTER wrote the cutover marker into
     the queue; the importer can still be far behind, and cutover cannot complete
     until it drains. This checks the importer side via queue_segment_meta's
-    per-segment imported flags (mirrors MetaDB.GetPendingSegments). The threshold
-    is <= 1, not 0: voyager finalizes the last open segment only during cutover,
-    so it stays unimported until then.
+    per-segment imported flags (mirrors MetaDB.GetPendingSegments).
     """
     meta_db = os.path.join(export_dir, "metainfo", "meta.db")
     query = (
@@ -322,7 +320,7 @@ def importer_caught_up(export_dir: str, leg: str) -> bool:
     if proc.returncode != 0:
         return False
     try:
-        return int(proc.stdout.strip()) <= 1
+        return int(proc.stdout.strip()) == 0
     except ValueError:
         return False
 
--- a/migtests/scripts/live-migration-resumption/helpers.py
+++ b/migtests/scripts/live-migration-resumption/helpers.py
@@ -293,6 +293,40 @@ def backlog_marker_present(export_dir: str) -> bool:
         return False
 
 
+_IMPORTER_CAUGHT_UP_PREDICATES = {
+    # cutover-to-target waits on the target importer draining source-exported segments
+    "forward": "exporter_role = 'source_db_exporter' AND imported_by_target_db_importer = 0",
+    # cutover-to-source waits on the source importer draining target-exported segments
+    "fallback": "exporter_role LIKE 'target_db_exporter%' AND imported_by_source_db_importer = 0",
+}
+
+
+def importer_caught_up(export_dir: str, leg: str) -> bool:
+    """
+    Return True once the streaming importer for *leg* ("forward" or "fallback") has
+    applied every closed queue segment, i.e. at most one segment is still pending.
+
+    backlog_marker_present() only proves the EXPORTER wrote the cutover marker into
+    the queue; the importer can still be far behind, and cutover cannot complete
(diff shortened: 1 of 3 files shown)


migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -293,6 +293,40 @@ def backlog_marker_present(export_dir: str) -> bool:
...
+    meta_db = os.path.join(export_dir, "metainfo", "meta.db")
+    query = (
+        "select count(*) from queue_segment_meta "
+        f"where {_IMPORTER_CAUGHT_UP_PREDICATES[leg]};"
+    )
+    proc = subprocess.run(["sqlite3", meta_db, query], capture_output=True, text=True)
+    if proc.returncode != 0:
+        return False
+    try:
+        return int(proc.stdout.strip()) <= 1
```
priyanshi-yb: why equal to 1? it should <1 no i.e. no segements left to be imported?

priyanshi-yb (approved):
LGTM, please ensure we run few run of this tests after latest changes

migtests/scripts/live-migration-resumption/helpers.py
```diff
@@ -293,6 +293,38 @@ def backlog_marker_present(export_dir: str) -> bool:
...
+    meta_db = os.path.join(export_dir, "metainfo", "meta.db")
+    query = (
+        "select count(*) from queue_segment_meta "
+        f"where {_IMPORTER_CAUGHT_UP_PREDICATES[leg]};"
+    )
+    proc = subprocess.run(["sqlite3", meta_db, query], capture_output=True, text=True)
+    if proc.returncode != 0:
+        return False
+    try:
+        return int(proc.stdout.strip()) == 0
```
makalaaneesh: [AI] **`== 0` contradicts the documented "at most one segment pending" (`<= 1`) contract and can hang the wait.**

The docstring above states this returns `True` "once the streaming importer … has applied every closed queue segment, **i.e. at most one segment is still pending**" and that it "mirrors `MetaDB.GetPendingSegments`". The PR description likewise says "**pending <= 1** since voyager finalizes the last open segment only during cutover." But the implementation returns `True` only when the pending count is `== 0`.

These disagree. A queue segment is marked processed only when the importer reads its EOF marker, and that marker is written by the exporter only when it closes/rotates a segment (rotation is byte-size driven). While the exporter is streaming, the currently-open segment always has `imported_by_* = 0`, so the pending count is `>= 1` and `== 0` is generally not reachable.

Failure case: on the **forward** leg the wait runs while both `export_data` and `import_data` resumptions are still active (they are stopped only after cutover). With the load generators already stopped there is no byte growth to force a rotation, so the exporter holds an open, unclosed segment → count stays `1` → the predicate never becomes true → `wait_for` raises `TimeoutError` after the 1800s window, failing the whole test. The only way count momentarily reaches 0 is a race during a resumption SIGTERM window — i.e. flaky.

Suggested fix: match the documented intent (and `GetPendingSegments`, whose result includes the open segment during streaming):

```python
return int(proc.stdout.strip()) <= 1
```

Per `migtests/.cursor/BUGBOT.md` ("Live Migration Tests"), streaming-phase wait conditions must be robust; a stricter-than-documented predicate defeats that.

---

makalaaneesh (commented):

---

migtests/tests/resumption/live-migration/fallback-unique-conflict-test/scenario.yaml.template
```diff
@@ -208,6 +208,13 @@ stages:
     condition: remaining_events_eq_0
     timeout_sec: 1800
 
+  # wait_backlog_zero only proves the exporter wrote the cutover marker; wait for
+  # the importer to drain the queue too, so cutover completes within its timeout.
+  - name: wait_importer_caught_up
+    action: wait_for
+    condition: forward_importer_caught_up
+    timeout_sec: 1800
+
```
priyanshi-yb: we should only wait for the iporter caught up[ and finished processing all the segments only after cutover to target is initiated. so please move this block to after `cutover to target`

priyanshi-yb (changes requested):

PR #3724, opened 2026-08-18
Fix unhandled and ignored errors surfaced by static-analysis audit

makalaaneesh:

While auditing why PR #3714's dropped `WaitUntilNoConflict` error got past CI, a static-analysis sweep (errcheck + staticcheck with a working config) surfaced several real bugs of the same family already on main. This PR fixes only the bug-shaped findings; the tooling changes that will prevent recurrence come in a follow-up PR.

- **`copyBatchCore` ignored `setTargetSchema`'s error** (`src/tgtdb/yugabytedb.go`): the check was dropped in PR #2498 when the COPY fast path was extracted. A failure to set the search path now fails the batch (wrapped as an import-batch error with a new `set_target_schema` step) instead of proceeding on a connection with the wrong schema.
- **All four `srcdb` drivers ignored `sql.Open`'s error** in `Connect()` (postgres/mysql/oracle/yugabytedb): on failure `db` is nil and the very next line (`db.SetMaxOpenConns`) panics. Now returns the error.
- **Callhome sent raw materialized-view names instead of the anonymized ones** (`cmd/callhome_payload_builder.go`): the loop anonymized the names into `objects` and then never used it. The mview-colocation payload now sends the anonymized names, matching the table-colocation block above it.
- **Start-clean ignored `UpdateMigrationStatusRecord`'s error** (`cmd/exportData.go`): a failed reset of MSR fields was silently swallowed; now exits with an error like its neighbors.
- **`test/containers` `Terminate()` dereferenced the receiver before its own nil check** (all four containers): the nil guard now comes first.
- `//lint:ignore SA5011 utils.ErrExit always exits` directives on the nil-MSR deref paths (`cmd/common.go`, `cmd/cutoverToTargetCmd.go`): `ErrExit` is a patchable func variable, so staticcheck cannot prove it terminates; the directives assert the production invariant without restructuring the code. `//lint:ignore SA4023` similarly keeps the conventional `if err != nil` after `StartMonitoring` (`cmd/importData.go`), which staticcheck proves is currently always true. If either invariant ever changes, staticcheck flags the directive as unmatched, forcing a revisit.
- Removed dead stores flagged by SA4006/SA4010 (`cmd/exportSchema.go`, `cmd/importData.go`, `cmd/importDataSequenceRestoration.go`).

No CLI, configuration, installation, or report changes. Failure behavior changes in two spots: import batches now fail loudly if setting the target schema on the connection fails, and source `Connect()` returns an error instead of panicking when `sql.Open` fails.

- `go build ./...`, `go vet ./...`, and `go test -tags unit ./...` all pass (zero failures).
- Verified with staticcheck 2025.1.1 that all targeted SA4006/SA4010/SA4023/SA5011 findings are resolved.
- The `setTargetSchema` path is exercised by the existing import-data integration tests; no new tests added since the fixes restore error propagation on paths whose happy path is already covered.

The payload *shape* is unchanged (no version bump needed); the mview-colocation `Objects` field now carries the anonymized names it was always supposed to carry.

No.

### DIFF
--- a/yb-voyager/cmd/callhome_payload_builder.go
+++ b/yb-voyager/cmd/callhome_payload_builder.go
@@ -479,7 +479,7 @@ func buildCallhomeSchemaOptimizationChanges() []callhome.SchemaOptimizationChang
 		schemaOptimizationChanges = append(schemaOptimizationChanges, callhome.SchemaOptimizationChange{
 			OptimizationType: MVIEW_COLOCATION_RECOMMENDATION_CHANGE_TYPE,
 			IsApplied:        schemaOptimizationReport.MviewColocationRecommendation.IsApplied,
-			Objects:          schemaOptimizationReport.MviewColocationRecommendation.ShardedObjects,
+			Objects:          objects,
 		})
 	}
 	if schemaOptimizationReport.SecondaryIndexToRangeChange != nil {
--- a/yb-voyager/cmd/common.go
+++ b/yb-voyager/cmd/common.go
@@ -870,14 +870,13 @@ func GetSourceDBTypeFromMSR() string {
 		utils.ErrExit("get migration status record: %v", err)
 	}
 	if msr == nil {
-		// utils.ErrExit is a patchable func variable, so it is not provably terminating; return explicitly.
 		utils.ErrExit("migration status record not found")
-		return ""
 	}
+	//lint:ignore SA5011 utils.ErrExit always exits
 	if msr.SourceDBConf == nil {
 		utils.ErrExit("source DB conf not found in migration status record")
-		return ""
 	}
+	//lint:ignore SA5011 utils.ErrExit always exits
 	return msr.SourceDBConf.DBType
 }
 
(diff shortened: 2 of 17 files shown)


makalaaneesh:

---

yb-voyager/cmd/callhome_payload_builder.go
```diff
@@ -479,7 +479,7 @@ func buildCallhomeSchemaOptimizationChanges() []callhome.SchemaOptimizationChang
 		schemaOptimizationChanges = append(schemaOptimizationChanges, callhome.SchemaOptimizationChange{
 			OptimizationType: MVIEW_COLOCATION_RECOMMENDATION_CHANGE_TYPE,
 			IsApplied:        schemaOptimizationReport.MviewColocationRecommendation.IsApplied,
-			Objects:          schemaOptimizationReport.MviewColocationRecommendation.ShardedObjects,
+			Objects:          objects,
```
ShivanshGahlot, line 482: Should we add some sort of tests around these payload builders to confirm that nothing is sending the real names?

ShivanshGahlot (approved):
LGTM with one question

yb-voyager/cmd/callhome_payload_builder.go
makalaaneesh: i think we may already have; this is probably a miss; agree that it's worth checking and adding. 
Can also ask AI to check callhome for any gaps in anonymization

priyanshi-yb (approved):
LGTM

PR #3510, opened 2026-04-17
Remove old archive changes implementation

AmeyaKasture:
### Describe the changes in this pull request

This PR removes the previous implementation of the `archive changes` command (`archiveChangesCommand.go`) and its associated helper functions (`registerCommonArchiveFlags`, `validateCommonArchiveFlags`, `validateMoveToFlag`) from `archiveCommand.go`. The old implementation included `EventSegmentCopier` and `EventSegmentDeleter` types that managed CDC event segment archiving and deletion during live migration. This logic has been superseded by the newer `segmentCleanupCommand.go`, which provides a cleaner, policy-based approach (delete/archive) for managing processed CDC event segments.

The `StopArchiverSignal` variable, previously declared in the deleted `archiveChangesCommand.go`, is moved to `segmentCleanupCommand.go` to maintain compilation. The parent `archive` command's short description is also updated to reference the `archive changes` subcommand for discoverability.

### Describe if there are any user-facing changes

1. **Command line:** The old `archive changes` command with its `--move-to`, `--delete-changes-without-archiving`, and `--fs-utilization-threshold` flags is removed. Users should use the new `archive changes` subcommand registered by `segmentCleanupCommand.go`, which uses a `--policy` flag instead.
2. **Configuration:** No configuration changes.
3. **Installation:** No installation changes.
4. **Reports:** No report changes.

### How was this pull request tested?

This change is a code removal/cleanup. Testing should verify:
- The `yb-voyager archive changes` command still works correctly via `segmentCleanupCommand.go`.
- The binary compiles successfully (`go build ./...`).
- Unit tests pass (`go test -tags unit ./...`).
- No references to the deleted types (`EventSegmentCopier`, `EventSegmentDeleter`) or functions (`registerCommonArchiveFlags`, `validateCommonArchiveFlags`, `validateMoveToFlag`) remain in the codebase.

### Does your PR have changes in callhome/yugabyted payloads? If so, is the payload version incremented?

No.

### Does your PR have changes to on-disk structures that can cause upgrade issues?

No.

---

#### On-disk structures potentially causing breaking changes:
| Component                                        |
| :----------------------------------------------: |
| MetaDB                                           |
| Name registry json                               |
| Data File Descriptor Json                        |
| Export Snapshot Status Json                       |
| Callhome Json                                    |
| Export Status Json                                |
| YugabyteD Tables                                 |
| TargetDB Metadata Tables                         |
| Schema Dump                                      |
| AssessmentDB                                     |
[...]

PR #3725, opened 2026-08-18
Re-enable staticcheck in CI and fix everything it surfaces

makalaaneesh:

CI's staticcheck step has been a silent no-op since `staticcheck.conf` was added: its `checks = ["-S1008"]` lacks `"inherit"`, which in staticcheck's config semantics *replaces* the default check list — leaving zero checks enabled. This is one of the gaps that let PR #3714's unchecked-error bug land. This PR (stacked on #3724, which fixes the real bugs the working checker found) turns the checker back on and makes the tree clean under it:

- `staticcheck.conf` → `checks = ["inherit", "-S1008", "-ST1005"]`, with comments explaining the `"inherit"` pitfall and why ST1005 (error-string capitalization) is excluded.
- CI (`go.yml`) now also runs `go vet`/`staticcheck` with `-tags unit`: every one of the 110 test files is tag-gated (enforced by CI itself), so the untagged run never analyzed any of them.
- Deleted ~430 lines of dead code (U1000). Every deletion was traced through git history to the commit that orphaned it — see the per-block breakdown below. Highlight: the gather-metadata copies in `cmd/` were correctly deleted in the move to `src/migassessment` (b63e70b74) but re-introduced caller-less by a conflict resolution in f03903790, after which the per-node pgss change had to be written into both copies. Silent double-maintenance like that is the concrete cost of the checker having been off.
- Split `cmd/importDataTestUtils.go`: shared helpers keep the file (now tagged `unit || integration || integration_voyager_command`); YugabyteDB-container helpers move to `importDataIntegrationTestUtils.go` (integration tags only). The file was untagged, so its integration-only helpers looked dead to single-tag analysis.
- Fixed assigned-then-overwritten errors in `importDataFileTaskPicker_test.go` (SA4006) and mechanical simplifications (S1000/S1005/S1011/S1021/S1023/S1028/S1031/S1034/S1039).
- `//go:build unit_test` → `unit` on `src/utils/type_mapping_test.go` — that test file was invisible to CI's `-tags unit` run and has never executed in CI; it passes.
- Deleted the stray `.github/workflows/Untitled` file; updated the lint documentation in `AGENTS.md`. (The pure-formatting `gofmt -w` sweep moved to stacked PR #3749 so this diff shows only real changes.)

<details>
<summary><b>Dead-code deletions: what each block did and why it is dead</b> (click to expand)</summary>

| Deleted block | What it did | Why it is dead |
|---|---|---|
| `cmd/assessMigrationCommand.go`: `gatherAssessmentMetadata`, `gatherMetadataStageConfig`, `parseExportedSchemaFileForAssessmentIfRequired`, `populateMetadataCSVIntoAssessmentDB` | The assess-migration metadata stage: source-DB metadata collection, exported-schema parsing, CSV → assessment-DB load | Moved (callers + definitions) to `src/migassessment/gather_metadata.go` in b63e70b74, then re-introduced caller-less by a conflict resolution in f03903790; the live copies in `migassessment` are the only ones referenced |
| `cmd/config.go`: `readAndValidateConfigFile`, `readConfigFileAndGetExportDataFromTargetKeys`, `readConfigFileAndGetImportDataToSourceKeys` | Re-read the config file to list which `export-data-from-target.*` / `import-data-to-source.*` keys the user set, for post-cutover commands | #3367 replaced the ad-hoc re-read with the `resolvedConfig` + `globalFlags` machinery and removed the only two callers; the inner reader was documented as "called only by" the two outer ones |
| `cmd/archiveCommand.go`: `moveDestination`, `deleteSegments`, `utilizationThreshold` | Backed the old `archive changes` flags `--move-to`, `--delete-changes-without-archiving`, `--fs-utilization-threshold` | #3510 deleted the entire old archive-changes implementation (flag bindings and all readers) in favor of the policy-based backend from #3459, leaving only the bare declarations |
| `cmd/archiveChangesCommand.go`: `waitForWorkflowEnd` | Wrapper calling `waitForWorkflowEndForDB` with the package globals | #3476 (archive changes across cutover iterations) needed explicit db/dir params and switched the sole caller to `waitForWorkflowEndForDB` in the same commit |
| `cmd/analyzeSchema.go`: `partitionTablesMap` | Marked partitioned tables so analyze-schema could flag FK references / before-row triggers on them | That state moved into `queryissue.ParserIssueDetector` in #2020; the same-named field there is a different symbol in a different package |
| `cmd/callhome_payload_builder.go`: `getAnonymizedConstraintNamesFromConstraints` | Anonymized constraint names for callhome payloads | Born dead in #3142 — only its sibling index-anonymizer was ever wired up |
| `cmd/common.go`: `type Alias NoteInfo` | The marshal-recursion-guard idiom inside `NoteInfo.MarshalJSON` | Never referenced: the method builds an explicit anonymous struct instead of embedding the alias (vestigial since #2999); marshaling behavior unchanged |
| `cmd/getDataMigrationReportCommand.go`: `getFinalRowCountWithSnapshotRows` | Twin of `getFinalRowCount` taking explicit snapshot-row overrides | Born dead in #3388 — both call sites always used the plain variant |
| `cmd/root.go`: `offlineCommands` | Decided which commands got `--config-file` un-hidden (`MarkHidden` branch) | #3367 made `--config-file` unconditional on every command and deleted that branch; the "used for registering" comment had been stale since |
| `src/migassessment/assessmentDB.go`: `columnExists` | `pragma_table_info` probe for backward compat against older assessment DBs | Its single caller (checking for the `source_node` column) was removed in #3225 when that column became unconditional |
| `src/migassessment/sizing.go`: `checkAndFetchIndexes`, `getMultiplicationFactorForImportTimeBasedOnIndexes` | O(n×m) per-table index scans used by sizing | #3155 added the map-backed `...FromMap` variants and rewrote all 9 call sites in one commit |
| `src/query/queryparser/helpers_protomsg.go`: `getOneofActiveNode` | Returned the active message of a protobuf `node` oneof | Born dead in #2108 — `GetStatementType` does its own `getOneofActiveField` walk |
| `src/srcdb/yugabytedb.go`: `getAllEnumTypesInSchema` | Listed enum types per schema for fall-forward/fall-back logical replication | #3242 folded enums into `getAllUserDefinedTypesInSchema` ("returns UDTs + Enums") and dropped the separate call |
| `src/schemadiff/tables_test.go`: `crossEngineSnapWithTables` | Test fixture building a mysql-typed snapshot for cross-engine gate tests | Landed already orphaned in #3648 — the tests using it were deleted during that PR's own API refactor (recorded in the file's comment) |
| `src/anon/sql_anonymizer_pg_objects_test.go`: `missingNodeCase.nodeType` field | Documentation-only field on a test-case struct | Never initialized and never read since introduction (#2904); all case literals are keyed, so removal cannot shift initializers |

Several of these were "born dead" (never had a caller in any commit) — staticcheck would have rejected them at introduction time had the config been working.

</details>

Verified the gate now works: temporarily reintroducing an `err = ...` assigned-but-never-checked shape makes `staticcheck` fail the build (SA4006), which it previously would not have.

No user-facing changes (no CLI, configuration, installation, or report changes; deleted code was unreachable).

- `go build ./...`, `go vet ./...` and `go vet -tags unit ./...` pass.
- `staticcheck ./...` and `staticcheck -tags unit ./...` (v2025.1.1, the CI-pinned version) report zero findings with the corrected config.
- `go test -tags unit ./...` passes with zero failures, now including the three previously-never-run tests in `type_mapping_test.go`.
- `go vet` compiles cleanly under `-tags integration`, `-tags integration_voyager_command`, and `-tags issues_integration` (relevant for the test-utils file split).
- Every U1000 deletion was verified unused under all build-tag combinations (helpers used only by tag-gated tests were tagged instead of deleted), and each was traced through git history to the commit that orphaned it.
- The `gofmt` sweep and its verification moved to stacked PR #3749.

No.

No.


makalaaneesh:

---

priyanshi-yb (approved):
LGTM

PR #3727, opened 2026-08-18
Fix: false negative case where subset of unique index columns being updated and some restructuring of code

priyanshi-yb:

Live import conflict detection missed unique-key conflicts when an UPDATE changed only a subset of a composite unique index (including partial unique indexes). Debezium's after-image (`Fields`) contains only changed columns, so the before-after check could not reconstruct the full index tuple and skipped the pair. Those events can land on different channels (different PKs), apply out of order, and fail with a unique-violation (`23505`).

This change builds a full after-image (`AfterFields`) at event unmarshal time by overlaying changed `Fields` on `BeforeFields` (`MergeBeforeAndChangedFields`), and uses that reconstructed tuple in the before-after conflict check. Unchanged index columns come from the before-image, so `REPLICA IDENTITY FULL` is required for this reconstruction to work. Incoming updates that do not touch any unique-index column still skip the before-after probe; they still run before-before (needed for partial unique indexes). Updates that only change the predicate and not UK columns continue to be indexed so those before-before matches can be found.

Conflict lookup is also restructured by event op (insert: before-after only; update: before-after then before-before; delete: no check), with NULLS DISTINCT tuples skipped before `computeConflictBucketKey`. Missing unique-index columns and DISTINCT NULLs that still reach key computation are now surfaced as errors instead of silently treated as non-indexable.

No command-line, configuration, installation, or report changes.

Live migration users with composite / partial unique indexes can now have previously missed UK conflicts serialized instead of failing import with a unique-violation when only some index columns change in an UPDATE.

Existing conflict-cache unit tests were updated to populate `AfterFields` the same way production unmarshal does (`withAfterFields`). They still cover before-after, before-before, composite indexes, NULLS [NOT] DISTINCT, same-PK exclusion, and missing-column error paths.

A new live-migration integration test covers the false-negative:

- `TestLiveMigrationWithSubsetOFPartialUNiqueIndexColumnsBeingChangedInUpdate` in `yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go`

That test uses a partial unique index `(c1, c2) WHERE most_recent`, recycles the same index slot across 500 UPDATE pairs where `c1` is unchanged (absent from `Fields`), and asserts via the `uniqueKeyConflictDetected=return("count")` failpoint that at least 500 conflicts are detected and import does not stop on `23505`. It requires `REPLICA IDENTITY FULL` on the source table so the unchanged index column is present in `before_fields`.

Conflict-cache unit tests were run with:
`go test -tags unit -run 'TestIndexTupleConflicts|TestEventsConfict|TestEventsConflict|TestConflictLookup|TestComputeConflictBucketKey|TestConflictWithMultipleIndexes|TestWaitUntilNoConflict' ./cmd/`

The new live-migration integration test was not run as part of writing this description.

Benchmark test 
1. with main
Screenshot 2026-08-24 at 5 45 19 PM

2. with this change
Screenshot 2026-08-24 at 5 45 12 PM

No.

No. `AfterFields` is computed in memory during JSON unmarshal and is not persisted.

---

| Component

### DIFF
--- a/yb-voyager/cmd/conflictDetectionCache.go
+++ b/yb-voyager/cmd/conflictDetectionCache.go
@@ -16,6 +16,7 @@ limitations under the License.
 package cmd
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 	"sync"
@@ -438,20 +439,22 @@ func (c *ConflictDetectionCache) checkBeforeAfterConflict(incomingEvent *tgtdb.E
 	}
 	key, err := computeConflictBucketKey(incomingEvent.TableNameTup, incomingEvent.AfterFields, index)
 	if err != nil {
-		return Conflict{}, err
+		return Conflict{}, fmt.Errorf("error computing conflict bucket key for before-after conflict for incoming event(vsn=%d) and index %s: %w", incomingEvent.Vsn, index.IndexName, err)
 	}
 	cachedEvents, err := c.getConflictingEventsWithDifferentPartitionKey(key, incomingEvent)
 	if err != nil {
-		return Conflict{}, err
+		return Conflict{}, fmt.Errorf("error getting conflicting events with different partition key for before-after conflict for incoming event(vsn=%d) and index %s: %w", incomingEvent.Vsn, index.IndexName, err)
 	}
-	if len(cachedEvents) > 0 {
-		for _, cachedEvent := range cachedEvents {
-			log.Debugf("conflict detected for table %s, index columns %v, between before value of cached-event1(vsn=%d, colVal=%s) and after value of incoming-event2(vsn=%d, colVal=%s)",
-				incomingEvent.TableNameTup.ForKey(), index.Columns,
-				cachedEvent.Vsn, formatUniqueIndexColumnValuesForLog(cachedEvent.BeforeFields, index.Columns),
-				incomingEvent.Vsn, formatUniqueIndexColumnValuesForLog(incomingEvent.AfterFields, index.Columns))
-		}
+	if len(cachedEvents) == 0 {
+		return Conflict{}, nil
 	}
+	for _, cachedEvent := range cachedEvents {
+		log.Debugf("conflict detected for table %s, index columns %v, between before value of cached-event1(vsn=%d, colVal=%s) and after value of incoming-event2(vsn=%d, colVal=%s)",
+			incomingEvent.TableNameTup.ForKey(), index.Columns,
+			cachedEvent.Vsn, formatUniqueIndexColumnValuesForLog(cachedEvent.BeforeFields, index.Columns),
+			incomingEvent.Vsn, formatUniqueIndexColumnValuesForLog(incomingEvent.AfterFields, index.Columns))
+	}
+
 	return Conflict{
 		indexName:         index.IndexName,
 		eventsConflicting: cachedEvents,
@@ -469,19 +472,20 @@ func (c *ConflictDetectionCache) checkBeforeBeforeConflict(incomingEvent *tgtdb.
 	}
 	key, err := computeConflictBucketKey(incomingEvent.TableNameTup, incomingEvent.BeforeFields, index)
 	if err != nil {
-		return Conflict{}, err
+		return Conflict{}, fmt.Errorf("error computing conflict bucket key for before-before conflict for incoming event(vsn=%d) and index %s: %w", incomingEvent.Vsn, index.IndexName, err)
 	}
 	cachedEvents, err := c.getConflictingEventsWithDifferentPartitionKey(key, incomingEvent)
 	if err != nil {
-		return Conflict{}, err
+		return Conflict{}, fmt.Errorf("error getting conflicting events with different partition key for before-before conflict for incoming event(vsn=%d) and index %s: %w", incomingEvent.Vsn, index.IndexName, err)
 	}
-	if len(cachedEvents) > 0 {
-		for _, cachedEvent := range cachedEvents {
-			log.Debugf("conflict detected for table %s, index columns %v, between before value of cached-event1(vsn=%d, colVal=%s) and before value of incoming-event2(vsn=%d, colVal=%s)",
-				incomingEvent.TableNameTup.ForKey(), index.Columns,
-				cachedEvent.Vsn, formatUniqueIndexColumnValuesForLog(cachedEvent.BeforeFields, index.Columns),
-				incomingEvent.Vsn, formatUniqueIndexColumnValuesForLog(incomingEvent.BeforeFields, index.Columns))
-		}
+	if len(cachedEvents) == 0 {
+		return Conflict{}, nil
+	}
+	for _, cachedEvent := range cachedEvents {
+		log.Debugf("conflict detected for table %s, index columns %v, between before value of cached-event1(vsn=%d, colVal=%s) and before value of incoming-event2(vsn=%d, colVal=%s)",
+			incomingEvent.TableNameTup.ForKey(), index.Columns,
+			cachedEvent.Vsn, formatUniqueIndexColumnValuesForLog(cachedEvent.BeforeFields, index.Columns),
+			incomingEvent.Vsn, formatUniqueIndexColumnValuesForLog(incomingEvent.BeforeFields, index.Columns))
 	}
 	return Conflict{
 		indexName:         index.IndexName,
@@ -567,7 +571,8 @@ func computeConflictBucketKey(table sqlname.NameTuple, fields map[string]*string
 	for _, column := range index.Columns {
 		val, exists := fields[column]
 		if !exists {
-			//In case the incoming event like update is not having this field in after fields then it is not indexable as the unique key is not changed so it won't conflict with anything
+			//this is not expected as now eveyrthing should be present in the fields as we pass before/after fields which are all column values so we should have that
+			//hence erroring out here
 			return "", goerrors.Errorf("column %s is missing from fields", column)
 		}
 		b.WriteByte(0) //separator for the field name and value
--- a/yb-voyager/cmd/conflictDetectionCache.go
+++ b/yb-voyager/cmd/conflictDetectionCache.go
@@ -217,12 +217,9 @@ func (c *ConflictDetectionCache) indexEventLocked(event *tgtdb.Event) error {
 			//		if uniqueINdex.NullsDistinct AND ANY unique key column value is NULL:
 			continue
 		}
-		key, ok, err := computeConflictBucketKey(event.TableNameTup, event.BeforeFields, index)
+		key, err := computeConflictBucketKey(event.TableNameTup, event.BeforeFields, index)
 		if err != nil {
 			return goerrors.Errorf("error computing conflict bucket key for table %s, index %s: %v", event.TableNameTup.ForKey(), index.IndexName, err)
-		}
-		if !ok {
-			continue
 		}
 		bucket := c.ukLookup[key]
 		if bucket == nil {
@@ -394,7 +391,7 @@ FindConflicts()
 */
 func (c *ConflictDetectionCache) findValueConflictLocked(incomingEvent *tgtdb.Event) ([]Conflict, error) {
 	if incomingEvent.Op == "d" {
-		return []Conflict{}, nil
+		return []Conflict{}, goerrors.Errorf("incoming event is a delete event")
 	}
 	uniqueIndexes, _ := c.tableToUniqueIndexes.Get(incomingEvent.TableNameTup)
 	if len(uniqueIndexes) == 0 {
@@ -439,12 +436,9 @@ func (c *ConflictDetectionCache) checkBeforeAfterConflict(incomingEvent *tgtdb.E
 		//		if uniqueINdex.NullsDistinct AND ANY unique key column value is NULL:
 		return Conflict{}, nil
 	}
-	key, ok, err := computeConflictBucketKey(incomingEvent.TableNameTup, incomingEvent.AfterFields, index)
+	key, err := computeConflictBucketKey(incomingEvent.TableNameTup, incomingEvent.AfterFields, index)
 	if err != nil {
 		return Conflict{}, err
-	}
-	if !ok {
-		return Conflict{}, nil
 	}
 	cachedEvents, err := c.getConflictingEventsWithDifferentPartitionKey(key, incomingEvent)
 	if err != nil {
@@ -473,12 +467,9 @@ func (c *ConflictDetectionCache) checkBeforeBeforeConflict(incomingEvent *tgtdb.
 		//		if uniqueINdex.NullsDistinct AND ANY unique key column value is NULL:
 		return Conflict{}, nil
 	}
-	key, ok, err := computeConflictBucketKey(incomingEvent.TableNameTup, incomingEvent.BeforeFields, index)
+	key, err := computeConflictBucketKey(incomingEvent.TableNameTup, incomingEvent.BeforeFields, index)
 	if err != nil {
 		return Conflict{}, err
-	}
-	if !ok {
-		return Conflict{}, nil
 	}
 	cachedEvents, err := c.getConflictingEventsWithDifferentPartitionKey(key, incomingEvent)
 	if err != nil {
@@ -565,9 +556,9 @@ const (
 //
 // Values are length-prefixed so distinct tuples never collide (e.g. {"ab",""} vs
 // {"a","b"}); NULLs under NULLS NOT DISTINCT use a dedicated sentinel.
-func computeConflictBucketKey(table sqlname.NameTuple, fields map[string]*string, index tgtdb.UniqueIndex) (string, bool, error) {
+func computeConflictBucketKey(table sqlname.NameTuple, fields map[string]*string, index tgtdb.UniqueIndex) (string, error) {
 	if fields == nil {
-		return "", false, goerrors.Errorf("fields are nil")
+		return "", goerrors.Errorf("fields are nil")
 	}
 	var b strings.Builder
 	b.WriteString(table.ForKey())
@@ -577,13 +568,13 @@ func computeConflictBucketKey(table sqlname.NameTuple, fields map[string]*string
 		val, exists := fields[column]
 		if !exists {
 			//In case the incoming event like update is not having this field in after fields then it is not indexable as the unique key is not changed so it won't conflict with anything
-			return "", false, goerrors.Errorf("column %s is missing from fields", column)
+			return "", goerrors.Errorf("column %s is missing from fields", column)
 		}
 		b.WriteByte(0) //separator for the field name and value
 		if val == nil {
 			if !index.NullsNotDistinct {
 				// default NULLS DISTINCT: NULLs never conflict,so it should never come into this compute
-				return "", false, goerrors.Errorf("column %s has null value but index is not NULLS NOT DISTINCT", column)
+				return "", goerrors.Errorf("column %s has null value but index is not NULLS NOT DISTINCT", column)
 			}
 			b.WriteString(conflictBucketNull) //placeholder for the null value
 			continue
@@ -593,7 +584,7 @@ func computeConflictBucketKey(table sqlname.NameTuple, fields map[string]*string
 		b.WriteByte(':')                       //separator for the length and value
 		b.WriteString(*val)                    //value
 	}
-	return b.String(), true, nil
+	return b.String(), nil
 }
 
 func (c *ConflictDetectionCache) RemoveEvents(events ...*tgtdb.Event) {
--- a/yb-voyager/cmd/conflictDetectionCache.go
+++ b/yb-voyager/cmd/conflictDetectionCache.go
@@ -167,6 +167,19 @@ func NewConflictDetectionCache(tableToUniqueIndexes *utils.StructMap[sqlname.Nam
 	return c
 }
 
+/*
+Put()
+
+	if ! U/D
+		return
+	if U:
+		if changedColumns <intersection> (UK columns <union> predictate columns) == EMPTY:
+		Can't add this check until we have predicate columns as we still need the event in index even if its unique columns are unchanged for predicate cases 
+			return
+		if uniqueINdex.NullsDistinct AND ANY unique key column value is NULL:
+			return
+		add-to-cache (uklookup) <---- computeKey (beforeFields) (all columns should be present)
+*/
 func (c *ConflictDetectionCache) Put(event *tgtdb.Event) error {
 	if event.BeforeFields == nil || event.Op == "c" {
 		//put can only insert an event in the cache if its delete/update event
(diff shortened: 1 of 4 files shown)


yb-voyager/src/tgtdb/event.go
```diff
@@ -87,6 +88,7 @@ func (e *Event) UnmarshalJSON(data []byte) error {
 	e.Fields = rawEvent.Fields
 	e.BeforeFields = rawEvent.BeforeFields
 	e.ExporterRole = rawEvent.ExporterRole
+	e.AfterFields = MergeBeforeAndChangedFields(e.Op, e.BeforeFields, e.Fields)
```
makalaaneesh: [AI] **`AfterFields` is heap-allocated per UPDATE event on the CDC hot path, even when conflict detection never uses it.**

`UnmarshalJSON` runs for every streamed event — a documented CDC hot path (root `.cursor/BUGBOT.md`, "Performance-Critical (Hot) Paths"). For `op=="u"`, `MergeBeforeAndChangedFields` allocates a brand-new map and copies all before-image columns plus changed columns into it, for **every** update event.

But `AfterFields` is consumed in exactly one place: `checkBeforeAfterConflict` in `conflictDetectionCache.go`. Conflict detection only runs for tables that have unique indexes and use `PARTITION_BY_PK`/`CUSTOM` (`shouldHandleConflicts`); it never runs for tables without unique indexes, for `PARTITION_BY_TABLE`, or for Oracle sources. So for the majority of update events this is a pure-waste per-event allocation that the old code (which used `incomingEvent.Fields` directly) did not incur.

Failure case: high-volume update stream on wide tables → one full map copy per event → GC pressure on the streaming import path.

Suggested fix: compute the after-image lazily inside `checkBeforeAfterConflict` (or memoize on the `Event` on first use) so the allocation only happens when a unique index actually needs it, instead of eagerly in `UnmarshalJSON`.

---

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -504,13 +577,13 @@ func computeConflictBucketKey(table sqlname.NameTuple, fields map[string]*string
 		val, exists := fields[column]
 		if !exists {
 			//In case the incoming event like update is not having this field in after fields then it is not indexable as the unique key is not changed so it won't conflict with anything
-			return "", false, nil
+			return "", false, goerrors.Errorf("column %s is missing from fields", column)
```
makalaaneesh: [AI] **A previously-tolerated "missing unique-index column" case is now a fatal error that halts live import.**

`computeConflictBucketKey` previously returned `("", false, nil)` when an index column was absent from the fields map, so the caller skipped that index and import continued — the known "needs `REPLICA IDENTITY FULL`" limitation degraded gracefully. This PR turns it into a hard error, which propagates: `findValueConflictLocked` → `WaitUntilNoConflict`/`Put` → `handleEvent` returns the error and stops the whole import. The updated unit test `TestEventsConfict_MissingColumnInEvent` codifies this new fatal behavior (was `require.Len(conflicts, 0)`, now `require.Error`).

Note the comment on the line just above still reads "…not indexable as the unique key is not changed so it won't conflict with anything" — i.e. it documents the old graceful-skip contract while the code now errors, so the two now contradict each other.

Reachability: under PostgreSQL the missing-column case can't occur because `REPLICA IDENTITY FULL` is an enforced live-migration prerequisite (`postgres.go`). But for **MySQL** sources nothing in the repo enforces `binlog_row_image=FULL`; with `binlog_row_image=minimal` the before-image (and thus `AfterFields`) lacks unchanged UK columns for a subset-column update, so a single such event would now abort a long-running migration where it previously continued.

Suggested fix: keep the missing-column case a non-fatal skip (`return "", false, nil`) — or gate the hard error behind a verified "full before-image" invariant — and update the stale comment accordingly. Reserve the error for genuinely impossible states. (The NULL-under-DISTINCT error just below is separately guarded by `anyUniqueIndexColumnValueIsNull` at every call site, so it is effectively defensive; a fatal error there is lower-risk.)

---

makalaaneesh (commented):

---

yb-voyager/src/tgtdb/event.go
priyanshi-yb: @makalaaneesh I think we discussed this and decided not to do it lazily, but Claude also suggesting the same to avoid unnecessary memory allocations. Let me know what you think

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: intentional

yb-voyager/src/tgtdb/event.go
makalaaneesh: can you run the benchmark tests to see if there is any meaningful drop-off?

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -347,73 +364,164 @@ func (c *ConflictDetectionCache) findConflictLocked(incomingEvent *tgtdb.Event)
...
+				check before-after conflicts <---- computeKey (afterFields) (possible that only a subset of columns are present)
+
+			if uniqueIndex has partial predicate: (before-before)
+				if uniqueINdex.NullsDistinct AND ANY unique key column value (before fields) is NULL:
+					skip-check
+				check before-before conflicts.  <---- computeKey (beforeFields) (all columns should be present)
+*/
 func (c *ConflictDetectionCache) findValueConflictLocked(incomingEvent *tgtdb.Event) ([]Conflict, error) {
+	if incomingEvent.Op == "d" {
+		return []Conflict{}, nil
```
makalaaneesh: error, because this is unexpected?

yb-voyager/cmd/conflictDetectionCache.go
makalaaneesh: do we really need the `bool` ?

makalaaneesh (commented):
early comments; still need to go through it in detail

yb-voyager/src/tgtdb/event.go
```diff
@@ -99,6 +101,27 @@ func (e *Event) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// MergeBeforeAndChangedFields builds the after-image of a CDC event:
+// insert ("c") is Fields, update ("u") is BeforeFields overlaid with changed Fields, delete ("d") is nil.
+func MergeBeforeAndChangedFields(op string, beforeFields, changeFields map[string]*string) map[string]*string {
```
makalaaneesh: nit: for c,d it's not doing any merge, so call this `generateAfterFields`

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -347,73 +361,158 @@ func (c *ConflictDetectionCache) findConflictLocked(incomingEvent *tgtdb.Event)
...
+	}
+	if len(cachedEvents) > 0 {
+		for _, cachedEvent := range cachedEvents {
+			log.Debugf("conflict detected for table %s, index columns %v, between before value of cached-event1(vsn=%d, colVal=%s) and after value of incoming-event2(vsn=%d, colVal=%s)",
+				incomingEvent.TableNameTup.ForKey(), index.Columns,
+				cachedEvent.Vsn, formatUniqueIndexColumnValuesForLog(cachedEvent.BeforeFields, index.Columns),
+				incomingEvent.Vsn, formatUniqueIndexColumnValuesForLog(incomingEvent.AfterFields, index.Columns))
+		}
+	}
+	return Conflict{
```
makalaaneesh, line 458: shouldn't this be inside the `if len(cachedEvents) > 0` block?

yb-voyager/src/tgtdb/event.go
makalaaneesh: pls put the before/after benchmarks in the descriptions of all PRs where we're making significant changes

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -347,73 +361,158 @@ func (c *ConflictDetectionCache) findConflictLocked(incomingEvent *tgtdb.Event)
...
+		indexName:         index.IndexName,
+		eventsConflicting: cachedEvents,
+		indexColumns:      index.Columns,
+		matchType:         "before-after",
+		matchedValue:      formatUniqueIndexColumnValuesForLog(incomingEvent.AfterFields, index.Columns),
```
makalaaneesh, line 463: [AI] **Hot-path regression: `matchedValue` is formatted on every probe, even when there are zero conflicts.**

Both `checkBeforeAfterConflict` and `checkBeforeBeforeConflict` call `formatUniqueIndexColumnValuesForLog` unconditionally in the returned `Conflict` literal. The pre-refactor code only built `matchedValue` inside `if len(cachedEvents) > 0`. The no-conflict case is the overwhelming majority, so every streamed insert/update on a unique-index table now pays 1–2 `strings.Builder` string builds per unique index per event that the old code did not (root `.cursor/BUGBOT.md` flags `conflictDetectionCache` as a previously-regressed hot path).

Failure case: 1M streamed updates on a table with 2 unique indexes → ~4M needless string allocations vs. the old code.

Suggested fix — return early before any formatting:
```go
if len(cachedEvents) == 0 {
    return Conflict{}, nil
}
for _, cachedEvent := range cachedEvents { log.Debugf(...) }
return Conflict{
    indexName:         index.IndexName,
    eventsConflicting: cachedEvents,
    indexColumns:      index.Columns,
    matchType:         "before-after",
    matchedValue:      formatUniqueIndexColumnValuesForLog(incomingEvent.AfterFields, index.Columns),
}, nil
```
Same change in `checkBeforeBeforeConflict`. Running `BenchmarkCDCIngest` (`go test -tags cdc_benchmark -bench CDCIngest ...`) before/after would confirm the impact.

---

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -347,73 +361,158 @@ func (c *ConflictDetectionCache) findConflictLocked(incomingEvent *tgtdb.Event)
...
 }
 
+func (c *ConflictDetectionCache) checkBeforeAfterConflict(incomingEvent *tgtdb.Event, index tgtdb.UniqueIndex) (Conflict, error) {
+	if !index.NullsNotDistinct && anyUniqueIndexColumnValueIsNull(incomingEvent.AfterFields, index.Columns) {
+		//		if uniqueINdex.NullsDistinct AND ANY unique key column value is NULL:
+		return Conflict{}, nil
+	}
+	key, err := computeConflictBucketKey(incomingEvent.TableNameTup, incomingEvent.AfterFields, index)
+	if err != nil {
+		return Conflict{}, err
```
makalaaneesh: [AI] **Before-after/before-before errors lose table and index context.**

`indexEventLocked` wraps `computeConflictBucketKey` errors with the table and index name, but `checkBeforeAfterConflict` and `checkBeforeBeforeConflict` return the bare error (`column c1 is missing from fields`), which `handleEvent` wraps with only the VSN. Since the missing-column case is now intentionally fatal (per the discussion on the other [AI] thread), this bare message is all the user gets when import aborts — no table, no index, no hint that `REPLICA IDENTITY FULL` on the source table is the likely fix.

Suggested fix (both helpers):
```go
if err != nil {
    return Conflict{}, goerrors.Errorf("computing conflict bucket key (before-after) for table %s, index %s: %v (is REPLICA IDENTITY FULL set on the source table?)",
        incomingEvent.TableNameTup.ForKey(), index.IndexName, err)
}
```

---

yb-voyager/cmd/conflictDetectionCache_test.go
```diff
@@ -76,47 +76,57 @@ func testTableTuple() sqlname.NameTuple {
 	return sqlname.NameTuple{CurrentName: oname, TargetName: oname}
 }
 
+// withAfterFields sets AfterFields the same way Event.UnmarshalJSON does:
+// insert ("c") -> Fields, update ("u") -> BeforeFields merged with Fields, delete ("d") -> nil.
+func withAfterFields(e *tgtdb.Event) *tgtdb.Event {
```
makalaaneesh, line 81: [AI] **No unit test exercises the actual fixed scenario (subset-of-composite-UK update).**

All unit-test changes here are mechanical (`withAfterFields` wraps, error-assertion flips). There is no cache-level test of the false negative this PR fixes: an incoming `"u"` whose `Fields` contains only a subset of a composite index's columns, with the unchanged column reconstructed from `BeforeFields`, matching a cached event's before-tuple. The only coverage of the `anyUniqueIndexColumnChanged` gate + `GenerateAfterFields` merge interplay is the new integration test. If someone later optimizes away the merge or flips the gate, `go test -tags unit` stays green.

Suggested test:
```go
func TestEventsConflict_SubsetOfCompositeUKColumnsChanged(t *testing.T) {
	cache := newConflictCacheForTest([][]string{{"c1", "c2"}})
	cached := withAfterFields(&tgtdb.Event{ // FREE: predicate-only update, slot (100,1000)
		Vsn: 1, Op: "u", TableNameTup: testTableTuple(),
		Key:          map[string]*string{"id": strPtr("1")},
		BeforeFields: map[string]*string{"c1": strPtr("100"), "c2": strPtr("1000")},
		Fields:       map[string]*string{"most_recent": strPtr("false")},
	})
	require.NoError(t, cache.Put(cached))
	incoming := withAfterFields(&tgtdb.Event{ // TAKE: c1 unchanged => absent from Fields
		Vsn: 2, Op: "u", TableNameTup: testTableTuple(),
		Key:          map[string]*string{"id": strPtr("2")},
		BeforeFields: map[string]*string{"c1": strPtr("100"), "c2": strPtr("21")},
		Fields:       map[string]*string{"c2": strPtr("1000"), "most_recent": strPtr("true")},
	})
	conflicts := findConflictForTest(t, cache, incoming)
	require.Len(t, conflicts, 1) // pre-fix: 0
	assert.Equal(t, int64(1), conflicts[0].Vsn)
}
```
A small table-driven test of `GenerateAfterFields` itself (c/u/d, nil maps, overlap) would also be cheap insurance.

---

yb-voyager/src/tgtdb/event.go
```diff
@@ -44,6 +44,7 @@ type Event struct {
 	Key                 map[string]*string
 	Fields              map[string]*string //all the column values of the row - worst
 	BeforeFields        map[string]*string //all the column values of the row - worst
+	AfterFields         map[string]*string //all the column values of the row - worst //all fields of the row adn columns that are chnges in fields are updated to fields value
```
makalaaneesh, line 47: [AI] **`Event.Copy()` and `Event.String()` silently drop the new `AfterFields`.**

`Copy()` reproduces every other field of `Event` but not `AfterFields`; `String()` doesn't print it either. Today the only `Copy()` caller (`ConflictDetectionCache.Put`) doesn't need the after-image, but `Copy` is the event's general deep-copy — the next caller that copies an event before conflict lookup gets a nil after-image and silently reintroduces the exact false negative this PR fixes (`anyUniqueIndexColumnValueIsNull` returns false on nil, then `computeConflictBucketKey` errors with "fields are nil").

Related inconsistency worth documenting: for `"c"` events `AfterFields` **aliases** `Fields` (same map reference), while for `"u"` it's a fresh map — so a later in-place mutation of `Fields` (e.g. `ConvertEvent`) changes `AfterFields` for inserts but not for updates.

Suggested fix: add `AfterFields: lo.MapEntries(e.AfterFields, idFn)` to `Copy()`, include it in `String()`, and document (or remove) the `"c"` aliasing in `GenerateAfterFields`. Also, this field's comment has typos ("adn", "chnges") and a copy-pasted "- worst"; consider: `// full after-image: Fields for inserts; BeforeFields overlaid with changed Fields for updates; nil for deletes. Derived in UnmarshalJSON, never serialized.`

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -2116,3 +2116,150 @@ func TestLiveMigrationCustomCdcPartitionKeyNoConflict(t *testing.T) {
 	err = lm.WaitForCutoverComplete(0, 30)
 	testutils.FatalIfError(t, err, "failed to wait for cutover complete")
 }
+
+
+func TestLiveMigrationWithSubsetOFPartialUNiqueIndexColumnsBeingChangedInUpdate(t *testing.T) {
```
makalaaneesh, line 2110: [AI] **New integration test has not been executed (per the PR description).**

The PR description states the new live-migration integration test was not run. It compiles (`go vet -tags failpoint_import` passes), but it is the only executable proof of the end-to-end fix — there is currently no cache-level unit test for the subset-column scenario either. The `>= 500` assertion thresholds, the failpoint plumbing, and the `REPLICA IDENTITY FULL` setup are all unverified until it runs.

Please run it (requires `failpoint-ctl enable` + built binary) and report the result before merge.

Minor, same function: the name has typos (`...SubsetOFPartialUNique...` → `...SubsetOfPartialUnique...`), the comment above `WaitForForwardStreamingComplete` says "500 loops x {1 insert, 2 updates, 1 delete}" but the loop performs no deletes (and the expectation asserts `Deletes: 0`), and the file is missing its trailing newline.

---

makalaaneesh (commented):
[AI] Automated review — 5 findings (0 critical, 5 warning). See inline comments. One additional hot-path finding (per-update `AfterFields` allocation in `UnmarshalJSON`) was skipped because an existing [AI] comment already covers it.

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -347,73 +361,158 @@ func (c *ConflictDetectionCache) findConflictLocked(incomingEvent *tgtdb.Event)
...
-						incomingEvent.TableNameTup.ForKey(), index.Columns,
-						cachedEvent.Vsn, formatUniqueIndexColumnValuesForLog(cachedEvent.BeforeFields, index.Columns),
-						incomingEvent.Vsn, formatUniqueIndexColumnValuesForLog(incomingEvent.Fields, index.Columns))
+			if len(conflict.eventsConflicting) > 0 {
+				totalConflictInfo = append(totalConflictInfo, conflict)
+			}
+		case "u":
+			if anyUniqueIndexColumnChanged(incomingEvent.Fields, index.Columns) {
+				conflict, err := c.checkBeforeAfterConflict(incomingEvent, index)
+				if err != nil {
```
makalaaneesh, line 416: in this function, we should add context (i.e. wrap error) before returning because there are various cases here (insert-before-after, update-before-after, etc)

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -504,13 +568,13 @@ func computeConflictBucketKey(table sqlname.NameTuple, fields map[string]*string
 		val, exists := fields[column]
 		if !exists {
 			//In case the incoming event like update is not having this field in after fields then it is not indexable as the unique key is not changed so it won't conflict with anything
```
makalaaneesh: fix comment

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -2116,3 +2116,150 @@ func TestLiveMigrationCustomCdcPartitionKeyNoConflict(t *testing.T) {
...
+
+	conflictStats, err := testutils.ReadUniqueKeyConflictStats(uniqueKeyConflictStatsPath)
+	testutils.FatalIfError(t, err, "failed to read unique key conflict stats")
+
+	// The whole point of this test: exactly one true conflict per loop (500).
+	// Pre-fix this is ~0 (all missed). Post-fix it is 500.
+	require.GreaterOrEqual(t, conflictStats.Total, 500,
+		"subset-column partial-index delta should detect 1 conflict per loop (500 loops)")
+	require.GreaterOrEqual(t, conflictStats.ByTable[`"test_schema"."test_false_negative"`], 500,
+		"test_false_negative should detect 500 subset-column UK conflicts")
```
makalaaneesh: isn't this prone to intermittent failures? i.e. expecting 500 exact conflicts (i don't understand why `GreaterOrEqual`)

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -2116,3 +2116,150 @@ func TestLiveMigrationCustomCdcPartitionKeyNoConflict(t *testing.T) {
...
+
+	conflictStats, err := testutils.ReadUniqueKeyConflictStats(uniqueKeyConflictStatsPath)
+	testutils.FatalIfError(t, err, "failed to read unique key conflict stats")
+
+	// The whole point of this test: exactly one true conflict per loop (500).
+	// Pre-fix this is ~0 (all missed). Post-fix it is 500.
+	require.GreaterOrEqual(t, conflictStats.Total, 0,
+		"subset-column partial-index delta should detect 1 conflict per loop (500 loops)")
+	require.GreaterOrEqual(t, conflictStats.ByTable[`"test_schema"."test_false_negative"`], 0,
+		"test_false_negative should detect 500 subset-column UK conflicts")
```
makalaaneesh: why `GreaterOrEqual` 0? 0 is not right; neither is 500 (because it could lead to race conditions).

priyanshi-yb: yeah, I have fixed it to be Greater than 0 check

yb-voyager/cmd/conflictDetectionCache.go
priyanshi-yb: done

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -547,10 +543,8 @@ func TestLiveMigrationWithUniqueKeyValuesWithPartialPredicateConflictDetectionCa
 	conflictStats, err := testutils.ReadUniqueKeyConflictStats(uniqueKeyConflictStatsPath)
 	testutils.FatalIfError(t, err, "failed to read unique key conflict stats")
 
-	require.LessOrEqual(t, conflictStats.Total, 3500, "true-positive delta should produce at most 2500 UK conflicts")
-	require.LessOrEqual(t, conflictStats.ByTable[`"test_schema"."test_live"`], 3500, "test_live should have at most 2500 UK conflicts")
-	require.GreaterOrEqual(t, conflictStats.Total, 0, "true-positive delta should produce at least 0 UK conflicts")
-	require.GreaterOrEqual(t, conflictStats.ByTable[`"test_schema"."test_live"`], 0, "test_live should have at least 0 UK conflicts")
+	require.Greater(t, conflictStats.Total, 0, "true-positive delta should produce UK conflicts")
+	require.Greater(t, conflictStats.ByTable[`"test_schema"."test_live"`], 0, "test_live should produce UK conflicts")
```
makalaaneesh, line 547: if we have an upper bound then IMO we should keep the `LessOrEqual` checks. It will guard against potential false-positives

yb-voyager/src/tgtdb/event.go
priyanshi-yb: added

makalaaneesh (approved):
LGTM! with minor comment on the test assertion bounds

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
priyanshi-yb: Got it, I will have to figure out if that works for all cases or not. I will check it out and make the changes in a separate PR if that's okay?

PR #3729, opened 2026-08-18
Custom key Follow up1: guardrail for checking if the columns in the key exist in the table and not allow custom key for tables having expression based unique indexes 

priyanshi-yb:

This PR adds import-start guardrails for tables routed by a **custom CDC partition key** (`--cdc-partition-key-overrides <table>:(col1,col2)`), the first follow-up to the custom-partition-key feature (see `cmd/cdc_partition_key_followups.md`, Follow-up 1). It hardens the configuration so misconfigurations are caught up front (before snapshot import) instead of failing per-event during streaming.

Specifically, in `live_migration_cdc_partition_strategy.go`:
- **Reject `custom` on expression-based unique index tables.** `resolveEffectiveCdcPartitionKeys` previously rejected only `pk` on such tables; it now rejects `custom` too, since both route by stored column *values* and neither can protect an index whose conflicting value is an expression output rather than a stored column.
- **Validate custom key columns exist.** A new `validateCustomPartitionKeyTables` hard-fails at import start if any custom key column does not exist on the table (queried via `tdb.GetListOfTableAttributes`), with an actionable error listing the missing and available columns.

These checks query the target DB, so they run only on the first import; on resume the config is unchanged (already enforced by `validateCdcPartitioningStrategyUnchanged`), so re-querying is skipped.

No new CLI flags or configuration — `--cdc-partition-key` / `--cdc-partition-key-overrides` already exist. Behavior change only:
1. Command line: `import data` now fails fast at import start with clear errors when a custom partition key is invalid — using `custom` on a table with an expression-based unique index, or naming a custom key column that doesn't exist on the table.

No changes to installation or reports.

- New unit test in `cmd/live_migration_test.go`:
  - `TestResolveEffectiveCdcPartitionKeys` subtest "rejects override custom on expression-UK table".
- New live integration tests (`//go:build failpoint_import`) in `src/testlivemigration/live_migration_uk_conflict_test.go`, run against real PostgreSQL source + YugabyteDB target containers:
  - `TestLiveMigrationCdcPartitionKeyRejectsCustomOnExpressionUniqueIndex` — exercises the real expression-UK detection on the target.
  - `TestLiveMigrationCdcPartitionKeyRejectsCustomKeyColumnNotOnTable` — column-existence hard-fail.
- `go vet -tags failpoint_import ./src/testlivemigration/` and unit tests pass.

No.

No. This follow-up adds validation logic and tests only; it does not change any on-disk structure (the `TableToCDCPartitionKey` MetaDB/Import Data State shape landed with the base feature already in `main`).


yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -181,9 +181,12 @@ func resolveEffectiveCdcPartitionKeys(
 
 	for _, t := range tableNames {
 		override, _ := result.Get(t)
-		if override.Strategy == PARTITION_BY_PK {
+		// Both pk and custom route by column *values*; an expression-based unique index
+		// cannot be protected by value routing (its conflicting value is the expression
+		// output, not a stored column), so neither strategy is safe on such tables.
+		if override.Strategy == PARTITION_BY_PK || override.Strategy == PARTITION_BY_CUSTOM {
```
makalaaneesh: nit: it's equivalent but != TABLE is a better check

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -447,3 +461,53 @@ func getExpressionUniqueIndexTables(tableNames []sqlname.NameTuple) ([]sqlname.N
 
 	return expressionUniqueIndexTables, nil
 }
+
+// validateCustomPartitionKeyTables runs the import-start guardrails for tables routed by a
+// custom partition key (see cdc_partition_key_followups.md, Follow-up 1):
+//   - hard-fail if any custom key column does not exist on the table, so misconfiguration is
+//     caught up front instead of erroring per-event in hashEvent, and
+//   - warn (do not refuse) when the custom key does not cover a unique index worth protecting:
```
makalaaneesh: [AI] **Docstring & call-site comment describe a "coverage warning" that isn't implemented.**

The function docstring here (line 469: `warn (do not refuse) when the custom key does not cover a unique index worth protecting`) and the call-site comment (line 295: `column existence + coverage warning`) both state that `validateCustomPartitionKeyTables` emits a coverage warning in addition to the column-existence hard-fail.

The implementation only performs the column-existence check — there is no `PrintAndLogfWarning`/warning code in the function (grep for `coverage`/`worth protecting`/`Warning` in this file finds only these two comment lines). The commit history shows the warning was intentionally dropped from this PR, but the comments were not updated.

Why it matters: per the root `.cursor/BUGBOT.md`, comments must describe actual behavior and leftover artifacts should be removed before merging. As written, a future maintainer will believe voyager warns users when a custom key fails to cover a protectable unique index, when no such warning is produced — a misunderstanding about what safety net exists for custom-key routing.

Suggested fix: drop the coverage-warning bullet from the docstring (line 469) and remove `+ coverage warning` from line 295, so both comments describe only the column-existence guardrail. If the coverage warning is planned for a later follow-up, replace the bullet with a `TODO (Follow-up 1.x): coverage warning not yet implemented` note rather than describing it as current behavior.

---

makalaaneesh (commented):

---

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -447,3 +461,53 @@ func getExpressionUniqueIndexTables(tableNames []sqlname.NameTuple) ([]sqlname.N
...
+		}
+		columnSet := make(map[string]bool, len(tableColumns))
+		for _, c := range tableColumns {
+			columnSet[c] = true
+		}
+		var missing []string
+		for _, c := range override.Columns {
+			if !columnSet[c] {
+				missing = append(missing, c)
+			}
```
makalaaneesh, line 500: How will this work with case sensitive column names? 
Shouldn't we use `AttributeNameRegistry` here to lookup/match/etc?

priyanshi-yb: Yeah we might have to use it but its not implemented yet, its a separate task for the case-sensitive column names

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
makalaaneesh: @priyanshi-yb can you address this?

makalaaneesh (approved):
LGTM. pls address the AI comment before merging

PR #3736, opened 2026-08-20
Document metrics-port in config templates

JoeSouthan:

`--metrics-port` is config-bindable for every export/import data section (allow-listed in `cmd/config.go`), but it was missing from all shipped YAML config templates under `config-templates/`. This adds a commented `metrics-port` block, following the existing template convention, to every export/import data section that accepts it: `export-data`/`export-data-from-source`, `export-data-from-target`, `import-data`/`import-data-to-target`, `import-data-to-source`, `import-data-to-source-replica`, and `import-data-file`. Each section uses a distinct example port so a config running multiple commands concurrently doesn't collide on the same port.

The deprecated `prometheus-metrics-port` alias is intentionally left undocumented since it's a hidden, deprecated flag.

No behavioral changes. Documentation-only: adds commented example config to the shipped templates. No changes to the command line, configuration semantics, installation, or reports.

Verified each edited template still parses as valid YAML (`yq e '.' <file>`), and confirmed via `git diff` that every change is a pure addition with no unrelated lines touched. Cross-checked the added `metrics-port` key against the allow-lists in `cmd/config.go` for each section.

No.

No.

### DIFF
--- a/yb-voyager/config-templates/live-migration-with-fall-back.yaml
+++ b/yb-voyager/config-templates/live-migration-with-fall-back.yaml
@@ -370,6 +370,10 @@ export-data-from-source:
   ### Default - info
   # log-level: info                                            
 
+  ### Port to expose Prometheus metrics on (0 disables). Serves GET /metrics.
+  ### Default - 0
+  # metrics-port: 9105
+
 # ---------------------------------------------------------
 # Import Data Configuration
 # ---------------------------------------------------------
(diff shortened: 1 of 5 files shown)


makalaaneesh:

---

yb-voyager/config-templates/live-migration-with-fall-back.yaml
```diff
@@ -370,6 +370,10 @@ export-data-from-source:
   ### Default - info
   # log-level: info                                            
 
+  ### Port to expose Prometheus metrics on (0 disables). Serves GET /metrics.
+  ### Default - 0
+  # metrics-port: 9105
```
makalaaneesh, line 375: nit: 
I would suggest the following ports just so that it maps mentally to the order in which they will be executed. 
- export-data-from-source, import-data-to-target: 9101,9102
- export-data-from-target: 9103
- import-data-to-source/import-data-to-source-replica: 9104 (they will never be used together since they are different workflows)

- import-data-file: 9105

PR #3737, opened 2026-08-21
Make yb-voyager log rotation configurable (log-max-size-mb, log-max-backups)

jasonriddell:

Addresses #3679 ("Make log settings configurable").

**Scope reduced from the previous revision of this PR.** `--log-dir` has been dropped
entirely per @ShivanshGahlot's review — there was no user or customer driver behind it, and
it was the source of most of the complexity (log-dir existence/permission checks, concurrent
runs colliding on one file name, `end migration`'s log backup reading the default location,
error messages elsewhere hardcoding `<export-dir>/logs`). If a real need for it appears, it
can be dug into properly in its own PR. The branch was rebuilt from `main` rather than
patched, so what's left is only the part that motivated the issue: not losing log messages
when files wrap around.

Two settings, each available as a CLI flag or a config-file key (global **or** per-command,
matching `log-level`):

| Setting | Default | Notes |
| --- | --- | --- |
| `--log-max-size-mb` | 200 | size threshold before a log file is rotated |
| `--log-max-backups` | 10 | rotated files retained; **`-1` retains all** |

Both default to the exact previously hardcoded values, so nothing changes for anyone not
setting them.

**Why `-1` and not `0` for "retain all":** lumberjack itself treats `MaxBackups == 0` as
"retain all files". Accepting `0` would therefore silently invert the intent of a user
passing it to *cap* disk usage — they'd get unbounded growth of 200 MB files instead. So
`ValidateLogSettings()` rejects `0` for both settings, and `LumberjackMaxBackups()`
translates the `-1` sentinel at the two points where a `lumberjack.Logger` is built.

Also in scope:

- **Debezium logs.** `debezium-<role>.log` carried the same hardcoded 200/10
  (`src/dbzm/dbzm.go`). It now honours these settings too — otherwise "retain all" would
  silently not hold for live migration, the most log-heavy flow. (Raised in review.)
- **Deduplicated** the 3x-copy-pasted `log-level` flag registration into `registerLogFlags()`
  (root.go / assess-migration-bulk / get-data-migration-report).
- **Validation ordering.** Collapsed the two copies of the pre-logging validation into
  `validateLogFlags()` and hoisted it above the iterative-cutover branch, which reaches
  `InitLogging()` via `resolveToActiveIterationIfRequired()` before the old call sites ran.
  (Both raised in review.)
- **Propagation.** The settings are forwarded to every spawned yb-voyager subprocess that
  already forwards `--log-level`: the iterative-cutover re-exec sites,
  `assess-migration-bulk`'s per-schema children, and the report `end migration` spawns
  (`endMigrationCommand.go:493`, raised in review). With `--log-dir` gone there is no
  same-file/two-writer question to reason about, so all of them are forwarded unconditionally.

- Two new CLI flags / config-file keys: `--log-max-size-mb`, `--log-max-backups`. Both
  default to the exact previous behavior.
- Documented in the global section of all 5 `config-templates/*.yaml`, alongside `log-level`.

- `src/config/config_test.go` — `ValidateLogSettings` boundaries and exact error messages
  (including that `0` is rejected rather than meaning unlimited), and `LumberjackMaxBackups`
  sentinel translation.
- `cmd/logging_test.go` — the values actually reaching the `lumberjack.Logger` (defaults,
  custom, and the sentinel), disabled logging, invalid level, `logSettingsCLIArgs`, and flag
  registration on the two commands that bypass `registerCommonGlobalFlags`.
  `TestInitLogging_UnlimitedBackupsRetainsRotatedFiles` drives *real* rotation over a 1 MB
  threshold and asserts rotated files are kept on disk (and still pruned when a limit is
  set), so a lumberjack upgrade changing the meaning of `MaxBackups` would be caught here
  rather than in production.
- `src/dbzm/dbzm_test.go` — the same rotation wiring for `debezium-<role>.log`.
- `cmd/config_test.go` — config-file binding, per-command section binding, CLI-overrides-config,
  defaults-when-unset (asserting the live `lumberjack.Logger`, not just the flag vars), and
  invalid values halting via `ErrExitPreLog` before logging is initialised.
- Mutation-checked the new tests: breaking the sentinel translation and loosening the
  validation each fail the suite at every layer (config, cmd, dbzm).
- Manual: built the binary and confirmed the help text, the rejection messages for
  `--log-max-size-mb 0` / `--log-max-backups 0` / `--log-max-backups -2`, and that `-1` is
  accepted. Separately confirmed empirically that lumberjack with `MaxBackups=0` retains
  every rotated file while `MaxBackups=2` keeps 3.
- `go build ./...`, `go vet ./...`, `staticcheck ./...`, `staticcheck -tags unit ./...` and
  `go test -tags unit ./...` all pass across the module, and the diff is gofmt-clean.

No.

No changes to MetaDB, AssessmentDB, or any other serialized/on-disk structure.

### DIFF
--- a/yb-voyager/cmd/config.go
+++ b/yb-voyager/cmd/config.go
@@ -52,7 +52,7 @@ var commandsUsingTargetConfig = []string{
 }
 
 var allowedGlobalConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"export-dir", "log-level", "send-diagnostics",
+	"export-dir", "log-level", "log-max-size-mb", "log-max-backups", "send-diagnostics",
 	"profile",
 	// environment variables keys
 	"control-plane-type", "java-home",
@@ -88,7 +88,7 @@ var allowedYBAeonControlPlaneConfigKeys = mapset.NewThreadUnsafeSet[string](
 )
 
 var allowedAssessMigrationConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level", "run-guardrails-checks",
+	"log-level", "log-max-size-mb", "log-max-backups", "run-guardrails-checks",
 	"iops-capture-interval", "target-db-version", "assessment-metadata-dir",
 	"invoked-by-export-schema",
 	// environment variables keys
@@ -96,14 +96,14 @@ var allowedAssessMigrationConfigKeys = mapset.NewThreadUnsafeSet[string](
 )
 
 var allowedAnalyzeSchemaConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level",
+	"log-level", "log-max-size-mb", "log-max-backups",
 	"output-format", "target-db-version",
 	// environment variables keys
 	"report-unsupported-plpgsql-objects",
 )
 
 var allowedExportSchemaConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level", "run-guardrails-checks",
+	"log-level", "log-max-size-mb", "log-max-backups", "run-guardrails-checks",
 	"use-orafce", "comments-on-objects", "object-type-list", "exclude-object-type-list",
 	"skip-colocation-recommendations", "assessment-report-path",
 	"skip-performance-recommendations",
@@ -113,7 +113,7 @@ var allowedExportSchemaConfigKeys = mapset.NewThreadUnsafeSet[string](
 )
 
 var allowedExportDataConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level", "run-guardrails-checks",
+	"log-level", "log-max-size-mb", "log-max-backups", "run-guardrails-checks",
 	"disable-pb", "exclude-table-list", "table-list", "exclude-table-list-file-path",
 	"table-list-file-path", "parallel-jobs", "export-type",
 	"allow-oracle-clob-data-export", "metrics-port",
@@ -122,7 +122,7 @@ var allowedExportDataConfigKeys = mapset.NewThreadUnsafeSet[string](
 )
 
 var allowedExportDataFromTargetConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level",
+	"log-level", "log-max-size-mb", "log-max-backups",
 	"disable-pb", "exclude-table-list", "table-list", "exclude-table-list-file-path",
 	"table-list-file-path", "transaction-ordering", "metrics-port",
 	// environment variables keys
@@ -130,18 +130,18 @@ var allowedExportDataFromTargetConfigKeys = mapset.NewThreadUnsafeSet[string](
 )
 
 var allowedImportSchemaConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level", "run-guardrails-checks",
+	"log-level", "log-max-size-mb", "log-max-backups", "run-guardrails-checks",
 	"continue-on-error", "object-type-list", "exclude-object-type-list", "straight-order",
 	"ignore-exist", "enable-orafce",
 )
 
 var allowedFinalizeSchemaPostDataImportConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level", "run-guardrails-checks",
+	"log-level", "log-max-size-mb", "log-max-backups", "run-guardrails-checks",
 	"continue-on-error", "ignore-exist", "refresh-mviews",
 )
 
 var allowedImportDataConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level", "run-guardrails-checks",
+	"log-level", "log-max-size-mb", "log-max-backups", "run-guardrails-checks",
 	"batch-size", "parallel-jobs", "adaptive-parallelism", "adaptive-parallelism-max",
 	"skip-replication-checks",
 	"disable-pb", "max-retries-streaming", "exclude-table-list", "table-list",
@@ -163,7 +163,7 @@ var allowedImportDataConfigKeys = mapset.NewThreadUnsafeSet[string](
 )
 
 var allowedImportDataToSourceConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level", "run-guardrails-checks",
+	"log-level", "log-max-size-mb", "log-max-backups", "run-guardrails-checks",
 	"parallel-jobs", "disable-pb", "prometheus-metrics-port", "metrics-port", "use-partition-root",
 	// environment variables keys
 	"num-event-channels", "event-channel-size", "max-events-per-batch",
@@ -172,7 +172,7 @@ var allowedImportDataToSourceConfigKeys = mapset.NewThreadUnsafeSet[string](
 )
 
 var allowedImportDataToSourceReplicaConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level", "run-guardrails-checks",
+	"log-level", "log-max-size-mb", "log-max-backups", "run-guardrails-checks",
 	"batch-size", "parallel-jobs", "truncate-tables", "disable-pb", "max-retries-streaming",
 	"prometheus-metrics-port", "metrics-port",
 	// environment variables keys
@@ -182,7 +182,7 @@ var allowedImportDataToSourceReplicaConfigKeys = mapset.NewThreadUnsafeSet[strin
 )
 
 var allowedImportDataFileConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level",
+	"log-level", "log-max-size-mb", "log-max-backups",
 	"disable-pb", "max-retries-streaming", "enable-upsert", "use-public-ip", "target-endpoints",
 	"batch-size", "parallel-jobs", "adaptive-parallelism", "adaptive-parallelism-max",
 	"format", "delimiter", "data-dir", "file-table-map", "has-header", "escape-char",
@@ -205,12 +205,12 @@ var allowedInitCutoverToSourceConfigKeys = mapset.NewThreadUnsafeSet[string](
 	"restart-data-migration-source-target",
 )
 var allowedArchiveChangesConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level",
+	"log-level", "log-max-size-mb", "log-max-backups",
 	"policy", "archive-dir", "fs-utilization-threshold",
 )
 
 var allowedEndMigrationConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"log-level",
+	"log-level", "log-max-size-mb", "log-max-backups",
 	"backup-schema-files", "backup-data-files", "save-migration-reports", "backup-log-files",
 	"backup-dir",
 )
--- a/yb-voyager/cmd/exportDataDebezium.go
+++ b/yb-voyager/cmd/exportDataDebezium.go
@@ -126,8 +126,12 @@ func prepareDebeziumConfig(partitionsToRootTableMap map[string]string, tableList
 		// dbzm does not support fatal/panic log levels
 		dbzmLogLevel = config.ERROR
 	}
+	// Read before the local `config` variable below shadows the config package.
+	dbzmLogMaxSizeMB, dbzmLogMaxBackups := config.LogMaxSizeMB, config.LogMaxBackups
 	config := &dbzm.Config{
 		LogLevel:           dbzmLogLevel,
+		LogMaxSizeMB:       dbzmLogMaxSizeMB,
+		LogMaxBackups:      dbzmLogMaxBackups,
 		MigrationUUID:      migrationUUID,
 		RunId:              runId,
 		SourceDBType:       source.DBType,
--- a/yb-voyager/cmd/logging.go
+++ b/yb-voyager/cmd/logging.go
@@ -20,9 +20,13 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
 	"gopkg.in/natefinch/lumberjack.v2"
+
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/config"
 )
 
 type MyFormatter struct{}
@@ -48,7 +52,7 @@ func (mf *MyFormatter) Format(entry *log.Entry) ([]byte, error) {
 	return []byte(msg), nil
 }
 
-func InitLogging(logDir string, logLevel string, disableLogging bool, cmdName string) error {
+func InitLogging(logDir string, logLevel string, disableLogging bool, cmdName string, logMaxSizeMB int, logMaxBackups int) error {
 	// Redirect log messages to ${logDir}/yb-voyager.log if not a status command.
 	if disableLogging {
 		log.SetOutput(io.Discard)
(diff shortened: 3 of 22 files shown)


yb-voyager/src/config/config.go
```diff
@@ -45,6 +53,16 @@ func ValidateLogLevel() error {
 	return nil
 }
 
+func ValidateLogSettings() error {
+	if LogMaxSizeMB <= 0 {
+		return goerrors.Errorf("invalid log-max-size-mb: %d. Must be a positive integer", LogMaxSizeMB)
+	}
+	if LogMaxBackups < 0 {
```
makalaaneesh: [AI] **`log-max-backups: 0` silently means "retain unlimited backups", not "keep none".**

`ValidateLogSettings` accepts `LogMaxBackups == 0` (only `< 0` is rejected), and the value flows straight into `lumberjack.Logger.MaxBackups`. In lumberjack, `MaxBackups = 0` means *retain all old log files* — with `MaxAge` also 0 (the default here) and no compression, rotated files are never deleted.

The flag help reads `"maximum number of rotated log files to retain"`, so a user who sets `--log-max-size-mb 200 --log-max-backups 0` to cap disk usage will get the opposite: unbounded growth of 200 MB log files until the disk fills. This is a plausible misconfiguration for a long-running migration, and nothing (help text, config templates, or validation error) warns about it.

If `0` is intended to mean "unlimited", please document that explicitly in the flag help and in the `config-templates/*.yaml` comments (e.g. "0 = retain all rotated files"). If it isn't intended, reject `0` in `ValidateLogSettings` the same way `log-max-size-mb` rejects it:

```go
if LogMaxBackups <= 0 {
    return goerrors.Errorf("invalid log-max-backups: %d. Must be a positive integer", LogMaxBackups)
}
```

(Note the existing test `"positive maxSizeMB and zero maxBackups are valid"` encodes the current allow-0 behavior, so it would need updating alongside whichever direction you choose.)

---

makalaaneesh (commented):

---

yb-voyager/cmd/logging.go
```diff
@@ -78,6 +99,43 @@ func InitLogging(logDir string, logLevel string, disableLogging bool, cmdName st
...
+	}
+	return args
+}
+
+// registerLogFlags registers the CLI flags that control yb-voyager's logging behavior
+// (level, destination directory, and rotation) on the given command.
+func registerLogFlags(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVarP(&config.LogLevel, "log-level", "l", "info",
+		"log level for yb-voyager. Accepted values: (trace, debug, info, warn, error, fatal, panic)")
+	cmd.PersistentFlags().StringVar(&config.LogDir, "log-dir", "",
```
ShivanshGahlot: end migration command still reads the default log location for backup log files
> _, err = os.ReadDir(filepath.Join(exportDir, "logs"))

Is this the intended behaviour? 
We might not need to back up the log files if they are already in a separate folder but we need to confirm the usecase once.

yb-voyager/cmd/logging.go
```diff
@@ -48,19 +51,37 @@ func (mf *MyFormatter) Format(entry *log.Entry) ([]byte, error) {
...
+}
+
+func InitLogging(exportDir string, logLevel string, disableLogging bool, cmdName string, fileSettings logFileSettings) error {
 	// Redirect log messages to ${logDir}/yb-voyager.log if not a status command.
 	if disableLogging {
 		log.SetOutput(io.Discard)
 		return nil
 	}
-	logFileName := filepath.Join(logDir, "logs", fmt.Sprintf("yb-voyager-%s.log", cmdName))
+	logDir := fileSettings.Dir
```
ShivanshGahlot: We never verify whether the log dir actually exists or it has permissions to be used. Check expor dir implementation for context.

yb-voyager/cmd/logging.go
```diff
@@ -48,19 +51,37 @@ func (mf *MyFormatter) Format(entry *log.Entry) ([]byte, error) {
...
+	if logDir == "" {
+		logDir = filepath.Join(exportDir, "logs")
+	}
+	logFileName := filepath.Join(logDir, fmt.Sprintf("yb-voyager-%s.log", cmdName))
```
ShivanshGahlot: Hypothetically if two concurrent runs of the same command are run, they will overlap in the same file. We should probably have a pid/migration uuid also attached to the name if we intend to use the same log dir for two separate concurrent runs.

yb-voyager/cmd/logging.go
```diff
@@ -48,19 +51,37 @@ func (mf *MyFormatter) Format(entry *log.Entry) ([]byte, error) {
 	return []byte(msg), nil
 }
 
-func InitLogging(logDir string, logLevel string, disableLogging bool, cmdName string) error {
+// logFileSettings groups the log file destination and rotation settings used by InitLogging.
+type logFileSettings struct {
```
ShivanshGahlot: I think none of these configs are reaching the Debezium log rotators. They will still be present in the default log dir with the default caps.

yb-voyager/cmd/root.go
```diff
@@ -122,12 +122,17 @@ Refer to docs (https://docs.yugabyte.com/preview/migrate/) for more details like
 				// logging is not initialized yet
 				utils.ErrExitPreLog("ERROR: %v", err)
 			}
+			err = config.ValidateLogSettings()
```
ShivanshGahlot: We need to probably place this above the `isLiveMigrationIterationCommand` condition above. We might be consuming the log settings before we validate in case `resolveToActiveIterationIfRequired` is run.

ShivanshGahlot: This will also collapse the two instances of the code into one.

yb-voyager/cmd/logging.go
ShivanshGahlot: At multiple places: 
`yb-voyager/src/srcdb/pg_dump_extract_schema.go:67, pg_dump_export_data.go:73,87, cmd/exportData.go:1733`
Error messages mention the default log directory even if `--log-dir `is used.

ShivanshGahlot: We should forward the log dir and settings in the case where get data-migration-report is run by end migration command too:

`strCmd := fmt.Sprintf("yb-voyager get data-migration-report --export-dir %s --log-level %s --output-format json", ...)`
at yb-voyager/cmd/endMigrationCommand.go:493

yb-voyager/cmd/config.go
```diff
@@ -52,7 +52,7 @@ var commandsUsingTargetConfig = []string{
 }
 
 var allowedGlobalConfigKeys = mapset.NewThreadUnsafeSet[string](
-	"export-dir", "log-level", "send-diagnostics",
+	"export-dir", "log-level", "log-dir", "log-max-size-mb", "log-max-backups", "send-diagnostics",
```
ShivanshGahlot: `log-level` is allowed at the global level as well as command levels too. We should do the same for these new flags.

yb-voyager/test/utils/errexit_helpers.go
```diff
@@ -0,0 +1,29 @@
+package testutils
```
ShivanshGahlot: Claude: 

>Three things on this new helper:
> 
> 1. src/utils/logging.go:227-245 already has exactly this shape (originalErrExit + MonkeyPatchUtilsErrExit + RestoreUtilsErrExit), living next to the function it patches. Putting the ErrExitPreLog twin in test/utils splits two halves of the same mechanism across packages — worth co-locating.
> 2. This is the only .go file in the tree without the Apache license header.
> 3. Small bug in the copied logic: MonkeyPatchUtilsErrExitPreLog overwrites the package-level originalErrExitPreLog on every call, so a second patch followed by RestoreUtilsErrExitPreLog() restores the intermediate function rather than the original. Harmless with the current single call site, but it's a trap for the next caller. Guarding with if originalErrExitPreLog == nil on set would fix it.

ShivanshGahlot (commented):
Can you check this once in the PR description: it says assess-migration-bulk's children "don't yet inherit these settings," but the code does forward both rotation flags — only --log-dir is withheld

ShivanshGahlot:
Overall Question @jasonriddell. What is the motivation behind adding the log dir flag? Why would a user want a separate log dir than export dir? Its a singular location where logs can be found. 
Why I am asking is that it is adding these complications and other fixes we need to do to support it as I have mentioned in the comments.

jasonriddell:
@ShivanshGahlot The log-dir was just me trying to make it complete, no customer or use driver that is pushing it. The most important parts are the size and the number of files retained; everything else is just a nice-to-have.

Please don't feel obligated to take anything from the PR; the goal was just to ensure we didn't lose log messages when the files wrapped around. Anything above that isn't needed. Thanks.

ShivanshGahlot:
Yeah @jasonriddell I guess in that case we should keep the `--log-dir` flag out of this PR. If we need it in the future we can properly dig and get it in in a separate PR.

jasonriddell:
@ShivanshGahlot Rebuilt this from `main` as a single commit with the reduced scope we agreed on.

**`--log-dir` is gone entirely.** That removes most of what you flagged — the log-dir existence/permission check, the concurrent-runs-sharing-a-file-name problem, `end migration` backing up from the default `<export-dir>/logs`, and the error messages in `pg_dump_extract_schema.go` / `pg_dump_export_data.go` / `exportData.go` that hardcode the default log directory. All of those were only problems because the directory could move; none apply now.

What's left is just the part that motivated the issue — `--log-max-size-mb` and `--log-max-backups`, with `-1` on backups meaning "retain all". I kept the four other points you raised, since they stand on their own:

- **Debezium rotators** — good catch, and it mattered more than the rest: `debezium-<role>.log` had the same hardcoded 200/10, so "retain all" would have silently not held for live migration. It now honours both settings.
- **Validation ordering** (root.go:125) — validation is hoisted above the `isLiveMigrationIterationCommand` branch, since `resolveToActiveIterationIfRequired()` reaches `InitLogging()` before the old call sites ran. This also collapses the two copies into one `validateLogFlags()`, as you suggested.
- **`end migration` → `get data-migration-report`** (endMigrationCommand.go:493) — now forwards the settings along with `--log-level`.
- **Command-level config keys** — both keys are allowed per-command as well as globally, matching `log-level`.

On @makalaaneesh's `log-max-backups: 0` finding: `0` is now rejected outright rather than documented, since lumberjack reads it as "retain all" and that's the opposite of what someone passing `0` to cap disk usage would expect. `-1` is the explicit sentinel instead.

Also moved the `ErrExitPreLog` test helper next to the existing `MonkeyPatchUtilsErrExit` in `src/utils/logging.go` instead of a new file under `test/utils`, with the license header and the nil-guard fix.

One test worth pointing at: `TestInitLogging_UnlimitedBackupsRetainsRotatedFiles` drives real rotation over a 1 MB threshold and asserts the files are actually kept on disk, so if a lumberjack upgrade ever changes what `MaxBackups == 0` means, it fails here rather than in a customer's migration.

The previous version of the branch is preserved locally if anything from it is worth pulling back in.

yb-voyager/cmd/exportDataDebezium.go
```diff
@@ -126,8 +126,12 @@ func prepareDebeziumConfig(partitionsToRootTableMap map[string]string, tableList
 		// dbzm does not support fatal/panic log levels
 		dbzmLogLevel = config.ERROR
 	}
+	// Read before the local `config` variable below shadows the config package.
+	dbzmLogMaxSizeMB, dbzmLogMaxBackups := config.LogMaxSizeMB, config.LogMaxBackups
 	config := &dbzm.Config{
```
makalaaneesh, line 131: [AI][Suggestion] **Local `config` variable shadows the `config` package; workaround is brittle for future additions.**

The current code pre-reads `dbzmLogMaxSizeMB, dbzmLogMaxBackups := config.LogMaxSizeMB, config.LogMaxBackups` two lines above the `config := &dbzm.Config{...}` shadow, with a comment warning the reader. Any future field that also needs a `config.*` lookup and gets added inside the struct literal will silently resolve to the shadowed `*dbzm.Config` — usually a compile error, but for fields that happen to exist on both types it would silently pick the wrong one. Renaming the local `dbzm.Config` variable removes the trap entirely.

Failure case: a follow-up patch adds e.g. `SendDiagnostics: config.SendDiagnostics` inside the `&dbzm.Config{...}` literal without noticing the shadow, and the compile fails only because `dbzm.Config` doesn't have `SendDiagnostics` — but a future dbzm field named the same as a `config` var would compile with the wrong value.

Suggested change: rename the local variable to `dbzmCfg` and update every subsequent `config.Field = ...` in this function to `dbzmCfg.Field = ...`.

---

yb-voyager/src/dbzm/dbzm.go
```diff
@@ -234,8 +235,8 @@ func (d *Debezium) setupLogFile() error {
 
 	logRotator := &lumberjack.Logger{
 		Filename:   logFilePath,
-		MaxSize:    200, // 200 MB log size before rotation
-		MaxBackups: 10,  // Allow upto 10 logs at once before deleting oldest logs.
+		MaxSize:    d.LogMaxSizeMB, // log size in MB before rotation
```
makalaaneesh, line 238: [AI][Suggestion] **`setupLogFile` silently accepts `LogMaxSizeMB == 0` and falls back to lumberjack's own 100 MB default.**

If any future caller constructs a `dbzm.Config` without setting `LogMaxSizeMB` (the zero value), `lumberjack.Logger` interprets `MaxSize == 0` as "use my default (100 MB)" — not the documented yb-voyager default (200 MB). Today only `exportDataDebezium.go` builds the config and it always populates the two fields, so this is not a live bug. A `config.LogMaxSizeMB` typo in a new caller would go unnoticed, though.

Failure case: a future refactor adds a second `dbzm.Config` construction and forgets the two log fields; behaviour diverges from the documented defaults with no error.

Suggested change: default zero values in `setupLogFile` to `config.DefaultLogMaxSizeMB` / `config.DefaultLogMaxBackups`.

---

yb-voyager/cmd/logging_test.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/natefinch/lumberjack.v2"
+
+	"github.com/yugabyte/yb-voyager/yb-voyager/src/config"
+)
+
+// restoreLogOutput restores logrus' global output after a test that calls InitLogging.
+func restoreLogOutput(t *testing.T) {
```
makalaaneesh, line 38: [AI][Suggestion] **`restoreLogOutput` does not restore `SetReportCaller` / `SetFormatter`.**

`InitLogging` calls `log.SetReportCaller(true)` and `log.SetFormatter(&MyFormatter{})`; `restoreLogOutput` restores only `SetOutput` and `SetLevel`. Any later test in the `cmd` package that inherits logrus's global state gets `MyFormatter` and caller reporting.

Failure case: a future test in this package that asserts on the default formatter's output starts failing after `TestInitLogging_RotationSettingsReachLumberjack` runs first.

Suggested change: also restore `Formatter` and `ReportCaller` in the cleanup.

---

makalaaneesh (commented):
[AI] Automated review — 3 findings (0 critical, 0 warning, 3 suggestion). See inline comments.

---

ShivanshGahlot (approved):
LGTM

ShivanshGahlot:
@jasonriddell we will be putting this in the current release. So I am merging this PR.

PR #3739, opened 2026-08-24
Block BETA_FAST_DATA_EXPORT for postgresql sources

### LINKED ISSUE #3743: Disable BETA_FAST_DATA_EXPORT for PostgreSQL
Jira Link: [DB-23167](https://yugabyte.atlassian.net/browse/DB-23167)

`BETA_FAST_DATA_EXPORT=1` routes the snapshot export through Debezium instead of pg_dump. It is a beta accelerator for Oracle and MySQL, but nothing enforced that, and CI ran the full PG offline suite under it.

It is broken on PostgreSQL: Debezium hands the value converters a JDBC object instead of postgres' own text, so an hstore column arrives as a java.util.Map and serializes to {k=v}, which no longer parses as hstore on import.

**yb-voyager**

- Reject BETA_FAST_DATA_EXPORT in export data when the source is postgresql.
- Leave live migration alone — it always uses Debezium, so the variable is a no-op there.
- Drop the BETA_FAST_DATA_EXPORT matrix dimension from the pg-13 and pg-17 workflows, keeping MySQL's.

**jenkins-helpers**

- Make run_tests_using_debezium skip the variable for postgres tests rather than applying it to every test, so one build can still use Debezium for oracle/mysql while running the postgres suite normally.
- Apply at all three export sites: yb-voyager-testing, yb-voyager-resumption-tests, and the shared common-helpers used by the rhel8/rhel9/apt release-test pipelines.

Update the parameter description to say it applies to Oracle and MySQL only.

ShivanshGahlot:

Fixes #3743.

`BETA_FAST_DATA_EXPORT` routes the snapshot export through debezium instead of the source's native dump tool. It is a beta feature for Oracle and MySQL, but nothing stopped it being used with postgresql, and the PG migtest workflows ran the whole offline suite under it.

- **Why it is wrong for PG** — debezium hands the value converters a JDBC object rather than postgres' own text. An hstore column arrives as a `java.util.Map` and serializes to `{k=v}`, which no longer parses as hstore on import.
- **Guardrail** — `export data` now rejects the combination for a postgresql source.
- **CI** — drops the `BETA_FAST_DATA_EXPORT` matrix dimension from the pg-13 and pg-17 workflows. MySQL and Oracle keep their coverage.
- **Live migration is unaffected** — it always uses debezium, so setting the variable there changes nothing and is still allowed.

`export data` with a postgresql source and `BETA_FAST_DATA_EXPORT` set now fails instead of exporting:

```
BETA_FAST_DATA_EXPORT is not supported for source database type "postgresql". It is
available only for oracle and mysql. Unset the BETA_FAST_DATA_EXPORT environment
variable and retry.
```

Manually, against a real PG source. `export data --export-type snapshot-only` with the variable set now fails with the message above. `--export-type snapshot-and-changes` with the variable still set proceeds into live migration untouched. No automated test: the check is a source-type comparison, and the PG workflows no longer run the combination at all.

No.

No.


makalaaneesh:

Reviewed the 3 changed files (+17/-12) against the PR base `main`, applying the root, `yb-voyager`, and `yb-voyager/cmd` BUGBOT review rules:

- **`validateBetaFastDataExportSupportedForSource` (exportData.go)** — the new helper *returns* a wrapped error rather than calling `utils.ErrExit` itself; the top-level `exportDataCommandPreRun` entry point calls `ErrExit`. This matches the cmd-package error-handling rule (new leaf helpers should return errors, not exit). Logic is correct: `useDebezium` is already resolved from `BETA_FAST_DATA_EXPORT` in `setExportFlagsDefaults()` and from change-streaming just above the call site, and the `!useDebezium || changeStreamingIsEnabled(exportType)` early return leaves all live-migration flows (which always use Debezium) unaffected, blocking only the PG snapshot-via-Debezium path — consistent with the stated intent.
- **CI workflows (pg-13/pg-17)** — dropping the `BETA_FAST_DATA_EXPORT: [0,1]` matrix dimension is consistent with the guardrail; MySQL/Oracle coverage is retained in their own workflows.

Remaining observations are Suggestion/Question-level only (the dropped guardrail unit test; now-unreachable `BETA_FAST_DATA_EXPORT == '1'` branches in `migtests/tests/pg/sequences/validate*`), which this automated sweep does not post.

---

priyanshi-yb (approved):
LGTM, please disallow this in jenkins runs also before merging the PR.

PR #3741, opened 2026-08-24
Use partition by table strategy for the tables having unique index on stored generated column

priyanshi-yb:

During live migration, CDC events are routed across parallel channels by hashing a routing key (`pk` or a `custom` column list). A table whose unique index covers a **STORED generated column** is unsafe under `pk`/`custom`: generated-column values are absent from Debezium events, so conflict detection can't see that unique key and colliding events can apply out of order (duplicate-key / lost-update). This forces such tables onto a single channel (`table`) and rejects the unsafe explicit configs.

**Detect (source-side, at export)**
- New `SourceDB.GetGeneratedStoredColumns` — PostgreSQL impl via `pg_attribute.attgenerated = 's'`; no-op for Oracle/MySQL/YB sources.
- `export data` captures each table's generated columns into a new MetaDB record `export_data_source_db_exporter_status`. Import reads it without needing a source connection.

**Resolve (import-side)**
- `getGeneratedStoredColumnsHybrid` intersects the source-captured generated columns with the target's unique-index columns (`GetTableToUniqueIndexesMap`).
- `auto` routes a table with a unique index on a generated column to `table` (same as expression-unique-index tables).
- An explicit `pk`/`custom` on such a table is rejected, pointing the user to `table`.
- A `custom` key that names a generated column is rejected.
- Recomputed every run (first run and resume); not persisted in the import status record.

**Refactor**
- The target unique-index map is fetched once and reused for both partitioning and conflict detection.

**Not handled**
- A primary key on a generated column stays out of scope. The event's routing key itself is incomplete, so no strategy fixes it. Documented as a known limitation.

- Default (`--cdc-partition-key auto`): a table with a unique index on a STORED generated column now routes with `table` instead of `pk`.
- New `import data` validation errors before snapshot: an explicit `pk`/`custom` on such a table, and a `custom` key column that is a generated column.
- No new CLI flags or configuration.

- Unit: `TestResolveEffectiveCdcPartitionKeys` (resolution matrix), `TestBuildGeneratedStoredColumns` (hybrid intersection), `TestGetGeneratedStoredColumns` (target catalog query).
- Live integration in `src/testlivemigration/live_migration_uk_conflict_test.go`: rejects the unsafe `custom` configs against a real YugabyteDB, then completes snapshot + streaming + cutover with the `table` fallback.

No.

Yes, but additive and backward-compatible. MetaDB gains a new key `export_data_source_db_exporter_status` written by export and read by import. An absent record (older export dir) resolves to no-op, so no upgrade break. `ImportDataStatusRecord` is unchanged.

---

| Component        


yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -2514,6 +2514,135 @@ SELECT
...
+	query := fmt.Sprintf(`
+	SELECT
+		n.nspname AS table_schema,
+		t.relname AS table_name,
+		a.attname AS column_name,
+		EXISTS (
+			SELECT 1 FROM pg_index idx
+			WHERE idx.indrelid = t.oid
+				AND idx.indisunique
+				AND a.attnum = ANY(idx.indkey)
```
makalaaneesh: [AI] **`in_unique_index` detection matches INCLUDE (covering) columns, not just key columns.**

The `EXISTS` subquery marks a stored generated column as `in_unique_index` whenever its `attnum` appears anywhere in `idx.indkey`:

```sql
AND a.attnum = ANY(idx.indkey)
```

For a unique index defined with an `INCLUDE` clause (supported here — see the existing `UNIQUE (code) INCLUDE (email)` test), `indkey` contains both the key columns (the first `indnkeyatts`) and the trailing covering columns. A stored generated column that appears only in the `INCLUDE` list does not participate in the uniqueness constraint, yet this check returns `true` for it.

Failure case: a table with `UNIQUE (id) INCLUDE (g)` where `g` is a `STORED` generated column. Uniqueness is on `id` only, so `pk`/`custom` routing (and conflict detection on `id`) is actually safe, but `GetGeneratedStoredColumns` reports `InUniqueIndex=true`. In `auto` this forces `PARTITION_BY_TABLE` unnecessarily; with `--cdc-partition-key pk` (or a custom override) `resolveEffectiveCdcPartitionKeys` rejects the valid config with the misleading error "table ... has a unique index on a stored generated column".

Suggested fix: restrict the membership check to the key columns of the index:

```sql
AND a.attnum = ANY((idx.indkey::int2[])[1:idx.indnkeyatts])
```

(Casting the `int2vector` to `int2[]` yields a 1-based array, so `[1:indnkeyatts]` selects the key columns; please confirm the slice bounds against your PG/YB versions.)

---

makalaaneesh (commented):
[AI] Automated review — 1 finding (0 critical, 1 warning). See inline comments.

---

yb-voyager/src/srcdb/postgres.go
```diff
@@ -812,6 +812,72 @@ func (pg *PostgreSQL) GetColumnsWithSupportedTypes(tableList []sqlname.NameTuple
...
+    n.nspname AS table_schema,
+    t.relname AS table_name,
+    a.attname AS column_name
+FROM pg_class t
+JOIN pg_namespace n ON t.relnamespace = n.oid
+JOIN pg_attribute a ON t.oid = a.attrelid
+WHERE t.relkind IN ('r', 'p')
+    AND a.attnum > 0
+    AND NOT a.attisdropped
+    AND a.attgenerated = 's'
```
makalaaneesh, line 849: [AI] **Export data hard-fails on PG 10/11 live-migration sources.**

`pg_attribute.attgenerated` exists only on PG 12+, but live migration supports PG 10+ (`MIN_SUPPORTED_PG_VERSION_LIVE = "10"`). `captureSourceGeneratedStoredColumns` runs on every PG live export, so on a PG 10/11 source this query errors with `column a.attgenerated does not exist` and `export data` exits fatally at startup — a regression vs main for supported sources.

The function docstring says "on older sources this simply returns an empty map", but nothing guards the query.

Suggested fix: version-gate the query (source major version < 12 → return empty map, which is correct since generated columns cannot exist there), or reuse the `attgenerated does not exist` error-sniffing guard this PR previously had in the target-side implementation (commit 5d27588ad) before it was removed.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -181,18 +191,60 @@ func resolveEffectiveCdcPartitionKeys(
...
+			return nil, goerrors.Errorf("cdc-partition-key %s is not allowed for table %s because it has an expression-based unique index; use table (via --cdc-partition-key or --cdc-partition-key-overrides)", override.Strategy, t.ForOutput())
+		}
+		switch override.Strategy {
+		case PARTITION_BY_PK:
+			//For PK, we need to check if the table has a unique index on a stored generated column
+			//and if so, we need to reject the strategy
+			if tableHasUniqueIndexOnGenerated(generatedStoredCols, t) {
+				return nil, goerrors.Errorf("cdc-partition-key %s is not allowed for table %s because it has an unique index on a stored generated column; use table (via --cdc-partition-key or --cdc-partition-key-overrides)", override.Strategy, t.ForOutput())
+			}
+		case PARTITION_BY_CUSTOM:
```
makalaaneesh: [AI] **`custom` strategy is allowed on a table with a unique index on a generated column → deterministic streaming crash.**

The rejection set here should mirror the auto bucket's force-table set, but the UK-on-generated condition is only enforced for `pk`:

| Condition | auto forces `table` | explicit `pk` rejected | explicit `custom` rejected |
|---|---|---|---|
| expression UK | yes | yes | yes |
| UK on generated column | yes | yes | **no — the gap** |
| custom key column is generated | n/a (auto never picks custom) | n/a | yes |

A custom-routed table keeps the UK-on-generated in the conflict detection cache (target UK map), and every event lacks the generated column's value in all images (pgoutput omits generated columns), so `computeConflictBucketKey` fails with `column X is missing from fields` — the first event on the table kills streaming. This is exactly the failure class this PR moves to upfront rejection, and expression-UK tables do reject `custom`.

Failure case: `--cdc-partition-key-overrides "s.audit:(customer_id)"` where `audit` has `UNIQUE(amount)` on generated `amount` → config accepted, mid-stream crash on the first `audit` event.

Suggested fix — hoist the check out of the switch so it structurally covers every non-table strategy:

```go
if override.Strategy == PARTITION_BY_TABLE {
    continue
}
if _, isExprUK := exprUKSet.Get(t); isExprUK {
    return nil, goerrors.Errorf(...)
}
if tableHasUniqueIndexOnGenerated(generatedStoredCols, t) { // now covers pk AND custom
    return nil, goerrors.Errorf(...)
}
if override.Strategy == PARTITION_BY_CUSTOM {
    if cols, ok := tableHasGeneratedColumn(generatedStoredCols, t, override.Columns); ok {
        return nil, goerrors.Errorf(...)
    }
}
```

Note the unit test currently enshrines the gap: the subtest named "UK on generated: auto uses table; pk and custom rejected; override table allowed" actually `require.NoError`s the custom-override-with-plain-column case — it should flip to `require.Error` once this is fixed.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -459,6 +533,87 @@ func getExpressionUniqueIndexTables(tableNames []sqlname.NameTuple) ([]sqlname.N
...
+// If the source capture is absent (export written by an older voyager, or capture not yet
+// persisted), it falls back to the legacy target-only detection (getGeneratedStoredColumns).
+func getGeneratedStoredColumnsHybrid(tableNames []sqlname.NameTuple, tableToUniqueIndexes *utils.StructMap[sqlname.NameTuple, []tgtdb.UniqueIndex]) (*utils.StructMap[sqlname.NameTuple, []GeneratedStoredColumn], error) {
+	result := utils.NewStructMap[sqlname.NameTuple, []GeneratedStoredColumn]()
+
+	exportStatus, err := metaDB.GetExportDataSourceDBExporterStatusRecord()
+	if err != nil {
+		return nil, fmt.Errorf("error getting export data source db exporter status record: %w", err)
+	}
+	if exportStatus == nil {
```
makalaaneesh, line 549: [AI] **Missing source-capture record silently disables all generated-column protection (and the docstring describes a fallback that no longer exists).**

An export dir written by an older voyager has no `export_data_source_db_exporter_status` record, so this returns nil and the import proceeds with zero protection — reintroducing the mid-stream `column X is missing from fields` crash for exactly the users this PR protects. (Hard-failing would be worse: it would break mid-migration upgrades for the majority with no generated columns.)

Meanwhile the function docstring still claims it "falls back to the legacy target-only detection (getGeneratedStoredColumns)" — that code was removed in the "changes as per new approach" commit, so the claim is now false.

Suggested fix: restore the target-side `GetGeneratedStoredColumns` from commit 3fc51fc30 as the nil-record fallback — it already handles old YB (`attgenerated does not exist` guard), it is a good proxy since voyager migrated the schema, and it makes the docstring true. At minimum, emit a `PrintAndLogfWarning` noting that restarting `export data` on the upgraded binary backfills the record (the capture runs on every export invocation).

yb-voyager/cmd/importData.go
```diff
@@ -1234,12 +1234,17 @@ func importData(importFileTasks []*ImportFileTask, errorPolicy importdata.ErrorP
 		utils.ErrExit("Failed to clean MigrationStatusRecord for import data start clean: %s", err)
 	}
 
+	tableToUniqueIndexes, err := tdb.GetTableToUniqueIndexesMap(importTableList)
```
makalaaneesh: [AI] **`GetTableToUniqueIndexesMap` now runs unconditionally on every import, including flows that never use it.**

Offline snapshot-only imports and `import data file` now pay a target catalog query at startup and gain a new fatal-failure surface (`utils.ErrExit` on query error) for a map that neither `prepareCdcPartitionKey` (gated to target-role PG live) nor `streamChanges` (never called in those flows) will use.

Suggested fix: gate the fetch on `changeStreamingIsEnabled(importType)` — that covers both consumers (strategy prepare + streaming init, including fall-back/fall-forward which are streaming imports) while sparing offline and file imports.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -181,18 +191,60 @@ func resolveEffectiveCdcPartitionKeys(
...
+
+func tableHasGeneratedColumn(generatedStoredCols *utils.StructMap[sqlname.NameTuple, []GeneratedStoredColumn], t sqlname.NameTuple, overrideColumns []string) ([]string, bool) {
+	cols, ok := generatedStoredCols.Get(t)
+	if !ok {
+		return nil, false
+	}
+	generatedCustomCols := make([]string, 0)
+	for _, col := range overrideColumns {
+		if lo.ContainsBy(cols, func(c GeneratedStoredColumn) bool {
+			return c.Name == col
```
makalaaneesh, line 220: [AI] **Cross-PR interaction with #3758: the generated-column check compares un-normalized custom-key names.**

`tableHasGeneratedColumn` compares raw config strings (`c.Name == col`) and runs before `validateCustomPartitionKeyTables` — which is exactly where PR #3758 (same base, in flight) makes custom-key columns case/quote-insensitive by normalizing them to catalog names.

At this head alone there is no hole: a case-mismatched custom column is later rejected by the exact-match existence check in `validateCustomPartitionKeyTables`. Once #3758 merges, `(AMOUNT)` for generated column `amount` passes this check (raw string mismatch), gets normalized to `amount` afterwards, and the table is routed on a generated column → mid-stream failure.

Suggested fix: whichever PR merges second must run this comparison on normalized column names (normalize before `resolveEffectiveCdcPartitionKeys`, or move the generated-column check after normalization).

makalaaneesh (commented):
[AI] Automated review — 7 findings (2 critical, 3 warning, 2 suggestion). See inline comments.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -350,21 +409,36 @@ func getExpressionUniqueIndexTablesIfRequired(tableNames []sqlname.NameTuple, ov
 	return exprUKSet, nil
 }
 
+// getGeneratedStoredColumnsIfRequired resolves the per-table STORED generated columns needed
+// for the CDC partitioning decision. It is intentionally NOT persisted in the import status
+// record: it is recomputed from the source-captured metaDB record (+ target UK/PK) on every
+// run (first run and resume alike) via getGeneratedStoredColumnsHybrid.
+func getGeneratedStoredColumnsIfRequired(tableNames []sqlname.NameTuple, overrides *utils.StructMap[sqlname.NameTuple, cdcPartitionKeyOverride], cdcPartitionKey string, tableToUniqueIndexes *utils.StructMap[sqlname.NameTuple, []tgtdb.UniqueIndex]) (*utils.StructMap[sqlname.NameTuple, []GeneratedStoredColumn], error) {
```
makalaaneesh, line 403: [AI] **The gate on the generated-column fetch is vestigial — and it muddies `checkIfNeedsExprUKCheck`.**

Gating this on `checkIfNeedsExprUKCheck` bought real savings for the expr-UK path (a target-DB query), but here it skips only a metaDB read plus an in-memory intersection with a UK map that `importData` already fetched unconditionally. The costs: `checkIfNeedsExprUKCheck` now gates two unrelated fetches under an expr-UK-specific name, and the added `|| cdcPartitionKey == PARTITION_BY_CUSTOM` arm is dead code — the global key can only be pk/table/auto (`validCdcPartitionKeys`, import.go), and custom overrides already trigger via the override loop.

Suggested fix: delete the `getGeneratedStoredColumnsIfRequired` wrapper, call `getGeneratedStoredColumnsHybrid` directly, and revert the `checkIfNeedsExprUKCheck` edit. Behavior is identical — when every table is table-partitioned the resolve loop ignores the set anyway. Caveat: if the nil-record target-side fallback is restored (see the other comment), a gate becomes worth keeping — in that case keep the wrapper but rename the shared gate (e.g. `needsValueRoutingSafetyChecks`).

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -56,18 +56,18 @@ func prepareCdcPartitionKey(tableNames []sqlname.NameTuple) error {
 		// On resume the per-table strategy map is already persisted. Rather than trusting
 		// a raw string comparison of the flags (done for the global key in
 		// validateCdcPartitionKeyFlags), re-resolve the effective per-table strategy from
-		// the current flags — reusing the expression-UK tables captured on the first run so
-		// no target-DB re-query is needed — and reject if it differs from what was
-		// persisted. This catches semantically-different overrides that a plain string
-		// compare would miss (ordering, spelling/quoting, whitespace).
-		if err := validateCdcPartitioningStrategyUnchanged(tableNames, importDataStatus); err != nil {
+		// the current flags — reusing the expression-UK and generated-stored-column tables
```
makalaaneesh, line 59: [AI] **Stale resume comment: "no target-DB re-query is needed" is no longer true.**

The UK map is now a fresh target query on every run (hoisted to `importData`), and the generated-column set is deliberately recomputed from it plus the export-captured record on every run rather than persisted. Suggest updating the comment to describe the actual inputs: persisted expr-UK list, fresh target UK map, export-captured generated-column record.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
makalaaneesh: don't think we should restore target-side fetching here; let's think about this more

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
makalaaneesh: not sure about the suggested fix, but let's think about this

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
priyanshi-yb: yeah initially Claude suggested to add faillback on target for me also, but I didn't add it intentionally, as this (Export status not found) is only possible in an upgrade scenario so i think its okay to not do anything that is current behaviour without the change.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
priyanshi-yb: for now, I have renamed the function to `needsNonTablePartitionKeyChecks` generic handling

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -225,13 +250,14 @@ func resolveCdcPartitionKeyOverrides(rawOverrides map[string]cdcPartitionKeyOver
 	return resolved, nil
 }
 
-func checkIfNeedsExprUKCheck(cdcPartitionKey string, overrides *utils.StructMap[sqlname.NameTuple, cdcPartitionKeyOverride]) (bool, error) {
-	// Collect expression-UK tables whenever we need them for auto resolution or pk-on-expr-UK validation.
-	needsExprUKCheck := cdcPartitionKey == "auto" || cdcPartitionKey == PARTITION_BY_PK
-	if !needsExprUKCheck {
+func needsNonTablePartitionKeyChecks(cdcPartitionKey string, overrides *utils.StructMap[sqlname.NameTuple, cdcPartitionKeyOverride]) (bool, error) {
```
makalaaneesh, line 260: /*
if cdcPartitionKey != table { return true }
if any override is NOT table {return true }
*/

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -160,6 +163,10 @@ func resolveEffectiveCdcPartitionKeys(
 			for _, t := range tableNames {
 				if _, ok := exprUKSet.Get(t); ok {
 					result.Put(t, cdcPartitionKeyOverride{Strategy: PARTITION_BY_TABLE})
+				} else if tableHasUniqueIndexOnGenerated(generatedStoredCols, t) {
```
makalaaneesh: let's have a single function that determines whether the table is allowed to have any config that is NOT PARTITION_BY_TABLE. 

auto and validation should re-use the same function

yb-voyager/cmd/live_migration.go
```diff
@@ -243,9 +244,9 @@ func streamChangesFromSegment(
 				we need to use the actual source db type at the moment.
 			*/
 			sourceDBTypeForConflictCache := lo.Ternary(isTargetDBExporter(event.ExporterRole), YUGABYTEDB, sourceDBType)
-			err = initializeConflictDetectionCache(evChans, sourceDBTypeForConflictCache, importTableList, tablePartitionKeyMap, tableToPKColumns)
+			err = initializeConflictDetectionCache(evChans, sourceDBTypeForConflictCache, importTableList, tablePartitionKeyMap, tableToPKColumns, tableToUniqueIndexes)
 			if err != nil {
-				return fmt.Errorf("error initializing conflict detection cache: %w", err)
+				return goerrors.Errorf("error initializing conflict detection cache: %v", err)
```
makalaaneesh: %w

yb-voyager/src/srcdb/mysql.go
```diff
@@ -511,6 +511,13 @@ func (ms *MySQL) GetPrimaryKeyColumns(tables []sqlname.NameTuple) (*utils.Struct
 	panic("not implemented")
 }
 
+// GetGeneratedStoredColumns is a no-op for MySQL: STORED generated columns and the CDC
+// custom/pk partitioning that consumes this info are only supported for a PostgreSQL source
+// (MySQL live migration always uses PARTITION_BY_TABLE).
+func (ms *MySQL) GetGeneratedStoredColumns(tableList []sqlname.NameTuple) (*utils.StructMap[sqlname.NameTuple, []string], error) {
```
makalaaneesh, line 517: let's just return a notImplementedError for mysql/oracle instead of en empty list

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -3248,3 +3238,94 @@ func TestLiveMigrationPartitionedTableChildPKWithCustomCdcPartitionKeyPKRecycleC
...
+	err = lm.SetupSchema()
+	testutils.FatalIfError(t, err, "failed to setup schema")
+
+	err = lm.StartExportData(true, nil)
+	testutils.FatalIfError(t, err, "failed to start export data")
+
+	// A custom key that is a stored generated column must fail during prepare.
+	err = lm.StartImportData(false, map[string]string{
+		"--cdc-partition-key":           "table",
+		"--cdc-partition-key-overrides": "test_schema.orders:(bucket)",
```
makalaaneesh, line 3300: can we also test for rejections when 
- cdc-partition-key=pk or override=pk

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -170,27 +185,44 @@ func resolveEffectiveCdcPartitionKeys(
...
+		if col.InUniqueIndex {
+			return true
 		}
 	}
-	return result, nil
+	return false
+}
+
+// Generated cols and override Columns are Unquoted cased names with proper casing
+func tableHasGeneratedColumn(generatedStoredCols *utils.StructMap[sqlname.NameTuple, []GeneratedStoredColumn], t sqlname.NameTuple, overrideColumns []string) ([]string, bool) {
```
makalaaneesh, line 212: this is actually `tableCustomPartitionKeyHasGeneratedCOlumn` right?

priyanshi-yb: function is generic where we are passing a table and checking if it has a generated column; we are using it to check for custom ones.

makalaaneesh (approved):
LGTM

PR #3746, opened 2026-08-25
custom follow up2: Add PK to the unique index list if the table is custom partitioned

priyanshi-yb:

Live-migration import distributes CDC events across parallel channels by hashing a routing key. Under the `custom` partition-key strategy, a table is routed by a user-chosen column list, so two events for the same primary key can carry *different* custom-key values, land on different channels, and be applied out of order. The most dangerous case is a recycled primary key (e.g. `DELETE id=5` then `INSERT id=5` with a different custom key): the re-insert can be applied before the delete, producing a duplicate primary key. `GetTableToUniqueIndexesMap` deliberately excludes the primary key (it filters out `contype='p'`) because under `pk` routing same-PK events always co-locate on one channel and can never race — but that assumption breaks under `custom` routing.

This PR makes custom-key routing correct by adding the primary key as a synthetic unique index to the per-table conflict-detection set for every custom-partitioned table (`addPrimaryKeyToConflictSetForCustomTables`, called from `initializeConflictDetectionCache`). Conflict detection then serializes PK-recycle races across channels. It skips tables where an existing unique index already covers exactly the PK columns (`uniqueIndexWithSameColumnsExists`) to avoid a redundant synthetic index, and errors if a custom-key table has no primary key. It is a no-op for `pk`/`table` routing.

To fetch primary keys efficiently for many tables at once, the `TargetDB` interface swaps the single-table `GetPrimaryKeyColumns` for a batched `GetPrimaryKeyColumnsForTables`, implemented with one catalog query (`(nspname, relname) IN (...)`) for YugabyteDB and PostgreSQL. The query keeps `ORDER BY array_position(i.indkey, a.attnum)` so PK column order is preserved — order matters because it is used to build conflict-bucket keys. Oracle implements the batched method by delegating to the per-table variant (Oracle targets are only used in fall-forward/fall-back, where the fast path doesn't apply). Existing callers (`runPKConflictModeGuardrails`, `getImportBatchArgsProto`) are updated to the batched API.

No new CLI flags, configuration, installation, or report changes. This is a correctness fix for the `custom` `--cdc-partition-key` / `--cdc-partition-key-overrides` routing: PK-recycle races on custom-partitioned tables are now detected and serialized, preventing duplicate-key / out-of-order application during live migration.

- New live-migration integration test `TestLiveMigrationCustomCdcPartitionKeyPKRecycleConflict` (in `src/testlivemigration/live_migration_uk_conflict_test.go`): a `custom`-routed table with **no** unique index other than the primary key, whose delta recycles each PK with a new custom-key value in a single transaction. Using the `uniqueKeyConflictDetected=return("count")` failpoint, it asserts a non-zero detected-conflict count (proving the synthetic PK guard fires) and validates target/source data consistency plus cutover.
- Existing conflict-detection unit and integration tests continue to exercise the batched `GetPrimaryKeyColumnsForTables` path and the non-custom routing (no-op) behavior.

No.

No. The only interface change (`GetPrimaryKeyColumns` → `GetPrimaryKeyColumnsForTables`) is an in-process Go interface; there are no on-disk/serialized-format changes in this PR.

---

| Component        


yb-voyager/cmd/live_migration.go
```diff
@@ -639,11 +639,104 @@ func initializeConflictDetectionCache(evChans []chan *tgtdb.Event, sourceDBTypeF
...
+	if err != nil {
+		return fmt.Errorf("get primary key columns for custom-key tables: %w", err)
+	}
+
+	for _, table := range customTables {
+		pkColumns, _ := tableToPKColumns.Get(table)
+		if len(pkColumns) == 0 {
+			// Live migration requires a primary key, so this is not expected; fail rather than
+			// adding an empty-column index that would match every row.
+			return goerrors.Errorf("no primary key columns found for custom-key table %s", table.ForOutput())
```
makalaaneesh, line 695: [AI] **PK-existence failure fires after the entire snapshot, not at the prepare-time guardrail.**

`addPrimaryKeyToConflictSetForCustomTables` errors with `"no primary key columns found for custom-key table"` here — but this runs inside `initializeConflictDetectionCache`, which is called from `streamChangesFromSegment` (`live_migration.go:243`), i.e. at **streaming start, after the whole snapshot import completes**. Follow-up 1 (#3729) deliberately placed custom-key guardrails *before* snapshot to fail fast; this check breaks that contract.

A PK-less table under `custom` routing is a real configuration: #3508 supports partitioned roots without a root PK (children carry PKs, `use-partition-root`), and followup1's `validateCustomPartitionKeyTables` only validates *column existence* — so the user imports the entire snapshot (hours/days) and then hard-fails on every streaming attempt, with no recovery except reconfigure + `--start-clean`.

Failure case: partitioned root without a root PK (#3508) + `--cdc-partition-key-overrides root:(region)` → snapshot completes → streaming init errors forever.

Suggested fix: also check PK existence in `validateCustomPartitionKeyTables` (prepare stage, which already queries the target per custom table), and keep this init-time error as defense in depth:

```go
// in validateCustomPartitionKeyTables, alongside the column-existence check
tableToPK, err := tdb.GetPrimaryKeyColumnsForTables(customTables)
...
if cols, _ := tableToPK.Get(t); len(cols) == 0 {
    return goerrors.Errorf("cdc-partition-key-overrides: table %s has no primary key; a custom partition key requires one (PK-recycle conflicts cannot be detected otherwise)", t.ForOutput())
}
```

---

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -600,39 +600,62 @@ outer:
...
+// ybTargetQueryGetPrimaryKeyColumnsForTables returns the PK columns of all tables in the
+// given list. ORDER BY array_position(indkey, attnum) is essential: (id, region) and
+// (region, id) are different keys, and the PK column order is used to build conflict-bucket
+// keys during live-migration conflict detection.
+var ybTargetQueryGetPrimaryKeyColumnsForTables = `
+SELECT n.nspname, c.relname, a.attname
+FROM pg_index i
+JOIN pg_class      c ON c.oid = i.indrelid
+JOIN pg_namespace  n ON n.oid = c.relnamespace
+JOIN pg_attribute  a ON a.attrelid = c.oid AND a.attnum = ANY(i.indkey)
```
makalaaneesh: [AI] **PK query includes INCLUDE (covering) columns, narrowing the synthetic index and missing conflicts.**

`a.attnum = ANY(i.indkey)` picks up *all* index columns. For `PRIMARY KEY (id) INCLUDE (val)` (PG 11+ syntax; the PG target driver carries the identical query), `val` enters the synthetic index's columns. Conflict buckets then key on `(id, val)`: a recycled `id` with a different `val` lands in a *different* bucket → the PK-recycle race this PR exists to catch goes undetected → out-of-order apply → duplicate-PK failure.

The sibling query in this same file (`pgQueryTmplForUniqIndexes`) already filters exactly this with a documented `indnkeyatts` clause — this new query should do the same.

Failure case: target PK declared with `INCLUDE`; `DELETE (id=5, val=1)` + `INSERT (id=5, val=2)` routed under different custom keys → no conflict detected.

Suggested fix (applies to the identical query in `postgres.go` too):

```sql
AND array_position(i.indkey, a.attnum) + 1 <= i.indnkeyatts  -- indkey is 0-indexed; exclude INCLUDE columns
```

---

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -337,43 +337,67 @@ outer:
...
-		JOIN pg_namespace  n ON n.oid = c.relnamespace
-		JOIN pg_attribute  a ON a.attrelid = c.oid AND a.attnum = ANY(i.indkey)
-		WHERE n.nspname = '%s'
-			AND c.relname  = '%s'
-			AND i.indisprimary;`, schemaName, tableName)
+// pgTargetQueryGetPrimaryKeyColumnsForTables returns the PK columns of all tables in the
+// given list. ORDER BY array_position(indkey, attnum) is essential: (id, region) and
+// (region, id) are different keys, and the PK column order is used to build conflict-bucket
+// keys during live-migration conflict detection.
+var pgTargetQueryGetPrimaryKeyColumnsForTables = `
```
makalaaneesh: [AI] **PK query + scan loop duplicated verbatim across both target drivers (with a third copy in srcdb).**

`pgTargetQueryGetPrimaryKeyColumnsForTables` (here) and `ybTargetQueryGetPrimaryKeyColumnsForTables` (`yugabytedb.go:607`) are byte-identical query vars, and the two `GetPrimaryKeyColumnsForTables` method bodies are identical ~35-line copies (also cloned from `src/srcdb/postgres.go`).

Per `yb-voyager/src/.cursor/BUGBOT.md` / root rules ("When the same SQL pattern is used for both PostgreSQL and YugabyteDB, extract it into a shared constant"; "consolidate duplicate logic into shared helpers") — and this package already sets the precedent with `pgQueryTmplForUniqIndexes` + `queryPGUniqueIndexesByCatalog` shared by both drivers. Concretely, the `indnkeyatts` fix suggested on the sibling comment now has to be applied in 2–3 places instead of one. Also, these should be `const`, not `var` (per `yb-voyager/.cursor/BUGBOT.md`).

Suggested fix: one `const pgQueryTmplPKColumnsForTables` + one package-level helper `queryPGPrimaryKeyColumnsByCatalog(queryFn, tables)` used by both drivers, mirroring the unique-indexes pattern.

---

makalaaneesh (commented):
[AI] Automated review — 3 findings (0 critical, 3 warning). See inline comments.

Scope: the batched PK-fetch + synthetic-PK-for-custom-routing change. All three are correctness/maintainability concerns on the custom `--cdc-partition-key` path. (A 4th manually-noted finding about an empty PR description was dropped — the description is now fully populated, including the testing statement.)

---

yb-voyager/cmd/live_migration.go
```diff
@@ -639,11 +639,104 @@ func initializeConflictDetectionCache(evChans []chan *tgtdb.Event, sourceDBTypeF
 		return fmt.Errorf("get table unique indexes map from target: %w", err)
 	}
 
+	// For custom-key tables the primary key must join the checked constraints (see
+	// cdc_partition_key_followups.md, Follow-up 3.2). GetTableToUniqueIndexesMap deliberately
```
makalaaneesh: pls fix up comments referring to files that won't be accessible

makalaaneesh: pls add an exact example case to explain why we need to add PKs to the list

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -600,39 +600,62 @@ outer:
...
+WHERE (n.nspname, c.relname) IN (%s)
+  AND i.indisprimary
+ORDER BY array_position(i.indkey, a.attnum);`
```
makalaaneesh: AI: 

rows from different tables interleave, and per-table order survives only because a sorted sequence's subsequence stays sorted while Go appends per table. ORDER BY   n.nspname, c.relname, array_position(...) would make it obviously right at no cost

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -71,6 +71,49 @@ func prepareCdcPartitionKey(tableNames []sqlname.NameTuple) error {
...
+	}
+
+	var tablesWithoutPK []sqlname.NameTuple
+	for _, t := range tableNames {
+		if pkColumns, _ := tableToPKColumns.Get(t); len(pkColumns) == 0 {
+			tablesWithoutPK = append(tablesWithoutPK, t)
+		}
+	}
+	if len(tablesWithoutPK) > 0 {
+		return nil, goerrors.Errorf("cdc-partition-key-overrides: table(s) %v have no primary key on the target; a custom partition key requires one so that PK-recycle conflicts can be detected. Use the 'table' strategy for these tables instead", tablesWithoutPK)
```
makalaaneesh: let's fix the error message; it should not talk about cdc-partition-key-overrides.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -71,6 +71,49 @@ func prepareCdcPartitionKey(tableNames []sqlname.NameTuple) error {
...
+not need to be persisted in metaDB.
+
+It also fails fast, before the snapshot import, if a table routed by a custom partition key
+has no primary key on the target: custom routing adds the PK as a synthetic unique index for
+conflict detection (a recycled PK across different custom keys must be serialized), so a
+custom-key table without a PK cannot be made correct. This is scoped to custom-key tables so
+that legitimately PK-less tables under pk/table routing (e.g. partitioned roots imported via
+--use-partition-root) are not blocked.
+*/
+func getPrimaryKeyColumnsForImportTables(tableNames []sqlname.NameTuple) (*utils.StructMap[sqlname.NameTuple, []string], error) {
```
makalaaneesh: let's move this func out of this file; importData.go?

yb-voyager/cmd/live_migration.go
```diff
@@ -632,18 +633,112 @@ func processEvents(chanNo int, evChan chan *tgtdb.Event, lastAppliedVsn int64, d
...
+
 	log.Infof("initializing conflict detection cache")
 	conflictDetectionCache = NewConflictDetectionCache(tableToUniqueIndexes, evChans, sourceDBTypeForConflictCache, tablePartitionKeyMap)
 	return nil
 }
 
+// addPrimaryKeyToConflictSetForCustomTables appends the primary key as a synthetic unique
+// index to the per-table conflict-detection set for every table routed by a custom partition
+// key. This is required for correctness under custom routing (see
+// initializeConflictDetectionCache and cdc_partition_key_followups.md, Follow-up 3.2); it is
```
makalaaneesh: fix comment

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -337,41 +337,126 @@ outer:
...
+	result := utils.NewStructMap[sqlname.NameTuple, []string]()
+	if len(tables) == 0 {
+		return result, nil
+	}
 
-	rows, err := pg.Query(query)
+	// getPartitionTableToRootTableMap returns, for every table whose root is in tables, a
+	// mapping of its catalog name ("schema.table") to its root's catalog name. This includes
+	// each leaf partition -> root, and each root/normal table -> itself.
+	tableToRootMap, err := getPartitionTableToRootTableMap(queryFn, tables)
```
makalaaneesh, line 384: Pls create a separate task to tackle this refactor as we discussed. This should be done outside of the target DB implementation as this is a special case only relevant for live migration

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -2430,4 +2430,144 @@ func TestLiveMigrationWithSubsetOFPartialUNiqueIndexColumnsBeingChangedInUpdate(
 
 	err = liveMigrationTest.WaitForCutoverComplete(0, 30)
 	testutils.FatalIfError(t, err, "failed to wait for cutover complete")
+}
+
+
+// TestLiveMigrationCustomCdcPartitionKeyPKRecycleConflict verifies the primary-key guard for
```
makalaaneesh, line 2436: let's add tests for custom partition key on 
- partitioned table
- partitioned table where root does not have PK but child partitions do.

yb-voyager/src/tgtdb/postgres.go
priyanshi-yb: yeah will do

makalaaneesh (approved):
LGTM

PR #3751, opened 2026-08-25
Support PostgreSQL 18 as source: raise version guardrail, install PG 18 client, add pg-18 migtests workflow

Abhi-yb:

- `MAX_SUPPORTED_PG_VERSION` 17 → 18 (`src/srcdb/postgres.go`).
- Installer: PG 18 client (`postgresql-client-18` apt / `postgresql18` yum) — required with the guardrail bump, since export checks that pg_dump >= source version. pg_dump 18 still dumps older sources.
- New `pg-18-migtests.yml` workflow mirroring pg-17 against a `postgres:18` source (contrib is part of the base package on 18; postgis installed in-container).
- `multiple-schemas`: expect 41 `schema2` routines when the source/source-replica is PG >= 18 (pgcrypto ships an extra function there). Gated on the *source* version, so it stacks with the existing YB-version gate rather than replacing it.

Notes:
- This PR's pg-18 matrix passed 15/15, incl. live migration on bundled debezium 2.5.2. Official debezium support for PG 18 lands with DB-23172 (3.x).
- Docs (supported versions page) tracked separately in the docs repo.
- The red `TestLatestStable` check is the pre-existing stale YB pin, unrelated to this PR.
- Jenkins-side PG 18 provisioning lives in jenkins-helpers#1568; that branch has been rebased onto master so it picks up jenkins-helpers#1545 (concurrent-DDL gflags for YB 2026.1+, GH #29856). Without those gflags the parallel migtest suites break on 2026.1 at `CREATE DATABASE` regardless of source PG version.

- Voyager accepts PostgreSQL 18 sources (previously rejected: "supported range 9 to 17").
- Installer provisions PG 18 client tools instead of 17.

- The new pg-18-migtests workflow on this PR (15/15 pass).
- Cross-verified on Jenkins real-agent infra: https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/1172/

No.

No.


ShivanshGahlot (approved):
LGTM

migtests/tests/pg/multiple-schemas/validate
```diff
@@ -69,6 +69,12 @@ def check_validations_per_schema(tgt, schema_name, threeDotVersion):
 	print(f"count of fetched procedures/functions - {len(fetched_procedures_functions)}")
 	# an extra routine ships from YB 2025.2.4, bumping the count from 40 to 41
 	schema2_count = 41 if (threeDotVersion and Version(threeDotVersion) >= Version("2025.2.4")) else 40
+	# pgcrypto ships an extra function on PG 18, so the fall-forward/fallback
```
makalaaneesh, line 72: [AI][Warning] **PG-18 count logic copy-pasted across two validate scripts instead of a shared lib helper.**

This identical 6-line PG-18 detection block is duplicated verbatim here and in `migtests/tests/pg/multiple-schemas/validateAfterChanges` (around line 106). `migtests/AGENTS.md` requires the opposite: "Reusable validation patterns should be library functions in `lib/yb.py`, not copy-pasted across test validate scripts" and "Fix shared library functions (`yb.py`) rather than patching individual test validation scripts."

It also creates a second source of truth for "is this a YB target": the new inline `"YB" not in version_string` string test sits next to the existing `get_three_dot_version` regex (`YB-…`) in `lib/yb.py` that already answers the same question — two detections for one fact, which the root AGENTS.md "One source of truth" rule warns against.

Failure case: if YB's `version()` string format changes (or another PG-major bump is needed), one copy is likely updated and the other left stale, and the YB-vs-plain-PG test drifts from `get_three_dot_version`.

Suggested fix: add one helper to `migtests/lib/yb.py` and call it from both scripts, e.g.:

```python
def plain_pg_major_at_least(version_string, major):
    # True only for a non-YB (plain PostgreSQL) server at >= the given major version.
    if "YB" in version_string:
        return False
    try:
        return int(version_string.split()[1].split(".")[0]) >= major
    except (IndexError, ValueError):
        return False
```

then in each validate script:

```python
if yb.plain_pg_major_at_least(tgt.get_target_version(), 18):
    schema2_count = 41
```

---

migtests/tests/pg/multiple-schemas/validate
```diff
@@ -69,6 +69,12 @@ def check_validations_per_schema(tgt, schema_name, threeDotVersion):
...
+	# checks against the source/source-replica also see 41 (threeDotVersion is
+	# empty there — it is derived from the YB-x.y.z token, absent on plain PG)
+	version_string = tgt.get_target_version()
+	if "YB" not in version_string and int(version_string.split()[1].split(".")[0]) >= 18:
```
makalaaneesh, line 76: [AI][Suggestion] **Brittle version parsing can raise instead of returning a clean result.**

`int(version_string.split()[1].split(".")[0])` assumes `version()` is always `PostgreSQL <numeric-major>.…`. A beta/RC image (`PostgreSQL 18beta1 …`) makes `int("18beta1")` raise `ValueError`, and any unexpected shape makes `split()[1]` raise `IndexError` — either aborts the whole validate run rather than degrading gracefully. The same line is duplicated in `validateAfterChanges` (around line 110).

Failure case: running the multiple-schemas fall-forward/fall-back validation against a PG beta/RC container crashes the validator with an unhandled exception instead of an assertion message.

Suggested fix: fold this into the `lib/yb.py` helper suggested above with a `try/except (IndexError, ValueError)` guard, so a malformed version string falls back to the default count path instead of crashing.

---

makalaaneesh (commented):
[AI] Automated review — 2 findings (0 critical, 1 warning, 1 suggestion). See inline comments.

---

PR #3752, opened 2026-08-26
Set idle_in_transaction_session_timeout on import-data connections

makalaaneesh:

Sets `idle_in_transaction_session_timeout = '5min'` on voyager's import-data connections.

**The problem.** If voyager exits abruptly (`kill -9`, OOM, container eviction) while one of its pooled connections is inside a transaction, the target backend does not notice the client is gone. The session stays `idle in transaction` and keeps holding its locks until TCP keepalive reaps it — roughly two hours by default. A later import that touches those rows wedges behind the orphan. This has been observed in the field: CDC apply wedged on a lock held by a stale connection left behind by a voyager restart.

**Why voyager has to set it.** YSQL defaults this GUC to `0` (off), so nothing bounds it today. Setting it server-side via `ysql_pg_conf_csv` or `ALTER DATABASE`/`ALTER ROLE` is not reliable either — a session-level value overrides the cluster default, and we have seen exactly that happen. Setting it on voyager's own sessions is what makes it stick.

**Scope.** Applied to the import-data connection pool for YugabyteDB and PostgreSQL targets, via `getYBSessionInitScript` / `getPGSessionInitScript`. Since `InitConnPool` is only ever called from `cmd/importData.go`, one change covers offline, live, changes-only, and iterative-cutover flows; fall-back / fall-forward to PG go through the PG builder.

Deliberately **not** applied to:
- **Export paths** — they legitimately hold long `pg_dump` snapshot transactions and must not be capped.
- **Oracle targets** — `TargetOracleDB.InitConnPool` does not use `SessionInitScript`, and Oracle has no equivalent GUC. Genuine no-op.
- **The single ad-hoc target connection** (`yb.conn_` / `pg.conn_`) — `connect()` does not run `initSession`; it serves metadata/DDL, not the apply path this addresses.

**Why 5min is safe.** Every gap voyager can produce *inside* a transaction is a single round trip — `importBatch`/`copyBatchCore` is BEGIN → open file → set schema → dedupe check → `CopyFrom` → record → COMMIT, and `ExecuteBatch` builds all statements before entering `WithConn`. During `COPY` the backend state is `active`, not `idle in transaction`, so a slow COPY is never a target. The one genuinely slow wait, `WaitUntilNoConflict`, runs in `handleEvent` outside any transaction. 5min is orders of magnitude above any real gap.

**Ordering.** The new statement is deliberately **first** in the session script. `ConnectionPool.initSession` returns early on a permission-denied error, silently skipping every remaining variable, and `session_replication_role` is the privileged one. Ordering this ahead of it means it can never be the one dropped. (The early-return itself is pre-existing and left alone here.)

The diff is intentionally minimal: one new constant, one gated append in each of the two session-script builders, and one entry in `defaultSessionVars` (the fallback the pool uses when no script is supplied).

No CLI, config, installation, or report changes.

Behavioural change worth knowing about: import-data sessions now self-terminate after 5 minutes idle **inside a transaction**, surfacing as `FATAL: 25P03: terminating connection due to idle-in-transaction timeout`. The connection pool already handles this — `WithConn` discards the failed connection, clears its prepared-statement cache, and the next caller reconnects with the session script re-applied.

An operator who needs a different value, or `0` to disable, can append a `SET` to `/etc/yb-voyager/ybSessionVariables.sql`; those entries are applied after the built-ins and the last `SET` wins. Note this file is read on the YB path only.

The effective value is visible in the logs via the existing `YBSessionInitScript: ...` line, which is useful for support.

**New integration tests** (`src/tgtdb/idle_in_transaction_timeout_test.go`, `-tags integration`, 2 tests):
- A connection handed out by the import pool actually reports `5min`.
- With a 1s timeout, an idle transaction is terminated, the caller sees SQLSTATE `25P03`, and the *next* `WithConn` succeeds on a fresh connection — exercising the pool's discard-and-reconnect path.

**Please note:** these two tests compile and vet cleanly but **I could not execute them locally** — this package's shared `TestMain` starts PostgreSQL, Oracle, a single-node YB and a 3-node YB cluster before any test runs, and the Oracle container failed on my machine (`ORA-12547: TNS:lost contact`) with ~8GB of Docker memory. They will run for the first time in the `integration` CI job. Reviewers may want to confirm that job goes green before merging, since these are the only tests covering the runtime behaviour.

**No unit tests.** `getYBSessionInitScript` / `getPGSessionInitScript` call `checkSessionVariableSupport`, which opens a real connection, so the script composition is not unit-testable without injecting that dependency. I had done exactly that in an earlier revision and reverted it as out of scope for this change — the refactor was larger than the fix it enabled. Coverage therefore lives entirely in the integration tier.

Existing tests were otherwise sufficient: full `-tags unit` suite passes (29 packages), `go build ./...`, `go vet ./...` and `go vet -tags integration ./src/tgtdb/` all clean, gofmt clean.

No manual end-to-end migration run was performed.

No changes to callhome or yugabyted payloads.

No. Nothing in the table below is touched. The change only adds a `SET` statement to the in-memory session script applied to import connections; `TargetConf.SessionVars` is an internal cache and is not persisted to the MSR or any on-disk structure.

### DIFF
--- a/yb-voyager/src/tgtdb/idle_in_transaction_timeout_test.go
+++ b/yb-voyager/src/tgtdb/idle_in_transaction_timeout_test.go
@@ -33,9 +33,8 @@ import (
 // transaction for longer than idle_in_transaction_session_timeout.
 const sqlStateIdleInTxnTimeout = "25P03"
 
-// The unit tests cover how the session script is composed; this covers the part
-// that only a real target can show: that a connection handed out by the import
-// pool actually carries the timeout.
+// Covers what only a real target can show: that a connection handed out by the
+// import pool actually carries the timeout.
 func TestPooledConnectionHasIdleInTransactionTimeoutSet(t *testing.T) {
 	yb, ok := testYugabyteDBTarget.TargetDB.(*TargetYugabyteDB)
 	require.True(t, ok, "expected a *TargetYugabyteDB")
--- a/yb-voyager/src/tgtdb/idle_in_transaction_timeout_test.go
+++ b/yb-voyager/src/tgtdb/idle_in_transaction_timeout_test.go
@@ -1,0 +1,101 @@
+//go:build integration
+
+/*
+Copyright (c) YugabyteDB, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package tgtdb
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgconn"
+	"github.com/jackc/pgx/v4"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// SQLSTATE raised when the server terminates a session that sat idle inside a
+// transaction for longer than idle_in_transaction_session_timeout.
+const sqlStateIdleInTxnTimeout = "25P03"
+
+// The unit tests cover how the session script is composed; this covers the part
+// that only a real target can show: that a connection handed out by the import
+// pool actually carries the timeout.
+func TestPooledConnectionHasIdleInTransactionTimeoutSet(t *testing.T) {
+	yb, ok := testYugabyteDBTarget.TargetDB.(*TargetYugabyteDB)
+	require.True(t, ok, "expected a *TargetYugabyteDB")
+
+	// Init() runs during suite setup, which is what populates SessionVars.
+	require.NotEmpty(t, yb.Tconf.SessionVars)
+
+	pool, err := NewConnectionPool(&ConnectionParams{
+		NumConnections:    1,
+		NumMaxConnections: 1,
+		ConnUriList:       []string{testYugabyteDBTarget.GetConnectionString()},
+		SessionInitScript: yb.Tconf.SessionVars,
+	})
+	require.NoError(t, err)
+
+	var timeout string
+	err = pool.WithConn(func(conn *pgx.Conn) (bool, error) {
+		return false, conn.QueryRow(context.Background(),
+			"SHOW idle_in_transaction_session_timeout").Scan(&timeout)
+	})
+	require.NoError(t, err)
+	assert.Equal(t, "5min", timeout)
+}
+
+// The whole point of the timeout is that the server kills an orphaned
+// in-transaction session. When that happens to a live pooled connection, the
+// caller must see 25P03 and the pool must recover: WithConn() discards the dead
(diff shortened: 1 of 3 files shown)


priyanshi-yb (approved):
LGTM

yb-voyager/src/tgtdb/idle_in_transaction_timeout_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/jackc/pgx/v4"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// SQLSTATE raised when the server terminates a session that sat idle inside a
+// transaction for longer than idle_in_transaction_session_timeout.
+const sqlStateIdleInTxnTimeout = "25P03"
+
+// The unit tests cover how the session script is composed; this covers the part
```
makalaaneesh: [AI] **Comment claims unit-test coverage that does not exist; the load-bearing "first" ordering invariant is untested in default CI.**

This comment states "The unit tests cover how the session script is composed", but there are no unit tests for the composition. The only tests touching `getYBSessionInitScript`/`getPGSessionInitScript` behavior are the two `//go:build integration` tests in this file — nothing in the `unit`-tagged files (`conn_pool_unit_test.go`, `parallelism_unit_test.go`, `attr_name_registry_test.go`) exercises composition. `go test -tags unit ./...` (CI's default tier per `go.yml`) asserts nothing about the session script.

This matters because the PR deliberately places `SET_IDLE_IN_TRANSACTION_SESSION_TIMEOUT` *first* in both `getPGSessionInitScript` and `getYBSessionInitScript`, with a `// first, so a permission-denied error on a later var cannot make initSession() skip it` rationale (`yb-voyager/src/tgtdb/yugabytedb.go:1681`, `:1696`). That ordering is an intentional, load-bearing invariant, yet nothing that runs without a live target DB pins it — a future refactor that reorders the vars would pass unit CI silently, and `TestPooledConnectionHasIdleInTransactionTimeoutSet` only checks the value is set, not that it is first.

Note the composition functions call `checkSessionVariableSupport`, which does a real `pgx.Connect`, so a straight unit test isn't feasible without a small refactor (e.g. injecting the support-check). At minimum, correct the comment so it doesn't overstate coverage; ideally, add a unit-tier assertion on ordering/inclusion.

Suggested fix: reword to "Integration tests cover how the session script is composed and applied;" — and, if the ordering invariant is worth protecting, factor the support-check behind a seam so a `//go:build unit` test can assert the timeout is emitted first.

---

makalaaneesh (commented):
[AI] Automated review — 1 finding (0 critical, 1 warning). See inline comments.

Scope reviewed: the whole PR diff (`conn_pool.go`, `yugabytedb.go`, new integration test) against all migration flows and the tgtdb/src/repo BUGBOT rules. The core change — bounding orphaned in-transaction sessions with `idle_in_transaction_session_timeout='5min'`, gated per-var and placed first — is sound: the 5min bound cannot abort normal imports (snapshot COPY and CDC `ExecuteBatch` keep the txn active with no >5min idle gap), the "first" ordering is safe because the timeout is a USERSET GUC, pool recovery on 25P03 is real (`WithConn` drops the dead conn), and PG/YB paths plus the `defaultSessionVars` fallback are all covered. Only Critical/Warning findings are posted.

---

PR #3755, opened 2026-08-27
Adds YugabyteDB 2026.1 as the latest stable version and marks XML datatype as supported

priyanshi-yb:

Adds YugabyteDB 2026.1 as the latest stable version and adapts issue reporting and tests to it.

- **Versions** — `yb-versions.json`: `2026.1.1.1-b2` becomes latest stable. Series refreshed; `2024.1` and `2.25` are dropped from the supported version matrix.
- **Version-gated datatype reporting** — datatype issues can now carry `MinimumVersionsFixedIn`: below the fixed-in version the reports show the regular "Unsupported datatype" issue; at or above it, that issue is filtered out and only the "Unsupported datatype for Live migration" caveat remains, since the CDC connector still cannot stream the type.
- **How the gate works** — the live-migration unsupported list is now the full debezium list (no set-difference with the offline list). A column whose type is in both lists produces both issues, and `ShouldFilterOutIssue` drops the live caveat whenever the offline issue survived version filtering. The same helper backs both analyze-schema (`finalizeIssues`) and the assessment report.
- **Gated types today** — xml (fixed-in 2026.1) and the six multirange types (fixed-in 2025.1). Multirange columns previously reported nothing at ≥ 2025.1 targets; they now correctly surface as live-migration caveats.
- **Live-migration issues for every debezium-unsupported type** — new `issues_migration_caveats.go` defines the live caveat for all 25 types, so the gate can route any of them.
- **Version-independent descriptions** — TP/EA and perf-recommendation notes no longer embed the target version number, so expected reports stop churning on every latest-stable bump.
- **Expected files** — assessment, analyze-schema, and callhome expectations updated for all of the above (xml and multirange flips, XML_FUNCTIONS dropped at ≥ 2026.1, complexity and InvalidCount recomputes). The pinned `assessment-report-test-with-tdb` (2.25.1.0) keeps the offline entries with the fixed-in map.
- **Test containers** — YugabyteDB testcontainers set the concurrent-DDL preview flags (`ysql_enable_concurrent_ddl`, `ysql_yb_ddl_transaction_block_enabled`, `enable_object_locking_for_table_locks`) on master and tserver, gated to 2026.1+ since older versions abort on unknown gflags.
- **Docs** — `update-yb-latest-stable` skill documents how report expectations drift on a latest-stable bump.

Assessment and analyze-schema reports change: xml columns flip from "unsupported datatype" to a live-migration caveat at target ≥ 2026.1, multirange columns now get live-migration caveats at ≥ 2025.1 targets (previously unreported), and TP/EA notes no longer print the target version number. No CLI or config changes.

`go test -tags unit ./...` passes. The analyze-schema migtest was validated locally end-to-end with a branch-built binary; the new assessment expected entries were cross-checked against the CI run's actual report. Container preview flags verified against `yugabytedb/yugabyte:2026.1.1.1-b2` (master + tserver) and confirmed skipped on 2025.2.

No — payload structure unchanged (expected callhome test payloads updated for the new issue content only).

No — Migration Assessment Report Json content changes, but reports are regenerated per run.

---

| Component        


yb-voyager/src/query/queryissue/detectors_ddl.go
```diff
@@ -351,6 +355,42 @@ func detectHotspotIssueOnConstraint(isPartitionedTable bool, constraintType stri
 	return reportHotspotsOnTimestampTypes(hotspotTypeName, obj.GetObjectType(), obj.GetObjectName(), col, false, usageCategory)
 }
 
+// versionGatedOfflineDatatypeIssues maps a base type name from srcdb.PostgresUnsupportedDataTypes
+// to its offline unsupported-datatype issue, for the types whose support in YugabyteDB is
+// version-gated (has MinimumVersionsFixedIn). Types absent here are unsupported on all versions.
+// Keys must use the exact casing of the entries in srcdb.PostgresUnsupportedDataTypes.
+var versionGatedOfflineDatatypeIssues = map[string]issue.Issue{
+	"XML": xmlDatatypeIssue,
```
makalaaneesh: [AI] **Version-gating requires two synchronized edits — miss this map entry and the live-migration caveat silently disappears.**

Correctness on 2026.1+ for `xml` hinges on both this map entry AND `xmlDatatypeIssue.MinimumVersionsFixedIn`. If a future contributor version-gates another entry in `srcdb.PostgresUnsupportedDataTypes` and only sets `MinimumVersionsFixedIn` on the issue (forgetting to add its uppercase key here), the failure mode is silent:

- `GetPGUnsupportedDatatypes(v)` still returns the type (map lookup misses ⇒ `!ok` path returns `true`).
- `TableIssueDetector.DetectIssues` (this file, ~line 208) takes the first `if isUnsupportedDatatype` branch and emits the *offline* issue — the live-migration variant is never emitted because of the `else if`.
- `getIssuesNotFixedInTargetDbVersion` in `parser_issue_detector.go` then filters that offline issue out because its `MinimumVersionsFixedIn` says it's GA in the target.
- Net: the assessment/analyze-schema report emits **no** caveat at all for that datatype on the new target — exactly the failure the XML gating was added to prevent.

`MinimumVersionsFixedIn` on the issue is already the source of truth for "fixed in version X". This map is a second source of truth kept in sync only by convention, and the invariant it embodies ("if the issue has `MinimumVersionsFixedIn`, it must be listed here with the exact `PostgresUnsupportedDataTypes` casing") is not enforced anywhere.

Suggested fixes (any one):

1. Add a `//go:build unit` test that iterates `srcdb.PostgresUnsupportedDataTypes`, maps each type-name to its `xxxDatatypeIssue` (the same table `ReportUnsupportedDatatypes` already uses in its switch), and asserts: every issue with a non-nil `MinimumVersionsFixedIn` appears as a key in `versionGatedOfflineDatatypeIssues`, and vice versa.
2. Build `versionGatedOfflineDatatypeIssues` at package init from a single typed list of `{TypeName, IssuePtr}` pairs so future additions are one edit.
3. Derive gating directly from a `TypeName -> *issue.Issue` lookup shared with `ReportUnsupportedDatatypes`, and drop this map entirely.

---

makalaaneesh (commented):

---

yb-voyager/src/query/queryissue/issues_ddl.go
```diff
@@ -807,6 +810,22 @@ func NewVectorDatatypeIssue(objectType string, objectName string, sqlStatement s
 	return newQueryIssue(issue, objectType, objectName, sqlStatement, map[string]interface{}{}, map[string]interface{}{})
 }
 
+var xmlLiveMigrationDatatypeIssue = issue.Issue{
```
makalaaneesh: it feels like migration caveats deserves a separate file

yb-voyager/src/query/queryissue/detectors_ddl.go
```diff
@@ -351,6 +355,42 @@ func detectHotspotIssueOnConstraint(isPartitionedTable bool, constraintType stri
...
+}
+
+// GetPGUnsupportedDatatypes returns the datatypes unsupported on the target YugabyteDB version:
+// srcdb.PostgresUnsupportedDataTypes minus the types whose datatype issue is already fixed in
+// targetDbVersion. A nil targetDbVersion returns the full static list.
+func GetPGUnsupportedDatatypes(targetDbVersion *ybversion.YBVersion) []string {
```
makalaaneesh: I don't understand why we need this. 
We've separated issue detection from target-db-version specific support by introducing MinVersionsFixedIn. It will get filtered before returning. 

So, why do we need to pass `targetDbVersion` to `GetPGUnsupportedDatatypes` ?

yb-voyager/src/query/queryissue/issues_ddl.go
priyanshi-yb: done

yb-voyager/src/query/queryissue/detectors_ddl.go
priyanshi-yb: As discussed, changed the approach for filtering the issues of migration caveat if an unsupported issue is detected.

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1324,23 +1326,26 @@ func addAssessmentIssuesForUnsupportedDatatypes(unsupportedDatatypes []utils.Tab
 			// Coneverting queryissue directly to AssessmentIssue would have lead to the creation of a new function which would have required a lot of cases to be handled and led to code duplication
 			// This converted AssessmentIssue is then appended to the assessmentIssues slice
 			queryissue := queryissue.ReportUnsupportedDatatypes(baseTypeName, colInfo.ColumnName, constants.COLUMN, qualifiedColName)
-			checkIsFixedInAndAddIssueToAssessmentIssues(queryissue)
-
-		default:
```
makalaaneesh, line 1329: why remove?

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1269,9 +1270,10 @@ func fetchColumnsWithUnsupportedDataTypes() ([]utils.TableColumnsDataTypes, []ut
 		// Array of enums are now supported with logical connector (default), so not including them as unsupported
 		isUnsupportedDatatypeInLiveWithFFOrFB := isUnsupportedDatatypeInLiveWithFFOrFBList || isUDTDatatype
 
-		switch true {
-		case isUnsupportedDatatype:
+		if isUnsupportedDatatype {
```
makalaaneesh, line 1273: seems safe, but let's try to ensure agents don't unnecessarily make random unrelated changes

yb-voyager/src/query/queryissue/detectors_ddl.go
```diff
@@ -203,7 +205,8 @@ func (d *TableIssueDetector) DetectIssues(obj queryparser.DDLObject) ([]QueryIss
 
 		if isUnsupportedDatatype {
 			issues = append(issues, ReportUnsupportedDatatypes(col.TypeName, col.ColumnName, obj.GetObjectType(), table.GetObjectName()))
-		} else if isUnsupportedDatatypeInLive {
```
makalaaneesh, line 206: is this intentional?

makalaaneesh (commented):
[AI] Automated review — 3 findings (0 critical, 3 warning). See inline comments. Suggestions/questions from the full review were not posted.

yb-voyager/src/query/queryissue/detectors_ddl.go
```diff
@@ -606,7 +737,9 @@ func ReportUnsupportedDatatypesInLiveWithFFOrFB(baseTypeName string, columnName
 
 //ForeignTableIssueDetector handles detection Foreign table issues
 
-type ForeignTableIssueDetector struct{}
+type ForeignTableIssueDetector struct {
+	ParserIssueDetector
```
makalaaneesh: [AI] **`ForeignTableIssueDetector` embeds `ParserIssueDetector` but never uses it.**

The struct gained an embedded `ParserIssueDetector`, and `GetDDLDetector` now copies the whole detector (`ParserIssueDetector: *p` — several maps/slices, shallow-copied per DDL statement) into it. Nothing in `ForeignTableIssueDetector.DetectIssues` references the embedded value — it only uses `srcdb.PostgresUnsupportedDataTypes` and package-level functions. This looks like a leftover from the earlier approach before the live-issue filtering moved to `finalizeIssues`.

Per the repo review rules (YAGNI / unused surface): suggest reverting to the empty struct until a consumer lands:

```go
type ForeignTableIssueDetector struct{}
...
case *queryparser.ForeignTable:
    return &ForeignTableIssueDetector{}, nil
```

yb-voyager/cmd/assessMigrationCommand.go
```diff
@@ -1509,7 +1519,7 @@ func addMigrationCaveatsToAssessmentReport(unsupportedDataTypesForLiveMigration
 				// Coneverting queryissue directly to AssessmentIssue would have lead to the creation of a new function which would have required a lot of cases to be handled and led to code duplication
 				// This converted AssessmentIssue is then appended to the assessmentIssues slice
 				queryIssue := queryissue.ReportUnsupportedDatatypesInLive(baseTypeName, colInfo.ColumnName, constants.COLUMN, qualifiedColName)
-				checkIsFixedInAndAddIssueToAssessmentIssues(queryIssue)
+				checkIsFixedInAndAddIssueToAssessmentIssues(queryIssue, issueTypeMap)
```
makalaaneesh, line 1525: [AI] **Old-style `MigrationCaveats` / `UnsupportedDataTypes` sections don't apply the new precedence filter — the yugabyted payload double-lists columns.**

`ShouldFilterOutIssue` precedence is applied only to `assessmentReport.Issues`. The `columns` slice feeding the `UNSUPPORTED_DATATYPES_LIVE_CAVEAT_FEATURE` objects list is appended unconditionally a few lines above, and `assessmentReport.UnsupportedDataTypes` (populated in `generateAssessmentReport`) is the raw general bucket.

Consequences:
- With `--target-db-version < 2026.1`, an xml/geometry/multirange column now appears in **both** `UnsupportedDataTypes` and `MigrationCaveats` (before this PR the two buckets were disjoint by construction in `fetchColumnsWithUnsupportedDataTypes`).
- With target >= 2026.1, `UnsupportedDataTypes` still lists XML columns even though the issues list (correctly) treats XML as GA and reports only the live-migration caveat.

These fields are `json:"-"` on the report (so migtests never compare them) but are shipped verbatim to the control plane via `yugabyted_event_builder.go` (`UnsupportedDataTypes: assessmentReport.UnsupportedDataTypes`), so the yugabyted UI will show the inconsistency.

Suggested fix: have `checkIsFixedInAndAddIssueToAssessmentIssues` return whether it actually added the issue, and only append to `columns` in that case; mirror the GA-drop for `assessmentReport.UnsupportedDataTypes` so all report surfaces stay consistent.

yb-voyager/src/query/queryissue/parser_issue_detector.go
```diff
@@ -1124,13 +1145,77 @@ func (p *ParserIssueDetector) getDDLIssues(query string) ([]QueryIssue, error) {
...
+			continue
+		}
+		filteredIssues = append(filteredIssues, i)
+	}
+
+	return filteredIssues
+
+}
+
+func ShouldFilterOutIssue(issue QueryIssue, issueTypeMap map[string]bool) bool {
```
makalaaneesh, line 1165: [AI] **No unit tests for `finalizeIssues` / `ShouldFilterOutIssue` or the 16 new live-migration issue types.**

The whole precedence mechanism (detect both general + live issue → version-filter → suppress the live twin when the general issue survived) lands with no Go tests; the behavior is pinned only indirectly by migtests expected JSONs, which don't run per-PR. Per `src/query/queryissue/.cursor/BUGBOT.md`: "Each new issue type must have unit tests verifying the issue is detected."

Failure mode this leaves open: a future edit that renames or version-gates one constant of a general/live pair (e.g. `UNSUPPORTED_DATATYPE_GEOMETRY`) silently breaks suppression — both issues get reported and nothing fails until the migtests pipeline runs.

Suggested: a table-driven unit test asserting, for a `CREATE TABLE` with an `xml` column: target `2025.2` → exactly `UNSUPPORTED_DATATYPE_XML`; target `2026.1` → exactly `UNSUPPORTED_DATATYPE_LIVE_MIGRATION_XML`; plus one multirange case (general issue GA-dropped → live caveat present — the false negative this PR fixes).

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: This is probably not worth fixing

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: It is required, so previously we had the diffing logic for unsupported or unsupported-live datatypes, but now, since we want to report all and then filter later, we need to have both reported from here also but earlier we had if else so we were doing another level of check here to just report unsupported if that is present so I removed that if/else and kept two ifs.

yb-voyager/src/query/queryissue/detectors_ddl.go
priyanshi-yb: yes as mentioned above

yb-voyager/cmd/assessMigrationCommand.go
priyanshi-yb: Yeah, these two are old fields that we don't support now, so not required to fix

makalaaneesh (approved):
LGTM

PR #3758, opened 2026-08-31
Custom key followup3: support case sensitive column names in the config

priyanshi-yb:

`cdc-partition-key-overrides` matched custom key column names by exact string. A wrong case or quoted spelling made a valid column look missing, and on resume it falsely tripped the change-guard.

- Custom key columns are now resolved against the table's actual columns (case-insensitive, quote-aware) via `FindBestMatchingColumnName`, and the canonical name is persisted so runtime event lookup matches.
- This resolution now runs on resume too, not just the first import, so the change-guard compares normalized names and no longer rejects an equivalent re-spelling (e.g. `customer_id` vs `"customer_id"`).

Custom key columns in `--cdc-partition-key-overrides` now accept any case and quoted identifiers:
```
--cdc-partition-key-overrides 'test_schema.orders:("CustomKey")'
```

Unit test `TestValidateCdcPartitioningStrategyUnchanged` (stubbed target DB) and new integration test `TestLiveMigrationCustomCaseSensitiveCdcPartitionKeyNoConflict`.

No.

No — the custom cdc-partition-key feature is new in this release, so the persisted names only differ within it.

---

| Component        


yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -292,13 +292,9 @@ func computeCdcPartitioningStrategyPerTable(tableNames []sqlname.NameTuple, isFi
...
 	}
 
-	// Import-start guardrails for custom-key tables (custom key column existence). These query
-	// the target DB, so only run them on the first import; on resume the config is unchanged
-	// (enforced by validateCdcPartitioningStrategyUnchanged) and re-querying is unnecessary.
-	if isFirstRun {
-		if err := validateCustomPartitionKeyTables(tableToPartitionKeyOverrideMap); err != nil {
-			return nil, nil, err
-		}
+	err = validateCustomPartitionKeyTables(tableToPartitionKeyOverrideMap)
```
makalaaneesh, line 295: what happened to the `isFirstRun` check? why is it okay to remove?

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -482,6 +478,7 @@ func validateCustomPartitionKeyTables(tableToPartitionKeyOverrideMap *utils.Stru
 	if len(customTables) == 0 {
 		return nil
 	}
+	attributeNameRegistry := tgtdb.NewAttributeNameRegistry(tdb, &tconf)
```
makalaaneesh: I thought `AttributeNameRegistry` is embedded in the target db. so you can use `tdb`  directly?

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -490,20 +487,24 @@ func validateCustomPartitionKeyTables(tableToPartitionKeyOverrideMap *utils.Stru
...
-		for _, c := range tableColumns {
-			columnSet[c] = true
-		}
+		overrideColumns := make([]string, 0, len(override.Columns))
 		var missing []string
 		for _, c := range override.Columns {
-			if !columnSet[c] {
+
+			bestMatchingColumnName, err := attributeNameRegistry.FindBestMatchingColumnName(c, tableColumns)
+			if err != nil {
```
makalaaneesh, line 496: we're making an assumption here that error in `FindBestMatchingColumnName` means that the column is missing; 

THat may not be true?

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -2106,6 +2106,181 @@ func TestLiveMigrationCustomCdcPartitionKeyNoConflict(t *testing.T) {
...
+		},
+		SchemaNames: []string{"test_schema"},
+		SchemaSQL: []string{
+			`CREATE SCHEMA IF NOT EXISTS test_schema;
+			CREATE TABLE test_schema.test_live (
+				id int PRIMARY KEY,
+				"CustomKey" int,
+				most_recent boolean
+			);
+			CREATE TABLE test_schema.test_live_multi_case (
```
makalaaneesh, line 2129: can we perhaps add a partitioned table as well ?

makalaaneesh:

---

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
priyanshi-yb: This was a bug where the validation of custom columns was only being done in the first run, but we should do the column validation every time and realised it now because this function is doing the case matching so we need the matching in re-run also.

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
priyanshi-yb: That function returns two errors only, which are when it is not able to figure out the column from the target columns, so I thought it might be okay to throw a missing or incorrect column name error but let me see if I can have specific errors returned from that and handle them

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -478,9 +478,14 @@ func validateCustomPartitionKeyTables(tableToPartitionKeyOverrideMap *utils.Stru
 	if len(customTables) == 0 {
 		return nil
 	}
-	attributeNameRegistry := tgtdb.NewAttributeNameRegistry(tdb, &tconf)
-
+	yb, ok := tdb.(*tgtdb.TargetYugabyteDB)
```
makalaaneesh: why do we need to type assert it to `yb`?

priyanshi-yb: The attribute name registry is only part of TargetYugabyteDB and not the TargetDB interface. So had to get yb and then use attribute name registry

makalaaneesh: it's supported by other target implementations but this is fine for now.

makalaaneesh (approved):
LGTM

PR #3759, opened 2026-08-31
Fix fall-back exporter crash on YB tablet split (DB-20886)

ShivanshGahlot:

A YB tablet split closes the replication stream that fall-back CDC reads from. Debezium treats the split error as retriable and restarts the connector. But the batch's offset flush then throws `This replication stream has been closed`, which is not retriable, so the engine dies before the restart runs and `export data from target` exits 1.

The offset commit now swallows that one error and lets the restart proceed. The batch is already fsynced to the export queue, so no events are lost. Debezium re-reads the batch after restarting. Any other error still propagates.

This mitigates the crash in Voyager. DB-20886 tracks the fix in YB.

No user-facing changes.

New unit test `YbExporterConsumerSplitGuardTest` (3 cases): `mvn -f debezium-server-voyager/debezium-server-voyagerexporter/pom.xml test -Dtest=YbExporterConsumerSplitGuardTest`.
Manually verified on a 6-node YB cluster with low split thresholds — tablets split 1 → 149 under load and the exporter survived every split.
No integration test: the repro needs a multi-node cluster with auto-splitting tuned down.

No.

No.

---

| Component        


debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -369,15 +369,67 @@ public void handleBatch(List<ChangeEvent<Object, Object>> changeEvents,
...
+    /**
+     * True if the throwable, or any cause in its chain, indicates the logical replication
+     * stream is already closed - e.g. after a YB tablet split closed the walsender COPY
+     * stream. Debezium surfaces this from V3PGReplicationStream.checkClose() as a
+     * ConnectException ("This replication stream has been closed") during an offset flush.
+     */
+    private boolean isReplicationStreamClosed(Throwable e) {
+        for (Throwable t = e; t != null; t = t.getCause()) {
+            String msg = t.getMessage();
+            if (msg != null && msg.toLowerCase().contains("replication stream has been closed")) {
```
makalaaneesh: [AI] **Fragile exception-message match with no test pinning the wording.**

The guard's *only* trigger is `msg.toLowerCase().contains("replication stream has been closed")`. That string is the current text of Debezium's `V3PGReplicationStream.checkClose()`, but nothing in this repo pins it: if a future Debezium/pgjdbc bump reworders (e.g. "The replication stream is closed", "Replication stream already closed", i18n) or wraps it in an outer `RuntimeException` whose own `getMessage()` doesn't include the phrase and whose `getCause()` returns `null` because the wrapper dropped the cause, `isReplicationStreamClosed(e)` silently returns `false` and every fall-back / fall-forward exporter goes back to crashing on tablet split — exactly the DB-20886 regression this PR fixes, but now invisible until QA hits a split again. The previous commit removed `YbExporterConsumerSplitGuardTest`, so nothing in CI would catch this drift.

Two concrete, cheap hardenings — either alone is enough:

1. **Match on the exception class as an additional signal.** `V3PGReplicationStream.checkClose()` throws `org.apache.kafka.connect.errors.ConnectException`; recognizing that class (walking the cause chain) makes the guard robust to message rewording, while the substring check stays as a narrowing filter so unrelated `ConnectException`s still propagate:

   ```java
   private boolean isReplicationStreamClosed(Throwable e) {
       for (Throwable t = e; t != null; t = t.getCause()) {
           if (t instanceof org.apache.kafka.connect.errors.ConnectException) {
               String msg = t.getMessage();
               if (msg != null
                       && msg.toLowerCase(java.util.Locale.ROOT).contains("replication stream")
                       && msg.toLowerCase(java.util.Locale.ROOT).contains("closed")) {
                   return true;
               }
           }
       }
       return false;
   }
   ```
   (Also switches to `Locale.ROOT` — a good habit for ASCII substring matching, even though the current phrase doesn't happen to hit the Turkish `I`/`ı` case.)

2. **Reintroduce a small unit test that asserts `isReplicationStreamClosed` returns `true` for the exact wording it depends on**, so a Debezium upgrade that renames the message fails at build time instead of at the next tablet split in production. The commit message notes the previous test was dropped to keep the helper private — a package-private test in the same package, or `@VisibleForTesting`, would preserve the guard without leaking API.

The manual 6-node cluster verification cited in the PR description is great for confirming today's behavior, but it doesn't protect against future Debezium bumps re-introducing the exact same crash.

---

makalaaneesh (commented):

---

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
ShivanshGahlot: Fair point on the fragile exact-sentence match — fixed in 325787a0: the match is now `"replication stream"` + `"closed"` (`Locale.ROOT`, cause chain still walked) instead of debezium's exact sentence, so a reworded message can't silently re-break DB-20886.

Skipped the `instanceof ConnectException`: the message check is still required alongside it, so the class test adds no rewording robustness — it only adds a second mandatory condition, making the guard *narrower* (it would stop matching a bare `PSQLException`, or a different wrapper class).

Skipped the unit test because it wouldn't run: every build path is `mvn -ntp clean install -Dquick` (`installer_scripts/install-yb-voyager:414,429,449,465`) and no workflow invokes `mvn` directly, so this module's tests never execute in CI. Making them run is the real fix and worth its own issue.

Also worth noting: the race now reproduces naturally (no injection) on a local YB with a lowered split threshold — 21 splits → 3 guard fires, 0 fatal exits, ~217k rows written after the first hit all replicated back to source. The unguarded build dies on its first split.

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -369,15 +370,60 @@ public void handleBatch(List<ChangeEvent<Object, Object>> changeEvents,
...
+        } catch (RuntimeException e) {
+            // A tablet split closes the stream, so this offset flush throws (DB-20886).
+            // Nothing is lost: handleBatchComplete() already fsynced the batch, and debezium
+            // is restarting the connector for the same split - propagating here would exit
+            // the engine first, making a recoverable split fatal. Skipping the commit only
+            // re-reads the batch on restart; VSN dedupes downstream. Anything else rethrows.
+            if (isReplicationStreamClosed(e)) {
+                LOGGER.warn("Skipping offset commit for this batch: the replication stream is closed "
+                        + "(typically a YB tablet split). The batch is already durably written to the export "
+                        + "queue; the connector will restart and resume from the last committed offset.", e);
```
makalaaneesh: this seems risky; why is it safe to skip committing offsets??
Can't we just write a retry loop so that it ultimately will succeed after the connector is restarted?

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -369,15 +370,60 @@ public void handleBatch(List<ChangeEvent<Object, Object>> changeEvents,
...
+        }
+    }
+
+    /**
+     * True if any cause in the chain reports a closed replication stream, which debezium
+     * raises from V3PGReplicationStream.checkClose() during an offset flush. Matches on
+     * "replication stream" plus "closed" rather than debezium's exact sentence, so that
+     * rewording it in a future bump cannot silently re-break DB-20886.
+     */
+    private boolean isReplicationStreamClosed(Throwable e) {
```
makalaaneesh, line 427: we should also make sure this is only run when exporter_role is target_db_exporter_*

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
ShivanshGahlot: The offsets are already committed to disk by the time this throws btw. They just have not been propagated to YB. markBatchProcessed tries to use the stream object to do this but throws the error which is caught here. Debezium reads the local offsets on start.

> markBatchProcessed does these things:
> 1. offsetWriter.doFlush(...) followed by Future.get(timeout) — writes offsets to data/offsets.target_db_exporter_fb.dat and waits for completion. This is what Debezium reads on restart.
> 2. task.commit() → PostgresStreamingChangeEventSource.commitOffset → stream.flushLsn() — sends the LSN to YB. This is the only call touching the replication stream, and the one that throws once a split has closed it

So the events are already in queue by this time and the offsets are durable on the disk.
The comment here is probably misleading. There are no duplicates produced since it is resuming from the offsets on disk. Claude also verified this in the split test I ran with it.

About the retry loop, the flushLsn() in the markBatchProcessed will keep on retrying on the same streaming object which is dead. Only a new streaming object can enable this to succeed and that restart also apparently runs on this same consumer thread. So, retrying would mean blocking that new stream object creation.

Also only markBatchFinished can raise this error within handleBatch as that is the only thing communicating with the stream object.

.

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
ShivanshGahlot: Yep makes sense.

makalaaneesh:
[AI] Automated review — no findings.

---

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -369,15 +372,106 @@ public void handleBatch(List<ChangeEvent<Object, Object>> changeEvents,
...
+            catch (RuntimeException e) {
+                // Only target-side exporters stream from YB, so only they can see a split;
+                // anywhere else, and for any other exception, this stays fatal.
+                if (!isTargetDbExporter() || !isReplicationStreamClosed(e)) {
+                    throw e;
+                }
+                if (attempt == OFFSET_COMMIT_MAX_ATTEMPTS) {
+                    LOGGER.error("Offset commit failed on all {} attempts; the replication stream "
+                            + "is still closed. The batch is durably written to the export queue, "
+                            + "but the offsets could not be committed.", OFFSET_COMMIT_MAX_ATTEMPTS, e);
```
makalaaneesh, line 435: [AI][Suggestion] **Terminal-attempt error message overstates what was lost.**

The message says "The batch is durably written to the export queue, but the offsets could not be committed." Per the javadoc a few lines above, `commitOffsets()` writes the offset file (`doFlush + Future.get`) *before* it calls `flushLsn()`, so on the failing attempt 1 the offsets are already durable on disk — what actually failed is the LSN flush to the source (the WAL-release hint). Calling out "offsets could not be committed" without that qualification makes it sound like resume state was lost, which triage-time readers may take at face value.

Suggested wording:

```java
LOGGER.error("Offset commit failed on all {} attempts; the replication stream "
        + "is still closed. The batch is durably written to the export queue and "
        + "the offset file is persisted, but the LSN flush to the source could not "
        + "complete — debezium will now surface this exception.", OFFSET_COMMIT_MAX_ATTEMPTS, e);
```

---

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -369,15 +372,106 @@ public void handleBatch(List<ChangeEvent<Object, Object>> changeEvents,
...
+        for (int attempt = 1; attempt <= OFFSET_COMMIT_MAX_ATTEMPTS; attempt++) {
+            try {
+                for (ChangeEvent<Object, Object> event : changeEvents) {
+                    committer.markProcessed(event);
+                }
+                committer.markBatchFinished();
+                LOGGER.debug("Committed batch complete with {} records", changeEvents.size());
+                return;
+            }
+            catch (RuntimeException e) {
```
makalaaneesh, line 426: [AI][Suggestion] **Catching bare `RuntimeException` is broader than the guard intends.**

The catch grabs every `RuntimeException`, then narrows via `isTargetDbExporter() && isReplicationStreamClosed(e)`. The rest of `commitBatchOffsets`' try body is a plain `for (event : changeEvents)` and two Debezium method calls whose declared checked exception is `InterruptedException` — a stray unrelated `RuntimeException` from any of these (e.g. an NPE on a malformed event) would still enter the catch, be re-evaluated by the guard, and rethrown. That's functionally correct today, but catching a narrower type (`Exception` from the committer calls, or a targeted `try` around just `markBatchFinished()` where the flush lives) would make it obvious that `markProcessed()` is not in scope of the retry and would reduce surprise on future edits to this try body.

Suggested restructure — narrow the try to just the flush:

```java
for (ChangeEvent<Object, Object> event : changeEvents) {
    committer.markProcessed(event);
}
try {
    committer.markBatchFinished();
    LOGGER.debug("Committed batch complete with {} records", changeEvents.size());
    return;
}
catch (RuntimeException e) {
    // ... existing guard + retry logic ...
}
```

Left as a suggestion rather than a warning because the current placement matches the javadoc's mental model of "the whole commit block" and the behavior difference is small.

---

makalaaneesh (commented):
[AI] Automated review — 2 findings (0 critical, 0 warning, 2 suggestion). See inline comments.

---

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
```diff
@@ -369,15 +372,106 @@ public void handleBatch(List<ChangeEvent<Object, Object>> changeEvents,
...
+                // Only target-side exporters stream from YB, so only they can see a split;
+                // anywhere else, and for any other exception, this stays fatal.
+                if (!isTargetDbExporter() || !isReplicationStreamClosed(e)) {
+                    throw e;
+                }
+                if (attempt == OFFSET_COMMIT_MAX_ATTEMPTS) {
+                    LOGGER.error("Offset commit failed on all {} attempts; the replication stream "
```
makalaaneesh: "The batch is durably written to the export queue, but the offsets could not be committed." is implementation detail of debezium; we should not mention that in our error message

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
makalaaneesh: can ignore IMO

makalaaneesh (approved):
LGTM with a minor error message comment

debezium-server-voyager/debezium-server-voyagerexporter/src/main/java/io/debezium/server/ybexporter/YbExporterConsumer.java
ShivanshGahlot: Done

PR #3760, opened 2026-08-31
errcheck 3a: panic-on-error helpers for cobra flag registration

makalaaneesh:

First of a six-PR stack that makes unchecked-error-return bugs (the PR #3714 class) unlandable by adopting errcheck via golangci-lint. This PR handles the cobra flag-registration bucket: `MarkFlagRequired` / `MarkHidden` / `MarkDeprecated` / `MarkPersistentFlagRequired` return an error only when the named flag does not exist (a typo, or a flag renamed without updating the Mark* call) — and all 47 call sites dropped it, so such a mistake silently un-requires/un-hides the flag. New `mustMarkFlag*` helpers panic at command-registration time instead, and every site is converted mechanically. All helpers take the `*cobra.Command` and name the command in the panic message.

**The helper caught a real instance on its first run**: `end migration` called `MarkFlagRequired("export-dir")`, which has *never worked* — `export-dir` is registered on `PersistentFlags` (via `registerExportDirFlag`), which the local-flag variant cannot see. Fixed to `mustMarkPersistentFlagRequired(endMigrationCmd, "export-dir")`, restoring the original intent (matching `archive changes`, which marks the same persistent flag). This is safe for config-file-driven runs: cobra v1.6.1 runs `ValidateRequiredFlags` *after* `PersistentPreRun` (command.go:890-908), where `initConfig` applies config values and marks flags Changed. `validateExportDirFlag` continues to provide post-config validation as before (it runs inside `PersistentPreRun`, i.e. still ahead of cobra's required-flag check).

**`export data from target` guard (per review — this branch initially introduced a startup panic there)**: `markFlagsRequired` runs at PreRun time, and `export data from target` borrows `exportDataCmd`'s PreRun while registering no `source-*` connection flags (its connection info comes from the MigrationStatusRecord). Pre-PR, the two unconditional Mark calls silently no-oped for that command; the must-helpers turned that into a panic on every fall-forward/fall-back export. `markFlagsRequired` now returns early for that command — matched by its full command path, as a literal because referencing `exportDataFromTargetCmd` there creates a package initialization cycle back through `exportDataCmd` — restoring the exact pre-PR behavior. (Guarding on `source.DBType == YUGABYTEDB` instead would have un-required `source-db-user` for plain `export data --source-db-type yugabytedb`, which is a supported flow.) A behavior audit of all 49 sites confirmed this was the only one where the marked flag can be absent: every other site either runs at init (a missing flag would panic on any invocation, including the unit-test run) or marks a flag registered a few lines above in the same function.

Also converted per review: the two remaining hand-rolled `MarkFlagRequired` + `ErrExit` sites in `importDataFileCommand.go` (`data-dir`, `file-table-map`) now use the must-helper (49 sites total).

None in normal operation. A future flag-name typo in registration code now panics at startup (immediately caught by any test or CI run) instead of silently no-oping.

- `go build ./...`, `go test -tags unit ./...` pass (the full cmd-package test run exercises every init-time registration — that is what caught the export-dir case).
- The `export data from target` guard was verified red/green by invoking `markFlagsRequired(exportDataFromTargetCmd)` locally: panic before the guard, clean after (the path literal was also asserted against `exportDataFromTargetCmd.CommandPath()`).
- errcheck production findings: 301 → 254.

No.

No.


yb-voyager/cmd/importDataFileCommand.go
```diff
@@ -532,7 +532,7 @@ func init() {
 		2. quote_char: 	character used to quote the values (default double quotes '"')
 		for eg: --file-opts "escape_char=\",quote_char=\"" or --file-opts 'escape_char=",quote_char="'`)
 
-	importDataFileCmd.Flags().MarkDeprecated("file-opts", "use --escape-char and --quote-char flags instead")
+	mustMarkFlagDeprecated(importDataFileCmd.Flags(), "file-opts", "use --escape-char and --quote-char flags instead")
```
makalaaneesh: [AI] **Migration to `mustMark*` helpers is incomplete in this file.**

Two `MarkFlagRequired` call sites remain unmigrated in the same `init()`:

```go
// lines 506-509
err := importDataFileCmd.MarkFlagRequired("data-dir")
if err != nil {
    utils.ErrExit("mark 'data-dir' flag required: %v", err)
}

// lines 515-518
err = importDataFileCmd.MarkFlagRequired("file-table-map")
if err != nil {
    utils.ErrExit("mark 'file-table-map' flag required: %v", err)
}
```

The PR's stated goal is to standardize cobra flag registration on the new panic-on-error helpers. Leaving these two behind means the file now mixes three patterns for the same class of programmer error (silent-ignore, `utils.ErrExit`, and `panic`), which is what the helpers are meant to eliminate. Please convert these to `mustMarkFlagRequired(importDataFileCmd, "data-dir")` and `mustMarkFlagRequired(importDataFileCmd, "file-table-map")` for consistency.

---

yb-voyager/cmd/endMigrationCommand.go
```diff
@@ -1108,11 +1108,10 @@ func init() {
...
 
-	endMigrationCmd.MarkFlagRequired("backup-schema-files")
-	endMigrationCmd.MarkFlagRequired("backup-data-files")
-	endMigrationCmd.MarkFlagRequired("save-migration-reports")
-	endMigrationCmd.MarkFlagRequired("backup-log-files")
-	endMigrationCmd.MarkFlagRequired("export-dir")
+	mustMarkFlagRequired(endMigrationCmd, "backup-schema-files")
+	mustMarkFlagRequired(endMigrationCmd, "backup-data-files")
+	mustMarkFlagRequired(endMigrationCmd, "save-migration-reports")
+	mustMarkFlagRequired(endMigrationCmd, "backup-log-files")
```
makalaaneesh, line 1116: [AI] **Silent drop of `MarkFlagRequired("export-dir")` — use the persistent variant instead.**

The pre-PR code had `endMigrationCmd.MarkFlagRequired("export-dir")` but this PR just deletes that line rather than migrating it. That deletion is not innocuous — it's covering up a pre-existing bug that the new panic-on-error helpers would have exposed.

`export-dir` is registered via `registerExportDirFlag` on `PersistentFlags()` (not `Flags()`). `cmd.MarkFlagRequired("export-dir")` looks up the name in `cmd.Flags()`, which does not contain persistent flags at registration time, so the call returns `"no such flag -export-dir"` and the required-mark is silently dropped. I verified this against `cobra v1.6.1` (the pinned version):

```
MarkFlagRequired on persistent flag: err=no such flag -export-dir
MarkPersistentFlagRequired on persistent flag: err=<nil>
```

Converting it verbatim to `mustMarkFlagRequired(endMigrationCmd, "export-dir")` would (correctly) panic at startup — which is exactly what these helpers are supposed to do. Instead of dropping the line, please use the persistent variant that this PR already introduces for `archiveChangesCommand.go`:

```go
mustMarkPersistentFlagRequired(endMigrationCmd, "export-dir")
```

That matches the `archiveChangesCmd` pattern (`mustMarkPersistentFlagRequired(archiveChangesCmd, "export-dir")`) and preserves the original author's intent. Otherwise `end migration` and `archive changes` now handle the same persistent flag inconsistently — one is cobra-required, the other only relies on `validateExportDirFlag()`. If the intent really is to rely solely on `validateExportDirFlag()`, please add a comment saying so and drop the mark from `archiveChangesCommand.go` too for consistency.

---

makalaaneesh (commented):

---

yb-voyager/cmd/importDataFileCommand.go
makalaaneesh: Done — both sites converted to `mustMarkFlagRequired`.

yb-voyager/cmd/endMigrationCommand.go
makalaaneesh: Adopted — now `mustMarkPersistentFlagRequired(endMigrationCmd, "export-dir")`. Verified safe for config-file-driven runs: cobra v1.6.1 runs ValidateRequiredFlags (command.go:908) *after* PersistentPreRun (command.go:890-896), where initConfig applies config values and marks flags Changed — matching how archive changes has always handled the same persistent flag.

ShivanshGahlot:
```
voyager_cmd_runner.go:43: [import data] Starting export data from target with command:
voyager_cmd_runner.go:43: [import data]  TARGET_DB_PASSWORD=*** yb-voyager export data from target --log-level info --export-dir /tmp/yb-voyager-export1330844664 --disable-pb=true --send-diagnostics=false
voyager_cmd_runner.go:43: [import data] Using export-dir: /tmp/yb-voyager-export1330844664
voyager_cmd_runner.go:43: [import data] panic: marking flag "source-db-type" required on command "target": no such flag -source-db-type
```
There seems to be a bug in export data from target.

yb-voyager/cmd/endMigrationCommand.go
```diff
@@ -1108,11 +1108,13 @@ func init() {
...
-	endMigrationCmd.MarkFlagRequired("save-migration-reports")
-	endMigrationCmd.MarkFlagRequired("backup-log-files")
-	endMigrationCmd.MarkFlagRequired("export-dir")
+	mustMarkFlagHidden(endMigrationCmd.Flags(), "send-diagnostics")
+
+	mustMarkFlagRequired(endMigrationCmd, "backup-schema-files")
+	mustMarkFlagRequired(endMigrationCmd, "backup-data-files")
+	mustMarkFlagRequired(endMigrationCmd, "save-migration-reports")
+	mustMarkFlagRequired(endMigrationCmd, "backup-log-files")
+	// export-dir lives on PersistentFlags, so the persistent variant is needed;
```
ShivanshGahlot: Nit: We can remove this comment.

yb-voyager/cmd/root.go
```diff
@@ -524,6 +524,37 @@ func BoolVar(flagSet *pflag.FlagSet, p *utils.BoolStr, name string, value bool,
...
+	}
+}
+
+func mustMarkPersistentFlagRequired(cmd *cobra.Command, name string) {
+	if err := cmd.MarkPersistentFlagRequired(name); err != nil {
+		panic(fmt.Sprintf("marking persistent flag %q required on command %q: %v", name, cmd.Name(), err))
+	}
+}
+
+func mustMarkFlagHidden(flagSet *pflag.FlagSet, name string) {
```
ShivanshGahlot: Nit: We won't be able to see the command name in `mustMarkFlagHidden` and `mustMarkFlagDeprecated` unlike the previous two functions.

yb-voyager/cmd/export.go
```diff
@@ -366,15 +366,15 @@ func getAndStoreTargetDBPasswordInSourceConf(cmd *cobra.Command) {
 
 func markFlagsRequired(cmd *cobra.Command) {
```
ShivanshGahlot, line 367: We can probably add a skip here if the db is YB?

ShivanshGahlot (changes requested):

yb-voyager/cmd/endMigrationCommand.go
makalaaneesh: Done — removed the comment; the rationale lives in the PR description.

yb-voyager/cmd/root.go
makalaaneesh: Done — `mustMarkFlagHidden` and `mustMarkFlagDeprecated` now take the `*cobra.Command` and include `cmd.Name()` in the panic message, plus a `mustMarkPersistentFlagHidden` variant for the one `PersistentFlags()` site (`cutoverToCmd` `--yes`).

yb-voyager/cmd/export.go
makalaaneesh: Good catch — this was a real regression: `export data from target` borrows `exportDataCmd`'s PreRun but registers no `source-*` flags, so the previously-silent no-op Mark calls became a startup panic on every fall-forward/fall-back export.

Fixed with a skip keyed on the command rather than on the DB type: `yugabytedb` is a supported `--source-db-type` for plain `export data`, so `source.DBType == YUGABYTEDB` would also have un-required `source-db-user` for that flow. `markFlagsRequired` now returns early for the `export data from target` command (matched by name — referencing `exportDataFromTargetCmd` directly creates a package initialization cycle back through `exportDataCmd`). Verified red/green with a throwaway unit test calling `markFlagsRequired(exportDataFromTargetCmd)`.

Also audited all 49 converted sites for this same silently-failing-Mark class: every other site either runs at init (a missing flag would panic on any invocation, including the unit-test run) or marks a flag registered in the same function, so this was the only reachable instance.

ShivanshGahlot (approved):
LGTM

PR #3761, opened 2026-08-31
errcheck 3b: handle file close and console flush errors

makalaaneesh:

Second PR of the errcheck stack: resolves every unchecked file `Close()`/`Flush()` in production code (~50 sites). Each site was individually classified **read vs write**, because the correct handling is opposite:

- **Write paths now CHECK `Close()`** — after writing, close is the last chance for the OS to report that data never reached disk; ignoring it is a silent data-loss bug in a migration tool. Pattern: a silent backstop `defer func() { _ = f.Close() }()` plus an explicit checked `Close` on the success path. Sites: bulk-assessment and compare-perf HTML reports, batch error-append files, sqlldr control file, `CopyFile`'s destination, `toc.txt`, error-stash files, `CreateTempFile`.
- **Notable find**: `processImportDirectives` (srcdb) built a tmp file and `os.Rename`d it **over the original schema file** while the tmp file's close was only a bare defer — an incomplete write could silently replace the source file. It now closes-and-checks *before* the rename.
- **Read paths and teardown** use a new `utils.CloseAndLogOnError(what, closer)` (close + warn-level log), so problems stay observable without inventing error paths for non-events: batch files opened for COPY/sqlldr, queue segments, status/config/report/schema reads, container temp schema files.
- Error-path cleanups and bench/mock teardown become explicit `_ = c.Close()` with a why-comment; console table redraws (uilive/tabwriter `Flush` in export/import stats reporters) become `_ = Flush()` — nothing actionable on a console redraw error.

The reviewing question for every hunk is simply: is the read/write classification right?

Failure-behavior change on write paths only: a failed close of a written artifact (report, control file, batch file) now surfaces as an error instead of being swallowed.

- `go build ./...`, `go test -tags unit ./...` pass (zero failures).
- errcheck production findings: 254 → 205.

No.

No.

### DIFF
--- a/yb-voyager/cmd/assessMigrationBulkCommand.go
+++ b/yb-voyager/cmd/assessMigrationBulkCommand.go
@@ -272,7 +272,7 @@ func parseFleetConfigFile(filePath string) ([]AssessMigrationDBConfig, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer file.Close()
+	defer utils.CloseAndLogOnError(filePath, file)
 
 	reader := csv.NewReader(file)
 	header, err := reader.Read()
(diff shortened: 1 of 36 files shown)


makalaaneesh:

---

priyanshi-yb (approved):
LGTM

PR #3763, opened 2026-08-31
errcheck 3c: propagate dropped errors in state, validation, and iteration paths

makalaaneesh:

Third PR of the errcheck stack: check-and-propagate (or `utils.ErrExit`, matching each call path's idiom) for dropped errors whose loss hides real failures.

- `metaDB.UpdateMigrationStatusRecord` ×7: rename-tables map, export table-list flag, source DB conf, on-PK-conflict action, archiving flags, and `EndMigrationRequested` — whose silent loss means data export never learns it should stop.
- `s3/gcs/az.ValidateObjectURL` ×3: **behavior change** — an invalid `--data-dir` bucket URL now fails fast at validation instead of silently passing and failing confusingly later.
- namereg persistence: `setDefaultSourceReplicaDBSchemaName` and `UnRegisterYBNames`' `reg.save()` now propagate.
- A true dropped-error bug: `sqlfiletransformer`'s IterKV callback can genuinely fail (index deparse error) and the outer error was discarded — a failure silently truncated the removed-redundant-indexes file.
- Also: end-migration report backup, per-table exported-snapshot rows in the data-migration report, `waitForDebeziumStartIfRequired`, csv row processor `Write`+`Error()`, `os.WriteFile` of `postdata.sql` (PG and ora2pg) and `uncategorized.sql`, task-picker chooser init, `closeBatch()` on CDC error paths (explicit best-effort; the original batch error takes precedence).
- `StructMap.IterKV` sites whose callbacks provably never return an error (display/message/callhome-payload assembly) are dismissed as `_ =` with a why-comment; they interleave with the checked sites in the same files, so they ride here rather than in the ignores PR.
- The deferred session-variable-reset IIFE in `importSchemaYugabyteDB.go` keeps its (inherently discarded) error, made explicit with `_ =`. Per review, `(*tgtConn)` remains an eagerly-evaluated defer-time argument — wrapping it in an outer closure would move the read to function-exit time, where the error paths set `(*tgtConn) = nil` and the closure's inverted guard would then nil-deref in `ResetSessionVariables`. The pre-existing inverted nil-guard (the reset body only ever runs when conn is nil, i.e. the closure has always been a no-op) is documented for follow-up rather than changed.

Failure behavior only: migration-state persistence failures and invalid cloud-storage `--data-dir` URLs now error out loudly instead of being swallowed.

- `go build ./...`, `go test -tags unit ./...` pass (zero failures).
- errcheck production findings: 205 → 164.

No (payload assembly sites only gained explicit `_ =` on never-erroring iterations; no shape or content change).

No.

### DIFF
--- a/yb-voyager/cmd/importSchemaYugabyteDB.go
+++ b/yb-voyager/cmd/importSchemaYugabyteDB.go
@@ -291,17 +291,27 @@ func executeSqlStmtWithRetries(tgtConn **ImportSchemaTargetConn, sqlInfo sqlInfo
 		return goerrors.Errorf("error applying session variable: %v", err)
 	}
 
-	defer func(conn *ImportSchemaTargetConn) error {
-		if conn != nil {
+	// NOTE (errcheck): the closure's error return has always been discarded — a
+	// deferred call's return value cannot be observed. Made explicit with `_ =`
+	// below. `(*tgtConn)` MUST stay a defer-time argument (evaluated eagerly,
+	// always non-nil here): reading it at exit time would arm the closure body
+	// on the error paths that set `(*tgtConn) = nil`, nil-dereferencing in
+	// ResetSessionVariables. Pre-existing oddity flagged for a follow-up: the
+	// `conn != nil` early-return means the reset body only ever runs when conn
+	// is nil, so the closure has always been a no-op.
+	defer func(conn *ImportSchemaTargetConn) {
+		_ = func() error {
+			if conn != nil {
+				return nil
+			}
+			// Reset all session variables on the connection for next stmt
+			log.Infof("Resetting all session variables on the connection for next stmt")
+			err = conn.ResetSessionVariables(sessionVariables)
+			if err != nil {
+				return goerrors.Errorf("error resetting all session variables on connection: %v", err)
+			}
 			return nil
-		}
-		// Reset all session variables on the connection for next stmt
-		log.Infof("Resetting all session variables on the connection for next stmt")
-		err = conn.ResetSessionVariables(sessionVariables)
-		if err != nil {
-			return goerrors.Errorf("error resetting all session variables on connection: %v", err)
-		}
-		return nil
+		}()
 	}((*tgtConn))
 
 	for retryCount := 0; retryCount <= DDL_MAX_RETRY_COUNT; retryCount++ {
--- a/yb-voyager/cmd/archiveChangesCommand.go
+++ b/yb-voyager/cmd/archiveChangesCommand.go
@@ -83,14 +83,21 @@ func archiveChangesCommandFn(cmd *cobra.Command, args []string) {
 		utils.ErrExit("the streaming phase of export data has not started yet — archive changes can only be run after streaming begins")
 	}
 
-	metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
+	err = metaDB.UpdateMigrationStatusRecord(func(record *metadb.MigrationStatusRecord) {
 		record.ArchivingEnabled = true
 		record.SegmentCleanupRunning = true
 	})
+	if err != nil {
+		utils.ErrExit("failed to mark archiving enabled in migration status record: %w", err)
+	}
 	resetArchiveChangesRunning := func() {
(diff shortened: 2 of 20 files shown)


yb-voyager/cmd/importSchemaYugabyteDB.go
```diff
@@ -291,18 +291,25 @@ func executeSqlStmtWithRetries(tgtConn **ImportSchemaTargetConn, sqlInfo sqlInfo
...
-		}
-		// Reset all session variables on the connection for next stmt
-		log.Infof("Resetting all session variables on the connection for next stmt")
-		err = conn.ResetSessionVariables(sessionVariables)
-		if err != nil {
-			return goerrors.Errorf("error resetting all session variables on connection: %v", err)
-		}
-		return nil
-	}((*tgtConn))
+		}((*tgtConn))
```
makalaaneesh: [AI] **Defer refactor changes when `(*tgtConn)` is captured — introduces a nil-deref panic on error paths.**

The comment on this defer claims the change "preserves behavior exactly," but wrapping the original single-argument-deferred call in an outer `defer func() { ... }()` shifts when `(*tgtConn)` is evaluated from **defer-registration time** to **defer-execution time**.

Before:
```go
defer func(conn *ImportSchemaTargetConn) error {
    if conn != nil { return nil }
    ...
    err = conn.ResetSessionVariables(sessionVariables) // conn captured at defer-time (non-nil)
    ...
}((*tgtConn))
```
Per the Go spec, deferred function arguments are evaluated when the `defer` statement executes. At that point `*tgtConn` is guaranteed non-nil (we just called `ApplySessionVariables` on it), so the inner closure always hits the `if conn != nil { return nil }` early-return — the reset body never runs. This matches the author's own comment ("the reset body only ever runs when conn is nil").

After:
```go
defer func() {
    _ = func(conn *ImportSchemaTargetConn) error {
        if conn != nil { return nil }
        ...
        err = conn.ResetSessionVariables(sessionVariables)
        ...
    }((*tgtConn)) // evaluated when the deferred outer func RUNS, not at defer-time
}()
```
The outer `defer func(){...}()` captures no args, so `(*tgtConn)` is now dereferenced at function-exit time. Several error paths in the retry loop below nil out `*tgtConn` before returning:

- L324-326: `(*tgtConn).Close(...); (*tgtConn) = nil; return fmt.Errorf("before index creation: %w", err)`
- L351-353: `(*tgtConn).Close(...); (*tgtConn) = nil; return goerrors.Errorf("extract qualified index name from DDL ...")`
- L360-362: `(*tgtConn).Close(...); (*tgtConn) = nil; return fmt.Errorf("drop invalid index ...")`

On any of these returns, the deferred code now reads `conn = nil`, skips the guard, and calls `conn.ResetSessionVariables(sessionVariables)` — a nil pointer dereference that panics the process. In the pre-PR code, `conn` had been captured at defer-time (non-nil), so the guard always fired and no panic occurred.

Suggested fix: preserve the original arg-captured semantics, e.g.

```go
defer func(conn *ImportSchemaTargetConn) {
    if conn == nil {
        return
    }
    log.Infof("Resetting all session variables on the connection for next stmt")
    if rerr := conn.ResetSessionVariables(sessionVariables); rerr != nil {
        log.Warnf("error resetting session variables on connection: %v", rerr)
    }
}((*tgtConn))
```

This keeps `conn` bound at defer-time (matching prior behavior), makes the guard direction consistent with its intent, and logs (instead of discards) the reset error. If the intent really is to run reset only when `conn` was nil at defer-time, keep the condition — but do not evaluate `(*tgtConn)` inside the deferred body.

---

makalaaneesh (commented):
[AI] Automated review — 1 findings (1 critical, 0 warning). See inline comments.

---

yb-voyager/cmd/importSchemaYugabyteDB.go
makalaaneesh: Fixed — the conn is back to an eagerly-evaluated defer-time argument, so the closure keeps its historical no-op behavior on the error paths that nil it out. The code comment now documents why the eager capture is load-bearing.

priyanshi-yb (approved):
LGTM

PR #3764, opened 2026-08-31
errcheck 3d: explicitly dismiss best-effort and teardown errors

makalaaneesh:

Fourth PR of the errcheck stack: the remaining production sites where ignoring the error IS the correct behavior, converted from bare calls to explicit `_ = call()` with a why-comment each. The point: errcheck distinguishes "author consciously discarded" (`_ =`) from "author possibly forgot" (bare call) — the #3714 bug was indistinguishable from intentional ignoring precisely because the codebase used bare calls for both. After this PR, a bare error-dropping call can only mean a mistake, which is what lets the gate (next PRs) flag it.

Sites: best-effort `SIGKILL` on a possibly-already-gone process, console progress stages in assess-migration, `cmd.Help()`, reaping Debezium after a deliberate stop (non-zero exit expected), `ImportSchemaTargetConn.Close` ×8 in import-schema retry/teardown paths (voyager-owned type, so the config-level pgx/sql exclusions do not cover it; the connection is being discarded in every case), and cdcbench best-effort directory cleanup.

After this PR every remaining production errcheck finding (150) is covered by the gate's documented `exclude-functions` policy list.

No user-facing changes; no behavior changes at all (`_ =` is semantically identical to the bare call).

- `go build ./...`, targeted unit tests pass.
- errcheck production findings: 164 → 150, all 150 exclusion-covered.

No.

No.

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -313,7 +313,7 @@ func assessMigration() (err error) {
 	parserIssueDetector.PopulateObjectUsages(objectUsagesStats)
 
 	// Stage 2: Assessing migration
-	tracker.StartStage("Assessing migration", 0, nil)
+	_ = tracker.StartStage("Assessing migration", 0, nil) // console progress UX; nothing actionable on error
 	err = runAssessment()
 	if err != nil {
 		tracker.FailStage()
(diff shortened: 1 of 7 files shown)


makalaaneesh:

---

makalaaneesh:
looks good from my side; no functional changes; just `_ =`

priyanshi-yb (approved):

PR #3765, opened 2026-08-31
errcheck 3e: fixes surfaced while validating the wider linter set

makalaaneesh:

Fifth PR of the errcheck stack. While validating the gate config, the candidate linters (govet's printf analyzer, per-build-tag staticcheck/ineffassign runs, and the next-wave errorlint/rowserrcheck/sqlclosecheck) surfaced findings beyond errcheck; the real ones are fixed here even though only part of the candidate set gets enabled by the gate PR.

- **govet printf ×5**: `utils.ErrExit(msg)` with a non-constant format string — a message containing `%` verbs would be misrendered; now `ErrExit("%s", msg)`.
- **Missed error checks in tag-gated test code** (same disease as the rest of the stack, but in files no tool had ever linted — they only compile under `integration_live_migration`/`failpoint`/`integration` tags): an SA5001 (`defer rows.Close()` *before* checking the Query error), four genuinely unchecked `err = lm.X(...)` calls in live-migration tests (plus two expected-abort calls made explicitly `_ =` with comments), an unchecked `os.Setenv` in a yugabyted test.
- **errorlint (partial, ahead of enabling it)**: sentinel comparisons → `errors.Is` (`io.EOF`, `sql.ErrTxDone`, `pgx.ErrTxClosed`), error type assertions → `errors.As` (`exec.ExitError`), `%v` → `%w` in test helpers; callhome's manual unwrap-tree walker keeps its deliberate per-node assertions with nolint reasons.
- **rowserrcheck/sqlclosecheck (partial)**: `rows.Err()` checks after scan loops in assess-migration and test utils; a missing `stmt.Close` in assessmentDB bulk insert; container Query helpers annotated (rows are returned to and closed by the caller).
- **ineffassign**: dead `matchedSchema` logic simplified in `sqlname.FindBestMatchingIdenitifier`; the final `idx++` in `SchemaOptimizationReport.Summary()` is kept for uniformity with its sibling sections and suppressed with `//nolint:ineffassign,staticcheck` (golangci flags the never-read increment via both linters; standalone staticcheck does not — same fact-divergence family as the `helpers_protomsg.go` exclusion in the gate config).
- **`-tags cdc_benchmark` does not compile on main**: the bench shim's `streamChanges` call was never updated for the `tableToPKColumns` parameter, and no CI workflow builds that tag. Fixed (caught by linting every tag group).

Adopted from review (round 1):
- `tgtdb/yugabytedb.go`: the twin `== io.EOF` comparison 12 lines below the converted one is now also `errors.Is` (a wrapped EOF would previously have busy-looped `importBatchFastRecover`), and the YB `ExecuteBatch` rollback defer's `pgx.ErrTxClosed` sentinel is converted to match its postgres/oracle siblings.
- `importDataFileTaskImporterErrorPolicy_test.go`: `assert.NoError` → `require.NoError` before `defer rows.Close()` (assert continues on failure; a nil `rows` would panic in the defer).
- `test/cdcbench/artifact.go`: the `%w` on the early-exit error is nil-guarded — `exec.Wait()` returns nil on a clean exit-0, which previously rendered as `%!w(<nil>)`.

Adopted from review (round 2) — pre-existing wrong-variable/inverted-condition bugs on lines adjacent to this PR's conversions, surfaced by review:
- `importBatchFastRecover` now wraps `readLinErr` in the read-failure batch error — the old code wrapped `err`, which is provably nil at that point, so genuine read failures produced an error with no cause. The third `== io.EOF` comparison (case 2, last line + EOF) is also converted to `errors.Is`, completing the loop.
- The `ExecuteBatch` rollback defers in all three target drivers (`yugabytedb.go`, `postgres.go`, `oracle.go`) logged the closure's `err` — the batch error or nil — instead of `errRollBack`; a real rollback failure was logged with the wrong (or no) cause.
- `assessmentDB.BulkInsert`'s rollback warning was fully inverted: it fired only on `sql.ErrTxDone` (the benign, by-design outcome after a successful commit) and never on real rollback failures. Guard flipped to `!errors.Is`, with its own `errRollBack` variable, logging that variable.
- Six `rows.Close()` defers in `assessMigrationCommand.go` (dating to #2549) logged the outer `err` instead of `closeErr`.

None. The `errors.Is`/`errors.As` conversions match wrapped errors where `==`/type-assertions matched only exact values — strictly more correct on the same sentinel/type. Rollback/close failure log lines now carry the actual rollback/close error instead of an unrelated (usually nil) one.

- `go build ./...`, full `go test -tags unit ./...` pass; `go vet` clean under every build-tag combination.
- golangci-lint (gate config) verified clean across all 13 tag groups on this branch merged with the gate PR (the two remaining SA4023 findings on this branch alone are at the `StartMonitoring` site whose suppression conversion lands in the gate PR).

No.

No.

### DIFF
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -392,6 +392,9 @@ func fetchObjectUsageStats() ([]*types.ObjectUsageStats, error) {
 		}
 		objectUsagesStats = append(objectUsagesStats, &objectUsage)
 	}
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("iterating object usage stats: %w", err)
+	}
 	return objectUsagesStats, nil
 }
 
@@ -652,6 +655,9 @@ func fetchRedundantIndexInfoFromAssessmentDB() ([]utils.RedundantIndexesInfo, er
 		redundantIndex.DBType = source.DBType
 		redundantIndexesInfo = append(redundantIndexesInfo, redundantIndex)
 	}
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("iterating redundant indexes: %w", err)
+	}
 
 	resolvedRedundantIndexes := getResolvedRedundantIndexes(redundantIndexesInfo)
 
@@ -729,6 +735,9 @@ func fetchColumnStatisticsInfo() ([]utils.ColumnStatistics, error) {
 		}
 		stat.DBType = source.DBType
 		columnStats = append(columnStats, stat)
+	}
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("iterating column statistics: %w", err)
 	}
 	return columnStats, nil
 }
--- a/yb-voyager/cmd/assessMigrationCommand.go
+++ b/yb-voyager/cmd/assessMigrationCommand.go
@@ -379,7 +379,7 @@ func fetchObjectUsageStats() ([]*types.ObjectUsageStats, error) {
 	defer func() {
 		closeErr := rows.Close()
 		if closeErr != nil {
-			log.Warnf("error closing rows while fetching object usage stats %v", err)
+			log.Warnf("error closing rows while fetching object usage stats %v", closeErr)
 		}
 	}()
 
@@ -639,7 +639,7 @@ func fetchRedundantIndexInfoFromAssessmentDB() ([]utils.RedundantIndexesInfo, er
 	defer func() {
 		closeErr := rows.Close()
 		if closeErr != nil {
-			log.Warnf("error closing rows while fetching redundant indexes %v", err)
+			log.Warnf("error closing rows while fetching redundant indexes %v", closeErr)
 		}
 	}()
 
@@ -722,7 +722,7 @@ func fetchColumnStatisticsInfo() ([]utils.ColumnStatistics, error) {
 	defer func() {
 		closeErr := rows.Close()
 		if closeErr != nil {
-			log.Warnf("error closing rows while fetching column statistics %v", err)
+			log.Warnf("error closing rows while fetching column statistics %v", closeErr)
 		}
 	}()
 
@@ -1005,7 +1005,7 @@ func fetchUnsupportedObjectTypes() ([]UnsupportedFeature, error) {
 	defer func() {
 		closeErr := rows.Close()
 		if closeErr != nil {
-			log.Warnf("error closing rows while fetching object type mapping metadata: %v", err)
+			log.Warnf("error closing rows while fetching object type mapping metadata: %v", closeErr)
 		}
 	}()
 
@@ -1141,7 +1141,7 @@ func fetchUnsupportedQueryConstructs() ([]utils.UnsupportedQueryConstruct, error
 	defer func() {
 		closeErr := rows.Close()
 		if closeErr != nil {
-			log.Warnf("error closing rows while fetching database queries summary metadata: %v", err)
+			log.Warnf("error closing rows while fetching database queries summary metadata: %v", closeErr)
 		}
 	}()
 
@@ -1228,7 +1228,7 @@ func fetchColumnsWithUnsupportedDataTypes() ([]utils.TableColumnsDataTypes, []ut
 	defer func() {
 		closeErr := rows.Close()
 		if closeErr != nil {
-			log.Warnf("error closing rows while fetching unsupported datatypes metadata: %v", err)
+			log.Warnf("error closing rows while fetching unsupported datatypes metadata: %v", closeErr)
 		}
 	}()
 
--- a/yb-voyager/cmd/exportSchemaOptimizationReport.go
+++ b/yb-voyager/cmd/exportSchemaOptimizationReport.go
@@ -116,7 +116,7 @@ func (s *SchemaOptimizationReport) Summary() string {
 	}
 	if s.UKRangeShardingChange != nil {
 		summary += fmt.Sprintf("%d. %s\n", idx, s.UKRangeShardingChange.Summary())
-		// idx deliberately not incremented after the last section; re-add when appending sections
+		idx++ //nolint:ineffassign,staticcheck // kept for uniformity with the sections above; a section appended below will read it
 	}
 	return summary
 }
--- a/yb-voyager/cmd/exportSchemaOptimizationReport.go
+++ b/yb-voyager/cmd/exportSchemaOptimizationReport.go
@@ -116,7 +116,7 @@ func (s *SchemaOptimizationReport) Summary() string {
 	}
 	if s.UKRangeShardingChange != nil {
 		summary += fmt.Sprintf("%d. %s\n", idx, s.UKRangeShardingChange.Summary())
-		idx++
+		// idx deliberately not incremented after the last section; re-add when appending sections
 	}
 	return summary
 }
--- a/yb-voyager/cmd/importDataFileTaskImporterErrorPolicy_test.go
+++ b/yb-voyager/cmd/importDataFileTaskImporterErrorPolicy_test.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/sourcegraph/conc/pool"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/constants"
 	"github.com/yugabyte/yb-voyager/yb-voyager/src/importdata"
@@ -59,8 +60,8 @@ func assertTableRowCount(t *testing.T, tableName string, expectedCount int64) {
 func assertTableIds(t *testing.T, tableName string, expectedIds []int64) {
 	var ids []int64
 	rows, err := tdb.Query(fmt.Sprintf("SELECT id FROM %s", tableName))
+	require.NoError(t, err) // must halt here: a nil rows would panic in the defer below
 	defer rows.Close()
-	assert.NoError(t, err)
 
 	for rows.Next() {
 		var id int64
--- a/yb-voyager/src/migassessment/assessmentDB.go
+++ b/yb-voyager/src/migassessment/assessmentDB.go
@@ -266,9 +266,9 @@ func (adb *AssessmentDB) BulkInsert(table string, records [][]string) error {
 	}
 
 	defer func() {
-		err = tx.Rollback()
-		if err != nil && errors.Is(err, sql.ErrTxDone) {
-			log.Warnf("error while rollback the BulkInsert txn: %v", err)
+		errRollBack := tx.Rollback()
+		if errRollBack != nil && !errors.Is(errRollBack, sql.ErrTxDone) {
+			log.Warnf("error while rollback the BulkInsert txn: %v", errRollBack)
 		}
 	}()
 
--- a/yb-voyager/src/migassessment/assessmentDB.go
+++ b/yb-voyager/src/migassessment/assessmentDB.go
@@ -280,6 +280,7 @@ func (adb *AssessmentDB) BulkInsert(table string, records [][]string) error {
 	if err != nil {
 		return fmt.Errorf("error preparing statement for bulk insert into %s: %w", table, err)
 	}
+	defer stmt.Close()
 
 	for rowNum := 1; rowNum < len(records); rowNum++ {
 		row := utils.ConvertStringSliceToInterface(records[rowNum])
--- a/yb-voyager/src/tgtdb/oracle.go
+++ b/yb-voyager/src/tgtdb/oracle.go
@@ -19,6 +19,7 @@ import (
 	"bufio"
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -533,7 +534,7 @@ func (tdb *TargetOracleDB) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBat
 		}
 		defer func() {
 			errRollBack := tx.Rollback()
-			if errRollBack != nil && errRollBack != sql.ErrTxDone {
+			if errRollBack != nil && !errors.Is(errRollBack, sql.ErrTxDone) {
 				log.Errorf("error rolling back tx for batch id (%s): %v", batch.ID(), err)
 			}
 		}()
--- a/yb-voyager/src/tgtdb/oracle.go
+++ b/yb-voyager/src/tgtdb/oracle.go
@@ -535,7 +535,7 @@ func (tdb *TargetOracleDB) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBat
 		defer func() {
 			errRollBack := tx.Rollback()
 			if errRollBack != nil && !errors.Is(errRollBack, sql.ErrTxDone) {
-				log.Errorf("error rolling back tx for batch id (%s): %v", batch.ID(), err)
+				log.Errorf("error rolling back tx for batch id (%s): %v", batch.ID(), errRollBack)
 			}
 		}()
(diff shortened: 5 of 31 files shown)


yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -960,7 +960,7 @@ func (yb *TargetYugabyteDB) importBatchFastRecover(conn *pgx.Conn, batch Batch,
 	}
 	for {
 		line, _, readLinErr := df.NextLine()
-		if readLinErr != nil && readLinErr != io.EOF {
+		if readLinErr != nil && !errors.Is(readLinErr, io.EOF) {
```
makalaaneesh, line 963: [AI] **Half-converted `io.EOF` check in `importBatchFastRecover` — the two sibling comparisons are still `==`, so a wrapped EOF now spins instead of erroring.**

This line was converted to `errors.Is`, but the other two EOF checks in the same loop were left as exact comparisons:

- line 975: `if readLinErr == io.EOF { break }` (handles case 1, empty line)
- line 1016: `if readLinErr == io.EOF { // handles case 2` (break after the last line)

That makes the loop's control flow inconsistent, and not in a safe direction. Before this change, a `readLinErr` that *wraps* `io.EOF` (rather than being it) hit the guard and returned an import-batch error immediately. Now the guard lets it through, and neither of the two `==` checks matches it.

Failure case: `df.NextLine()` returns `("", n, fmt.Errorf("read: %w", io.EOF))`.
- Old: guard is true → return `newImportBatchErrorPgYb(...)`.
- New: guard is false → `line == ""` → `readLinErr == io.EOF` is false → `continue` → `NextLine()` returns the same wrapped EOF again → **infinite loop** in the fast-recover path of `import data`.

Today `CsvDataFile`/`TextDataFile`/`SqlDataFile` all return bare `io.EOF` (`src/utils/csv/reader.go:101`, `bufio.ReadString`), so this is latent rather than live. But the PR description states the conversions are "strictly more correct on the same sentinel/type" — that only holds if all comparisons on a given sentinel are converted together. A partial conversion in one loop is worse than either consistent state.

Suggested fix — convert all three in the loop:

```go
if readLinErr == io.EOF {           // line 975
```
```go
if errors.Is(readLinErr, io.EOF) {
```
and likewise at line 1016. (Same applies to line 1195 in this file — see the comment on `postgres.go`.)

---

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -884,7 +884,7 @@ func (pg *TargetPostgreSQL) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
 		}
 		defer func() {
 			errRollBack := tx.Rollback(ctx)
-			if errRollBack != nil && errRollBack != pgx.ErrTxClosed {
+			if errRollBack != nil && !errors.Is(errRollBack, pgx.ErrTxClosed) {
```
makalaaneesh, line 911: [AI] **The identical rollback sentinel check in `TargetYugabyteDB.ExecuteBatch` was not converted — YB is the primary import target, so the one that matters most is still `==`.**

This PR converts the `ExecuteBatch` rollback guard in two of the three target drivers:

- `src/tgtdb/postgres.go:887` — converted (this line)
- `src/tgtdb/oracle.go:533` — converted (`sql.ErrTxDone`)
- `src/tgtdb/yugabytedb.go:1195` — **not converted**, still `if errRollBack != nil && errRollBack != pgx.ErrTxClosed {`

`yugabytedb.go` is already touched by this PR (lines 963 and 1429), so it isn't a "different file, different PR" boundary. Per the root `.cursor/BUGBOT.md` rule on source/target heterogeneity — "when changing one source implementation, check if the same change is needed in others" — leaving the YB driver on the old comparison means the CDC/live-migration rollback path most users actually exercise keeps the exact-match semantics the PR is arguing against, and `errorlint` will flag it again the moment the gate PR turns the linter on.

Suggested fix — `src/tgtdb/yugabytedb.go:1195`:

```go
if errRollBack != nil && !errors.Is(errRollBack, pgx.ErrTxClosed) {
```

(`errors` is already imported in that file.)

---

yb-voyager/cmd/importDataFileTaskImporterErrorPolicy_test.go
```diff
@@ -59,8 +59,8 @@ func assertTableRowCount(t *testing.T, tableName string, expectedCount int64) {
 func assertTableIds(t *testing.T, tableName string, expectedIds []int64) {
 	var ids []int64
 	rows, err := tdb.Query(fmt.Sprintf("SELECT id FROM %s", tableName))
-	defer rows.Close()
 	assert.NoError(t, err)
+	defer rows.Close()
```
makalaaneesh, line 64: [AI] **Reordering the `defer` past `assert.NoError` silences SA5001 but does not fix the nil-deref it was pointing at.**

`assert.NoError` records a failure and *continues*; it does not stop the goroutine. So when `tdb.Query(...)` fails, `rows` is `nil` (`TargetYugabyteDB.Query` at `src/tgtdb/yugabytedb.go:87` returns `yb.db.Query(query)` verbatim, and `database/sql` returns a nil `*sql.Rows` on error), and execution falls straight into `rows.Next()` / the deferred `rows.Close()` — both panic on a nil `*sql.Rows`. That is the same outcome as before the reorder, so the linter finding is suppressed without the underlying defect changing.

Failure case: the target table has been dropped by a prior sub-test, `SELECT id FROM <table>` errors, `assert.NoError` marks the test failed, then `rows.Next()` panics — the test aborts with a nil-pointer stack trace instead of the readable "relation does not exist" failure.

`yb-voyager/.cursor/BUGBOT.md` covers this directly: *"Use `testify/require` for setup steps that must succeed for the test to be meaningful."*

Suggested fix:

```go
rows, err := tdb.Query(fmt.Sprintf("SELECT id FROM %s", tableName))
require.NoError(t, err)
defer rows.Close()

for rows.Next() {
    var id int64
    err = rows.Scan(&id)
    testutils.FatalIfError(t, err)
    ids = append(ids, id)
}
require.NoError(t, rows.Err())
```

The trailing `rows.Err()` also keeps this helper consistent with the `rows.Err()` checks this PR adds in `assessMigrationCommand.go` and `test/utils/testutils.go` — without it, a mid-iteration driver error silently turns into a short `ids` slice and a confusing `ElementsMatch` diff.

---

yb-voyager/test/cdcbench/artifact.go
```diff
@@ -409,7 +409,7 @@ func pollUntil(timeout, interval time.Duration, procDone <-chan error, cond func
 		}
 		select {
 		case err := <-procDone:
-			return fmt.Errorf("export data exited early: %v", err)
+			return fmt.Errorf("export data exited early: %w", err)
```
makalaaneesh, line 416: [AI] **`%w` here is applied to an error that is routinely `nil`, producing `%!w(<nil>)` and a non-wrapping error.**

`procDone` is fed by `go func() { exportDone <- export.Wait() }()` (line 226). `exec.Cmd.Wait()` returns `nil` when the process exits 0 — which is exactly the "export data exited early" case this branch exists to report (the export command finished cleanly before the poll condition was met). `%v` rendered that as `<nil>`; `%w` renders it as `%!w(<nil>)`, and `fmt.Errorf` does not produce a wrapping error when the `%w` operand isn't a non-nil `error`, so the `%w` buys nothing even on the non-nil path here.

Failure case: `export data` exits 0 before the first queue segment appears (e.g. an empty table list). `pollUntil` returns `export data exited early: %!w(<nil>)`, which is then wrapped again by `fail("waiting for streaming phase", err)` — the artifact-generation failure the developer sees is a printf-verb error rather than a description of what happened.

Suggested fix:

```go
case err := <-procDone:
    if err == nil {
        return fmt.Errorf("export data exited early with status 0")
    }
    return fmt.Errorf("export data exited early: %w", err)
```

---

makalaaneesh (commented):
[AI] Automated review — 4 findings (0 critical, 4 warning). See inline comments.

---

yb-voyager/src/tgtdb/yugabytedb.go
makalaaneesh: Fixed — the twin comparison 12 lines below now uses errors.Is as well, so wrapped EOFs cannot busy-loop importBatchFastRecover.

yb-voyager/src/tgtdb/postgres.go
makalaaneesh: Fixed — the YB ExecuteBatch rollback defer now uses errors.Is(errRollBack, pgx.ErrTxClosed), matching the postgres and oracle siblings.

yb-voyager/cmd/importDataFileTaskImporterErrorPolicy_test.go
makalaaneesh: Fixed — now require.NoError before the defer, with a comment on why it must halt.

yb-voyager/test/cdcbench/artifact.go
makalaaneesh: Fixed — nil is now handled separately ("export data exited early with status 0") and %w only applies to a non-nil error.

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -960,7 +960,7 @@ func (yb *TargetYugabyteDB) importBatchFastRecover(conn *pgx.Conn, batch Batch,
...
 			return 0, newImportBatchErrorPgYb(err, batch,
```
ShivanshGahlot: This should be `readLinErr` and not `err`?

yb-voyager/src/migassessment/assessmentDB.go
ShivanshGahlot: Claude pointed out that the condition should be `!errors.Is(err, sql.ErrTxDone)`. Since this is what t.RollBack returns on a successful commit.

Image

yb-voyager/src/tgtdb/postgres.go
```diff
@@ -908,7 +908,7 @@ func (pg *TargetPostgreSQL) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
...
 				log.Errorf("error rolling back tx for batch id (%s): %v", batch.ID(), err)
```
ShivanshGahlot: `errRollBack` and not `err` in the log

yb-voyager/cmd/assessMigrationCommand.go
ShivanshGahlot: Claude: We are logging the wrong error in 6 such instances in this file.

yb-voyager/src/tgtdb/yugabytedb.go
ShivanshGahlot: We missed handling this too similar to case 1

yb-voyager/cmd/exportSchemaOptimizationReport.go
```diff
@@ -116,7 +116,7 @@ func (s *SchemaOptimizationReport) Summary() string {
 	}
 	if s.UKRangeShardingChange != nil {
 		summary += fmt.Sprintf("%d. %s\n", idx, s.UKRangeShardingChange.Summary())
-		idx++
+		// idx deliberately not incremented after the last section; re-add when appending sections
```
ShivanshGahlot: Didn't understand this.

yb-voyager/src/tgtdb/oracle.go
```diff
@@ -533,7 +534,7 @@ func (tdb *TargetOracleDB) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBat
 		}
 		defer func() {
 			errRollBack := tx.Rollback()
-			if errRollBack != nil && errRollBack != sql.ErrTxDone {
+			if errRollBack != nil && !errors.Is(errRollBack, sql.ErrTxDone) {
 				log.Errorf("error rolling back tx for batch id (%s): %v", batch.ID(), err)
```
ShivanshGahlot: `errRollBack` and not `err` in the log

yb-voyager/src/tgtdb/yugabytedb.go
```diff
@@ -1197,7 +1197,7 @@ func (yb *TargetYugabyteDB) ExecuteBatch(migrationUUID uuid.UUID, batch *EventBa
 		}
 		defer func() {
 			errRollBack := tx.Rollback(ctx)
-			if errRollBack != nil && errRollBack != pgx.ErrTxClosed {
+			if errRollBack != nil && !errors.Is(errRollBack, pgx.ErrTxClosed) {
 				log.Errorf("error rolling back tx for batch id (%s): %v", batch.ID(), err)
```
ShivanshGahlot: `errRollBack` instead of `err`

yb-voyager/src/tgtdb/yugabytedb.go
makalaaneesh: Yes — fixed in 05e363a44. `err` is provably nil at that point (both earlier error paths returned), so a genuine read failure produced a batch error with no cause. Now wraps `readLinErr`.

yb-voyager/src/tgtdb/yugabytedb.go
makalaaneesh: Fixed in 05e363a44 — the case-2 comparison now uses `errors.Is` like its two siblings, completing the loop conversion.

yb-voyager/src/migassessment/assessmentDB.go
makalaaneesh: Fixed in 05e363a44 — the guard was fully inverted: it warned only on `sql.ErrTxDone` (the benign post-commit outcome) and stayed silent on real rollback failures. Now `errRollBack != nil && !errors.Is(errRollBack, sql.ErrTxDone)`, with its own `errRollBack` variable (instead of reusing the function's `err`), and the log prints that variable.

yb-voyager/src/tgtdb/postgres.go
makalaaneesh: Fixed in 05e363a44 — logs `errRollBack` now (same fix applied in oracle.go and yugabytedb.go; at defer time `err` held the batch error or nil, not the rollback failure).

yb-voyager/src/tgtdb/oracle.go
makalaaneesh: Fixed in 05e363a44 — logs `errRollBack` now.

yb-voyager/src/tgtdb/yugabytedb.go
makalaaneesh: Fixed in 05e363a44 — logs `errRollBack` now.

yb-voyager/cmd/assessMigrationCommand.go
makalaaneesh: Fixed in 05e363a44 — all six `rows.Close()` defers in this file (lines 382, 642, 725, 1008, 1144, 1231) now log `closeErr` instead of the outer `err`. Pre-existing since #2549, not introduced by this stack.

yb-voyager/cmd/exportSchemaOptimizationReport.go
makalaaneesh: Fair — the comment was confusing. Context: each section prints its number (`idx`) and then increments it for the next section. On the last section the increment computes a value nothing reads (there is no next section), which ineffassign flags — the numbering itself was always correct. This PR originally resolved that by deleting the last `idx++`, but that made the final block asymmetric and left a footgun (someone appending a section must remember to re-add the increment).

Reworked in 05e363a44: the `idx++` is restored for uniformity and suppressed with `//nolint:ineffassign,staticcheck` plus a short comment. (Both linters are needed: golangci's staticcheck also reports it as SA4006, even though standalone staticcheck was green on this exact code — same golangci-vs-standalone divergence as the helpers_protomsg exclusion in the gate config.)

ShivanshGahlot (approved):
LGTM

PR #3766, opened 2026-08-31
errcheck 3f: adopt golangci-lint as the CI lint gate

makalaaneesh:

Top of the errcheck stack: the gate itself. golangci-lint runs **errcheck, govet, staticcheck, ineffassign, and unused** in one invocation with one config. errcheck is the piece that closes the PR #3714 bug class end to end: a bare call statement dropping an error return is caught by no other tool, and signature-widening refactors (void → error) become unlandable without updating every call site.

**`yb-voyager/.golangci.yml`:**
- staticcheck check list mirrors the standalone default gate exactly: all minus staticcheck's default ST excludes (incl. `ST1023`, verified default-off in 2025.1.1 both by a scratch-module test and by a pre-existing violation on staticcheck-green main), minus the project's pre-existing `-S1008`/`-ST1005` opt-outs, minus `-QF*` (2025.1.1 ships no QF checks; excluded defensively for future golangci-bundled versions). **`SA4023` is enforced** — it is part of the standalone default set, so excluding it would have weakened the gate (fixed per review). Two accompaniments: the deliberate always-true `err != nil` after `StartMonitoring` carries two `//nolint:staticcheck` markers with one explanatory comment (golangci emits two diagnostics for one SA4023 — the finding at the `if` and the related-information position at the call — and each needs its own suppression; the old `//lint:ignore` was dropped for simplicity, so bare standalone staticcheck flags this one site while golangci, the actual gate, is clean), and `helpers_protomsg.go` gets a path+check-scoped exclusion for five defensive protoreflect nil-checks that golangci's fact computation flags but standalone staticcheck provably does not on identical code.
- The errcheck `exclude-functions` list is deliberate policy with per-entry rationale in the file: sql/pgx connection + rows teardown closes (pgx **v4, v5, and pgconn** — the codebase uses all three), deferred `Tx.Rollback` (returns `ErrTxDone` on the success path by design), `resp.Body.Close`, console `fmt.Fprint*`/color printers. Anything not listed must handle its error explicitly. Per review, every excluded staticcheck check now carries its own one-line rationale in the file, and the `hash.Write` entries were dropped as provably inert: under the pinned golangci v2.13.2, errcheck produces no finding for the fnv call site even with an empty exclude-functions list. Post-merge follow-up (agreed in review): replace the Close/Rollback exclusions with explicit `utils.CloseAndLogOnError`-style handling at the ~90 call sites, shrinking exclude-functions to the console printers only.
- Temporary rule: `_test.go` files skip **errcheck only** (~380 legacy findings, dominated by `os.Setenv`/`os.RemoveAll` → `t.Setenv`/`t.TempDir`; removed by the planned test-modernization PR). Cross-tag shared test-helper files skip **unused only** (no single-tag lint run can see all their consumers).
- TODO(lint) documents the next wave — errorlint (~15 remaining findings), rowserrcheck (~40), sqlclosecheck (~8) — each to be enabled by its own sweep-then-enable PR.

**`.github/workflows/lint.yml`:** a single job, one golangci-lint install (pinned via `versions/ci-config.json` → `golangci_lint`, replacing the staticcheck pin end to end: `versions.go`, `prepare-versions.yml`), then **one named step per build-tag group** (13 groups). The groups cannot be combined into one run — test files of different tags redeclare symbols (e.g. `TestMain`) in the same package, so a combined-tags build does not compile. `if: !cancelled()` lets every group report even when an earlier one fails. go.yml's Vet and staticcheck steps are removed (both now run inside golangci-lint). `scripts/lint.sh` mirrors the CI steps for local runs; AGENTS.md and the update-yb-latest-stable skill doc are updated.

No user-facing changes (CI/tooling only).

- `golangci-lint run` (v2.13.2, uncapped output) verified **0 issues across all 13 build-tag groups** locally, on this exact stack.
- `go build ./...` and full `go test -tags unit ./...` pass.
- The negative test of the gate: reverting the #3714 one-line fix makes errcheck fail the lint run, which no previous CI configuration could do.

No.

No.

### DIFF
--- a/yb-voyager/.golangci.yml
+++ b/yb-voyager/.golangci.yml
@@ -1,21 +1,43 @@
 # golangci-lint v2 configuration — the single lint entry point for this module.
-# CI runs it once per build-tag group (see .github/workflows/lint.yml); the tag
-# groups cannot be combined because test files of different tags redeclare
-# symbols (e.g. TestMain) in the same package.
+# The binary is pinned in versions/ci-config.json (`golangci_lint`).
+#
+# CI (.github/workflows/lint.yml) and scripts/lint.sh run it once per build-tag
+# group; the tag groups cannot be combined into one run because test files of
+# different tags redeclare symbols (e.g. TestMain) in the same package, so a
+# combined-tags build does not compile. The union of the groups covers every Go
+# file in the module.
 #
 # The errcheck exclude-functions list below is deliberate policy, not noise
 # suppression: an entry means "this method's error is safely ignorable at every
 # call site", with the rationale alongside. Anything not listed must handle its
 # error explicitly (check it, or `_ =` with a why-comment).
+
+# Config schema version for the golangci-lint v2.x binary (a v1 config is
+# rejected at startup).
 version: "2"
 
 linters:
+  # Deny-by-default: `none` plus an explicit enable list means upgrading
+  # golangci-lint can never silently add or remove linters from the gate.
+  # The gate is exactly the set below until someone edits this file.
   default: none
   enable:
+    # A bare call statement that drops an error return (the PR #3714 bug
+    # class) is caught by no other tool. Also makes void->error signature
+    # widenings unlandable without updating every call site.
     - errcheck
+    # Previously go.yml's standalone `go vet` step; runs inside the gate now.
     - govet
+    # Ineffectual assignments; overlaps staticcheck's SA4006 but catches
+    # shapes it misses, and is essentially free.
     - ineffassign
+    # The SA bug class (incl. SA4006 "err assigned, never read"). Replaces
+    # the standalone staticcheck CI step; check selection is under
+    # settings.staticcheck below.
     - staticcheck
+    # Dead-code detection (U1000). Standalone staticcheck bundles this in its
+    # default run, but golangci splits it into a separate linter — enabling
+    # staticcheck alone would silently narrow the gate vs. the standalone tool.
     - unused
     # TODO(lint): enable the next-wave linters, one PR each, after sweeping
     # that linter's remaining findings (measured Aug 2026: errorlint ~15,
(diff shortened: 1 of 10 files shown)


yb-voyager/.golangci.yml
```diff
@@ -0,0 +1,96 @@
...
+        - -ST1016
+        - -ST1020
+        - -ST1021
+        - -ST1022
+        - -ST1023
+        - -S1008
+        # QF* (quickfix) and SA4023 are not part of standalone staticcheck's
+        # default set (which main was gated on); excluded to keep parity.
+        - -QF*
+        - -SA4023
```
makalaaneesh: [AI] **Regression in staticcheck strictness: `-SA4023` and `-ST1023` were enforced by main's staticcheck 2025.1.1.**

The comment right above says QF* and SA4023 are "not part of standalone staticcheck's default set (which main was gated on); excluded to keep parity." That claim is correct for QF* (staticcheck 2025.1.1 doesn't ship any QF checks — `staticcheck -explain QF1001` returns "Couldn't find check"), but it is **not** correct for SA4023, and the silently-added `-ST1023` on line 41 has no such comment at all.

Verified empirically against `staticcheck@2025.1.1` and `golangci-lint@v2.13.2` (the pinned versions):

- With `checks: [all]`, golangci-lint's staticcheck reports both `SA4023` and `ST1023`.
- With the PR's exclusion list, both are silenced (`0 issues`).
- Standalone staticcheck 2025.1.1's default set is `["all", "-ST1000", "-ST1003", "-ST1016", "-ST1020", "-ST1021", "-ST1022"]` — SA4023 and ST1023 are **not** in the excluded list, so `staticcheck.conf = ["inherit", "-S1008", "-ST1005"]` (still on `main` and unchanged in this PR) enforces both.

Net effect: this lint gate is **less strict** than the one it replaces. SA4023 in particular is a correctness check ("Impossible comparison of interface value with untyped nil") — the sort of check we should be adding, not silencing. `staticcheck.conf` is now out of sync with the yaml, which contradicts the AGENTS.md line ("staticcheck.conf … must stay in sync with the yaml's staticcheck checks").

Suggested fix: drop `-ST1023` (line 41) and `-SA4023` (line 46), keep `-QF*` (defensive; 2025.1.1 has none, but future golangci-lint bundles will), and update the comment to say only QF* is being excluded for parity. If SA4023 or ST1023 later turns out to have a false positive in this codebase, exclude it in a follow-up with a concrete pointer to the offender.

```yaml
      checks:
        - all
        - -ST1000
        - -ST1003
        - -ST1005
        - -ST1016
        - -ST1020
        - -ST1021
        - -ST1022
        - -S1008
        # QF* (quickfix) checks are not part of standalone staticcheck's
        # default set (which main was gated on); excluded to keep parity.
        - -QF*
```

---

makalaaneesh (commented):
[AI] Automated review — 1 findings (0 critical, 1 warning). See inline comments.

---

yb-voyager/.golangci.yml
makalaaneesh: Half adopted. SA4023: you are right — it fired under the inherit config during the #3725 work, so it is default-enforced on main; -SA4023 is removed and SA4023 is now enforced (with a nolint at the deliberate StartMonitoring check, and a path+check-scoped exclusion for five helpers_protomsg.go findings that golangci flags but standalone staticcheck 2025.1.1 provably does not on identical code — a fact-computation divergence). ST1023: keeping the exclusion — it is default-OFF in standalone 2025.1.1, verified two ways: (1) a scratch module containing `var X int = 5` passes standalone staticcheck@2025.1.1 clean with no conf, and (2) a pre-existing ST1023 violation at src/tgtdb/postgres.go:2050 (not touched by this stack) exists on staticcheck-green main — if ST1023 were default-enforced, main would be red. The `checks: [all]` test in the comment proves ST1023 exists in all, not in the default set that parity targets. Config comment rewritten to state all of this precisely.

priyanshi-yb (approved):
LGTM

PR #3768, opened 2026-09-01
Fix conflict-detection log message content and log conflict resolution

makalaaneesh:

Log-only changes in the live-migration conflict-detection path (`ConflictDetectionCache` + the two target drivers' unique-index discovery), found while reviewing every log statement in that flow.

**Log-content fixes**

1. `formatUniqueIndexColumnValuesForLog` joined nothing between columns, so a two-column unique index rendered as `c1=abcc2=def` — unreadable, and ambiguous (`{"ab",""}` and `{"a","b"}` rendered identically). Values are now joined with `", "` and bracketed at the call sites.
2. The `conflict detected` log reported the index *columns* but not the index *name*, even though `Conflict.indexName` already carried it. A table can have several unique indexes over overlapping columns, so the name is what identifies the constraint that was serialized on.
3. `eventsConfict` passed the `NameTuple` itself to `%s`, which expands to the verbose `[CurrentName=(..) SourceName=(..) TargetName=(..)]` form.
4. Both target drivers (`TargetYugabyteDB` / `TargetPostgreSQL`) logged the table list as `%v` over `[]sqlname.NameTuple` and the discovered indexes as `%v` over the `*utils.StructMap`. What that actually produced:

   ```
   getting unique indexes from target for tables: [[CurrentName=(users) SourceName=(users) TargetName=(users)] [CurrentName=(orders) ...]]
   unique indexes from target for tables: &{map["public"."users":{0x1400037f2b0 0x1400037f2b0 0x1400037f2b0} "sales"."orders":{0x1400037f380 ...}] map["public"."users":[{users_email_key [email] false} ...]]}
   ```

   The table list is 3x redundant *and* drops the schema; the index map leads with a pointer-address dump of the `StructMap`'s key map (`NameTuple` has `*ObjectName` fields, and `fmt` can't call `String()` through unexported struct fields). Now:

   ```
   getting unique indexes from target for tables: public.users, sales.orders
   unique indexes from target for tables: public.users: users_email_key(email), users_tenant_login_key(tenant_id, login); sales.orders: orders_ref_key(ref) [nulls not distinct]
   ```

   The table list reuses the existing `sqlname.NameTupleListToStrings`; the index map goes through one new `formatTableToUniqueIndexesForLog` in `postgres.go`, alongside the other helpers shared by both PG/YB target drivers.
5. Table names throughout this path are now rendered with `ForOutput()` (`public.users`) rather than `ForKey()` (`"public"."users"`).

**New log: conflict resolution + wait duration**

`WaitUntilNoConflict` now reports when a conflict clears, and how long the event was blocked. Previously the log showed an event getting *blocked* but never showed whether it was released in milliseconds or is still stuck — the difference between a healthy migration and a stalled one. It is emitted only for events that actually waited, so the no-conflict path stays silent.

Before / after, for one blocked event:

```
# before
INFO conflict detected: event(vsn=2) on table "public"."users", unique index columns [tenant_id email] with matched value tenant_id=42email=a@example.com (before-after conflict) waiting for in-flight event(s) [1] to be applied
(nothing further)

# after
INFO conflict detected: event(vsn=2) on table public.users, unique index users_tenant_email_key (columns tenant_id, email) with matched value [tenant_id=42, email=a@example.com] (before-after conflict) waiting for in-flight event(s) [1] to be applied
INFO conflict cleared: event(vsn=2) on table public.users proceeding after waiting 301ms
```

No behavior change outside logging.

This PR deliberately does **not** address the other findings from the same log review (log-level demotions for the per-event DEBUG lines, per-conflict INFO rate-limiting/aggregation, the unreachable `checkUniqueIndex*Conflict` functions and their two dead INFO lines, cache-size/stall WARNs, conflict metrics) — those are follow-ups.

Only the content of the conflict-detection lines in `yb-voyager-import-data*.log`, plus one new INFO line per blocked event when its conflict clears. No CLI, config, installation, or report changes.

Existing tests were enough — this is a log-message change with no behavior change, and the existing `cmd` conflict-detection suite already covers the surrounding logic (`go test -tags unit ./cmd/... ./src/tgtdb/...` passes).

Manually verified the rendered lines through a throwaway in-package test (outputs quoted above): the detected/cleared pair for a blocked event, silence for a non-conflicting event, and the current-vs-new rendering of both target-driver logs.

`go vet ./...`, `go vet -tags unit ./...`, `staticcheck ./...`, `staticcheck -tags unit ./...` are all clean; touched files are gofmt-clean.

No.

No.

### DIFF
--- a/yb-voyager/cmd/conflictDetectionCache.go
+++ b/yb-voyager/cmd/conflictDetectionCache.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	goerrors "github.com/go-errors/errors"
 	"github.com/samber/lo"
@@ -200,7 +201,7 @@ func (c *ConflictDetectionCache) Put(event *tgtdb.Event) error {
 			return err
 		}
 	}
-	log.Debugf("adding event vsn(%d) to conflict cache for table %s", event.Vsn, event.TableNameTup.ForKey())
+	log.Debugf("adding event vsn(%d) to conflict cache for table %s", event.Vsn, event.TableNameTup.ForOutput())
 	return nil
 }
 
@@ -257,6 +258,7 @@ func (c *ConflictDetectionCache) WaitUntilNoConflict(incomingEvent *tgtdb.Event)
 	c.Lock()
 	defer c.Unlock()
 
+	waitStartTime := time.Now()
 	conflictLogged := false
 	for {
 		conflictInfo, err := c.findConflictLocked(incomingEvent)
@@ -264,7 +266,7 @@ func (c *ConflictDetectionCache) WaitUntilNoConflict(incomingEvent *tgtdb.Event)
 			return err
 		}
 		if len(conflictInfo) == 0 {
-			return nil
+			break
 		}
 
 		// flushing all the batches in channels instead of waiting for MAX_INTERVAL_BETWEEN_BATCHES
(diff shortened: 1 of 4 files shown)


yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -482,8 +492,8 @@ func (c *ConflictDetectionCache) checkBeforeBeforeConflict(incomingEvent *tgtdb.
 		return Conflict{}, nil
 	}
 	for _, cachedEvent := range cachedEvents {
-		log.Debugf("conflict detected for table %s, index columns %v, between before value of cached-event1(vsn=%d, colVal=%s) and before value of incoming-event2(vsn=%d, colVal=%s)",
-			incomingEvent.TableNameTup.ForKey(), index.Columns,
+		log.Debugf("conflict detected for table %s, unique index %s (columns %s), between before value of cached-event1(vsn=%d, colVal=[%s]) and before value of incoming-event2(vsn=%d, colVal=[%s])",
+			incomingEvent.TableNameTup.ForOutput(), index.IndexName, strings.Join(index.Columns, ", "),
```
priyanshi-yb: nit: index.String() ? and in other logs above also

priyanshi-yb (approved):
LGTM

makalaaneesh:

---

PR #3769, opened 2026-09-01
cmd: add `schema detect-drift` command

ShivanshGahlot:

Adds the CLI that reports how the PostgreSQL source schema changed while a migration was running.

- **Command** — new `schema` parent command, with `schema detect-drift` under it.
- **Read-only** — compares the snapshots captured during the migration and writes reports to `<export-dir>/reports/`. It never modifies migration state.
- **Table universe** — the union of the live catalog, the stored snapshots, and the live read. A table dropped from the source but still present in history is compared.
- **Scope** — `--exclude-*` flags become positive keep-sets here, which is what the collapsed `schemadiff.Scope` expects.
- **Config file** — the command can be driven from a config file like the others.
- **Live read label** — `LabelDetectDrift` is added for the in-memory read. It is never persisted, but `Capture` requires a label from the known vocabulary.
- **Prerequisite** — capture is off by default, so this command reports no drift unless the export commands ran with `--disable-schema-snapshot-capture=false`. The `--help` text says so.

Fourth PR in the stack; it builds on the drift report package added by the previous three.

New command:

```
yb-voyager schema detect-drift --export-dir <dir> [--output-format html,json]
```

It also accepts `--table-list` / `--exclude-table-list` and `--object-type-list` / `--exclude-object-type-list`. Reports are written to `<export-dir>/reports/drift_analysis_report.{html,json}`.

`go test -tags unit ./cmd/ -run Drift` — unit tests for flag validation, object-type parsing, the table universe, and summary wrapping. Gap: no end-to-end test of the command against a real database. The engine has one (`TestDiff_EndToEnd`), but the command itself is unit tested only. An end-to-end test is the last PR in this stack.

No. The command joins the `--send-diagnostics` allowlist, but the payload shape is unchanged.

No. The command only reads snapshots, and the live read is held in memory and never saved.


ShivanshGahlot:
Superseded by a five-PR stack, so each layer can be reviewed on its own:

- #3811 `schemadiff`: collapse Scope to one allow-list per dimension
- #3812 `schemadiff/driftreport`: drift report model, assembler, and severity
- #3813 `schemadiff/driftreport`: HTML and JSON renderers
- #3814 `cmd`: add `schema detect-drift` command (the content of this PR)
- #3815 `cmd`: point users at `schema detect-drift` on failure and at cutover

Same code — the five branches reassemble to a tree identical to this PR's. GitHub would not let this PR's base move because it is registered in a stack, so #3814 replaces it. No review comments were lost; there were none.

PR #3770, opened 2026-09-01
Add custom-cdc-partition-key-test for --cdc-partition-key-overrides live migration

Abhi-yb:
Adds `custom-cdc-partition-key-test`, a two-phase resumption migtest for `--cdc-partition-key-overrides` (no soak-level coverage existed for the feature).

- **Random pick, two levels**: each run randomly picks one of 12 candidate tables to route by a custom key, then randomly samples 1-2 of that table's columns (random order) from a per-table pool — so runs vary across table, column count, column combination and key datatype (varchar, text, int, bigint, numeric, boolean, date, timestamp, jsonb). Pools hold only immutability-safe columns whose value is identical across each conflict pair's rows; PK columns, DML-updated columns and uuid-default columns are excluded.
- **Assertions**: zero conflict-cache activity for the picked table (same-key events already share a channel), while other tables keep detecting real conflicts.
- **PK-recycle candidate**: one candidate (fixed `(email)` key) reuses the same PK with a different key value; when picked, the test asserts conflicts ARE still detected (the synthetic-PK guard from #3746).
- **Immutability**: new `exclude_columns_from_update` generator option keeps the random generator off whatever columns get sampled.
- **Resume guardrails**: mid-run, the importer is restarted with a changed global `--cdc-partition-key` and a changed per-table override; both must be rejected (new `voyager_import_start_expect_fail` action).
- **Fallback phase** (mirrors `fallback-unique-conflict-test`): full four-conflict-type DML on target, fallback resumptions, zero-conflicts assertion on import-to-source (table routing forced; forward custom keys must not leak), cutover-to-source, row/hash/sequence validations.
- **Schema**: 13-14 columns per table, shaped like production tables (audit backbone + realistic type mix incl. `numeric(38,6)`, jsonb, uuid, arrays).

Development of this test also surfaced a guardrail gap for PK-less-root partitioned tables under default `--use-partition-root=true` (silent acceptance, then a permanent `42P10` wedge on a mid-batch importer restart) — being filed as a DB-21011 follow-up; that shape's supported path is covered by `use-partition-root-false-test`.

No user-facing changes (test infra and migtest harness scripts only).

Repeated full local runs (both phases) across different random table/column picks, including the PK-recycle candidate and pool-sampled keys — conflict assertions, guardrail rejections, cutover-to-source and row/hash/sequence validations all passed.

No.

No.

https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/1278/.
https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-live-migration-resumption/1279/.

### DIFF
--- a/migtests/scripts/live-migration-resumption/orchestrator.py
+++ b/migtests/scripts/live-migration-resumption/orchestrator.py
@@ -387,15 +387,24 @@ def pick_random_custom_key_action(stage: Dict[str, Any], ctx: Any) -> None:
     orchestrator plumbing changes are needed for the pick to take effect.
 
     Required stage key:
-      - candidates: list of {table: "schema.table", columns: [col, ...],
-        expect_conflicts: bool (optional, default false)}. Every candidate's
-        conflict DML must already be safe to route by those columns (i.e.
-        never appears in an UPDATE for that table) -- this action only
-        performs the pick and the wiring, not that verification.
+      - candidates: list of {table: "schema.table", expect_conflicts: bool
+        (optional, default false)} with EITHER:
+          - columns: [col, ...] -- a fixed custom key, OR
+          - random_columns_pool: [col, ...] -- the key's columns are ALSO
+            randomized: 1..2 columns (or min_columns..max_columns) are sampled
+            from the pool, in random order, so successive runs route the same
+            table by different columns and column counts.
+        Every fixed column and every pool column must already be safe to route
+        by (never appears in an UPDATE for that table in the conflict DML, and
+        its value must be identical across the rows of each conflict pair --
+        e.g. a DML-churned key column, or a column the DML leaves at a
+        transaction-constant default) -- this action only performs the pick
+        and the wiring, not that verification.
         `expect_conflicts: true` marks a candidate whose DML deliberately
         creates conflicts that MUST still be detected when it is picked (e.g.
         a PK-recycle pattern); `validate_picked_custom_key_conflicts` uses it
-        to decide which assertion to run.
+        to decide which assertion to run. Such candidates should use fixed
+        `columns` -- their assertion depends on the specific key.
 
     Optional stage key:
       - generator_key: which generator config block to inject
@@ -419,12 +428,22 @@ def pick_random_custom_key_action(stage: Dict[str, Any], ctx: Any) -> None:
 
     choice = random.choice(candidates)
     table = choice["table"]
-    columns = choice["columns"]
+    pool = choice.get("random_columns_pool")
+    if pool:
+        if choice.get("columns"):
+            raise ValueError(f"pick_random_custom_key: candidate {table} must not set both 'columns' and 'random_columns_pool'")
+        min_cols = int(choice.get("min_columns", 1))
+        max_cols = min(int(choice.get("max_columns", 2)), len(pool))
+        num_cols = random.randint(min_cols, max_cols)
+        columns = random.sample(pool, num_cols)
+    else:
+        columns = choice["columns"]
     expect_conflicts = bool(choice.get("expect_conflicts", False))
     ctx.picked_custom_key = {"table": table, "columns": columns, "expect_conflicts": expect_conflicts}
     H.log(
         f"pick_random_custom_key: selected table={table} columns={columns} "
-        f"expect_conflicts={expect_conflicts} (out of {len(candidates)} candidates)"
+        f"expect_conflicts={expect_conflicts} "
+        f"({'sampled from pool of ' + str(len(pool)) if pool else 'fixed'}; out of {len(candidates)} candidates)"
     )
 
     override = f"{table}:({','.join(columns)})"
--- a/migtests/scripts/live-migration-resumption/orchestrator.py
+++ b/migtests/scripts/live-migration-resumption/orchestrator.py
@@ -372,9 +372,9 @@ def validate_no_conflicts_detected_action(stage: Dict[str, Any], ctx: Any) -> No
 
 @action("pick_random_custom_key")
 def pick_random_custom_key_action(stage: Dict[str, Any], ctx: Any) -> None:
-    """Randomly select one table/column(s) to route by `--cdc-partition-key-overrides`
-    instead of the hardcoded default, and wire that pick through everywhere it
-    needs to land:
+    """Randomly select ONE table/column(s) to route by `--cdc-partition-key-overrides`
+    this run -- mirroring how a real user opts specific tables into custom-key
+    routing -- and wire that pick through everywhere it needs to land:
 
       - appends to `voyager.import_data.flags.cdc-partition-key-overrides`.
       - adds the picked column(s) to the named generator's
@@ -386,17 +386,22 @@ def pick_random_custom_key_action(stage: Dict[str, Any], ctx: Any) -> None:
     orchestrator plumbing changes are needed for the pick to take effect.
 
     Required stage key:
-      - candidates: list of {table: "schema.table", columns: [col, ...]}. Every
-        candidate's conflict DML must already be safe to route by those
-        columns (i.e. never appears in an UPDATE for that table) -- this
-        action only performs the pick and the wiring, not that verification.
+      - candidates: list of {table: "schema.table", columns: [col, ...],
+        expect_conflicts: bool (optional, default false)}. Every candidate's
+        conflict DML must already be safe to route by those columns (i.e.
+        never appears in an UPDATE for that table) -- this action only
+        performs the pick and the wiring, not that verification.
+        `expect_conflicts: true` marks a candidate whose DML deliberately
+        creates conflicts that MUST still be detected when it is picked (e.g.
+        a PK-recycle pattern); `validate_picked_custom_key_conflicts` uses it
+        to decide which assertion to run.
 
     Optional stage key:
       - generator_key: which generator config block to inject
         exclude_columns_from_update into (default: "generator").
 
-    The pick is stored on `ctx.picked_custom_key` (`{"table", "columns"}`) for
-    `validate_no_conflicts_for_picked_custom_key` to consume later.
+    The pick is stored on `ctx.picked_custom_key` for
+    `validate_picked_custom_key_conflicts` to consume later.
     """
     if ctx.picked_custom_key is not None:
         raise RuntimeError(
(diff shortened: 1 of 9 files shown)


makalaaneesh:

---

migtests/tests/resumption/live-migration/custom-cdc-partition-key-test/scenario.yaml.template
```diff
@@ -0,0 +1,262 @@
...
+  # --cdc-partition-key-overrides this run -- mirroring how a real user opts a
+  # specific table into custom-key routing. Every candidate's conflict DML
+  # (source_dml.sql) is already immutability-safe for whichever one gets
+  # picked. Also injects the pick into the generator's
+  # exclude_columns_from_update so the random generator respects it too.
+  # single_unique_constraint's DML is a deliberate PK-recycle pattern (see
+  # source_dml.sql section 1), so when IT is picked, conflicts MUST still be
+  # detected (expect_conflicts: true) -- every other candidate must show none.
+  - name: pick_random_custom_key
+    action: pick_random_custom_key
```
priyanshi-yb: I think what we wanted from this test is one more step ahead, to randomly also select the columns of a table to be custom partitioned, so that we keep on testing a variety of columns in runs.

Abhi-yb: refactored

migtests/scripts/live-migration-resumption/orchestrator.py
```diff
@@ -321,19 +358,236 @@ def validate_no_conflicts_detected_action(stage: Dict[str, Any], ctx: Any) -> No
...
+    pool = choice.get("random_columns_pool")
+    unique_index_columns = choice.get("unique_index_columns") or [natural_columns]
+    if pool and random.random() >= float(choice.get("natural_key_probability", 0.5)):
+        min_cols = int(choice.get("min_columns", 1))
+        max_cols = min(int(choice.get("max_columns", 2)), len(pool))
+        columns = random.sample(pool, random.randint(min_cols, max_cols))
+        how = f"sampled from pool of {len(pool)}"
+    else:
+        columns = natural_columns
+        how = "natural unique-key columns"
```
priyanshi-yb: what is natural columns?
what are we doing here?

migtests/scripts/live-migration-resumption/orchestrator.py
```diff
@@ -321,19 +358,236 @@ def validate_no_conflicts_detected_action(stage: Dict[str, Any], ctx: Any) -> No
...
+    # index; if the custom key uses only such columns they are equal on the key
+    # too, so they share a channel and the conflict cache never has to serialize
+    # them -- structurally, for any traffic (deterministic DML or the random
+    # generator). Any other key can legitimately see cross-channel collisions
+    # (e.g. two rows with the same check_id but different status), which the
+    # cache MUST serialize -- so zero conflicts is only asserted in the first
+    # case. The synthetic PK index the importer adds for custom-key tables is
+    # deliberately not part of this check: nothing but the expect_conflicts
+    # PK-recycle candidate ever reuses a primary key.
+    zero_conflicts_guaranteed = all(set(columns) <= set(idx) for idx in unique_index_columns)
```
priyanshi-yb: How are zero conflicts guaranteed?

migtests/tests/resumption/live-migration/common-sql/verify_sequence_restored_after_cutover.sql
```diff
@@ -0,0 +1,6 @@
+-- Honest insert: relies on cutover_table's SERIAL default (nextval) instead of
+-- supplying an id. If the cutover did not restore the sequence past the migrated
+-- rows, nextval returns an already-used id and this fails with a duplicate-key
+-- error, surfacing the missed sequence bump. Run against the DB just cut over to.
+INSERT INTO public.cutover_table(status)
+VALUES ('verify sequence restored after cutover');
```
priyanshi-yb: where are we using this and why are we doing this?

migtests/tests/resumption/live-migration/custom-cdc-partition-key-test/scenario.yaml.template
```diff
@@ -0,0 +1,477 @@
...
+    action: voyager_import_start_expect_fail
+    flags:
+      cdc-partition-key: table
+    error_contains: "not allowed after the import data has started"
+
+  - name: guardrail_reject_picked_override_change
+    action: voyager_import_start_expect_fail
+    flags:
+      cdc-partition-key-overrides: "{picked_table}:pk"
+    error_contains: "not allowed after the import data has started"
```
priyanshi-yb: I don't think this is the right place to put these guadrails validations, it should be in unit/integration tests which I have added in go side tests.

migtests/tests/resumption/live-migration/custom-cdc-partition-key-test/scenario.yaml.template
```diff
@@ -0,0 +1,477 @@
...
+        random_columns_pool: [email, phone_number, status, metadata, amount, due_date, seq_no, is_active, retry_count, created_at, updated_at, deleted_at]
+      - table: public.subset_columns_unique_constraint_and_index
+        columns: [first_name, last_name]
+        unique_index_columns: [[first_name, last_name], [first_name, last_name, phone_number]]
+        random_columns_pool: [first_name, last_name, phone_number, status, description, amount, due_date, seq_no, is_active, metadata, created_at, updated_at, deleted_at]
+      - table: public.test_partial_unique_index
+        columns: [check_id]
+        unique_index_columns: [[check_id]]
+        random_columns_pool: [check_id, status, description, amount, due_date, seq_no, is_active, metadata, created_at, updated_at, deleted_at]
+      - table: public.partitioned_unique_conflict
```
priyanshi-yb, line 193: from AI:

[CRITICAL] scenario.yaml.template:~150 (partitioned candidate) / orchestrator.py:485-495 — Sampled custom key on partitioned_unique_conflict is not protected on leaf partitions → unrecoverable importer crash loop
- Problem/Suggestion: pick_random_custom_key injects exclude_columns_from_update keyed by the bare root name (partitioned_unique_conflict). But the event generator lists tables via information_schema.tables WHERE table_type = 'BASE TABLE', which includes the leaf partitions (..._east, ..._west, ..._default) as independent tables. If this candidate is picked with a sampled key (e.g. [status], ~50% of the times it's picked), the generator will freely UPDATE status on leaf rows. Debezium emits the event on the leaf, import renames it leaf→root (SourceRenameTablesMap, common.go:1062), and it then hits the root's PARTITION_BY_CUSTOM strategy with a changed key column — a hard error at live_migration.go:520 ("custom partition key column %q is required to be immutable"). The event replays on every resume, so the run enters a permanent crash loop. Roughly 1/12 × 1/2 ≈ 4% of runs of a multi-hour pipeline die this way. The natural key [region] is safe only by luck — it's part of every leaf's PK, so pk_set excludes it.
- Code line: migtests/scripts/live-migration-resumption/orchestrator.py:485-495 (exclusion injection), migtests/tests/resumption/live-migration/custom-cdc-partition-key-test/scenario.yaml.template (partitioned_unique_conflict candidate's random_columns_pool)
- Failure Code Case: pick = public.partitioned_unique_conflict, sampled key = [status]; generator runs UPDATE partitioned_unique_conflict_east SET status = ... → importer dies on every resume.

migtests/scripts/live-migration-resumption/orchestrator.py
priyanshi-yb: If we are randomly sampling from the pool, why do we need to use the set of columns from columns/unique-index-columns?

Abhi-yb: Simplified — we now just sample 1-2 random columns from the pool

migtests/tests/resumption/live-migration/custom-cdc-partition-key-test/scenario.yaml.template
Abhi-yb: refactored

migtests/tests/resumption/live-migration/common-sql/verify_sequence_restored_after_cutover.sql
Abhi-yb: Actually I was in other branch commited this by mistake,
This was other pr change my bad

migtests/tests/resumption/live-migration/custom-cdc-partition-key-test/scenario.yaml.template
Abhi-yb: Fixed — the partitioned candidate now declares its leaf partitions, and the generator's column exclusion is applied to the root and all leaves.

migtests/scripts/live-migration-resumption/orchestrator.py
```diff
@@ -321,19 +357,156 @@ def validate_no_conflicts_detected_action(stage: Dict[str, Any], ctx: Any) -> No
...
+    candidates = stage.get("candidates")
+    if not candidates:
+        raise ValueError("pick_random_custom_key: 'candidates' stage key is required and must be non-empty")
+
+    choice = random.choice(candidates)
+    table = choice["table"]
+    pool = choice.get("random_columns_pool")
+    if pool:
+        if choice.get("columns"):
+            raise ValueError(f"pick_random_custom_key: candidate {table} must not set both 'columns' and 'random_columns_pool'")
```
priyanshi-yb: why do we have this check here? in case the pool is present why do we need to check for columns?

migtests/tests/resumption/live-migration/custom-cdc-partition-key-test/scenario.yaml.template
```diff
@@ -0,0 +1,419 @@
...
+      wait_duration_seconds: 1
+      table_weights: {}
+      operation_weights: {INSERT: 3, UPDATE: 2, DELETE: 1}
+      random_seed: 123
+      faker_seed: 123
+      insert_rows: 3
+      update_rows: 2
+      delete_rows: 1
+      insert_max_retries: 5
+      update_max_retries: 1
```
priyanshi-yb, line 61: We need to have a configuration for the partitioned table to list the column overrides and values of the partition keys. Do we not need it here?

migtests/tests/resumption/live-migration/custom-cdc-partition-key-test/source_dml.sql
```diff
@@ -0,0 +1,295 @@
...
+-- 5. multi_unique_index (id PK, UNIQUE INDEX(first_name, last_name)) -- custom-key candidate: (first_name, last_name)
+-- ============================================================
+BEGIN;
+INSERT INTO multi_unique_index (id, first_name, last_name) VALUES
+    (:base + 40001, ('IdxJohn' || :cycle), ('Doe' || :cycle));
+
+-- DELETE-INSERT
+DELETE FROM multi_unique_index WHERE id = :base + 40001;
+INSERT INTO multi_unique_index (id, first_name, last_name) VALUES (:base + 40101, ('IdxJohn' || :cycle), ('Doe' || :cycle));
+COMMIT;
```
priyanshi-yb, line 161: In these DML operations, we are not inserting data into all the columns of the table. I think we should set these, since we are randomly choosing columns, and if we pick columns with null values, all those statements will go to the same channel. And also, these conflict cases won't trigger. So that we can add all the column data to these statements

priyanshi-yb (approved):
LGTM

PR #3773, opened 2026-09-02
Wrap errors with %w instead of %v/%s in ErrExit and goerrors.Errorf

makalaaneesh:

Changes `%v`/`%s` to `%w` at 433 call sites (61 files) where the formatted argument is statically an `error`, so the error chain stays unwrappable.

**Why this matters.** Callhome derives every structured field on an error via `errors.As` over the unwrap chain:

| Field | Extractor | Looks for |
|---|---|---|
| `pg_error_code` | `addPostgreSQLErrorContext` | `*pgconn.PgError` / `*pgconnv5.PgError` |
| `step`, `flow` | `addImportBatchErrorContext` | `errs.ImportBatchError` |
| `ddl` | `addExecuteDDLErrorContext` | `errs.ExecuteDDLError` |
| `stack_trace` | `addStackTrace` → `findInnermostGoError` | deepest `*goerrors.Error` |

`utils.ErrExit` builds the error it hands to callhome with `goerrors.Errorf`, which is `Wrap(fmt.Errorf(format, args...), 1)`. So the verb at the call site decides whether the chain survives: `%w` yields a `*fmt.wrapError` that `errors.As` can walk, while `%v`/`%s` yields a plain `errorString` with no `Unwrap()` and the chain dead-ends.

Two positions in the chain were affected:

- **The sink** — `utils.ErrExit(...)` in `cmd/`. Severs everything below it, including leaf stacks that were already being captured correctly.
- **The leaves** — `goerrors.Errorf(...)` that create the `*goerrors.Error` (so the stack is captured in the right place) but format the underlying library error with `%v`/`%s`, leaving the driver's `*pgconn.PgError` unreachable.

`prepareTargetDBForImport` shows both, stacked:

```go
// leaf: stack captured here, but the underlying error is unreachable
return goerrors.Errorf("Failed to initialize the target DB connection pool: %v", err)
...
// sink: discards the leaf's goerrors.Error entirely
utils.ErrExit("Failed to prepare target DB for import: %s", err)
```

Its four leaves each identify a different sub-step (connection pool, adaptive parallelism, health monitoring, metadata schema). The single `%s` above them collapsed all four into one error carrying no SQLSTATE and a stack pointing only at the exit site.

**Breakdown:** 287 `%v` → `%w`, 146 `%s` → `%w`.

None. `ErrExit` already rewrites `%w` to `%s` before formatting the message it prints:

```go
ErrExitErr = goerrors.Errorf(formatString, args...)
message := fmt.Sprintf(formatString, args...)
```

and `fmt` renders an error identically for `%v`, `%s` and `%w`. Console output, log output and exit codes are unchanged.

The change is observable only in callhome payloads, which should now carry `pg_error_code` / `step` / `flow` / `ddl` and a root-cause stack trace for error paths that previously reported none.

Existing tests were sufficient — this changes no behaviour, only error wrapping.

- `go build ./...` — clean
- `go vet ./...` and `go vet -tags unit ./...` — clean
- `staticcheck ./...` and `staticcheck -tags unit ./...` (2025.1.1, pinned in `versions/ci-config.json`) — clean
- `go test -tags unit ./...` — all passing
- `gofmt` — no new offenders. Six touched files are listed by `gofmt -l`, but all six were already unformatted at HEAD (33 files in the tree are); reformatting them is out of scope for this PR.

The rewrite was applied mechanically with a `go/ast` + `go/types` pass rather than a regex, so each verb was mapped positionally to its actual argument and rewritten only when that argument's static type implements `error`. This matters for multi-verb formats — for example `ErrExit("get table import state: %s: %s", task.TableNameTup, err)` has only its second verb changed, and `ErrExit("error executing batch on channel %v: %v", chanNo, err)` keeps the first `%v` for the channel number.

Guards applied by the rewriter:
- skips `%!w`-unsafe operands: `nil`, and anything whose static type is `string` (so `err.Error()` is never rewritten)
- skips flagged verbs (`%+v`), explicit argument indexes (`%[2]d`) and `*` width/precision, where positional mapping is unreliable
- skips calls with a `...` spread
- refuses to edit unless the bytes at the computed offset are exactly the expected `%` + verb

The resulting diff was then verified line-by-line to confirm every changed line differs from its original *only* at verb positions — 432 lines checked, no other edits.

No schema change, so no version bump. The payload shape is untouched; the existing `error` field's already-defined sub-fields (`pg_error_code`, `step`, `flow`, `ddl`, `stack_trace`) will simply be populated on paths where they previously came back empty.

No. Nothing in this PR touches on-disk structures.

### Follow-ups deliberately left out of scope

This PR is strictly the verb change. Related gaps found while investigating, each worth its own PR:

1. **10 `ErrExit` sites pass `err.Error()`** (`cmd/importData.go:140,145,155`, `cmd/exportData.go:122,1607,1770`, `cmd/importSchema.go:68`, `cmd/importDataFileCommand.go:247`, `cmd/assessMigrationBulkCommand.go:58`, `cmd/endMigrationCommand.go:56`). These need the `.Error()` call dropped as well, since `%w` on a `string` renders `%!w(string=...)` — a verb-only change would break them, so they were left untouched.
2. **Constructor-built error types capture no stack.** `errs.NewImportBatchError` and `errs.NewExecuteDDLError` carry rich typed context but are plain structs, so `findInnermostGoError` falls back to the `ErrExit`-site stack for the two highest-volume error buckets.
3. **Format strings that interpolate a value before the first `:`.** `SanitizeErrorMsg` keeps only the segment before the first colon, so `ErrExit("error executing batch on channel %v: ...", chanNo, ...)` puts the channel number into the callhome error message and fragments one failure mode into many single-occurrence buckets. `cmd/live_migration.go:609` does the same with a channel number and a VSN.
4. **A lint rule** rejecting `%s`/`%v` applied to an `error` in `ErrExit` / `goerrors.Errorf` / `fmt.Errorf`, plus `.Error()` passed to any of them, would keep this from regressing; extending it to flag a `%` verb before the first `:` would cover (3) in the same pass.


makalaaneesh:

---

priyanshi-yb (approved):
LGTM

PR #3779, opened 2026-09-03
ci: fail-fast:false on migtest matrices, and distinct workflow names

makalaaneesh:

Two small, independent GitHub Actions changes. Neither changes *what* is tested — only how much signal survives a failure, and how readable the results are.

> **Note on history:** this branch originally also added `concurrency` groups with `cancel-in-progress`. That has been dropped (see the last commit) — superseding an in-progress run also destroys its record, and the state of a previous run is often how you distinguish an infrastructure flake from a real regression. The net diff against `main` is just the two changes below; the repo squash-merges, so the add-then-remove pair collapses on merge.

**1. `fail-fast: false` on the three migtest matrices** (`pg-13`, `pg-17`, `mysql`)

These declare a `matrix` without `fail-fast`, so it defaults to `true` and one failing leg cancels every sibling.

Measured over 1 Jun – 3 Sep 2026 (4,984 workflow runs, job/step detail for all 1,186 failures):

| | Failed jobs | Cancelled jobs |
|---|---|---|
| PG 17 migtests | 86 | **771** |
| All migtests | 205 | **1,744** |

One flaky leg destroys the result for the other four YB versions, and the only recovery is re-running all 10 legs. Since most of these failures are infrastructure flakes rather than version-specific bugs (77% of manually re-run workflows — 201 of 262 — passed unchanged), the cancelled siblings would almost always have been green.

This also makes GitHub's "Re-run failed jobs" useful: today a flake usually means re-running all 10 legs, where it should mean re-running 1.

`integration-tests`, `issue-tests` and `failpoint-tests` already set `fail-fast: false`. `misc-migtests` has no matrix. `build-voyager` is deliberately untouched — it is tag-triggered and has a separate pre-existing failure mode.

**2. Give the three workflows named `Go` distinct names**

`go.yml`, `integration-tests.yml` and `issue-tests.yml` all declared `name: Go`, so the Actions UI and the PR checks list collapsed three unrelated workflows under one heading. A red `Go` could be a unit-test failure, a testcontainers integration failure, or a version-gated issue test — and the three have very different failure rates over the same window (**14.6% / 26.8% / 78.3%**).

| File | Was | Now |
|---|---|---|
| `go.yml` | `Go` | `Build, Unit Tests & Static Checks` |
| `integration-tests.yml` | `Go` | `Integration Tests` |
| `issue-tests.yml` | `Go` | `Issue Tests (All YB Versions)` |

**This is safe for branch protection.** Required status checks match on check-run names, which come from job ids/names — not the workflow's display name. `main` currently requires `build-and-test`, `integration-tests (integration | integration_live_migration | integration_voyager_command)`, and `Failpoint: Cutover | Export | Import`. This PR changes no job id and no job `name:`, so every one of those contexts still resolves.

None. CI configuration only — no changes to the CLI, configuration, installation, or reports.

- All ten workflow files parse as valid YAML.
- Verified the net diff against `origin/main` is exactly three `name:` lines and three `fail-fast: false` lines — no job id or job `name:` changes, so no required status-check context changes.
- Verified every PR-triggered matrix job now resolves `fail-fast: false`, and that all ten workflow names are unique.

The behaviour is observable on this PR: a single failing matrix leg should now leave its siblings running to completion.

No.

No.

### DIFF
--- a/.github/workflows/failpoint-tests.yml
+++ b/.github/workflows/failpoint-tests.yml
@@ -6,6 +6,14 @@ on:
   push:
     branches: [ main, '*.*-dev', '*.*.*-dev' ]
 
+# Supersede in-progress runs for the same PR; main / *-dev pushes are never
+# cancelled so every merged commit keeps a full CI record. The group is a literal
+# slug, not ${{ github.workflow }}, because go.yml, integration-tests.yml and
+# issue-tests.yml are all named "Go" and would otherwise share a single group.
+concurrency:
+  group: failpoint-tests-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   prepare-versions:
     uses: ./.github/workflows/prepare-versions.yml
--- a/.github/workflows/failpoint-tests.yml
+++ b/.github/workflows/failpoint-tests.yml
@@ -7,11 +7,9 @@ on:
     branches: [ main, '*.*-dev', '*.*.*-dev' ]
 
 # Supersede in-progress runs for the same PR; main / *-dev pushes are never
-# cancelled so every merged commit keeps a full CI record. The group is a literal
-# slug rather than ${{ github.workflow }} so that editing a workflow's display
-# name can never silently change how runs are grouped and cancelled.
+# cancelled so every merged commit keeps a full CI record.
 concurrency:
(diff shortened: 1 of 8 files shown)


makalaaneesh:

---

priyanshi-yb (approved):
LGTM

PR #3784, opened 2026-09-04
custom-cdc-partition-key-test follow-up: production-shaped schema, resume guardrail checks, fallback phase

Abhi-yb:
Stacked on #3770 (base branch is that PR's branch; will retarget to main once it merges).

Follow-up hardening of `custom-cdc-partition-key-test` before it joins nightly runs.

- **Schema**: every table widened to 13-14 columns shaped like production tables (audit-timestamp backbone; varchar/text/timestamp-heavy mix with boolean, int, bigint, smallint, date, `numeric(38,6)`, jsonb, uuid, arrays, double precision spread across tables). Keys/unique indexes unchanged, so the conflict DML is untouched. The picked custom-key table now keeps receiving real UPDATE traffic on its non-key columns.
- **Resume guardrails**: new `voyager_import_start_expect_fail` orchestrator action. Mid-run, the scenario stops the importer and asserts two illegal config changes are rejected — changing the global `--cdc-partition-key` default, and changing the picked table's override to `pk` — then resumes with the original flags and completes.
- **Fallback phase** mirroring `fallback-unique-conflict-test`: cutover-to-target with `prepare-for-fall-back`, target-side random + conflict generators (full four conflict types — custom keys don't apply on this leg), resumptions on export-from-target/import-to-source, a zero-conflicts assertion on the fallback importer (table routing forced; forward-leg custom-key config must not leak), cutover-to-source, then row-count/row-hash/sequence validations.

During development this test also surfaced a guardrail gap for partitioned tables whose root lacks a PK under default `--use-partition-root=true` (silently accepted, then permanently wedges on a mid-batch importer restart with SQLSTATE 42P10) — that table was moved out of this test (its supported path is covered by `use-partition-root-false-test`) and the gap is being filed as a follow-up to DB-21011.

No user-facing changes (test infra and migtest harness scripts only).

Full two-phase local runs (forward + fallback) passed repeatedly across different random picks, including the PK-recycle candidate; guardrail rejection stages verified against both config-change variants.

No.

No.

### DIFF
--- a/migtests/scripts/live-migration-resumption/orchestrator.py
+++ b/migtests/scripts/live-migration-resumption/orchestrator.py
@@ -4,6 +4,7 @@ import os
 import sys
 import glob
 import argparse
+import copy
 import random
 import subprocess
 import time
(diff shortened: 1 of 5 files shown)


Abhi-yb:
Folding into #3770 per discussion — same PR.

PR #3786, opened 2026-09-07
implement api handling for amp

AmeyaKasture:
- Let external orchestration supply a valid `YB_VOYAGER_MIGRATION_UUID` for a new migration. Invalid or absent values still generate a UUID.
- Forward that variable through the Docker wrapper and safely handle export directories with no Docker metadata files.
- Clean up assessment-test containers per test to avoid cross-test leakage.

New optional environment variable: `YB_VOYAGER_MIGRATION_UUID`.

Added `TestInitMigrationStatusRecordMigrationUUID`.
`go test -tags unit ./src/metadb`

No.

No.

---

| Component |
| MetaDB |
| Name registry json |
| Data File Descriptor Json |
| Callhome Json |
| Export Status Json |
| YugabyteD Tables |
| TargetDB Metadata Tables |
| Schema Dump |
| AssessmentDB |
| Import Data State |
| Data .sql files of tables |

### DIFF
--- a/yb-voyager/src/metadb/migrationStatus.go
+++ b/yb-voyager/src/metadb/migrationStatus.go
@@ -17,6 +17,7 @@ package metadb
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"time"
 
@@ -119,7 +120,6 @@ type MigrationStatusRecord struct {
 	//Iteration specific details
 	ExportDataFromSourceStarted bool `json:"ExportDataFromSourceStarted"`
 	ImportDataToTargetStarted   bool `json:"ImportDataToTargetStarted"`
-
 }
 
 type CutoverTimingRecord struct {
@@ -146,7 +146,10 @@ type CutoverTimingRecord struct {
 	ExportFromTargetFallBackStartedAt    time.Time `json:"ExportFromTargetFallBackStartedAt,omitempty"`
 }
 
-const MIGRATION_STATUS_KEY = "migration_status"
+const (
+	MIGRATION_STATUS_KEY    = "migration_status"
+	migrationUUIDEnvVarName = "YB_VOYAGER_MIGRATION_UUID"
+)
 
 func (m *MetaDB) UpdateMigrationStatusRecord(updateFn func(*MigrationStatusRecord)) error {
 	return UpdateJsonObjectInMetaDB(m, MIGRATION_STATUS_KEY, updateFn)
@@ -165,6 +168,19 @@ func (m *MetaDB) GetMigrationStatusRecord() (*MigrationStatusRecord, error) {
 }
 
 func (m *MetaDB) InitMigrationStatusRecord(cfgFile string) error {
+	record, err := m.GetMigrationStatusRecord()
+	if err != nil {
+		return err
+	}
+	if record != nil && record.MigrationUUID != "" {
+		return nil // already initialized
+	}
+
+	migrationUUID, err := externalOrNewMigrationUUID()
+	if err != nil {
+		return err
+	}
+
 	return m.UpdateMigrationStatusRecord(func(record *MigrationStatusRecord) {
 		if record != nil && record.MigrationUUID != "" {
 			return // already initialized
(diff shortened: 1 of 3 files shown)


yb-voyager/cmd/assessMigrationCommand_test.go
```diff
@@ -39,20 +38,19 @@ import (
...
 	// to avoid info level logs flooding the test output
 	log.SetLevel(log.WarnLevel)
 
-	exitCode := m.Run()
-
-	// cleaning up all the running containers
-	testcontainers.TerminateAllContainers()
-
-	os.Exit(exitCode)
+	t.Cleanup(testcontainers.TerminateAllContainers)
```
makalaaneesh: [AI] **Per-test `TerminateAllContainers` breaks the container-singleton reuse pattern.**

`testcontainers.NewTestContainer` is a singleton keyed by `<dbType>-<version>` (see `containerRegistry` in `test/containers/testcontainers.go`, lines 16-21, 83-131). The first test to request `NewTestContainer("postgresql", nil)` creates the entry; subsequent calls reuse it, and `PostgresContainer.Start` short-circuits when the container is already running.

The previous `TestMain` deliberately called `TerminateAllContainers` once, *after* `m.Run()`, so the registry stayed valid throughout the test run. Moving it into `t.Cleanup` runs it after every test, but:

1. `TerminateAllContainers` does not delete entries from `containerRegistry` (see `testcontainers.go:145-154`).
2. `PostgresContainer.Terminate` does not clear `pg.container` (`postgres_container.go:132-144`).

So after the first test finishes, the registry still holds a `PostgresContainer` whose `pg.container` is a *terminated* testcontainers-go container. When the next test calls `NewTestContainer("postgresql", nil)`, it gets that same object back, and `Start` (postgres_container.go:30-49) sees `pg.container != nil && !IsRunning()` and falls into the "restart in place" branch — calling `Start` on a container that was already removed. The second-and-later `Test_AssessMigration_*` tests in this file will fail here (or on the next DB operation), even though each ran fine in isolation.

Failure case: run the file end-to-end (`go test -tags integration_voyager_command ./yb-voyager/cmd/...` restricted to these tests). `Test_AssessMigration` passes, then `Test_AssessMigration_UsageCategory` (and the two after it) fails when trying to bring the "restarted" postgres container back up.

Fix options:
- Drop the per-test `t.Cleanup(TerminateAllContainers)` and keep a single-shot teardown (a package-level `TestMain` or one-time `sync.Once` from an init helper) so the singleton pattern the registry expects is preserved.
- Or make `TerminateAllContainers` (and each `Terminate` implementation) reset the container reference and delete the registry entry, so a fresh container is created on the next `NewTestContainer` call.

Also worth noting: `yb-voyager/src/.cursor/BUGBOT.md` explicitly says "Organize `TestMain` to start containers before tests and terminate them after all tests complete (via `m.Run()`)."

---

yb-voyager/src/metadb/migrationStatus.go
```diff
@@ -177,10 +180,18 @@ func (m *MetaDB) InitMigrationStatusRecord(cfgFile string) error {
...
 		}
 
-		record.MigrationUUID = uuid.New().String()
+		record.MigrationUUID = externalOrNewMigrationUUID()
 	})
 }
 
+func externalOrNewMigrationUUID() string {
+	migrationUUID, err := uuid.Parse(os.Getenv(migrationUUIDEnvVarName))
+	if err != nil {
```
makalaaneesh, line 211: [AI] **Silently discards an invalid `YB_VOYAGER_MIGRATION_UUID` and falls back to a random UUID.**

If a user (or wrapping automation) explicitly sets `YB_VOYAGER_MIGRATION_UUID` but the value fails `uuid.Parse` — a typo, a non-UUID identifier like `run-42`, whitespace, wrong casing/format — this helper silently generates a fresh UUID with no log line. The whole point of the env var is to let an external caller correlate a migration by UUID, and the caller has no way to notice its value was rejected. The correlation is lost until someone reads the meta.db and cross-checks.

Per the repo root `.cursor/BUGBOT.md` and `yb-voyager/.cursor/BUGBOT.md`: "Never silently swallow errors. If a function returns an error, either handle it, return it, or log it with sufficient context." The unset case is fine (env var is empty and we intentionally fall back), but the set-but-invalid case should be surfaced.

Failure case: a CI pipeline exports `YB_VOYAGER_MIGRATION_UUID=my-run-42` before invoking voyager. Voyager silently rolls a random UUID; the pipeline's downstream log/metric aggregation can't join by the UUID it thought it injected, and nothing in the voyager output points at the typo.

Suggested change — only fall back silently when the env var is unset; log a warning when it's set but unparseable:

```go
func externalOrNewMigrationUUID() string {
    envVal := os.Getenv(migrationUUIDEnvVarName)
    if envVal == "" {
        return uuid.New().String()
    }
    migrationUUID, err := uuid.Parse(envVal)
    if err != nil {
        log.Warnf("ignoring invalid %s=%q: %v; generating a new UUID", migrationUUIDEnvVarName, envVal, err)
        return uuid.New().String()
    }
    return migrationUUID.String()
}
```

---

makalaaneesh (commented):
[AI] Automated review — 2 findings (1 critical, 1 warning). See inline comments.

---

yb-voyager/cmd/assessMigrationCommand_test.go
AmeyaKasture: Good catch. This test change was unrelated to the PR (AMP migration UUID handling) and, as you noted, breaks the container-singleton reuse pattern and contradicts the BUGBOT.md guidance. Reverted the file to `main` — the original single-shot `TestMain` teardown is restored.

yb-voyager/src/metadb/migrationStatus.go
AmeyaKasture: Fixed. `externalOrNewMigrationUUID` now returns `(string, error)`: it generates a fresh UUID only when the env var is unset, and returns an error when it is set but unparseable, so `InitMigrationStatusRecord` fails fast instead of silently rolling a random UUID. Nothing is persisted for a rejected value. Updated the unit test accordingly (`errors when environment is set but invalid`).

priyanshi-yb (approved):
LGTM

PR #3798, opened 2026-09-11
Enhance assertions of the live migration unique key conflicts related tests

priyanshi-yb:

Strengthens the live-migration unique-key conflict tests (live_migration_uk_conflict_test.go).

- Adds an upper-bound assertion on detected conflict counts to all 10 conflict tests. Previously they only asserted Total > 0, so runaway double-counting or dedup regressions would pass unnoticed.
- Each bound is counted from that test's delta: the number of cached delete/update events a different-partition-key incoming event can catch. A cached event pairs with at most one incoming event, so timing can only lower the count — the bounds cannot flake.
- New test TestLiveMigrationCustomCdcPartitionKeyMutationFailsImport: an update that mutates one column of a multi-column custom partition key must fail the import with the "required to be immutable" error. The guard had no coverage.
- New test TestLiveMigrationCustomCdcPartitionKeyUniqueKeyConflictDetection: conflicts on a real unique index (not the custom key column) across different custom keys are detected and serialized. Covers both single-column and multi-column custom keys — the first conflict true-positive test for a multi-column key.

No user-facing changes.

Ran all 12 affected tests locally with go test -tags failpoint_import against real PG/YB containers — all pass. The NULLS NOT DISTINCT run recorded 2500 conflicts against its 4000 bound.

No.

No.


yb-voyager/src/testlivemigration/live_migration_testing_framework.go
```diff
@@ -1501,9 +1501,14 @@ func (lm *LiveMigrationTest) GetDataMigrationReport() (*DataMigrationReport, err
 		if lm.sourceReplicaContainer != nil {
 			reportArgs = append(reportArgs, "--source-replica-db-password", lm.sourceReplicaContainer.GetConfig().Password)
 		}
-		err := testutils.NewVoyagerCommandRunner(nil, "get data-migration-report", reportArgs, nil, true).WithT(lm.t).Run()
+		err := testutils.NewVoyagerCommandRunner(nil, "get data-migration-report", reportArgs, nil, false).WithT(lm.t).Run()
```
makalaaneesh, line 1506: [AI][Suggestion] **Retry-on-command-failure has no comment naming the transient failure it exists for.**

Both `AGENTS.md` layers that apply here call this out. `yb-voyager/AGENTS.md` (Test Determinism): "poll for the condition with a timeout, and **comment what is being waited for**." `yb-voyager/src/testlivemigration/AGENTS.md`: "Never use a fixed `time.Sleep` to wait for a migration phase; use the framework's polling helpers with an explicit timeout." The commit message says the retry is "because of timing", but the code itself carries no comment naming which transient error is being absorbed. A future reader (or a real regression) sees only "sleep 2 s and try again up to 5 times" and cannot tell whether a genuine failure is being masked.

Failure case: if `get data-migration-report` starts consistently failing for a real reason (bad flag, missing meta-DB entry), the test now spends ~10 s retrying and then reports the *last* wrapped error, hiding that the failure was reproducible.

Suggested fix: add a comment naming the concrete transient error class being retried (e.g. meta-DB lock contention), and bail out immediately on any error class the retry is not meant to absorb.

---

yb-voyager/src/testlivemigration/live_migration_testing_framework.go
```diff
@@ -1501,9 +1501,14 @@ func (lm *LiveMigrationTest) GetDataMigrationReport() (*DataMigrationReport, err
...
-			return nil, goerrors.Errorf("get data-migration-report command failed: %w", err)
+			maxRetry--
+			if maxRetry <= 0 {
+				return nil, goerrors.Errorf("failed to get data migration report: %w", err)
+			}
+			time.Sleep(2 * time.Second)
+			continue
 		}
 
 		reportFilePath := filepath.Join(lm.exportDir, "reports", "data-migration-report.json")
```
makalaaneesh, line 1517: [AI][Suggestion] **Stale `!FileOrFolderExists` retry branch after switching to synchronous execution.**

(Anchored here — the `if !utils.FileOrFolderExists(reportFilePath) { ... }` block starts on the next line, which is just outside the diff hunk.)

Before this PR the command was async, so the child could still be writing `data-migration-report.json` when `Run()` returned; the file-existence retry was needed. Now that the runner is synchronous, `Run() == nil` implies the child process exited cleanly and therefore that the report file exists (the child writes it before exit). The `!utils.FileOrFolderExists(reportFilePath)` branch then rerunning the whole command is effectively dead code. Per the root `AGENTS.md` "Fail loudly on unexpected state", if this branch is reachable it should error out, not retry.

Failure case: if the CLI ever ships a bug where it exits 0 without writing the report, the helper will retry silently for ~10 s and then return a generic `"report file does not exist"` error, obscuring the CLI bug.

Suggested fix: replace the retry branch with `return nil, goerrors.Errorf("get data-migration-report succeeded but %s is missing", reportFilePath)`.

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -348,6 +348,19 @@ func TestLiveMigrationWithMultiColumnUniqueIndexConflictDetectionCases(t *testin
 	require.Greater(t, conflictStats.Total, 0, "true-positive delta should produce UK conflicts")
 	require.Greater(t, conflictStats.ByTable[`"test_schema"."test_multi_column_unique_index"`], 0, "test_multi_column_unique_index should produce UK conflicts")
```
makalaaneesh: [AI][Suggestion] **Lower bounds are still `> 0` on counts expected in the hundreds/thousands.**

The PR correctly adds workload-derived upper bounds. The `require.Greater(conflictStats.Total, 0, …)` / `require.Greater(conflictStats.ByTable[t], 0, …)` companions were not tightened. Per `yb-voyager/src/testlivemigration/AGENTS.md`: "assert a justified range (both lower and upper bound), derived from the workload and explained in a comment. `>= 0`-style bounds are vacuous." `> 0` on an expected count of, e.g., 500-1500 per table is functionally the same as the vacuous case it calls out. The upper-bound comments already contain the derivation ("3 catchable events per iteration × 500"), so a matching lower bound is a small change of the same shape.

Failure case: a regression that silently reduces conflict detection to a handful of events per test (say the reader stops caching after the first batch) would still satisfy `> 0` on every table and pass. The upper bounds do not catch this direction.

Suggested fix: pick a conservative floor derived from the same workload analysis. For example, replace `require.Greater(t, conflictStats.ByTable[…], 0, …)` with `require.GreaterOrEqual(t, conflictStats.ByTable[…], 500, "at least 1 catchable event per iteration × 500 iterations")`, and similarly for each of the other affected tests.

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -545,6 +558,13 @@ func TestLiveMigrationWithUniqueKeyValuesWithPartialPredicateConflictDetectionCa
 
 	require.Greater(t, conflictStats.Total, 0, "true-positive delta should produce UK conflicts")
 	require.Greater(t, conflictStats.ByTable[`"test_schema"."test_live"`], 0, "test_live should produce UK conflicts")
+	// Upper bound, counted from the delta: all 5 updates + 2 deletes per iteration carry
```
makalaaneesh, line 561: [AI][Suggestion] **Comments justify "indexable" without accounting for the partial-index `WHERE most_recent` predicate.**

In `TestLiveMigrationWithUniqueKeyValuesWithPartialPredicateConflictDetectionCases` the comment reads "all 5 updates + 2 deletes per iteration carry before-image check_id=20 (indexed, non-NULL)". The index is *partial* (`… WHERE most_recent`) — some of those 5 updates have before-image `most_recent=false` and are therefore not in the index at before-image time. The same slip appears in the null-partial test's comment. The bounds `<= 3500` and `<= 2500` are still valid — they are just looser than they read. The concern is that a future reader relies on the comment when tuning the assertion or writing a similar test.

Failure case: if the workload changes (e.g. one of the "most_recent = true" flips is dropped) the true ceiling can drop below the 3500 / 2500 assertion, and the comment's overstated derivation will suggest the assertion is still tight when it is not.

Suggested fix: rephrase the reasoning as "at most 5 updates + 2 deletes per iteration are potentially catchable (the WHERE most_recent predicate excludes those whose before-image is most_recent=false, so the true max is lower)" and keep the numeric bound.

---

makalaaneesh (commented):
[AI] Automated review — 4 findings (0 critical, 0 warning, 4 suggestion). See inline comments.

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
makalaaneesh: Interesting suggestion; is this possible to do reliably (without it becoming flaky)? If not, we can keep it at > 0

yb-voyager/src/testlivemigration/live_migration_testing_framework.go
```diff
@@ -1501,9 +1501,14 @@ func (lm *LiveMigrationTest) GetDataMigrationReport() (*DataMigrationReport, err
...
-		err := testutils.NewVoyagerCommandRunner(nil, "get data-migration-report", reportArgs, nil, true).WithT(lm.t).Run()
+		err := testutils.NewVoyagerCommandRunner(nil, "get data-migration-report", reportArgs, nil, false).WithT(lm.t).Run()
 		if err != nil {
-			return nil, goerrors.Errorf("get data-migration-report command failed: %w", err)
+			maxRetry--
+			if maxRetry <= 0 {
+				return nil, goerrors.Errorf("failed to get data migration report: %w", err)
+			}
+			time.Sleep(2 * time.Second)
+			continue
```
makalaaneesh, line 1514: will this also solve flakiness problems in other tests?

priyanshi-yb: yes i think it should now

makalaaneesh (approved):
LGTM

PR #3799, opened 2026-09-11
Added couple of more tests around custom partition key 

priyanshi-yb:

Adds two live-migration tests closing coverage gaps around custom CDC partition keys.

- `TestLiveMigrationCustomCdcPartitionKeyMutationFailsImport` — the custom-key immutability guard had no test at any level. An update that mutates one column of a multi-column custom key `(ck1, ck2)` must kill the import with the "required to be immutable" error naming the column. The test then restarts with `--start-clean` and default routing to confirm the migration recovers through cutover.
- `TestLiveMigrationCustomCdcPartitionKeyUniqueKeyConflictDetection` — no test verified conflict detection on a real unique index for a custom-routed table. The delta frees and reclaims a unique value across rows carrying different custom keys, producing an UPDATE→INSERT and a DELETE→UPDATE conflict per iteration that must be detected. Covers both a single-column and a multi-column custom key. This is the first conflict true-positive test on a multi-column key.

No user-facing changes.

Both new tests pass locally against real PG/YB containers: `go test -tags failpoint_import -run 'TestLiveMigrationCustomCdcPartitionKey(MutationFailsImport|UniqueKeyConflictDetection)'` (~44s each).

No.

No.

---

| Component        


yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -3494,3 +3494,285 @@ func TestLiveMigrationCdcPartitionKeyRejectsCustomKeyOnGeneratedStoredColumnWith
...
+		3*time.Minute, 2*time.Second,
+		"import should exit after streaming an update that mutates a custom partition key column")
+
+	output := lm.GetImportCommandStderr() + lm.GetImportCommandStdout()
+	require.Contains(t, output, "is required to be immutable",
+		"expected the custom-key immutability error, got: %s", output)
+	require.Contains(t, output, `"ck2"`,
+		"the immutability error should name the mutated key column, got: %s", output)
+
+	//overrides with pk strategy but table has a unique index on a stored generated column
```
makalaaneesh: [AI][Warning] **Stale copy-pasted comment above start-clean restart block.**

The comment `//overrides with pk strategy but table has a unique index on a stored generated column` sits above the second `StartImportData` call, but that call has no `pk` override, the table has no unique index, and there is no generated column. It is copy-pasted from `TestLiveMigrationCdcPartitionKeyRejectsCustomKeyOnGeneratedStoredColumnWithUniqueIndex` and describes code that isn't there.

Root `AGENTS.md` `Comments` explicitly rules this out: "A stale comment is worse than no comment: it reads as authoritative while describing code that no longer exists."

Failure case: a future reader debugging this test edits the wrong flags or wastes time wondering why the `pk` path is not being exercised.

Suggested fix: delete the stale comment, or replace it with a one-liner describing what the restart actually does (e.g. `// Restart import fresh, without the custom-key override, so the mutation replay streams cleanly under default routing.`).

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -3494,3 +3494,285 @@ func TestLiveMigrationCdcPartitionKeyRejectsCustomKeyOnGeneratedStoredColumnWith
...
+		`"test_schema"."test_live_multi"`:  {Inserts: 500, Updates: 1000, Deletes: 500},
+	}, 200, 5)
+	testutils.FatalIfError(t, err, "failed to wait for forward streaming complete")
+
+	require.False(t, lm.GetImportRunner().IsStopped(),
+		"import should keep running during count failpoint mode")
+
+	conflictStats, err := testutils.ReadUniqueKeyConflictStats(uniqueKeyConflictStatsPath)
+	testutils.FatalIfError(t, err, "failed to read unique key conflict stats")
+	require.Greater(t, conflictStats.Total, 0, "custom-routed UK delta should produce conflicts")
```
makalaaneesh, line 3747: [AI][Warning] **Vacuous `>0` lower bound on unique-key conflict count.**

`require.Greater(t, conflictStats.Total, 0, ...)` and the two per-table `Greater(..., 0)` lower bounds are asserted on a workload the test itself upper-bounds at 2000 (2 × 500 catchable events × 2 tables). A `>0` lower bound is exactly the vacuous pattern that `yb-voyager/AGENTS.md` (Test Determinism and Flakiness) and `yb-voyager/src/testlivemigration/AGENTS.md` (Assertions) call out: "a vacuous lower bound (`>= 0`, `>= 1` on a count expected in the hundreds) asserts nothing."

Failure case: the custom-key exclusion regresses and suppresses all but a handful of UK conflicts on either table (or on the total); conflict detection is effectively broken for the case this test exists to pin, but a single stray conflict still passes `>0`.

Suggested fix: derive a justified per-table lower bound from the delta and use `require.GreaterOrEqual(t, conflictStats.ByTable[...], N, ...)` where `N` is a defensible fraction of the 1000-per-table upper bound (e.g. a few hundred), and update `Total` to `>= 2*N`. If truly timing-dependent, at minimum use `Greater(..., 100)` per table so a 99% suppression regression fails.

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -3494,3 +3494,285 @@ func TestLiveMigrationCdcPartitionKeyRejectsCustomKeyOnGeneratedStoredColumnWith
...
+		"import should exit after streaming an update that mutates a custom partition key column")
+
+	output := lm.GetImportCommandStderr() + lm.GetImportCommandStdout()
+	require.Contains(t, output, "is required to be immutable",
+		"expected the custom-key immutability error, got: %s", output)
+	require.Contains(t, output, `"ck2"`,
+		"the immutability error should name the mutated key column, got: %s", output)
+
+	//overrides with pk strategy but table has a unique index on a stored generated column
+	err = lm.StartImportData(true, map[string]string{
```
makalaaneesh, line 3579: [AI][Suggestion] **Second import run implicitly relies on `--start-clean` wiping the persisted custom-key overrides.**

After the mutation-driven abort, the restart passes only `{--start-clean, --truncate-tables}` with no `--cdc-partition-key-overrides`. Correctness here depends on `--start-clean` also erasing `TableToCDCPartitionKey`; otherwise the second run reuses the same `PARTITION_BY_CUSTOM` strategy from the metaDB, hits the same immutability guard, and the second `WaitForForwardStreamingComplete` fails. Nothing in the test asserts that erasure.

Failure case: a future refactor keeps `TableToCDCPartitionKey` across `--start-clean`; this test now fails at streaming with the same immutability error, and it looks like a regression of the guard rather than a regression of the `--start-clean` wipe semantics.

Suggested fix: make the routing intent explicit — either pass `"--cdc-partition-key": "auto"` on the restart, or assert after the restart that the persisted per-table row is cleared or overwritten (e.g. `assert.NotEqual(t, cmd.PARTITION_BY_CUSTOM, ...)` / `assert.Empty(t, importDataStatus.TableToCDCPartitionKey[...].Columns)`).

---

yb-voyager/src/testlivemigration/live_migration_testing_framework.go
```diff
@@ -899,7 +899,7 @@ func (lm *LiveMigrationTest) WaitForSnapshotComplete(expectedData map[string]int
 	ok := utils.RetryWorkWithTimeout(1, snapshotTimeout, func() bool {
 		ok, err := lm.snapshotPhaseCompleted(expectedData)
 		if err != nil {
-			testutils.FatalIfError(lm.t, err, "failed to get data migration report")
+			lm.t.Logf("failed to get snapshot data report: %v", err)
```
makalaaneesh, line 913: [AI][Suggestion] **Retry callbacks now swallow real inner errors behind a generic timeout.**

In `WaitForSnapshotComplete` / `WaitForForwardStreamingComplete` (and pre-existing `WaitForFallbackStreamingComplete`) the inner error from `snapshotPhaseCompleted` / `streamingPhaseCompleted` is now logged via `t.Logf` and the retry continues. The outer helper only ever returns `"phase did not complete within Xs"`, so a genuine bug (e.g. a name-tuple key mismatch) surfaces as a timeout rather than as the underlying error.

Root `AGENTS.md` tells authors to fail loudly on unexpected state; this change drops that loud-failure guarantee.

Failure case: `snapshotPhaseCompleted` starts returning a real error on every call (typo in a table key). All three sisters silently retry until timeout; the test fails as "did not complete within Xs" and the operator has to dig through `t.Logf` output to find the actual cause.

Suggested fix: if the retry is intended to tolerate transient errors, add a one-liner doc-comment stating that contract on each helper (`// Retry callback treats inner errors as transient: log and keep polling.`). Better still, if the inner helpers can distinguish `phase-not-ready` from `real-failure`, propagate the real failure out of the retry loop.

---

makalaaneesh (commented):
[AI] Automated review — 4 findings (0 critical, 2 warning, 2 suggestion). See inline comments.

---

yb-voyager/src/testlivemigration/live_migration_testing_framework.go
makalaaneesh: out of scope?

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -3494,3 +3494,285 @@ func TestLiveMigrationCdcPartitionKeyRejectsCustomKeyOnGeneratedStoredColumnWith
...
+		`"test_schema"."test_live"`: 5,
+	}, 120)
+	testutils.FatalIfError(t, err, "failed to wait for snapshot complete")
+
+	err = lm.ExecuteSourceDelta()
+	testutils.FatalIfError(t, err, "failed to execute source delta")
+
+	// The mutation event must kill the import: hashEvent -> customPartitionKeyColumnValue
+	// errors, streamChanges propagates it, and the process ErrExits.
+	require.Eventually(t, func() bool { return lm.GetImportRunner().IsStopped() },
```
makalaaneesh: nit: a helper here in the framework could prove handy. Something like `WaitForImportDataToCrashWithError(...)`

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
makalaaneesh: this doesn't make sense. This is fine.

priyanshi-yb: yeah

yb-voyager/src/testlivemigration/live_migration_testing_framework.go
priyanshi-yb: yeah

makalaaneesh (approved):
LGTM

PR #3802, opened 2026-09-11
Correct sequence restoration when the exported value lags the data

Abhi-yb:

At cutover, sequences are restored by calling `setval()` with a value read from `export_status.json`. That value is a running maximum of the values seen in the change stream, maintained by the exporter. It can lag the data that was actually migrated — if the last events before cutover are not folded into it before it is read, `setval()` leaves the sequence at or below a value already present in the table. The next insert that relies on the sequence default then fails with a duplicate key error.

**Fix:** restore to `GREATEST(exported, max value present in the column(s) the sequence feeds)` instead of blindly trusting the exported value, so restoration is correct regardless of *why* the exported value lagged. Applies to both the live and offline restore paths, and therefore to all target types, with no `tgtdb` changes.

The correction is never silent:
- it logs a warning naming the sequence, the exported value and the real max;
- `YB_VOYAGER_STRICT_SEQUENCE_RESTORE` turns it into a hard failure, so a lag can be caught in tests rather than absorbed.

The lookups this needs are **best-effort** — an unresolvable name, a non-numeric column or an unavailable table only costs the max-value check for that column and falls back to the exported value. This runs for every migration, so failing there would turn a working migration into a cutover failure.

**Test coverage** is added in the resumption framework rather than per scenario: each cutover records the role it switched to, and `row_hash_validations` drains those roles once the hashes are compared, inserting into `cutover_table` and letting the `SERIAL` default supply the id. Every other sequence-backed table is populated by the event generator with random ids and retry-on-conflict, which hides a sequence restored behind the data; `cutover_table` only ever receives these honest inserts, so its ids stay dense and a stale sequence collides immediately. Draining *after* the hashes matters — the insert adds a row on one side only. No scenario files change, so every resumption test (forward, fall-back and fall-forward) picks this up automatically.

- **New opt-in env var `YB_VOYAGER_STRICT_SEQUENCE_RESTORE`** (default off). When enabled, a lagging exported sequence value fails the import instead of being corrected with a warning. Intended for tests/CI.
- **New warning** during sequence restoration when a correction is applied, naming the sequence and both values. Previously this situation produced no diagnostic at all — it surfaced later as an unexplained duplicate key error.
- No command line, configuration, installation or report changes.

**Unit** — added `importDataSequenceRestoration_test.go` covering identifier quoting. `go build`, `go vet -tags unit` and `gofmt` are clean.

**Deterministic before/after (offline migration, lag forced by editing `postdata.sql` from the true `10` down to `3`, 10 rows imported):**

| | with fix | without fix |
| --- | --- | --- |
| warning | `exported last value 3 is behind the maximum value 10 … restoring it to 10 instead` | *(none)* |
| sequence after restore | **10** | **3** |
| insert relying on the default | **id=11**, succeeds | `ERROR: duplicate key value violates unique constraint "probe_tbl_pkey"` |

**End-to-end** — `partition-iteration-test` (live migration with fall-back, resumptions enabled) passed: 52 stages OK, 0 failed. Both framework probes fired and succeeded, on the target after cutover-to-target and on the source after cutover-to-source:

```
  1 | Last event before cutover
  2 | verify sequence restored after cutover   <- target probe
102 | Last event before cutover
103 | verify sequence restored after cutover   <- source probe
```

No `skipping max value check` warnings appeared, confirming every sequence's name resolved and every `MAX(col)` query ran — i.e. the correction path executes rather than silently degrading.

**Not reproduced:** the natural race was not reproduced on demand. The exported value was accurate in the end-to-end runs, so the correction did not need to engage there. The original failure is documented in `aa37d091` ("Fix duplicate key in cutover_table during fallback marker insert"), whose workaround — truncating `cutover_table` before the fallback marker insert — was removed in `f2b19fb4` on review, leaving the underlying behaviour unguarded. The fix stands on its own merit regardless: trusting a value that can lag the data is incorrect even if the window is narrow.

No changes to callhome or yugabyted payloads.

No. `export_status.json` is read, not written, and its shape is unchanged. No changes to MetaDB, the name registry, or any other on-disk structure.

### DIFF
--- a/migtests/scripts/live-migration-resumption/orchestrator.py
+++ b/migtests/scripts/live-migration-resumption/orchestrator.py
@@ -254,17 +254,20 @@ def wait_for_action(stage: Dict[str, Any], ctx: Any) -> None:
 @action("cutover_to_target")
 def cutover_to_target_action(_stage, ctx: Any) -> None:
     H.initiate_cutover(ctx.cfg, ctx.env, "target")
+    ctx.pending_sequence_checks.add("target")
 
 
 @action("cutover_to_source")
 def cutover_to_source_action(_stage, ctx: Any) -> None:
     H.initiate_cutover(ctx.cfg, ctx.env, "source")
+    ctx.pending_sequence_checks.add("source")
 
 
 @action("cutover_to_source_replica")
 def cutover_to_source_replica_action(_stage, ctx: Any) -> None:
     """Initiate cutover back to the source-replica database."""
     H.initiate_cutover(ctx.cfg, ctx.env, "source-replica")
+    ctx.pending_sequence_checks.add("source_replica")
 
 
 @action("row_count_validations")
@@ -287,6 +290,13 @@ def row_hash_validations_action(stage: Dict[str, Any], ctx: Any) -> None:
         H.run_sql_file(ctx, sql_path, target=role, use_admin=False)
 
     H.run_segment_hash_validations(ctx, left_role, right_role)
+
+    # A cutover restores sequences; confirm an insert relying on a sequence default
+    # still works. Done here rather than in the cutover action because it adds a row
+    # on one side only, which would fail the hash above.
+    for role in sorted(ctx.pending_sequence_checks):
+        H.verify_sequence_restored(ctx, role)
+    ctx.pending_sequence_checks.clear()
 
 
 def _conflict_log_table_markers(table: str | None) -> list[str] | None:
--- a/yb-voyager/cmd/importDataSequenceRestoration.go
+++ b/yb-voyager/cmd/importDataSequenceRestoration.go
@@ -109,6 +109,11 @@ func restoreSequencesInOfflineMigration(msr *metadb.MigrationStatusRecord, impor
 		}
 	}
 
+	err = correctSequenceLastValuesAgainstTableMax(sequenceNameTupleToLastValueMap)
+	if err != nil {
+		return fmt.Errorf("failed to correct sequence last values: %w", err)
+	}
+
 	//restore the sequence last value
 	err = tdb.RestoreSequences(sequenceNameTupleToLastValueMap)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


yb-voyager/cmd/importDataSequenceRestoration.go
```diff
@@ -131,13 +136,171 @@ func restoreSequencesInLiveMigration(sequenceLastValue map[string]int64) error {
...
+func maxValueAcrossSequenceColumns(sequenceTuple sqlname.NameTuple, ownerColumns []sequenceOwnerColumn) (int64, error) {
+	var maxValue int64
+	for _, ownerColumn := range ownerColumns {
+		if !ownerColumn.table.TargetTableAvailable() {
+			continue
+		}
+		query := fmt.Sprintf(`SELECT COALESCE(MAX(%s), 0) FROM %s`,
+			quoteIdentifierIfUnquoted(ownerColumn.column), ownerColumn.table.ForUserQuery())
+		var columnMax int64
+		if err := tdb.QueryRow(query).Scan(&columnMax); err != nil {
```
makalaaneesh, line 241: [AI][Warning] **`Scan` error swallow hides real failures, contradicting its own docstring.**

`maxValueAcrossSequenceColumns` swallows every error from `tdb.QueryRow(...).Scan(&columnMax)` and logs a warning that says the column is skipped because it is "non-numeric". In reality the same branch also catches `column "X" does not exist`, `relation "..." does not exist`, permission-denied, connection-reset, etc. That both silently disables the correction for legitimate misconfigurations and misleads the reader/log-viewer about *why* the correction was skipped. It also violates the repo-root `AGENTS.md` "Fail loudly on unexpected state" rule ("no silent fallback to a weaker mechanism").

Failure case: Oracle to YB (or any migration whose target column name differs in case from the source-side mapping key, e.g. Oracle key `"SAKILA"."T1".COL` -> we run `SELECT COALESCE(MAX("COL"), 0) FROM "sakila"."t1"` on YB, which errors `column "COL" does not exist`). The correction is silently disabled for every sequence-bound column, and the log line calls it a "non-numeric column" though it is not.

Suggested fix: Narrow the tolerated case to the actually-tolerated one (invalid textual input to bigint), and otherwise return the error so it propagates out of `restoreSequencesIn{Offline,Live}Migration`:

```go
if err := tdb.QueryRow(query).Scan(&columnMax); err != nil {
    if isNonNumericColumnErr(err) { // check pq errcode "22P02" / "22003" etc.
        log.Warnf("column %s of %s of sequence %s is non-numeric; skipping max check: %v",
            ownerColumn.column, ownerColumn.table.ForKey(), sequenceTuple.ForKey(), err)
        continue
    }
    return 0, fmt.Errorf("query max for column %s of %s: %w",
        ownerColumn.column, ownerColumn.table.ForKey(), err)
}
```

---

yb-voyager/cmd/importDataSequenceRestoration.go
```diff
@@ -131,13 +136,171 @@ func restoreSequencesInLiveMigration(sequenceLastValue map[string]int64) error {
...
+}
+
+// fetchSequenceToColumnListMap maps each sequence to the columns it feeds, using the
+// column-to-sequence mapping recorded for the database being imported into.
+func fetchSequenceToColumnListMap(msr *metadb.MigrationStatusRecord) (*utils.StructMap[sqlname.NameTuple, []sequenceOwnerColumn], error) {
+	columnToSequenceMapping := msr.SourceColumnToSequenceMapping
+	if importerRole == TARGET_DB_IMPORTER_ROLE && len(msr.TargetColumnToSequenceMapping) > 0 {
+		columnToSequenceMapping = msr.TargetColumnToSequenceMapping
+	}
+	if len(columnToSequenceMapping) == 0 {
```
makalaaneesh, line 263: [AI][Warning] **`columnToSequenceMapping` fallback chain is ambiguous and can select the wrong mapping.**

The chain is: default to `SourceColumnToSequenceMapping`; if role is `TARGET_DB_IMPORTER_ROLE` **and** `TargetColumnToSequenceMapping` is non-empty, prefer target; finally, "if the mapping so far is empty, use target". The first override fires after an iterative cutover has populated the target mapping (fall-back / fall-forward exporter), silently switching a `TARGET_DB_IMPORTER_ROLE` importer from the source-side mapping to the target-side one - the two are only interchangeable if the target schema is identical to the source schema, which is not a guaranteed invariant across iterations. The last `if len == 0` fallback is unmotivated for `SOURCE_DB_IMPORTER_ROLE` / `SOURCE_REPLICA_DB_IMPORTER_ROLE` and violates the repo-root "one source of truth" rule.

Failure case: An iterative cutover where a table was added/renamed on target after the initial import: on the second cutover-to-target, the target-side mapping is preferred and the source-side mapping is ignored, so any sequence whose source-to-target mapping differs is not compared. Reverse case for source-importer where source mapping is empty and the code silently falls back to the target's mapping, which lists target-only tables that may not exist on source.

Suggested fix: Select strictly by role, without an implicit source-to-target fallback, and comment the intent.

---

yb-voyager/cmd/importDataSequenceRestoration.go
```diff
@@ -131,13 +136,171 @@ func restoreSequencesInLiveMigration(sequenceLastValue map[string]int64) error {
...
+		sequenceNameTupleToLastValueMap.Put(sequenceTuple, tableMax)
+		return true, nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to compare sequence values against the data: %w", err)
+	}
+
+	if strict && len(laggedSequences) > 0 {
+		return goerrors.Errorf("exported sequence values lagged behind the migrated data for %d sequence(s): %s (%s is enabled)",
+			len(laggedSequences), strings.Join(laggedSequences, ", "), StrictSequenceRestoreEnvVar)
```
makalaaneesh, line 224: [AI][Warning] **Strict-mode error message enumerates sequences in Go map order (nondeterministic).**

`laggedSequences` is appended inside `IterKV`, which iterates a Go map with unspecified order, and then `strings.Join(laggedSequences, ", ")` is embedded in the returned error. When the strict env var is enabled (its stated purpose in the docstring is "Tests enable it") any test asserting on the error text will be intermittently flaky, and grep-based CI checks over the error string will match different substrings run-to-run.

Suggested fix: `sort.Strings(laggedSequences)` before joining.

---

yb-voyager/cmd/importDataSequenceRestoration_test.go
```diff
@@ -0,0 +1,54 @@
+//go:build unit
```
makalaaneesh, line 1: [AI][Warning] **New Go logic ships with no behavior test.**

The core of the PR - `correctSequenceLastValuesAgainstTableMax`, `maxValueAcrossSequenceColumns`, `fetchSequenceToColumnListMap` - has no unit coverage. The single test only pins the three-branch behavior of the trivial `quoteIdentifierIfUnquoted`. Per `yb-voyager/AGENTS.md` Testing rules, new logic needs tests that assert behavior, must include cases for case-sensitive names, and should exercise the exported entry point. In particular the interesting corners here (lastValue==0 skip, tableMax==lastValue no-op, tableMax>lastValue raises value + warning, strict mode error text, non-3-part column-key skip, unresolved-table skip) are all reachable without a live DB by factoring `tdb.QueryRow` behind a small interface.

Suggested fix: Add a table-driven test around `correctSequenceLastValuesAgainstTableMax` with the DB dependency injected covering: (1) lastValue == 0; (2) tableMax <= lastValue; (3) tableMax > lastValue; (4) strict mode + multiple lagged sequences (sorted); (5) case-sensitive column name; (6) malformed 3-part key.

---

yb-voyager/cmd/importDataSequenceRestoration.go
```diff
@@ -131,13 +136,171 @@ func restoreSequencesInLiveMigration(sequenceLastValue map[string]int64) error {
...
 		return fmt.Errorf("failed to restore sequences: %w", err)
 	}
 	return nil
 }
 
+// StrictSequenceRestoreEnvVar, when enabled, turns the "exported sequence value lagged
+// behind the data" correction below into a hard failure instead of a warning. Tests
+// enable it so that a regression in the exported sequence max is caught loudly instead
+// of being silently absorbed by the correction.
+const StrictSequenceRestoreEnvVar = "YB_VOYAGER_STRICT_SEQUENCE_RESTORE"
```
makalaaneesh, line 154: [AI][Warning] **`StrictSequenceRestoreEnvVar` says "Tests enable it" but no test does.**

The docstring claims the env var is enabled by tests so a regression is caught loudly, but a grep across the repo finds no test or CI configuration that sets `YB_VOYAGER_STRICT_SEQUENCE_RESTORE=true`. Either the env is dead surface (YAGNI per root `AGENTS.md`) or a companion test change is missing.

Suggested fix: Either export the env in the live-migration-resumption env (e.g. in `orchestrator.py` or the Jenkins env), or soften the docstring to describe when the operator/tester is expected to enable it, and add a Go unit test that flips it via `t.Setenv`.

---

yb-voyager/cmd/importDataSequenceRestoration.go
```diff
@@ -131,13 +136,171 @@ func restoreSequencesInLiveMigration(sequenceLastValue map[string]int64) error {
...
+			continue
+		}
+		ownerColumn := sequenceOwnerColumn{table: tableNameTuple, column: parts[2]}
+		ownerColumns, _ := sequenceToColumns.Get(sequenceTuple)
+		sequenceToColumns.Put(sequenceTuple, append(ownerColumns, ownerColumn))
+	}
+	return sequenceToColumns, nil
+}
+
+func quoteIdentifierIfUnquoted(identifier string) string {
```
makalaaneesh, line 297: [AI][Suggestion] **`quoteIdentifierIfUnquoted` belongs alongside the rest of `sqlname`.**

Per `yb-voyager/AGENTS.md` ("The cmd/ package should ideally contain only command-handling logic... Core business logic should reside in dedicated packages under src/") and the "Object Names" rule, a generic PG identifier-quoter is out of place in `cmd/`. `sqlname` already has related quoters (`quote2`, `minQuote2`, `IsQuoted`) - put it there so future callers can find it and its case-preservation semantics are unified with the rest.

Suggested fix: Move the helper to `yb-voyager/src/utils/sqlname/sqlname.go` and call it as `sqlname.QuoteIfUnquoted(parts[2])`.

---

yb-voyager/cmd/importDataSequenceRestoration.go
```diff
@@ -131,13 +136,171 @@ func restoreSequencesInLiveMigration(sequenceLastValue map[string]int64) error {
...
+			continue
+		}
+		query := fmt.Sprintf(`SELECT COALESCE(MAX(%s), 0) FROM %s`,
+			quoteIdentifierIfUnquoted(ownerColumn.column), ownerColumn.table.ForUserQuery())
+		var columnMax int64
+		if err := tdb.QueryRow(query).Scan(&columnMax); err != nil {
+			// A non-numeric column fed by a sequence (for example a text column whose
+			// default embeds nextval) cannot be compared; skip it rather than failing
+			// the whole restoration.
+			log.Warnf("skipping max value check for column %s of sequence %s: %v",
```
makalaaneesh, line 245: [AI][Suggestion] **Warning lines omit the table name.**

Per `yb-voyager/AGENTS.md` Logging ("Include the table name, file path, or operation context in log messages"), the "skipping max value check for column X of sequence Y" line does not identify the table (e.g. multiple tables share a column name `id`). Same for the two `Warnf`s in `fetchSequenceToColumnListMap` - the operator sees "skipping ... for sequence <x>" without knowing which table's mapping triggered it.

Suggested fix: Include `ownerColumn.table.ForKey()` (or `tableName`) in each of the three warning messages.

---

yb-voyager/cmd/importDataSequenceRestoration.go
```diff
@@ -131,13 +136,171 @@ func restoreSequencesInLiveMigration(sequenceLastValue map[string]int64) error {
...
+		tableMax, err := maxValueAcrossSequenceColumns(sequenceTuple, ownerColumns)
+		if err != nil {
+			return false, err
+		}
+		if tableMax <= lastValue {
+			return true, nil
+		}
+		laggedSequences = append(laggedSequences,
+			fmt.Sprintf("%s (exported=%d, data max=%d)", sequenceTuple.ForKey(), lastValue, tableMax))
+		utils.PrintAndLogfWarning("sequence %s: exported last value %d is behind the maximum value %d already present in the data; "+
```
makalaaneesh, line 212: [AI][Suggestion] **Warning message repeats the same corrected value twice.**

The line already says "maximum value %d already present in the data" and then repeats it as "restoring it to %d instead". Slightly noisy; the reader has already been told the number.

Suggested fix:

```go
utils.PrintAndLogfWarning("sequence %s: exported last value %d is behind the maximum value %d already present in the data; restoring the sequence to that maximum instead.",
    sequenceTuple.ForKey(), lastValue, tableMax)
```

---

yb-voyager/cmd/importDataSequenceRestoration.go
```diff
@@ -131,13 +136,171 @@ func restoreSequencesInLiveMigration(sequenceLastValue map[string]int64) error {
...
+	}
+	sequenceToColumns, err := fetchSequenceToColumnListMap(msr)
+	if err != nil {
+		return fmt.Errorf("failed to fetch sequence to column list map: %w", err)
+	}
+
+	strict := utils.GetEnvAsBool(StrictSequenceRestoreEnvVar, false)
+	var laggedSequences []string
+
+	err = sequenceNameTupleToLastValueMap.IterKV(func(sequenceTuple sqlname.NameTuple, lastValue int64) (bool, error) {
```
makalaaneesh, line 194: [AI][Suggestion] **Use `IterKVSorted` for a deterministic pass.**

The correction visits sequences in Go map order, which means warning-log ordering (and, once the strict-mode message is sorted per W3, the `Put` ordering inside `sequenceNameTupleToLastValueMap`) varies run-to-run. `StructMap.IterKVSorted` exists in `src/utils/struct_map.go` and gives a stable iteration order at trivial cost since sequence maps are small.

Suggested fix: Use `IterKVSorted` with a `NameTuple.ForKey()` comparator.

---

migtests/scripts/live-migration-resumption/orchestrator.py
```diff
@@ -288,6 +291,13 @@ def row_hash_validations_action(stage: Dict[str, Any], ctx: Any) -> None:
 
     H.run_segment_hash_validations(ctx, left_role, right_role)
 
+    # A cutover restores sequences; confirm an insert relying on a sequence default
```
makalaaneesh, line 294: [AI][Suggestion] **`verify_sequence_restored` only runs when a hash-validation action is queued after the cutover.**

If a scenario cuts over and then triggers only `row_count_validations` (not `row_hash_validations`), the enqueued sequence check is never drained; if the scenario ends before the next hash validation, it is silently skipped.

Suggested fix: Also drain (and clear) `ctx.pending_sequence_checks` inside `row_count_validations_action`, or add a first-class `verify_sequence_restored` action to `orchestrator.py`.

---

makalaaneesh (commented):
[AI] Automated review — 10 findings (0 critical, 5 warning, 5 suggestion). See inline comments.

---

yb-voyager/cmd/importDataSequenceRestoration.go
```diff
@@ -131,13 +138,190 @@ func restoreSequencesInLiveMigration(sequenceLastValue map[string]int64) error {
...
+		columnMax, isInteger := parseSequenceColumnMax(raw)
+		if !isInteger {
+			log.Warnf("column %s of table %s feeding sequence %s does not hold integer values (max is %q); skipping it",
+				ownerColumn.column, ownerColumn.table.ForKey(), sequenceTuple.ForKey(), raw)
+			return 0, false
+		}
+		return columnMax, true
+	}
+
+	log.Warnf("could not read the maximum of column %s of table %s feeding sequence %s, so its exported value is used as is: %v",
```
makalaaneesh, line 279: [AI][Warning] **Max-value read silently swallows *all* query errors, re-opening the bug this PR fixes.**

In `maxValueOfSequenceColumn` (lines 262–281) the loop treats *any* error from `sequenceColumnMaxQuerier` as "try the next spelling, then give up" and returns `(0, false)`. Only `column-does-not-exist` (SQLSTATE `42703`) is the expected reason to try the other spelling. Every other error — permission denied, a dropped connection, an uncastable type, a transient failure — is also swallowed. When that happens the column's `tableMax` stays `0`, so `correctSequenceLastValuesAgainstTableMax` skips the correction and restores the *exported* (possibly lagging) value — exactly the duplicate-key scenario this PR set out to prevent, now reintroduced silently.

It also defeats `YB_VOYAGER_STRICT_SEQUENCE_RESTORE`: `laggedSequences` never records the sequence (tableMax `0` ≤ lastValue), so even strict mode (used by the new tests) will not catch the regression.

Root `CLAUDE.md` → "Fail loudly on unexpected state": *"Do not silently fall back to a weaker mechanism … Either the fallback is a designed, documented behavior, or the missing input is an error."* The `42703` fallback is designed; other errors are not, and conflating them hides real failures.

Suggested fix: only fall through to the next spelling / skip on undefined-column; surface other errors so the caller (which already returns `error`) fails loudly. For example detect `42703` explicitly and `break` (propagating the last error) on anything else, then have `maxValueOfSequenceColumn` return that non-`42703` error rather than logging and returning `(0, false)`. Add a test asserting a table-missing / permission error is *not* swallowed.

---

yb-voyager/cmd/importDataSequenceRestoration.go
```diff
@@ -131,13 +138,190 @@ func restoreSequencesInLiveMigration(sequenceLastValue map[string]int64) error {
...
+			"restoring the sequence to that maximum instead.", sequenceTuple.ForKey(), lastValue, tableMax)
+		sequenceNameTupleToLastValueMap.Put(sequenceTuple, tableMax)
+		return true, nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to compare sequence values against the data: %w", err)
+	}
+
+	if utils.GetEnvAsBool(StrictSequenceRestoreEnvVar, false) && len(laggedSequences) > 0 {
+		sort.Strings(laggedSequences)
```
makalaaneesh, line 214: [AI][Suggestion] **`sort.Strings(laggedSequences)` is redundant.**

`laggedSequences` is appended inside `IterKVSorted(byKey, ...)`, where `byKey` sorts by `ForKey()`, and each entry is prefixed with `sequenceTuple.ForKey()`. The slice is therefore already in `ForKey` (lexicographic) order, so this extra `sort.Strings` never reorders anything.

Suggested change: drop the `sort.Strings(laggedSequences)` line for clarity (or keep it only as an explicitly-noted defensive guard). Harmless either way.

---

yb-voyager/cmd/importDataSequenceRestoration.go
```diff
@@ -131,13 +138,190 @@ func restoreSequencesInLiveMigration(sequenceLastValue map[string]int64) error {
...
+		}
+		columnMax, ok := maxValueOfSequenceColumn(sequenceTuple, ownerColumn)
+		if ok && columnMax > maxValue {
+			maxValue = columnMax
+		}
+	}
+	return maxValue
+}
+
+/*
```
makalaaneesh, line 242: [AI][Suggestion] **Multi-paragraph doc comments exceed the comment-length guidance.**

The doc blocks on `correctSequenceLastValuesAgainstTableMax` and `maxValueOfSequenceColumn` (242–255) run several paragraphs. The content is legitimate non-obvious "why" (the name registry does not map columns; cross-source spelling), which root `CLAUDE.md` permits — but that same file says: *"A surviving comment is one or two lines. If it needs a paragraph, the explanation is wrong or the code is."*

Suggested change: trim each block to the load-bearing sentence (why both spellings are attempted; why a lagging value is corrected) and move the rest to the PR description. E.g. reduce `maxValueOfSequenceColumn`'s comment to ~2 lines.

---

makalaaneesh (commented):
[AI] Automated review (incremental re-review of new commits since the last AI review) — 3 findings (0 critical, 1 warning, 2 suggestion). See inline comments.

---

PR #3812, opened 2026-09-16
schema/schemadrift: drift model, assembler, and classification

ShivanshGahlot:

Adds the package that owns schema drift above the diff engine. `schemadiff` compares two snapshots and stops; this decides which snapshots can be compared and what each change means for the migration in flight. Library only — no CLI, no output formats.

- **Model** — `Report`, `DriftEntry`, `CapturePoint`, `DriftInfo`. JSON-serialisable, and the shape is contractual.
- **Assembler** — walks the snapshots oldest-first and diffs each usable pair.
- **Bridging** — a failed capture, or one that did not cover every requested schema, is skipped without becoming a boundary, so the comparison reaches back past it. Both are recorded on their `CapturePoint` so the gap stays visible.
- **Coverage, not equality** — a capture covering *more* than was asked for is compared normally, and the extra schemas' findings are filtered by `Scope.Schemas`. Requiring equal schema sets rejected comparable pairs and lost whole intervals.
- **Phase** — labels each interval with what the migration was doing then, from the pair of snapshot labels.
- **Classification** — severity plus the Impact & action note in one map per change type. They have to agree: parallel maps keyed by the same `DiffType` let them drift apart silently. Severity answers "what does the migration do about this", not "how alarming is the DDL".
- **`ComparedIntervalCount`** — how many intervals were actually diffed. Zero changes over zero intervals is not a clean run, it is one that examined nothing, and `StoredCaptureCount` cannot say so because a failed capture still counts toward it.
- **`LiveCompared`** means the live read was actually diffed, not merely that one was taken.
- **`LabelSourceLive`** added to `schemasnapshot`. The live read's label is its timeline identity, so nothing carries a second one.

Renderers are the next PR in the stack.

No user-facing changes.

`go test -tags unit ./src/schema/schemadrift/` — 21 tests covering the walk, bridging, schema coverage, interval counting, phase mapping, and classification for all 16 change types. Includes a test that every classified type carries non-empty Impact and Action.

No.

No. The report is an output file, not persisted migration state, and `LabelSourceLive` is never written.


yb-voyager/src/schema/schemadrift/build.go
```diff
@@ -0,0 +1,217 @@
...
+			Schemas:             p.Schemas,
+			Tables:              p.Tables,
+			TablesFiltered:      p.TablesFiltered,
+			ObjectTypes:         p.ObjectTypes,
+			ObjectTypesFiltered: p.ObjectTypesFiltered,
+		},
+		Summary: Summary{
+			ChangeCount:  len(diffs),
+			CaptureCount: lo.CountBy(p.Snapshots, func(s SnapshotInput) bool { return s.Series != SeriesSourceLive }),
+			LiveCompared: lo.ContainsBy(p.Snapshots, func(s SnapshotInput) bool { return s.Series == SeriesSourceLive }),
```
makalaaneesh: [AI][Warning] **`Summary.LiveCompared` is computed from presence, not from an actual comparison.**

`LiveCompared` is set via `lo.ContainsBy(p.Snapshots, s.Series == SeriesSourceLive)` — true whenever a live snapshot is *present* in the input, regardless of whether it was actually diffed. If the live pair is schema-scope-skipped (its `Header.Schemas` differ from the preceding capture — the skip branch at lines 92–101), or the live read is the only content-bearing snapshot (nothing precedes it), no comparison happens, yet `LiveCompared` still reports `true`. In the skip case the report even carries a `SkippedInterval` for that same live pair, contradicting the flag.

Root `CLAUDE.md`: *"One source of truth. Do not persist the same fact two ways"* and *"Avoid in-band sentinel values … must not carry two different meanings."* The flag claims a fact it did not verify.

Suggested fix: track whether a live pair was actually diffed inside the loop (set a bool when a comparison with `next.Series == SeriesSourceLive` runs) and use that; or, if the field is meant to signal "a live read was taken", rename it (e.g. `LiveCaptured`) so the name matches the computation.

---

yb-voyager/src/schema/schemadrift/build.go
```diff
@@ -0,0 +1,217 @@
...
+// splitIdentity maps a table-scoped identity (a column) to its parent table as
+// Object and its own name as SubObject; a table maps to Object alone.
+func splitIdentity(id schemadiff.ObjectIdent) (obj schemasnapshot.ObjectRef, subObject string) {
+	switch it := id.(type) {
+	case schemasnapshot.ObjectRef:
+		return it, ""
+	case schemasnapshot.TableScopedObjectRef:
+		return it.Table, it.Name
+	default:
+		return schemasnapshot.ObjectRef{}, ""
```
makalaaneesh, line 209: [AI][Warning] **`splitIdentity` default branch silently emits an empty identity instead of failing loudly.**

The `default:` case of the type switch returns `schemasnapshot.ObjectRef{}, ""` for any `ObjectIdent` that is neither `ObjectRef` nor `TableScopedObjectRef` (and for a nil identity, which `displayIdentity` can return if both `ObjectA` and `ObjectB` are nil). This produces a `DiffEntry` with an all-empty `Object` (schema `""`, name `""`) with no signal that anything went wrong. `difftypes.go` already lists INDEX/SEQUENCE/VIEW/FUNCTION/TYPE as object types to be re-enabled; when a new `ObjectIdent` implementation lands, its findings will silently get empty objects until someone remembers to update this switch.

Root `CLAUDE.md` → "Fail loudly on unexpected state": *"When code reaches a state that 'shouldn't happen', return an error — do not … silently skip the work"*, and no in-band sentinel values.

Suggested fix: make the "shouldn't happen" case loud — since `BuildReport` returns no error, `panic(fmt.Sprintf("schemadrift: unexpected ObjectIdent type %T", id))` — or thread an error return through `splitIdentity`/`BuildReport`.

---

yb-voyager/src/schema/schemadrift/build.go
```diff
@@ -0,0 +1,217 @@
...
+		return "since last capture"
+	default:
+		return ""
+	}
+}
+
+// isExportDataRunningStart reports whether series is one of the two labels
+// that may precede a LabelExportDataFromSourcePeriodic capture while export
+// data is running: the initial start capture, or a prior periodic capture.
+func isExportDataRunningEnd(series string) bool {
```
makalaaneesh, line 183: [AI][Warning] **Doc comment sits on the wrong function.**

The doc comment at lines 179–181 begins *"isExportDataRunningStart reports whether series is one of the two labels that may precede a LabelExportDataFromSourcePeriodic capture … the initial start capture, or a prior periodic capture"* — describing Start-or-Periodic — but it is attached to `func isExportDataRunningEnd`, whose body checks `Periodic || Exit`. The correctly-named `isExportDataRunningStart` (line 186), whose body actually is `Start || Periodic`, has no doc comment. The comment describes the sibling function, and a Go doc comment must start with the name of the function it documents.

Root `CLAUDE.md` → §Comments: *"A stale comment is worse than no comment: it reads as authoritative while describing code that no longer exists."*

Suggested fix: move the existing comment onto `isExportDataRunningStart` and write a correct one for `isExportDataRunningEnd` (periodic or exit label ends the export-data span).

---

yb-voyager/src/schema/schemadrift/build.go
```diff
@@ -0,0 +1,217 @@
...
+)
+
+// SnapshotInput is one point in the chronological sequence of schemas fed
+// into BuildReport: either a stored snapshot or the live read of the source.
+type SnapshotInput struct {
+	Header schemasnapshot.SnapshotHeader
+	// nil for a failed capture: still a Capture on the timeline, never diffed.
+	Content *schemasnapshot.SnapshotContent
+	// Series is the timeline identity phaseFor and the renderers key off, and
+	// is not always Header.Label: the live read borrows LabelDetectDrift only to
```
makalaaneesh: [AI][Warning] **Comment references an undefined `LabelDetectDrift` and contradicts the actual live-read usage.**

The `SnapshotInput.Series` comment says *"the live read borrows `LabelDetectDrift` only to satisfy label validation, so its Series is `SeriesSourceLive` instead."* `LabelDetectDrift` is not defined anywhere in the codebase — only the four labels in `schemasnapshot/labels.go` exist, and `ValidateLabelReason` would reject any other. The live-read fixtures in `build_test.go` (e.g. lines 239, 269, 307) set the header `Label` to `""`, not to any borrowed label — so the comment does not match how the code is actually exercised.

Root `CLAUDE.md` → §Comments: a dangling reference reads as authoritative but cannot be resolved.

Suggested fix: if this refers to a caller/label landing in a later stacked PR, defer the sentence until that symbol exists; otherwise describe the actual mechanism (live header label is empty; `Series` is set to `SeriesSourceLive`).

---

yb-voyager/src/schema/schemadrift/classification.go
```diff
@@ -0,0 +1,136 @@
...
+	StatusPotentialImpact     Status = "potential_impact"
+	StatusBreaksRecoverable   Status = "breaks_migration_recoverable"
+	StatusBreaksUnrecoverable Status = "breaks_migration_unrecoverable"
+)
+
+// classification is everything the report says about a kind of change: how the
+// migration is affected, and the "Impact & action" note explaining it. Severity and
+// wording have to agree, so they are declared together rather than in parallel maps.
+//
+// `backticks` become inline code in HTML (see codeSpans) and stay literal in JSON.
```
makalaaneesh, line 39: [AI][Warning] **"see codeSpans" points to a function that does not exist.**

The `classification` doc says *"backticks become inline code in HTML (see codeSpans) and stay literal in JSON."* There is no `codeSpans` anywhere in the repo, and this package contains no HTML/JSON renderer at all (no `embed`, no template, no `Render*`). The reference dangles.

Root `CLAUDE.md` → §Comments: a forward/dangling reference reads as authoritative but cannot be resolved.

Suggested fix: drop the "(see codeSpans)" pointer and the HTML/JSON-rendering claim until the renderer lands, or move this note into the PR that introduces `codeSpans`.

---

yb-voyager/src/schema/schemadrift/report.go
```diff
@@ -0,0 +1,122 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package schemadrift converts a sequence of captured schema snapshots (plus
+// optionally one live snapshot) into a "schema drift" report: a
+// JSON-serializable model, and JSON/HTML renderers for it.
```
makalaaneesh, line 17: [AI][Suggestion] **Package doc describes renderers/HTML template that are not in this PR.**

The package doc says the package produces *"a JSON-serializable model, and JSON/HTML renderers for it"* (line 17) and *"nothing here does I/O beyond the embedded HTML template"* (line 20). This PR contains only the model, assembler, and classification — there is no renderer, no `embed`, and no HTML template. `build.go:32` similarly refers to "the renderers key off". These over-describe the current package.

Suggested fix: scope the package doc to what this PR ships (model + assembler + classification) and add the renderer/template sentences in the PR that introduces them. (Same theme as the `codeSpans` / `LabelDetectDrift` forward references.)

---

yb-voyager/src/schema/schemadrift/build_test.go
```diff
@@ -0,0 +1,474 @@
...
+	a := fixtureContent(fixtureTable("1", "public", "orders"))
+	b := fixtureContent(
+		fixtureTable("1", "public", "orders"),
+		fixtureTable("2", "public", "customers"),
+	)
+
+	p := BuildParams{
+		Snapshots: []SnapshotInput{
+			{Header: fixtureHeader(schemasnapshot.LabelExportSchema, t1(), "public"), Content: a, Series: schemasnapshot.LabelExportSchema},
+			{Header: fixtureHeader(schemasnapshot.LabelExportDataFromSourceStart, t2(), "other"), Content: b, Series: schemasnapshot.LabelExportDataFromSourceStart},
```
makalaaneesh: [AI][Suggestion] **No test exercises a DROP/rename path (`displayIdentity` ObjectA fallback) or a column-scoped SubObject through a change.**

Every fixture in `build_test.go` produces `TABLE_ADDED` / `COLUMN_ADDED` findings, so `ObjectB` is always non-nil. The `displayIdentity` fallback `return d.ObjectA` (build.go:195, taken for `*_DROPPED` where `ObjectB` is nil) and `splitIdentity`'s `TableScopedObjectRef` → `(Table, Name)` mapping for a *dropped/renamed* column are never covered.

Suggested fix: add a case that drops a table/column between two snapshots and assert the resulting `DiffEntry.Object` / `SubObject` (and `Status`) come out correct via the `ObjectA` path.

---

makalaaneesh (commented):
[AI] Automated review — 7 findings (0 critical, 5 warning, 2 suggestion). See inline comments.

---

yb-voyager/src/schema/schemadrift/build.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	return Report{
+		Report:      "schema_drift",
+		Version:     1,
+		GeneratedAt: time.Now().UTC(),
+		Source:      p.Source,
+		Window:      reportWindow,
+		Comparing: Comparing{
+			Schemas: p.Scope.Schemas,
```
makalaaneesh: [AI][Suggestion] **Schema names rendered raw while table names use engine-aware quoting.**

`Comparing.Schemas` is populated with `p.Scope.Schemas` (raw catalog strings), while `Comparing.Tables` on the following lines renders each ref through `r.ForDisplay(p.Source.DatabaseType)`. The same raw-string schemas are also joined verbatim into the `Excluded` reason (build.go:82). For a case-sensitive schema (e.g. `MixedCase`), the report shows the schema unquoted in `Comparing.Schemas`/exclusion reasons but the table fully-qualified-and-quoted, so the two halves of one report disagree on identifier rendering.

Failure case: a run over a quoted/mixed-case schema produces a report where `comparing.schemas` lists `MixedCase` but `comparing.tables` lists `"MixedCase"."orders"`.

Root `CLAUDE.md` (Case-sensitive identifiers) wants user-facing identifiers rendered consistently through the sqlname machinery.

Suggested fix: render each schema through the same display path used for tables (e.g. a schema-only display helper), so `Comparing.Schemas` and the exclusion reason are quoted consistently with `Comparing.Tables`. If raw is deliberate (because `Scope.Schemas` matches on raw catalog names), document that choice.

---

yb-voyager/src/schema/schemadrift/build.go
```diff
@@ -0,0 +1,211 @@
...
+	var drifts []DriftEntry
+	liveCompared := false
+	prevIdx := -1
+	for i := range p.Snapshots {
+		switch {
+		case p.Snapshots[i].Content == nil:
+			capturePoints[i].Excluded = "the capture failed, so this point holds no schema"
+			continue
+		case !CoversSchemas(p.Snapshots[i].Header, p.Scope.Schemas):
+			missing, _ := lo.Difference(p.Scope.Schemas, p.Snapshots[i].Header.Schemas)
```
makalaaneesh: [AI][Suggestion] **`lo.Difference` recomputed one line after `CoversSchemas` computes the same set.**

Line 79 calls `CoversSchemas(header, p.Scope.Schemas)`, which internally computes `lo.Difference(requested, h.Schemas)`; line 80 immediately recomputes the identical `lo.Difference(p.Scope.Schemas, header.Schemas)` just to name the missing schemas. The work is duplicated. This is a cold path, so purely a tidiness item.

Suggested fix: have `CoversSchemas` return the missing slice (e.g. `func CoversSchemas(...) (missing []string, ok bool)`) and use it directly in the exclusion branch, so the difference is computed once.

---

makalaaneesh (commented):
[AI] Automated review — 2 findings (0 critical, 0 warning, 2 suggestion). See inline comments.

---

makalaaneesh:
[AI] Automated re-review of the current head (`dc2f5bd`) — the branch was force-pushed since the last AI review, so this was a full re-review rather than incremental. No new findings: the open items are the ones already posted inline and still anchored to unchanged code (doc comment sitting on the wrong function in `build.go`; package doc in `report.go` describing JSON/HTML renderers that land in a later PR). Nothing further to add this pass.

---

yb-voyager/design-specs/schema-detect-drift.md
```diff
@@ -85,82 +85,68 @@ The schema filter is applied AFTER diffing, never by narrowing the snapshot cont
...
-	Schemas             []string        // display only
-	Snapshots           []SnapshotInput // oldest first; the live read, if any, is last
-	Scope               schemadiff.Scope
-	Tables              []string        // what was compared, resolved; see Comparing
-	TablesFiltered      bool
-	ObjectTypes         []string
+	Snapshots           []schemasnapshot.SchemaSnapshot // oldest first; the live read, if any, is last
+	Scope               schemadiff.Scope // the exact sets compared; Comparing is rendered from it
+	TablesFiltered      bool             // whether the user narrowed each dimension
 	ObjectTypesFiltered bool
```
priyanshi-yb: why do we need these two ?

yb-voyager/src/schema/schemadrift/build.go
```diff
@@ -0,0 +1,210 @@
...
+		},
+		Drifts:        drifts,
+		CapturePoints: capturePoints,
+	}
+}
+
+// coversSchemas reports whether h was captured with every schema in requested, so
+// that a table missing from its content is genuinely absent rather than never
+// looked for. Capturing MORE than was requested still covers it.
+func coversSchemas(h schemasnapshot.SnapshotHeader, requested []string) bool {
```
priyanshi-yb: nit: we can directly pass h.Schemas so that it is consistent with the function name

yb-voyager/src/schema/schemadrift/build.go
```diff
@@ -0,0 +1,210 @@
...
+			TablesFiltered: p.TablesFiltered,
+			ObjectTypes: lo.Map(p.Scope.ObjectTypes, func(t schemadiff.ObjectType, _ int) string {
+				return string(t)
+			}),
+			ObjectTypesFiltered: p.ObjectTypesFiltered,
+		},
+		Summary: Summary{
+			ChangeCount:           len(drifts),
+			ComparedIntervalCount: comparedIntervals,
+			StoredCaptureCount:    lo.CountBy(p.Snapshots, func(s schemasnapshot.SchemaSnapshot) bool { return s.Header.Label != schemasnapshot.LabelSourceLive }),
```
priyanshi-yb, line 144: what does this storedCaptureCount mean?

yb-voyager/src/schema/schemadrift/classification.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// classify falls back to SeverityAdvisory for an unmapped DiffType, the zero value
+// included, rather than dropping the change. The zero Impact/Action that comes with
+// it is deliberate: the report omits the note when there is nothing useful to say.
+func classify(t schemadiff.DiffType) DriftInfo {
+	if c, ok := infoByDiffType[t]; ok {
+		return c
+	}
+	return DriftInfo{Severity: SeverityAdvisory}
```
priyanshi-yb, line 141: Do we support other drift types/object types than these Table/Column changes?

yb-voyager/src/schema/schemadrift/report.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// CapturePoint is one point on the report's timeline: a moment at which the
+// source schema was captured, or capture was attempted. Three kinds appear -- a
+// stored capture, a stored placeholder (the capture failed, so no schema content
+// exists behind it), and the live read (Series == schemasnapshot.LabelSourceLive,
+// never persisted). It is a projection: the snapshot content itself stays out of
+// the report.
+type CapturePoint struct {
+	Series     string    `json:"series"`
```
makalaaneesh: [AI][Suggestion] **`CapturePoint.Series` is a leftover concept name for what is now just a snapshot Label.**

`CapturePoint.Series` (JSON `"series"`) is populated directly from `s.Header.Label` (`build.go:63`) and compared against `schemasnapshot.Label*` constants throughout `phaseFor` (`build.go:163-172`) — the value *is* a label. The `SnapshotInput.Series` field that once distinguished a stored label from `SeriesSourceLive` was removed when `SnapshotInput` was dropped, but `CapturePoint` kept the vestigial name. A reader sees a field named `Series` holding what the rest of the codebase — and this struct's own doc comment — calls a `Label`. Per root `CLAUDE.md` "Name to avoid collisions and ambiguity," and because `series` is part of the contractual JSON shape, it's worth settling before the shape ships.

Suggested fix: rename the field to `Label` (JSON `"label"`) unless the report's public contract deliberately calls it `series`, in which case a one-line "why" comment would prevent the confusion.

---

yb-voyager/src/schema/schemadrift/build_test.go
```diff
@@ -0,0 +1,676 @@
...
+	}
+
+	report := buildUnfiltered(p)
+
+	assert.Equal(t, 2, report.Summary.StoredCaptureCount, "StoredCaptureCount counts the stored records only, not the live read")
+	assert.Equal(t, 2, report.Summary.ChangeCount, "one TABLE_ADDED per interval (customers, then invoices)")
+	assert.True(t, report.Summary.LiveCompared)
+}
+
+func TestBuildReport_SchemaFilterKeepsOnlyRequestedSchemas(t *testing.T) {
```
makalaaneesh, line 347: [AI][Suggestion] **New schema-coverage/filter paths have no case-sensitive-identifier test.**

The schema-filtering / `coversSchemas` bridging logic (`build.go:80-84`, `154-157`) compares schema names as raw strings via `lo.Difference`, and `Comparing.Tables` renders through `ForDisplay(dbType)`. Root `CLAUDE.md` ("Case-sensitive identifiers") requires testing new table/column/schema handling with case-sensitive (quoted) names, but every fixture here uses lowercase unquoted names (`public`, `sales`, `orders`), so a mixed-case schema/table never exercises the coverage check, the schema filter, or `Comparing` rendering.

Raw string comparison of catalog names is itself correct (since `ObjectRef` stores raw names); the gap is purely coverage. Suggested fix: add a fixture with a mixed-case schema and/or table (e.g. `"Sales"` vs `sales`, or table `"Orders"`) to one of the coverage/filter tests and assert the covering check and `Comparing.Tables` rendering (`public."Orders"`) behave correctly.

---

makalaaneesh (commented):
[AI] Automated review (re-review of the current head — the previous AI-reviewed commit was force-pushed away, so this was a full re-review) — 2 new findings (0 critical, 0 warning, 2 suggestion) below. Of the previously posted [AI] findings, `Summary.LiveCompared` semantics and the `LabelDetectDrift` comment are now addressed; the others (e.g. the `splitIdentity` silent-empty default, and the `codeSpans`/HTML-renderer forward references in `report.go`/`classification.go`) remain posted inline and still anchored, so they are not repeated here.

---

yb-voyager/src/schema/schemadrift/build.go
ShivanshGahlot: Already fixed

ShivanshGahlot: liveCompared is now set inside the loop, only when an interval ending at the live read is actually diffed (build.go:95-97).

yb-voyager/src/schema/schemadrift/build.go
ShivanshGahlot: Fixed: the label is now schemasnapshot.LabelSourceLive, which is defined in this PR, and the comment refers to it.

yb-voyager/design-specs/schema-detect-drift.md
ShivanshGahlot: Update: I have removed these flags now to reduce scope. They were there to basically help with the report generation. TablesFiltered was the only way we could tell in the report whether the list passed to it is the complete universe of tables or a subset. It was attaching a "Filtered" tag in case if it was a subset.

yb-voyager/src/schema/schemadrift/classification.go
ShivanshGahlot: Not yet. Current version only emits TABLE and COLUMN findings, because that's what schemasnapshot captures. INDEX, SEQUENCE, VIEW, FUNCTION and TYPE are declared but commented out in schemadiff/difftypes.go.

yb-voyager/src/schema/schemadrift/build.go
ShivanshGahlot: How many snapshots the run read from the export dir. That's every stored capture, failed ones included, since a failed capture still leaves a row. Only the live read is left out. The console prints it as 'Captures stored'. It sits next to ComparedIntervalCount because stored captures overstate what was actually compared.

yb-voyager/src/schema/schemadrift/build.go
ShivanshGahlot: Fixed: `liveCompared` is now set inside the loop, only when an interval ending at the live read is actually diffed (`build.go:95-97`).

---
_automated · Claude Code (Opus 5.5)_

yb-voyager/src/schema/schemadrift/build.go
ShivanshGahlot: Fixed: the label is now `schemasnapshot.LabelSourceLive`, which is defined in this PR, and the comment refers to it.

---
_automated · Claude Code (Opus 5.5)_

yb-voyager/design-specs/schema-detect-drift.md
ShivanshGahlot: Agreed they didn't belong there. `Scope` can't carry them, so only the caller knew them, and they only changed the report's wording. Removed them from the input, the report, the renderers and callhome in 83c87997. The report now lists the exact sets compared. If we want the bit back later, adding it is backward-compatible.

---
_automated · Claude Code (Opus 5.5)_

yb-voyager/src/schema/schemadrift/classification.go
ShivanshGahlot: Not yet. v1 only emits TABLE and COLUMN findings, because that's what `schemasnapshot` captures. INDEX, SEQUENCE, VIEW, FUNCTION and TYPE are declared but commented out in `schemadiff/difftypes.go`, and get enabled as the engine learns them. If a type ever reaches `getDriftInfo` without an entry, it's reported as Advisory rather than dropped.

---
_automated · Claude Code (Opus 5.5)_

yb-voyager/src/schema/schemadrift/classification.go
ShivanshGahlot: Intentional. #3812 is the base of a stack, and `codeSpans` arrives with the renderers in #3813. The doc describes the package as the stack ships it, so it stays the same in every PR rather than being rewritten in each.

---
_automated · Claude Code (Opus 5.5)_

yb-voyager/src/schema/schemadrift/report.go
ShivanshGahlot: Same as the `codeSpans` thread: the renderers and embedded template land in #3813, and the package doc describes the finished package on purpose.

---
_automated · Claude Code (Opus 5.5)_

yb-voyager/src/schema/schemadrift/build.go
ShivanshGahlot: It's now missingSchemas(requested, captured []string) []string, so it takes the schema lists directly and returns the missing ones.

yb-voyager/src/schema/schemadrift/build.go
ShivanshGahlot: Fixed in 69a6a945, together with the nit on the same helper. missingSchemas computes the missing schemas once, and the loop uses them for both the check and the exclusion reason.

yb-voyager/src/schema/schemadrift/build.go
ShivanshGahlot: Fixed in 41db7595. Comparing.Schemas and the exclusion reason now quote schema names the same way ForDisplay quotes tables, so a mixed-case schema reads "Sales" everywhere. Matching still uses the raw catalog names.

yb-voyager/src/schema/schemadrift/build_test.go
ShivanshGahlot: Added in 93295eb2. One interval drops a table and a column, and the test checks each entry in full: Object/SubObject come from the old side, and OldValue is the dropped table or column.

yb-voyager/src/schema/schemadrift/build.go
ShivanshGahlot: Fixed in a233e743. Removed it: both function names already say which labels start and end the running span.

yb-voyager/src/schema/schemadrift/build.go
ShivanshGahlot: Fixed in 7e064692. The default branch now returns an error naming the unexpected type, and BuildReport returns (Report, error) to carry it up. The command handles it in #3814 (ccad72be) and exits 2 as an operational error, instead of writing an entry on an empty object.

yb-voyager/src/schema/schemadrift/report.go
ShivanshGahlot: Agreed. It is Header.Label verbatim, and the JSON hasn't shipped. Renamed to Label / "label" in 6dc39b93, with the renderer following in #3813 (f6fc9df6). The HTML column was already titled "Capture label", so the page is unchanged.

yb-voyager/src/schema/schemadrift/build_test.go
ShivanshGahlot: Added in 41db7595. TestBuildReport_CaseSensitiveSchemaAndTable runs a "Sales"."Orders" fixture through the coverage check, bridging, the schema filter, and Comparing's quoted rendering.

yb-voyager/src/schema/schemadrift/report.go
ShivanshGahlot: This is implemented in the next PR

yb-voyager/src/schema/schemadrift/classification.go
ShivanshGahlot: This is implemented in the next PR

makalaaneesh (approved):
LGTM!

PR #3813, opened 2026-09-16
schema/schemadrift: HTML and JSON renderers

ShivanshGahlot:

Adds the two output formats for the report built by the previous PR in the stack.

- HTML from an embedded template, plus JSON.
- **Minimal quoting** — printed object paths are valid, copy-pasteable SQL: `sales."MixedCase"` for a case-sensitive or space-containing name, `public.orders` otherwise.
- **Excluded captures are marked** — a point the assembler bridged appears on the timeline, with its reason in the footer table. Leaving it out shows a stretch of quiet timeline a reader takes for "nothing changed here". This also makes a *failed* capture visible in the report for the first time.
- **Empty report renders** instead of panicking, so a run that finds no drift still writes one.
- The view model consumes the typed `schemadiff` facets directly, so no finding is matched by round-tripping a constant through `string()`.

No user-facing changes yet. The command that writes these reports lands in the next PR in the stack.

`go test -tags unit ./src/schema/schemadrift/` — 7 render tests: JSON shape and round-trip, the flattened `DriftInfo` embed, HTML render, minimal quoting, the excluded-capture marker, and the empty report.

No.

No.


yb-voyager/src/schema/schemadrift/templates/drift_report.html
```diff
@@ -0,0 +1,247 @@
...
+    <dl class="facts">
+      <div><dt>Source</dt><dd class="mono">{{.SourceLine}}</dd></div>
+      <div><dt>Window</dt><dd>{{.WindowLine}}</dd></div>
+      <div class="facts-scope"><dt>Comparing</dt><dd>
+        <details class="scope">
+          <summary>{{.ComparingSummary}}</summary>
+          <div class="scope-body">
+            {{range .ComparingScope}}
+            <div class="scope-row"><span class="scope-k">{{.Label}}</span>
+              {{if .Chips}}{{range .Chips}}<span class="scope-chip">{{.}}</span>{{end}}{{else}}<span class="scope-chip">all</span>{{end}}</div>
```
makalaaneesh, line 188: [AI][Warning] **Empty scope row renders "all", contradicting the Go summary's "no tables".**

The dropdown row falls back to a single `all` chip when `.Chips` is empty (`{{else}}<span class="scope-chip">all</span>{{end}}`). But an empty scope dimension does not mean "all": `schemadiff.FilterByScope`/`passesObjectTypeFilter` treat an empty include-set as keep-nothing, and `comparingSummary` correctly renders the same empty set as "no tables"/"no object types" (render.go:207-208). So an empty Tables/ObjectTypes produces a report whose summary says "no tables" while the expanded row says "Tables (0)" + an "all" chip — two renderings of the same fact that disagree, and the "all" one is semantically wrong.

Failure case: `Comparing{Schemas:["public"], Tables:nil, ObjectTypes:nil}` renders summary "no tables · no object types" but dropdown rows "Tables (0) [all]" / "Object types (0) [all]".

Suggested fix: drop the `all` fallback so an empty dimension renders no chip (matching "no tables"), or make scopeCountLabel/scopeRowLabel agree — pick one. E.g.:

```
{{if .Chips}}{{range .Chips}}<span class="scope-chip">{{.}}</span>{{end}}{{else}}<span class="scope-chip note">none</span>{{end}}
```

---

yb-voyager/src/schema/schemadrift/render.go
```diff
@@ -0,0 +1,598 @@
...
+		return strings.Join(parts, ", ")
+	case schemasnapshot.Column:
+		return stringifyColumnDef(v)
+	case schemasnapshot.Table:
+		if len(v.Columns) == 0 {
+			return "no columns"
+		}
+		parts := make([]string, len(v.Columns))
+		for i, c := range v.Columns {
+			parts[i] = c.Name + " " + stringifyColumnDef(c)
```
makalaaneesh, line 546: [AI][Warning] **Column names in an added/dropped table's definition list are not min-quoted.**

In `stringifyValue`'s `schemasnapshot.Table` branch, each column is rendered as `c.Name + " " + stringifyColumnDef(c)`, concatenating the raw column name. Everywhere else the PR min-quotes identifiers via `sqlname` (objectPath/minQuoted, TestObjectPathMinQuotesIdentifiers). A case-sensitive or space-containing column name renders unquoted and ambiguous, against root CLAUDE.md's "never manual string concatenation" / "always test case-sensitive identifiers" and yb-voyager/AGENTS.md's "use the sqlname package for all object name handling".

Failure case: a TABLE_ADDED whose NewValue is a `schemasnapshot.Table` with a column named `MixedCase` renders the chiplet as `MixedCase integer NOT NULL, ...` instead of `"MixedCase" integer NOT NULL, ...`.

Suggested fix: `dbType` is already in scope — quote the name:

```go
parts[i] = minQuoted(c.Name, dbType) + " " + stringifyColumnDef(c)
```

---

yb-voyager/src/schema/schemadrift/render_test.go
```diff
@@ -0,0 +1,328 @@
+//go:build unit
```
makalaaneesh, line 1: [AI][Warning] **Several non-trivial formatting functions have no tests.**

Tests cover JSON shape, objectPath quoting, and a few HTML substrings, but the most branch-heavy Go logic is untested: `stringifyValue` (9 branches), `stringifyColumnDef`, `sourceLine` (partial-field assembly), `comparingSummary`/`comparingScope`/`cappedChips` (including the `+N more` cap and the empty-vs-"all" case flagged elsewhere in this review), `severityLabel` fallback, and `deriveEvent`'s Series/Reason vocabulary. yb-voyager/AGENTS.md's Testing section asks for behavior-asserting tests and case-sensitive coverage where applicable; the `stringifyValue` Table branch in particular would have surfaced the unquoted-column-name issue above.

Suggested fix: add a table-driven `TestStringifyValue` covering each dynamic type (including a case-sensitive Column/Table and the AttrNullability bool), a `TestSourceLine` with missing host/port/db/version, and a small `TestComparingScope` pinning the `+N more` cap and the empty-dimension rendering.

---

yb-voyager/src/schema/schemadrift/render.go
```diff
@@ -0,0 +1,598 @@
...
+	}
+	return buf.Bytes(), nil
+}
+
+// formatTime renders the zero time.Time as "-" (e.g. an empty report window).
+func formatTime(t time.Time) string {
+	if t.IsZero() {
+		return "-"
+	}
+	return t.Format("2006-01-02 15:04:05 UTC")
```
makalaaneesh, line 59: [AI][Suggestion] **`formatTime` hardcodes the "UTC" label without converting to UTC.**

`t.Format("2006-01-02 15:04:05 UTC")` emits the literal text `UTC` (not a zone token) and formats `t` in whatever location it carries. CapturedAt/GeneratedAt are UTC by contract today, so it is currently correct — but it will silently print local wall-clock time labeled "UTC" the moment any non-UTC time reaches it. Cheap to make robust.

Failure case: a `CapturePoint.CapturedAt` in `time.Local` prints e.g. `14:00:00 UTC` for a value that is actually 14:00 local / 09:00 UTC.

Suggested fix:

```go
return t.UTC().Format("2006-01-02 15:04:05 UTC")
```

---

yb-voyager/src/schema/schemadrift/render.go
```diff
@@ -0,0 +1,598 @@
...
+}
+
+// snapshotRows builds the footer's "Snapshots used" table rows from
+// Report.CapturePoints. The live read (Series == schemasnapshot.LabelSourceLive) shows "—" for its
+// sequence number, since it is never persisted/numbered like a stored
+// snapshot.
+func snapshotRows(capturePoints []CapturePoint) []snapshotRow {
+	rows := make([]snapshotRow, len(capturePoints))
+	for i, c := range capturePoints {
+		seq := fmt.Sprintf("%d", i+1)
```
makalaaneesh, line 576: [AI][Suggestion] **`snapshotRows` sets the live-read seq in two places.**

The switch sets `seq = "—"` for the live read, but a live read that is also excluded takes the `c.Excluded != ""` case first, so a trailing `if c.Excluded != "" && c.Series == LabelSourceLive { seq = "—" }` patches the seq back. Correct, but easy to misread as duplicated logic. Compute the seq once from a single predicate.

Suggested fix:

```go
seq := fmt.Sprintf("%d", i+1)
if c.Series == schemasnapshot.LabelSourceLive {
    seq = "—"
}
note := ""
switch {
case c.Excluded != "":
    note = "not compared: " + c.Excluded
case c.Series == schemasnapshot.LabelSourceLive:
    note = "read fresh at report time · not stored"
case c.Reason != "":
    note = "reason: " + c.Reason
}
```

---

makalaaneesh (commented):
[AI] Automated review — 5 findings (0 critical, 3 warning, 2 suggestion). See inline comments.

---

yb-voyager/src/schema/schemadrift/render.go
```diff
@@ -0,0 +1,598 @@
...
+	case schemadiff.OpDropped:
+		return "k-rem"
+	default:
+		return "k-chg"
+	}
+}
+
+// kindLabel renders a DriftEntry.Type (e.g. "COLUMN_TYPE_CHANGED") as its
+// lowercase, space-separated display label ("column type changed").
+func kindLabel(diffType schemadiff.DiffType) string {
```
makalaaneesh, line 450: [AI][Suggestion] **A couple of doc comments restate the one-line body (comments policy).**

Under the root `CLAUDE.md`/`AGENTS.md` comment policy ("default: no comments" except a non-obvious *why*), a couple of doc comments here narrate the *what* the one-line body already shows: `kindLabel` (lines 427-428) restates the lowercase / underscore→space transform its single `strings.ToLower(strings.ReplaceAll(...))` line performs, and `kindClass` (lines 413-415) largely restates its own switch. Most comments in this file do earn their place (the `codeSpans` escape-order note, the `buildTimeline` opening-capture rationale, the `groupByInterval`/severity notes) — this is only about the ones that paraphrase the body.

Suggested — drop the `kindLabel`/`kindClass` doc comments (the names are self-describing) and keep the genuine *why* comments elsewhere. Low priority.

---

makalaaneesh (commented):
[AI] Automated re-review of the current head (`e47db4b`) — the branch was force-pushed since the last AI review, so this was a full re-review. 1 new finding (0 critical, 0 warning, 1 suggestion) below. The other open items (the "all" scope chip on an empty dimension; the two-step live-read seq assignment in `snapshotRows`) are already posted inline and still anchored, so they are not repeated here.

---

yb-voyager/src/schema/schemadrift/render.go
```diff
@@ -0,0 +1,598 @@
...
+		}
+		b.WriteString(s.Database)
+	}
+	if s.DatabaseVersion != "" {
+		if b.Len() > 0 {
+			b.WriteString(" · ")
+		}
+		b.WriteString(s.DatabaseVersion)
+	}
+	return b.String()
```
priyanshi-yb: nit: this is a one-time thing, so maybe fmt.Sprintf() is more readable?

makalaaneesh:
[AI][Warning] **The renderer has no error path below `template.Execute`, so every "cannot happen" state degrades silently.**

`newReportView` and everything under it return values only. Root `CLAUDE.md` → "Fail loudly on unexpected state": a state that shouldn't happen should return an error, not fall through to a default. `build.go` already does this in one place (`splitIdentity` returns an error for an unknown identity type and `BuildReport` carries it up); the renderer should follow the same pattern. Cases found in `render.go`, worst first:

| Where | Unexpected state | Today | Should |
|---|---|---|---|
| `buildTimeline` | a drift group's `Window.From` matches no capture point | group is never emitted; its findings vanish from the HTML while the banner still counts them. This is the bug `14baefb85` fixed, and there is no guard against it recurring | error |
| `buildTimeline` | a group's `Window.To` matches no capture point | zero `CapturePoint` used; block renders as non-live with no closing info | error |
| `buildTimeline` | two capture points share a `CapturedAt` | later one silently overwrites the earlier in `capturePointAt` | error |
| `deriveEvent` | label or reason outside the known vocabulary | no marker, indistinguishable from a periodic capture that legitimately has none | explicit `false` for periodic and live; error on `default` |
| `objectPath` | COLUMN finding with empty `SubObject`, or any finding with empty `Object.Name` | renders `public.orders.` with nothing after the dot | error |
| `stringifyValue` | value of a type `Difference` does not document | `fmt.Sprintf("%v")` prints Go syntax into the report | error |
| `kindClass`, `newFindingView` | `Operation` that is none of ADDED / DROPPED / CHANGED | treated as CHANGED | make CHANGED the explicit case, error on `default` |
| `severityLabel` | `Severity` not in `severityLabelText` | raw string shown | error; `getDriftInfo` can only produce the four known values |

Two fallbacks are designed and fine: `codeSpans` renders an odd number of backticks verbatim, and `formatTime` prints `-` for a zero time.

Suggested fix: `newReportView` returns `(reportView, error)`; `buildTimeline`, `newIntervalView`, `newFindingView`, `objectPath`, `stringifyValue`, and `deriveEvent` each return an error from their `default` branch; `RenderHTML` propagates it. `cmd.writeDriftReports` (#3814) already treats a render error as exit 2, so nothing changes above the package. For the first row, the cheapest guard is a count: after the walk, if the number of interval blocks emitted is not `len(groups)`, return an error naming the orphaned window.

---
_automated · Claude Code (Fable 5.1)_

yb-voyager/src/schema/schemadrift/render.go
makalaaneesh: this seems good to address

yb-voyager/src/schema/schemadrift/render_test.go
makalaaneesh: we don't necessarily need tests for all of these; can ignore IMO

ShivanshGahlot:
Sample report generated from the current stack tip (`2ee67729`) by a real run against PG 16 with `public` and a mixed-case `"Sales"` schema: export schema → schema changes → export data → more changes → `schema detect-drift`. It shows 8 changes across two intervals (the second against the live source) and exits 1. The screenshot has every finding expanded; the zip holds the HTML and JSON as written. 
[drift_analysis_report_sample.zip](https://github.com/user-attachments/files/32560249/drift_analysis_report_sample.zip)

drift_report_screenshot

yb-voyager/src/schema/schemadrift/templates/drift_report.html
ShivanshGahlot: Fixed in 36e7da13. An empty dimension now renders no chips, so the dropdown agrees with the summary's "no tables". This only happens when the compared schemas have no tables at all: an exclude list that empties the set already errors out in the command.

yb-voyager/src/schema/schemadrift/render.go
ShivanshGahlot: Fixed in f880d305 with the suggested minQuoted(c.Name, dbType). An added table now reads "CouponCode" text NOT NULL, pct integer DEFAULT 0, as the sample report on this PR shows. TestRenderHTML_AddedTableColumnsAreMinQuoted pins it.

yb-voyager/src/schema/schemadrift/render_test.go
ShivanshGahlot: Agreed, I'm not adding the full suite. Each fix from this review has one test: TestRenderHTML_EmptyScopeDimensionShowsNoChip and TestRenderHTML_AddedTableColumnsAreMinQuoted. The new error paths are covered by TestRenderHTML_RejectsReportsItCannotDisplay.

yb-voyager/src/schema/schemadrift/render.go
ShivanshGahlot: Done in 346947ad: it converts to UTC before formatting. Every time that reaches it is already UTC today, so no output changes, but the label can no longer be wrong.

yb-voyager/src/schema/schemadrift/render.go
ShivanshGahlot: Done in 90126c1f. The seq is set once from the live-read check, and the switch now only picks the note.

yb-voyager/src/schema/schemadrift/render.go
ShivanshGahlot: Dropped both in 631f1d4b. kindClass is gone entirely since e9e14119: its switch moved into newFindingView, which now errors on an unknown operation.

ShivanshGahlot:
https://github.com/yugabyte/yb-voyager/pull/3813#issuecomment-5792998874:

Addressed in e9e14119. The spec came first in 0eadc305: a §3.5 sentence and a §7 failure-mode row. RenderHTML → newReportView → buildTimeline → newIntervalView → newFindingView now return errors, and #3814 already reports a render error as exit 2.

None of these can happen with today's code: each input is built by this binary from a closed set of values. They're guarded anyway, so a future change that breaks one of those assumptions fails loudly instead of rendering a report that looks complete but is wrong.
- An interval that no capture opens or closes. BuildReport takes every window from capture timestamps, so this can't happen by construction. It's checked before the walk anyway, and the error names the window, because this exact failure (findings silently missing from the HTML) already happened once, in 14baefb85.
- Two intervals opening at the same capture. Each capture opens at most one interval, so this can't happen. It errors anyway, since the second interval would otherwise vanish.
- An unknown operation. The engine only emits ADDED, DROPPED and CHANGED. Those three are now explicit cases, anything else errors, and kindClass is folded into that switch.
- An unknown severity. getDriftInfo only produces the four known severities, and an unclassified DiffType already falls back to advisory. Anything else now errors.
- An undocumented value type. Old and new values only come from the snapshot types stringifyValue handles. Anything else now errors instead of printing Go syntax.
- An empty schema or name, or a column finding without a column. PostgreSQL identities always have both, splitIdentity already errors on an unknown identity, and the engine always sets the column name. Each now errors anyway.

Two cases are deliberately not errors:
- Two captures sharing a CapturedAt: unchanged. Timestamps are stored to the nanosecond, so they don't collide in practice.
- deriveEvent: periodic and live are now explicit no-marker cases, but an unknown label or reason still gets no marker rather than an error. Unlike the cases above, this one can happen: an export dir written by a newer voyager can carry such a label. Failing the whole report over a missing timeline marker would throw away every finding. The footer still lists the capture, and phaseFor makes the same choice.

TestRenderHTML_RejectsReportsItCannotDisplay covers each error, and TestRenderHTML_UnknownCaptureLabelStillRenders covers the exception.

makalaaneesh (approved):
LGTM

PR #3814, opened 2026-09-16
cmd: add `schema detect-drift` command

ShivanshGahlot:

Adds the CLI that reports how the PostgreSQL source schema changed while a migration was running.

- **Command** — new `schema` parent command, with `schema detect-drift` under it.
- **Read-only** — compares the snapshots captured during the migration and writes reports to `<export-dir>/reports/`. It never modifies migration state.
- **Table universe** — the union of the live catalog, the stored snapshots, and the live read. A table dropped from the source but still present in history is compared.
- **Scope** — resolves `--exclude-*` into positive keep-sets, and passes the whole universe when a dimension is unfiltered, which is what the collapsed `schemadiff.Scope` expects.
- **Schema scope** — `--source-db-schema` is carried into `Scope`, so drift outside it is neither reported nor counted toward the exit code.
- **Nothing-compared guard** — a run that examined no interval exits 2 rather than writing an empty report, because empty reads as "no drift". The message names which of the three causes actually occurred, read off the report: nothing stored, nothing usable, or only one usable capture.
- **Exit path** — the command function returns its result and the cobra wrapper sets the exit code. Exiting from inside skipped the deferred disconnect and every handler registered with `atexit`, including the terminal restore that follows a password prompt.
- **List flags normalised** — `--table-list "  "` used to count as filtered while resolving to no names.
- **Config file** — the command can be driven from a config file like the others.
- **Prerequisite** — capture is off by default and cannot be turned on retroactively. A migration that ran without it has no history to compare, so the command exits 2 instead of reporting a misleading "no drift". The `--help` text says so.

Fourth PR in the stack.

New command:

```
yb-voyager schema detect-drift --export-dir <dir> [--output-format html,json]
```

It also accepts `--table-list` / `--exclude-table-list` and `--object-type-list` / `--exclude-object-type-list`. Reports are written to `<export-dir>/reports/drift_analysis_report.{html,json}`. Exit codes: `0` no drift, `1` drift found, `2` operational error — including a run that had nothing to compare.

`go test -tags unit ./cmd/ -run Drift` — seven test functions covering flag validation and normalisation, object-type parsing, the table universe, and summary wrapping. Gap: no end-to-end test against a real database; that is #3817.

No. The command registers `--send-diagnostics` because every command does, but sends nothing. A payload is added later in this stack.

No. The command only reads snapshots, and the live read is held in memory and never saved.


yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,851 @@
...
+	// so set it here before any sqlname use; otherwise GetAllTableNames panics
+	// with "invalid source db type" on the --table-list/--exclude-table-list path.
+	sqlname.SourceDBType = source.DBType
+
+	if err := source.DB().Connect(); err != nil {
+		return false, fmt.Errorf("failed to connect to source database: %w", err)
+	}
+	defer source.DB().Disconnect()
+
+	source.FetchSourceInfo() // best-effort; populates source.DBVersion (used non-fatally elsewhere)
```
makalaaneesh, line 509: [AI][Warning] **`FetchSourceInfo()` is not best-effort — it exits 1 on a version-query failure, contradicting the comment and the exit-code contract.**

The call is annotated `// best-effort; ... used non-fatally`, but `source.FetchSourceInfo()` → `GetVersion()` calls `utils.ErrExit` (`src/srcdb/postgres.go:246`), and `utils.ErrExit` exits with code **1** (`src/utils/logging.go:59`). The command reserves exit 1 for "success, drift found" and exit 2 for operational errors (Long help, lines 104-105). So a source whose `server_version` query fails aborts the run with the *drift-found* code, the deferred `source.DB().Disconnect()` (line 514) is skipped, and the "best-effort / non-fatal" comment is false.

Failure case: a restricted role or transient error on the `server_version` read after `Connect()` succeeds → exit 1; a CI gate reading `$? == 1` treats a connection failure as "drift detected."

Suggested — the version only feeds `Source.DatabaseVersion` (report header + callhome, both tolerant of empty), so either drop the fetch, read it through an error-returning path mapped to `exitDriftOperationalError` (exit 2), or at minimum correct the comment so it doesn't claim best-effort behavior the code doesn't provide.

---

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,851 @@
...
+	ref  schemasnapshot.ObjectRef
+	name *sqlname.ObjectName
+}
+
+// buildDriftTableCandidates reads the live catalog and hands the union off to
+// unionDriftTableCandidates. Must run after source.Schemas is resolved, which is
+// what decides the schemas GetAllTableNames() queries.
+func buildDriftTableCandidates(defaultSchema string, snapshotContents []*schemasnapshot.SnapshotContent, liveContent *schemasnapshot.SnapshotContent) []driftTableCandidate {
+	var liveRefs []schemasnapshot.ObjectRef
+	for _, n := range source.DB().GetAllTableNames() {
```
makalaaneesh: [AI][Warning] **`GetAllTableNames()` exits 1 mid-run, skipping the deferred disconnect and colliding with the drift-found exit code.**

`GetAllTableNames()` ErrExits (exit 1) internally on any query/scan error (`src/srcdb/postgres.go:339,354`) — unlike the sibling `GetAllSchemaNamesIdentifiers()` (line 518) which returns an error that `detectDrift` handles gracefully. So a catalog-read failure during scope resolution exits with the drift-found code, skips `defer source.DB().Disconnect()`, and bypasses the return-not-exit pattern this PR deliberately built.

Failure case: a `pg_class`/`pg_namespace` query fails mid-run (permissions, connection reset) → exit 1, misread as "drift found."

Suggested — surface the limitation: add an error-returning full-catalog variant in `srcdb` and map its failure through `exitDriftOperationalError` (exit 2); at minimum document (here and in the `detectDrift` docstring at 492-496, which claims defers unwind "on every path") that `GetAllTableNames` can still exit 1.

---

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,851 @@
...
+
+	PreRun: func(cmd *cobra.Command, args []string) {
+		resolveDetectDriftFlagDefaults()
+		validateDetectDriftFlags()
+		// Resolve the source password from the --source-db-password flag, the
+		// SOURCE_DB_PASSWORD env var, or an interactive prompt -- exactly as the
+		// export commands do. Without this, source.Password stays empty and the
+		// connect in detectDrift() fails SASL auth against any password-protected
+		// source.
+		getAndStoreSourceDBPasswordInSourceConf(cmd)
```
makalaaneesh, line 115: [AI][Warning] **Password resolution in `PreRun` exits 1 for an operational error the contract labels exit 2.**

`getAndStoreSourceDBPasswordInSourceConf(cmd)` calls `utils.ErrExit` on failure (`cmd/export.go:427`) → exit **1**. A missing/failed password prompt is an operational error; per the command's own contract (lines 104-105) it should be exit 2. Because it runs in `PreRun` (before the return-not-exit wrapper), it also can't be routed through `exitDriftOperationalError`.

Failure case: no `--source-db-password`, no `SOURCE_DB_PASSWORD`, non-interactive stdin → `getPassword` errors → exit 1, indistinguishable from "drift found."

Suggested — resolve the password inside `detectDrift()` via the error-returning `getPassword(...)` and map failure through `exitDriftOperationalError` (exit 2):
```go
pwd, err := getPassword(cmd, "source-db-password", "SOURCE_DB_PASSWORD")
if err != nil {
    return false, fmt.Errorf("failed to resolve source password: %w", err)
}
source.Password = pwd
```

---

yb-voyager/cmd/root.go
```diff
@@ -397,6 +397,13 @@ var noLockNeededList = []string{
 	"yb-voyager initiate",
 	"yb-voyager end",
 	"yb-voyager archive",
+	"yb-voyager schema",
+	// detect-drift is read-only w.r.t. migration state (it only writes report
+	// files under <export-dir>/reports), so it does not take the export-dir
+	// lock. This also means its non-zero exit codes (1 = drift found, 2 =
+	// operational error) can be raised via a direct os.Exit call without ever
```
makalaaneesh, line 404: [AI][Suggestion] **Comment says "direct os.Exit call" but the code path uses `atexit.Exit`.**

This new no-lock justification says the exit codes "can be raised via a direct os.Exit call." The implementation uses `atexit.Exit(1)` / `atexit.Exit(2)` (`detectDrift.go:126,655`), which runs registered handlers before exiting — the opposite of a direct `os.Exit`. Per the repo comments policy an inaccurate comment is worse than none, and this one could mislead a future reader into using a real `os.Exit`, which would skip callhome/child-cleanup handlers.

Suggested — reword to something like: "…can be raised via `atexit.Exit` without skipping any lockfile-unlock step, since no lock is acquired for it."

---

makalaaneesh (commented):

---

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,851 @@
...
+//
+// Table-list validation (whether a pattern actually matches a table) needs the
+// source's table list and so happens later, inside resolveDriftScope.
+func validateDetectDriftFlags() {
+	if source.DBType != POSTGRESQL {
+		exitDriftOperationalError("schema detect-drift currently supports PostgreSQL sources only (got --source-db-type=%q)", source.DBType)
+	}
+
+	if driftTableList != "" && driftExcludeTableList != "" {
+		exitDriftOperationalError("--table-list and --exclude-table-list are mutually exclusive. Use only one of them.")
```
priyanshi-yb, line 200: Out of scope: just putting it down if we want to keep the behaviour of these flags similar to the export data's table list flags, then it does allow both in cases of glob patterns, etc.

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,851 @@
...
+
+// exitDriftOperationalError prints the given error to stderr (and the log) and
+// exits with code 2, the contractual exit code for detect-drift operational
+// errors (bad flags, unreachable source, unsupported source type, etc.).
+//
+// It deliberately does not use utils.ErrExit, which exits with code 1 -- that
+// would collide with detect-drift's own "success, drift found" exit code. It
+// still leaves via atexit, so the handlers main.go registers (callhome, child
+// cleanup, terminal restore after a password prompt) run as they do for every
+// other command.
```
priyanshi-yb, line 644: Why do we want to have this behaviour to error out if drift is found? Can't we return success by saying drift was found and created report successfully? and all other places use errExit?

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,851 @@
...
+		if !lo.Contains(driftValidOutputFormats, f) {
+			return goerrors.Errorf("invalid report output format: %s. Supported formats are %v", f, driftValidOutputFormats)
+		}
+		if seen[f] {
+			return goerrors.Errorf("duplicate report output format: %s", f)
+		}
+		seen[f] = true
+	}
+	return nil
+}
```
priyanshi-yb, line 237: Can't we generate both by default, as we do for some other commands like Assess-migration? If not, output-format should be just one input and not a comma-separated list, I think

yb-voyager/cmd/root.go
```diff
@@ -397,6 +397,13 @@ var noLockNeededList = []string{
...
+	// skipping a lockfile-unlock step, since no lock is ever acquired for it.
```
makalaaneesh, line 405: this command does write reports to some destination, no? Is skipping lock file safe in that case?

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,841 @@
...
+	}
+	var liveContent *schemasnapshot.SnapshotContent
+	if live != nil {
+		liveContent = live.Content
+	}
+	// noDefaultSchema is carried rather than discarded: without a default, an
+	// unqualified --table-list pattern can match nothing at all, and every other
+	// caller of GetDefaultPGSchema treats that as an error rather than a silent
+	// no-match.
+	defaultSchema, noDefaultSchema := GetDefaultPGSchema(source.Schemas)
```
makalaaneesh, line 438: [AI][Warning] **Unqualified `--table-list` never matches when the single `--source-db-schema` is mixed-case.**

`GetDefaultPGSchema` returns `schema[0].MinQuoted` for a single schema, so for `--source-db-schema Sales` it returns `"Sales"` with the quotes. `unionDriftTableCandidates` (line 315) passes that to `sqlname.NewObjectName`, which sets `FromDefaultSchema` by raw string equality against the catalog name `Sales`. It is false for every candidate, so `MatchesPattern` rejects every unqualified pattern.

Confirmed with a scratch test at head (473d14f):

```
defaultSchema="\"Sales\"" noDefault=false
FromDefaultSchema=false
unqualified:  --table-list orders        -> unknown table name(s) [orders] in --table-list
qualified:    --table-list Sales.orders  -> [{Sales orders}]
lowercase:    --source-db-schema sales --table-list orders -> [{sales orders}]
```

So `--source-db-schema Sales --table-list orders` exits 2 with a misleading "unknown table" error for a table that exists. The case-sensitive test in `TestResolveDriftTableRefs` only covers a mixed-case table under a `public` default, which is why it did not catch this. Spec §3.6's new "Table scope" row promises unqualified entries work whenever there is a default schema.

Suggested fix: hand `NewObjectName` the raw name, and add the single mixed-case schema case to `TestResolveDriftTableRefs`.

```go
defaultSchema, noDefaultSchema := GetDefaultPGSchema(source.Schemas)
// NewObjectName compares raw catalog names; GetDefaultPGSchema returns the min-quoted form.
defaultSchema = strings.Trim(defaultSchema, `"`)
```

---
_automated · Claude Code (Fable 5.1)_

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,841 @@
...
+
+	PreRun: func(cmd *cobra.Command, args []string) {
+		resolveDetectDriftFlagDefaults()
+		validateDetectDriftFlags()
+		// Resolve the source password from the --source-db-password flag, the
+		// SOURCE_DB_PASSWORD env var, or an interactive prompt -- exactly as the
+		// export commands do. Without this, source.Password stays empty and the
+		// connect in detectDrift() fails SASL auth against any password-protected
+		// source.
+		getAndStoreSourceDBPasswordInSourceConf(cmd)
```
makalaaneesh, line 115: [AI][Warning] **Three operational failures inside the command still exit 1, the "drift found" code.**

The return-not-exit refactor covered `detectDrift`'s own paths, but these three still reach `utils.ErrExit` (exit 1):

| Where | Call | Exits 1 when |
|---|---|---|
| `PreRun`, line 115 | `getAndStoreSourceDBPasswordInSourceConf` | no `--source-db-password`, no `SOURCE_DB_PASSWORD`, non-interactive stdin |
| `detectDrift`, line 497 | `CreateMigrationProjectIfNotExists` | `mkdir` or metaDB init fails |
| `detectDrift`, line 509 | `FetchSourceInfo` → `GetVersion` | the `server_version` query fails after a successful connect |

The comment on line 509 still says `best-effort`, which `GetVersion` is not. Spec §3.6 reserves exit 1 for "drift found" and exit 2 for "operational error (flags, connection, unreadable snapshot)", and the `detectDrift` doc comment says defers unwind "on every path". A script gating cutover on `$? == 1` reads any of these as drift. The password and `FetchSourceInfo` cases were raised in the previous review round and are unaddressed at head; only the `GetAllTableNames` one was fixed.

Suggested fix: resolve the password inside `detectDrift` through the error-returning `getPassword` and return the error; replace `FetchSourceInfo` with the one field the report uses (`Source.DatabaseVersion`) read through an error-returning path, or drop it, since the report and callhome both tolerate an empty version; for line 497 see the separate comment.

```go
pwd, err := getPassword(cmd, "source-db-password", "SOURCE_DB_PASSWORD")
if err != nil {
	return false, fmt.Errorf("resolve source password: %w", err)
}
source.Password = pwd
```

---
_automated · Claude Code (Fable 5.1)_

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,841 @@
...
+// exits: returning lets its defers unwind and leaves the exit code to the caller,
+// which is also what gives the atexit handlers (callhome among them) a chance to
+// run on every path. driftFound is only meaningful when err is nil, and by then
+// the report is already on disk.
+func detectDrift() (driftFound bool, err error) {
+	// CreateMigrationProjectIfNotExists is idempotent: it's a no-op (aside from
+	// mkdir -p) if this export-dir already has a migration project. detect-drift
+	// only ever writes to <export-dir>/reports/ afterwards; it never touches
+	// migration state (MigrationStatusRecord, table lists, etc.).
+	metaDB = CreateMigrationProjectIfNotExists(source.DBType, exportDir)
```
makalaaneesh, line 497: [AI][Warning] **Not read-only: an uninitialised export dir gets a full project skeleton and a new `meta.db`.**

`detect-drift` is not in `exportDirInitialisedCheckNeededList` (`cmd/root.go`), so `PersistentPreRun` only checks that the directory exists. `CreateMigrationProjectIfNotExists` then creates `schema/`, `data/`, `metainfo/meta.db`, `assessment/`, `temp/` and the rest before `ListSnapshots` finds nothing and the run exits 2 with "holds no schema snapshots". Spec §3.6 says `read-only; writes only under reports/`, and the PR description says the same.

```
yb-voyager schema detect-drift --export-dir /tmp/empty --source-db-... 
# -> /tmp/empty is now a migration project, exit 2
```

On the happy path the call is also redundant: `PersistentPreRun` already sets `metaDB` via `initMetaDB` when `metaDBIsCreated(exportDir)`.

Suggested fix: add `"yb-voyager schema detect-drift"` to `exportDirInitialisedCheckNeededList`, delete this line, and rely on `writeDriftReports`'s existing `os.MkdirAll(reportsDir, 0755)` for the one directory the command does write. That also removes one of the three exit-1 paths flagged separately.

---
_automated · Claude Code (Fable 5.1)_

makalaaneesh (commented):
[AI] Automated review — 3 findings (0 critical, 3 warning, 0 suggestion). See inline comments.

yb-voyager/cmd/detectDrift.go
makalaaneesh, line 497: this seems wrong; we should error out if export dir is not created already; not create it on demand. 
It is only expected that this command will be run when an export-dir already exists.

yb-voyager/cmd/detectDrift_test.go
makalaaneesh, line 1: too many random tests again in cmd.go. 
Pls remember to revisit this post this PR

yb-voyager/cmd/detectDrift.go
makalaaneesh: +1 utils.ErrExit solves this problem, no?

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,841 @@
...
+
+	Run: func(cmd *cobra.Command, args []string) {
+		driftFound, err := detectDrift()
+		if err != nil {
+			exitDriftOperationalError("%v", err)
+		}
+		if driftFound {
+			// Not a failure: the report is already on disk and says so. Exiting
+			// here rather than inside detectDrift lets its defers unwind first.
+			atexit.Exit(1)
```
makalaaneesh, line 126: Why 1? 1 is usually for errors. Here we successfully wrote the report, right? so why 1? exit code 0 is usually used for success

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,841 @@
...
+type driftTableCandidate struct {
+	ref  schemasnapshot.ObjectRef
+	name *sqlname.ObjectName
+}
+
+// buildDriftTableCandidates reads the live catalog and hands the union off to
+// unionDriftTableCandidates. It lists tables through an error-returning call
+// rather than GetAllTableNames, which exits 1 on a query failure -- the code this
+// command reserves for "drift found".
+func buildDriftTableCandidates(listTables func(schema string) ([]string, error), schemas []string, defaultSchema string,
```
makalaaneesh, line 283: why are we not using nameregistry here? seems like a good use-case for that? 

Overall, at a high-level, shouldn't we be re-using some of the logic in export-data? 
Both export-data and schema-detect-drift have the same use-case: getting from flags -> list of tables that are validated.

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,841 @@
...
+	}
+	defer source.DB().Disconnect()
+
+	source.FetchSourceInfo() // best-effort; populates source.DBVersion (used non-fatally elsewhere)
+
+	allSchemas, err := source.DB().GetAllSchemaNamesIdentifiers()
+	if err != nil {
+		return false, fmt.Errorf("failed to fetch schema names from source: %w", err)
+	}
+	source.Schemas, err = namereg.SchemaNameMatcher(source.DBType, allSchemas, source.SchemaConfig)
```
makalaaneesh, line 515: [AI][Suggestion] **`--source-db-schema` is not trimmed here, so a space after a comma fails where export accepts it.**

The export commands run `validateSourceSchema` (`cmd/export.go:347`) before `SchemaNameMatcher`; it splits on commas with `utils.CsvStringToSlice`, trims each entry, and writes the cleaned list back to `source.SchemaConfig`. Detect-drift passes the raw flag value straight in, and `SchemaNameMatcher` splits with a plain `strings.Split`, so an entry keeps its leading space and `matchName` finds nothing for it.

```
yb-voyager export schema        --source-db-schema "public, sales"   # works
yb-voyager schema detect-drift  --source-db-schema "public, sales"   # exit 2: Following schemas are not present in source database:  sales
```

Small, but spec §1 says scoping mirrors the export commands, and the same value in a config file behaves the same way.

Suggested fix: normalise the flag alongside the four list flags in `resolveDetectDriftFlagDefaults`, which already owns "the only place that writes flag state":

```go
source.SchemaConfig = normalizeDriftListFlag(source.SchemaConfig)
```

---
_automated · Claude Code (Fable 5.1)_

makalaaneesh (commented):
[AI] Automated review — 1 finding (0 critical, 0 warning, 1 suggestion). See inline comment.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Fixed in eac0c73c. Exit codes are now 0 (the report was written, with or without drift) and 1 (any error). So FetchSourceInfo failing with exit 1 is the correct error exit. The "best-effort" comment is gone.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Fixed in 7cd2ac53. The universe is built with GetAllTableNamesRaw, which returns an error. detectDrift returns that error, so the deferred disconnect runs before the exit.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Fixed by eac0c73c: exit 2 is gone. A missing password is an error, and 1 is now the error code, so ErrExit in PreRun exits correctly.

yb-voyager/cmd/root.go
ShivanshGahlot: Fixed in eac0c73c. The comment no longer mentions how the exit happens.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Agreed, changed in eac0c73c. Drift found now exits 0 with the report written. Every failure goes through utils.ErrExit with exit 1. 
Removed the whole exit 0,1,2 contract.
Had kept it for now to ask for your review on it. It was suggested by claude.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Fixed by c6bc3bcd. Table lists now go through an in-memory name registry and export data's extractTableListFromString. The default schema is therefore the raw name. With --source-db-schema Sales, all of orders, Sales.orders, Invoices, "Invoices" and ord* resolve.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: With eac0c73c, exit 1 is the error code, so the password step and FetchSourceInfo now exit correctly. The "best-effort" comment is removed. The CreateMigrationProjectIfNotExists point is in the thread on line 527.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Agreed, changed in eac0c73c. A written report now exits 0 whether or not it found drift. 1 means an error. Drift is read from summary.change_count in the JSON report.
It was something suggested by claude. I had kept it in for reviews. Makes sense to remove it and simplify.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Done in c6bc3bcd.
- The table universe goes into an in-memory source name registry (namereg.NewInMemorySourceNameRegistry).
- Both flags are resolved with export data's extractTableListFromString, so matching, globs and the unknown-table error are the same as export data's.
- The registry is never written. name_registry.json can't name a table created after export data's first run.
- A matched partitioned table also brings in every partition beneath it, as export data does.

yb-voyager/cmd/detectDrift_test.go
ShivanshGahlot: Yeah will address that in a separate PR in the stack later.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Agreed, fixed in 84b36ed5. detect-drift is now on the list of commands that need a started migration. An uninitialised export dir fails with the standard "Migration has not started yet" error, and CreateMigrationProjectIfNotExists is gone.

ShivanshGahlot: Fixed in 84b36ed5, same change as the thread above. An empty export dir now errors, and no meta.db or project directories are created.

yb-voyager/cmd/root.go
ShivanshGahlot: Good point, fixed in 84b36ed5. It now takes its own per-command lock, so two detect-drift runs can't overwrite each other's report. The lock file is named per command, so it doesn't block export or import.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Fixed in 84b36ed5. --source-db-schema is trimmed the same way as the list flags, so public, sales matches both.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: So by default if we don't pass the flag it should generate both the reports. I more or less matched analyze schema's format here. One extra thing that analyze-schema does is allow "" as a value which also translates to both.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Globs already behave as in export data, since both use extractTableListFromString. If we want to allow both flags we can do it here or in a follow up depending on requirement. LMK.

yb-voyager/cmd/detectDrift.go
priyanshi-yb: Then why do we allow multiple formats in this argument? Allowing one should be enough, as the default anyways gives both

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,850 @@
...
+}
+
+// captureLiveSnapshotForDrift attempts a best-effort, in-memory-only schema
+// capture of the source for comparison against the historical snapshot chain.
+// Unlike CaptureAndSaveSnapshot, the result is never persisted. Any failure
+// (unsupported source type, capture error) is logged as a note and yields a nil
+// result -- the source being briefly unreachable (or the capture racing DDL)
+// must never fail the whole command, since the snapshot-only comparison is
+// still useful on its own.
+func captureLiveSnapshotForDrift(schemas []string) *schemasnapshot.SchemaSnapshot {
```
makalaaneesh: why is this best-effort? why do we log-warn and continue? why not error out?

makalaaneesh (commented):
[AI] Automated review — 4 findings (0 critical, 0 warning, 4 suggestion). See inline comments.

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,850 @@
...
+	pg, ok := source.DB().(*srcdb.PostgreSQL)
+	if !ok {
+		utils.PrintAndLogfWarning("Note: live schema capture is only supported for PostgreSQL sources; skipping live comparison.\n")
+		return nil
+	}
+	db := pg.GetDB()
```
makalaaneesh: [AI][Suggestion] **Two impossible states in `captureLiveSnapshotForDrift` are handled as warnings.**

The `*srcdb.PostgreSQL` type assertion on line 638 cannot fail: `validateDetectDriftFlags` has already rejected every source type other than PostgreSQL. `GetDB()` on line 643 cannot be nil: `Connect` returned nil a few lines earlier in `detectDrift`. Both branches print a note and continue history-only, which leaves `LiveCompared: false` in the report with no indication that something impossible happened.

Root `CLAUDE.md`, "Fail loudly on unexpected state": for each defensive branch, ask when it is reachable; if the answer is never, error out. The third branch, a real `Capture` failure after connecting, is the designed fallback in spec §7 and should stay a warning.

Suggested fix: return `(*schemasnapshot.SchemaSnapshot, error)`, error on the two impossible branches, keep warn-and-nil for the capture failure:

```go
pg, ok := source.DB().(*srcdb.PostgreSQL)
if !ok {
	return nil, goerrors.Errorf("live capture: source is %T, expected *srcdb.PostgreSQL", source.DB())
}
db := pg.GetDB()
if db == nil {
	return nil, goerrors.Errorf("live capture: no database handle after a successful connect")
}
```

---
_automated · Claude Code (Fable 5.1)_

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,850 @@
...
+
+	printDriftSummaryField("Comparison window", utils.PrintAndLogf,
+		"%s -> %s", formatDriftTimestamp(report.Window.From), formatDriftTimestamp(report.Window.To))
+	// Two numbers, not one: a stored capture that failed or did not cover the
+	// requested schemas is bridged, so the count of captures on disk overstates
+	// what was actually examined.
+	printDriftSummaryField("Captures stored", utils.PrintAndLogf, "%d", report.Summary.StoredCaptureCount)
+	printDriftSummaryField("Intervals compared", utils.PrintAndLogf,
+		"%d (live source comparison: %t)", report.Summary.ComparedIntervalCount, report.Summary.LiveCompared)
+	printDriftSummaryField("Schemas", utils.PrintAndLogf, "%s", joinOrAllDrift(report.Comparing.Schemas))
```
makalaaneesh: [AI][Suggestion] **`joinOrAllDrift` prints "all" for an empty schema list that cannot be empty.**

`--source-db-schema` is required and is resolved against the catalog before this point, so `Comparing.Schemas` is never empty here. The `"all"` fallback is dead code, and it reintroduces the empty-means-all reading that the stack removed from `schemadiff.Scope` (spec §3.1: "Empty means empty, not all"). The renderer PR (#3813) dropped the same fallback from its scope chip for the same reason.

Suggested fix: render the line the way the Tables line is rendered, and delete `joinOrAllDrift`:

```go
printDriftSummaryField("Schemas", utils.PrintAndLogf, "%s", driftScopeLine(report.Comparing.Schemas))
```

---
_automated · Claude Code (Fable 5.1)_

yb-voyager/cmd/detectDrift.go
```diff
@@ -0,0 +1,850 @@
...
+			// Unreachable: already validated in validateDriftOutputFormat.
+			return nil, goerrors.Errorf("unsupported output format %q", f)
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to render %s drift report: %w", f, err)
+		}
+
+		path := filepath.Join(reportsDir, fmt.Sprintf("%s.%s", DRIFT_REPORT_FILE_NAME, f))
+		if utils.FileOrFolderExists(path) {
+			fmt.Printf("\n%s already exists, overwriting it with a new generated report\n", filepath.Base(path))
```
makalaaneesh: [AI][Suggestion] **Overwrite notice goes to stdout only.**

This is the one user-facing line in the file that uses `fmt.Printf`; every other one goes through `utils.PrintAndLogf` or a sibling. The log file therefore does not record that a previous report was replaced, which is the kind of fact someone reading the log after the fact wants ("was this report from the run I think it was?").

Suggested fix:

```go
utils.PrintAndLogf("\n%s already exists, overwriting it with a new generated report\n", filepath.Base(path))
```

---
_automated · Claude Code (Fable 5.1)_

yb-voyager/design-specs/schema-detect-drift.md
```diff
@@ -284,8 +288,9 @@ cmd.detectDrift()
  │      └─ schemasnapshot.Capture(ctx, db, CaptureParams{Label: LabelSourceLive})        cmd → schemasnapshot
  │         its Header.Label IS the timeline identity; appended as the last input
  │
- ├─ cmd.buildDriftTableCandidates(contents, live)             → table universe                                §5.5
- ├─ cmd.resolveDriftTableRefs / complementDriftTableRefs      → []ObjectRef, then schemadiff.Scope            §5.5
+ ├─ cmd.driftTableUniverse(contents, live)                    → table universe                                §5.5
+ ├─ namereg.NewInMemorySourceNameRegistry(universe)           → []NameTuple, cmd.extractTableListFromString,
+ │      cmd.expandDriftPartitions / complementDriftTableRefs  → []ObjectRef, then schemadiff.Scope            §5.5
  │
  ├─ schemadrift.BuildReport(DetectionConfig)                  → Report, error            cmd → schemadrift
```
makalaaneesh: [AI][Suggestion] **§5.1 still omits the nothing-compared guard, and "operational error" survives without a meaning.**

The flow was updated this round for the registry step but still goes from `BuildReport` straight to `writeDriftReports`. The `ComparedIntervalCount == 0` guard between them is contractual behaviour with its own §7 row, and it is a decision point two implementations could place differently (before or after the report is built), so it belongs in the flow with a § tag. Suggested line after `BuildReport`'s subtree:

```
 ├─ if Summary.ComparedIntervalCount == 0 → nothingComparedError(report) → error, exit 1    §7
```

Separately, "operational error" still appears in §3.6 (Source type row) and three times in the §5.5 resolution table. It used to mean exit 2; now that every failure is exit 1 the phrase has no distinct referent. Replace with "error, exit 1" to match §7.

---
_automated · Claude Code (Fable 5.1)_

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Fair point, changed in e22c09c1. --output-format now takes one value, html or json, as analyze-schema does. Left unset, it writes both.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Fixed in e22c09c1. It now goes through utils.PrintAndLogf, so the log records it.

yb-voyager/design-specs/schema-detect-drift.md
ShivanshGahlot: Fixed in 53607ce9. §5.1 now shows the nothing-compared guard (error, exit 1, no report written), and the leftover 'operational error' wording now reads 'error, exit 1'.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Hmm so more or less this was built thinking that if source db is not reachable we should still calculate the drift with the already captured snapshots. But now after the implementation that state is technically never reachable as we require source db connection to even fetch the table list. Changed it now.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Fixed in 139056a2, along with the thread above. Every failure path in the function now returns an error, including these two branches.

yb-voyager/cmd/detectDrift.go
ShivanshGahlot: Fixed in 139056a2. The Schemas line now uses driftScopeLine like Tables, and joinOrAllDrift is gone.

yb-voyager/config-templates/offline-migration.yaml
```diff
@@ -535,6 +535,36 @@ finalize-schema-post-data-import:
   ### Default - info
   # log-level: info                                     
 
+# Schema Drift Detection Configuration
```
makalaaneesh: - let's add [TECH PREVIEW] here
- let's uncomment command + one flag
- let's add [TECH PREVIEW] also the command help

makalaaneesh:
Refactors for a subsequent PR: 

1. reuse functions from exportData:
func: flags(table-list, exclude-table-list), universeTables -> Nametuples[]
- glob patterns
- same UX (being able to use both include,exclude)
- partitions handling

export-data:
	universeTables = fetch from source
schema detect-drift
	universeTables = fetch from source UNION tables from all older snapshots
export-schema

2: no separate in-memory constructor for NameRegistry. 
- let's allow for configuring persistence in NameRegistryParams
- do we need a new role for schema-drift?
- do we need to implement a SourceDBInterface because the current constructor can't take table names? or a new constructor? 

3. Re-use ux package for high-level strucuture for consistent styling and headers/footers

makalaaneesh (approved):
LGTM with one nit about configs/command-help

ShivanshGahlot:
**Follow-up, not in this PR: iterative cutover doesn't carry a CLI capture opt-out.**

Schema-snapshot capture is now on by default (5196fba0). When import data starts the next iteration's exporter (`startExportDataFromSourceOnNextIteration`, `cmd/importData.go`), it passes on only `export-dir`, `config-file`, `disable-pb` and the global flags from the CLI. A `--disable-schema-snapshot-capture=true` or `--schema-snapshot-capture-interval` given on the command line is dropped, so the next iteration captures at the defaults. The config-file keys do carry over.

Fix: forward both flags when spawning the next iteration. The spec's migration-flow matrix (§6, iterative cutover row) records the gap until then.

---
_automated · Claude Code (Opus 5.5)_

PR #3815, opened 2026-09-16
cmd: point users at `schema detect-drift` on failure and at cutover

ShivanshGahlot:

Points users at `schema detect-drift` from the three places where checking for source schema drift is worth suggesting.

- **`export data` fails** — source exporter only. Covered on both the explicit failure branch and the ~95 `utils.ErrExit` paths in that file, which unwind nothing and so need the process exit handler.
- **`import data` fails** — forward path only, via the same exit handler. `import data to source` and `to source-replica` are excluded: they run only after cutover to target, past the last source capture, which §1 of the design spec puts out of scope.
- **Cutover to target** — before the confirmation prompt, because cutover to target ends schema capture on the source. Suppressed on a re-run and under `--yes`, where the user cannot act on it.
- **All three require a snapshot to exist.** Capture is off by default, so without that check the common case was pointing users at a command that exits 2 with "holds no schema snapshots". A command that dies before metaDB is opened prints nothing, which drops config and validation failures that have no bearing on drift.
- **Helpers live in `cmd/common.go`**, below the exit handler that is their main caller. They could not land a PR ahead of that caller — `unused` fails the lint job in every build-tag group.

Nothing is captured at the cutover prompt itself. The snapshot the recommendation refers to is the source exporter's exit capture, written from another process after the user confirms. `detect-drift`'s live source read is what covers the window up to that point.

Fifth PR in the `schema detect-drift` stack.

New hint lines in `export data`, `import data`, and cutover output, pointing at `schema detect-drift`. No change to export, import, or cutover behaviour.

`TestSchemaDriftErrorHintLeadIn` pins which commands qualify, including the post-cutover importers and the target exporter that must not. `TestSchemaDriftGuidanceIsUsefulWithoutMetaDB` covers the pre-metaDB exit. Plus `go build ./...`, the existing `-tags unit` suite, and `scripts/lint.sh` across all 13 build-tag groups. The rendered hint text itself is not asserted.

No.

No.


yb-voyager/cmd/common.go
```diff
@@ -1881,6 +1882,12 @@ func createCallhomePayload(migrationUUID uuid.UUID) callhome.Payload {
 }
 
 func PackAndSendCallhomePayloadOnExit() {
+	// Must stay above the payload-sent guard: that flag is only set once a payload
```
makalaaneesh, line 1885: [AI][Warning] **Inaccurate rationale comment on the payload-sent guard.**

The comment says the `callHomeErrorOrCompletePayloadSent` flag "is only set once a payload actually goes out, so anything below it differs with `--send-diagnostics` off and on." That premise is wrong: `callhome.SendPayload` returns `nil` when diagnostics are disabled (`src/callhome/diagnostics.go`, `if !SendDiagnostics { return nil }`), and each `packAndSend...Payload` sets the flag on `err == nil && status in {COMPLETE, ERROR}` — so the flag is set the same way whether `--send-diagnostics` is on or off. The exit-path behavior does not actually differ by that flag.

The genuine reason to place the hint above the guard is that on an `ErrExit` path a payload may already have been sent earlier in the run, and the hint must still print. Per the root `CLAUDE.md` Comments rule, a comment that is wrong about a subtle invariant is worse than none.

Suggested fix:
```go
	// Must run before the payload-sent guard's early return: on an ErrExit path a
	// payload may already have been sent earlier in the run, and the hint must still print.
	if utils.ErrExitErr != nil {
		printSchemaDriftErrorHintOnExit(currentCommand)
	}
```

---

yb-voyager/cmd/common.go
```diff
@@ -1881,6 +1882,12 @@ func createCallhomePayload(migrationUUID uuid.UUID) callhome.Payload {
...
+	// actually goes out, so anything below it differs with --send-diagnostics off and on.
+	if utils.ErrExitErr != nil {
+		printSchemaDriftErrorHintOnExit(currentCommand)
```
makalaaneesh, line 1888: [AI][Warning] **Exit-handler prints a false "export data exited with an error" hint when a post-success downstream spawn ErrExits.**

On the `export data` / `export data from source` success path, `startFurtherCommandsAfterCurrentExportData()` runs `startFallBackSetupIfRequired()` / `startNextIterationImportDataToTarget()`, which call `utils.ErrExit(...)` in several places (e.g. `exportData.go:2271`, `:2286`) *after* export already succeeded and its `COMPLETE` payload was sent. When that happens, `ErrExitErr != nil` while `currentCommand` is still `export data`/`export data from source` on the source role, so `printSchemaDriftErrorHintOnExit` emits "export data exited with an error. If the source schema may have changed…" even though the export itself succeeded and the failure was in fallback/next-iteration setup. The same risk exists for `import data to target` if it ErrExits after streaming completes.

Failure case: `export data from source` completes (fall-back enabled) → `startFallBackSetupIfRequired()` hits `utils.ErrExit("failed to get import data status record: %w", err)` → the user sees "export data exited with an error."

Suggested fix: gate the exit-handler hint on the data phase itself having failed rather than on any ErrExit while the command runs — e.g. set a package-level `dataPhaseFailed` flag only in the genuine export/import failure branches and require it in `printSchemaDriftErrorHintOnExit`. At minimum, record this edge in the design spec so the wording is a conscious tradeoff.

---

yb-voyager/cmd/exportData.go
```diff
@@ -242,6 +242,11 @@ func exportDataCommandFn(cmd *cobra.Command, args []string) {
 		color.Red("Export of data failed! Check %s/logs for more details.", exportDir)
 		log.Error("Export of data failed.")
 		sendPayloadAsPerExporterRole(ERROR, nil)
+		if exporterRole == SOURCE_DB_EXPORTER_ROLE {
+			// The exit handler cannot cover this: it is reached with ErrExitErr unset,
+			// which reads there as a clean exit rather than a failure.
+			printSchemaDriftErrorHint(exportDataDriftHintLeadIn)
```
makalaaneesh, line 248: [AI][Suggestion] **Duplicate "export failed" wording in the explicit export-failure branch.**

This branch prints `color.Red("Export of data failed! …")` just above, then immediately prints a hint whose lead-in (`exportDataDriftHintLeadIn`) is "export data exited with an error." The user sees two back-to-back "export failed" statements; the lead-in adds nothing here since the red line already stated it.

Suggested fix: in this explicit branch, print a lead-in-less variant (or pass an empty lead-in and have `printSchemaDriftErrorHint` skip the leading sentence when it is empty), so only the "If the source schema may have changed…" advisory follows the red failure line. Keep the full lead-in for the exit-handler (`ErrExit`) path, where no preceding failure sentence was printed.

---

makalaaneesh (commented):
[AI] Automated review — 3 findings (0 critical, 2 warning, 1 suggestion). See inline comments.

---

PR #3817, opened 2026-09-16
testschemadrift: end-to-end test for `schema detect-drift`

ShivanshGahlot:

Adds the end-to-end test for `schema detect-drift`. It changes the source schema in the middle of a real migration, then runs the real binary and asserts what the command reports and the exit code it reports it with.

- **Coverage** — the only test that drives the whole chain through the CLI: the capture hooks on export schema / export data, snapshot persistence in metaDB, the diff engine, the report assembler, and the renderers. Every layer already has unit tests; none of them prove the command wires them together.
- **Exit codes** — pins the contract scripts and CI pipelines branch on, which nothing covered before: `0` report written and no drift, `1` report written and drift found, `2` operational failure. Asserting the code matters, because treating any non-zero exit as failure would call a correct drift-detected run broken.
- **Report assertions** — decodes the JSON into `driftreport.Report` instead of scraping rendered text, since that JSON shape is contractual for downstream tooling.
- **Live read is not persisted** — asserting exactly three stored captures pins that the command's own live read never becomes a fourth snapshot.
- **Harness** — `isSourceCmd` now includes `schema detect-drift`, so the runner passes it source connection flags. It needs them for the live read.

The drift is deliberately one added column. The point is that the command detects, reports, and signals a change at all. What each change type means is pinned far more cheaply in `driftreport`'s unit tests.

No user-facing changes. Test only.

It is the test. Run against a real PostgreSQL container and a binary built from this branch:

```
go test -tags integration_voyager_command -run TestDetectDriftEndToEnd ./src/testschemadrift/
```

All four subtests pass (~19s). The whole package was also run to confirm this test and `TestSchemaSnapshotCaptureHooksFireDuringRealCommands` coexist — they use different container configs on purpose, since the registry is keyed by config and each test terminates its own container.

No.

No.


makalaaneesh:
[AI] Automated review — no findings.

Reviewed the end-to-end `schema detect-drift` test and the one-line `isSourceCmd` harness change against the base branch. `go vet -tags integration_voyager_command`, `go vet ./test/utils/...`, and `gofmt -l` are clean; exit-code assertions, container isolation, and the deterministic `StoredCaptureCount == 3` reasoning all check out, and the harness change affects no other command. No Critical/Warning/Suggestion issues.

---

PR #3818, opened 2026-09-16
Add tests for the guardrails for changing the cdc-partition-key on resumption 

priyanshi-yb:

Adds unit and end-to-end tests for the CDC partition-key resume guardrails. Test-only; no production code changes.

- **Unit (`TestCustomKeyResumeGuardChangeCases`)** — exercises the full resume change matrix in 13 cases against both guard entry points. Rejected: custom↔pk, table↔custom, changed custom key columns, override on a leaf partition name, global key change with an override present, and import-table-list grow/shrink. Accepted: equivalent overrides spelling, a redundant override dropped with an unchanged effective map, and a custom key equal to the PK columns.
- **Asymmetry pinned** — global `auto`→`pk` is rejected even though it resolves identically. The global key is raw-string compared, while overrides are compared semantically.
- **End-to-end (`TestLiveMigrationCustomKeyResumeGuardChangeCases`)** — runs a real custom-key live migration, attempts each disallowed resume change (all rejected before snapshot), then resumes with an equivalent overrides spelling and completes through cutover.

No user-facing changes.

The tests are the change. Both pass locally: `go test -tags unit -run TestCustomKeyResumeGuardChangeCases ./yb-voyager/cmd/` (13 subtests) and `go test -tags failpoint_import -run TestLiveMigrationCustomKeyResumeGuardChangeCases ./yb-voyager/src/testlivemigration/` (against real PostgreSQL + YugabyteDB containers).

No.

No.

---

| Component        


yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -4192,3 +4192,228 @@ SELECT i, 'user_' || i || '@example.com' FROM generate_series(1, 20) as i;`,
...
+	// Accepted: equivalent overrides spelling (quoting + whitespace) -> same per-table map.
+	equivalentOverrides := `  "test_schema"."by_pk":(id) ; "test_schema"."orders":( customer_id ) `
+	err = lm.ResumeImportData(true, map[string]string{
+		"--cdc-partition-key":           "pk",
+		"--cdc-partition-key-overrides": equivalentOverrides,
+	})
+	testutils.FatalIfError(t, err, "resume with semantically-equivalent overrides should succeed")
+
+	// Let the resumed import run prepareCdcPartitionKey; it must not trip the change-guard.
+	time.Sleep(20 * time.Second)
```
makalaaneesh: [AI][Warning] **Fixed `time.Sleep` + unsynchronized read of an async process's output buffers.**

The accepted-resume path resumes the import *asynchronously* (`ResumeImportData(true, ...)`), then `time.Sleep(20 * time.Second)` and reads command output via `GetImportCommandStderr()+GetImportCommandStdout()` while the import process is still running. This trips two loaded standards:

- root `CLAUDE.md` / `yb-voyager/AGENTS.md`: *"No fixed `time.Sleep` to wait for asynchronous work — poll for the condition with a timeout."*
- `testlivemigration/AGENTS.md`: *"Never use a fixed `time.Sleep` to wait for a migration phase"* and *"After an async start, do not read the process's output buffers or shared files until the run completes or a synchronized signal fires — concurrent reads race with the writer."*

The read is a genuine data race: `VoyagerCommandRunner.Stdout()/Stderr()` call `bytes.Buffer.String()` with no mutex while os/exec's copy goroutine writes the same buffer — `go test -race` would flag it. The 20s sleep is also flaky (the guard may run later under load) and largely redundant: the subsequent `WaitForForwardStreamingComplete` already proves the guard did not trip (a tripped guard exits the import, so streaming would never complete).

Suggested fix: drop the sleep + buffer read and rely on a synchronized wait as the pass signal, e.g. poll for streaming mode (`WaitForStreamingMode(90*time.Second, 2*time.Second)`), then keep the persisted-map assertions that follow.

(The same anti-pattern already exists elsewhere in this file, e.g. ~line 1597; this finding is scoped to the newly added line.)

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -2667,7 +2667,7 @@ func TestLiveMigrationWithSubsetOFPartialUNiqueIndexColumnsBeingChangedInUpdate(
 	testutils.FatalIfError(t, err, "failed to wait for streaming complete")
 
 	// Import must not crash on a 23505: with the bug the conflict is missed, the two
-	// events race, and import errors out instead of serializing them.
+	// events race, and import errors out instead of serializing them.
```
makalaaneesh: [AI][Warning] **Incidental trailing-whitespace edit breaks gofmt-clean.**

The only change to this pre-existing comment is trailing whitespace appended after "serializing them." It's an unrelated drive-by edit, and it makes the file *not* gofmt-clean: `gofmt -l` reports `live_migration_uk_conflict_test.go`, and `gofmt -d` shows it wants to strip exactly these trailing spaces.

Root `CLAUDE.md`: *"Keep new/edited files `gofmt`-clean."*

Suggested fix: revert this line to its original (remove the trailing spaces) so the diff touches only the new test.

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -4192,3 +4192,228 @@ SELECT i, 'user_' || i || '@example.com' FROM generate_series(1, 20) as i;`,
...
+		"by_pk custom key equal to the PK columns should be accepted and persisted")
+	assertStrategyInMap(t, importDataStatus.TableToCDCPartitionKey, "metrics", cmd.PARTITION_BY_PK)
+
+	err = lm.StopImportData()
+	testutils.FatalIfError(t, err, "failed to stop import data")
+
+	// expectResumeRejected resumes without --start-clean and asserts the attempt is rejected
+	// before snapshot with an output matching check. State is left intact for the next attempt.
+	expectResumeRejected := func(desc string, args map[string]string, check func(string) bool) {
+		_ = lm.ResumeImportData(false, args)
```
makalaaneesh: [AI][Suggestion] **Rejected-resume helper discards the command error, asserting only on stderr text.**

`expectResumeRejected` does `_ = lm.ResumeImportData(false, args)` and asserts only that the combined output contains the guard message. Since these are "must be rejected before snapshot" cases and the synchronous run returns a non-nil error on non-zero exit, also asserting the returned error strengthens the signal — it proves the process actually failed (rejected) rather than emitting the text and continuing.

`testlivemigration/AGENTS.md`: *"Assert on the strongest available signal."*

Suggested change:
```go
expectResumeRejected := func(desc string, args map[string]string, check func(string) bool) {
	err := lm.ResumeImportData(false, args)
	assert.Error(t, err, "%s: resume should have failed", desc)
	out := lm.GetImportCommandStderr() + lm.GetImportCommandStdout()
	assert.True(t, check(out), "%s; got output: %s", desc, out)
}
```

---

yb-voyager/cmd/live_migration_test.go
```diff
@@ -1321,3 +1321,287 @@ func TestBuildGeneratedStoredColumns(t *testing.T) {
...
+
+	// --- custom key column / strategy transitions (semantic guard) ---
+
+	t.Run("changed custom key column list is rejected", func(t *testing.T) {
+		cdcPartitionKey = PARTITION_BY_PK
+		cdcPartitionKeyOverrides = "test_schema.orders:(region)"
+		err := validateCdcPartitioningStrategyUnchanged(tableNames, customStored("customer_id"), emptyUK)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "changing cdc-partition-key")
+		assert.Contains(t, err.Error(), "custom key columns")
```
makalaaneesh, line 1447: [AI][Suggestion] **Error-path assertions use substring `assert.Contains` rather than the whole message.**

The rejection subtests assert with multiple `assert.Contains` on `err.Error()`. `yb-voyager/cmd/AGENTS.md`: *"Validate the entire expected value in tests, not just a substring,"* and `yb-voyager/AGENTS.md`: *"Assert on the strongest available signal … prefer exact values … over substring containment."*

This is low priority — some error strings embed dynamic table output and the file already uses this convention throughout (and correctly uses `assert.Equal` for the resolved-value assertions). Where the full error string is deterministic, consider `assert.EqualError` / `assert.Equal(t, expected, err.Error())`; otherwise leave as-is.

---

makalaaneesh (commented):
[AI] Automated review — 4 findings (0 critical, 2 warning, 2 suggestion). See inline comments.

---

yb-voyager/cmd/live_migration_cdc_partition_strategy.go
```diff
@@ -244,7 +244,7 @@ func resolveCdcPartitionKeyOverrides(rawOverrides map[string]cdcPartitionKeyOver
 			return nil, goerrors.Errorf("cdc-partition-key-overrides: table %q not found in name registry: %w", tableSpec, err)
 		}
 		if _, ok := importTableSet.Get(tuple); !ok {
-			return nil, goerrors.Errorf("cdc-partition-key-overrides: table %q is not in the import table list", tableSpec)
+			return nil, goerrors.Errorf("cdc-partition-key-overrides: table '%s' is not in the import table list", tableSpec)
```
makalaaneesh, line 247: [AI][Suggestion] **Quote-style inconsistency across the three touched error messages.**

This PR changes three guardrail messages away from `%q`, but not consistently: this line wraps the raw table spec in single quotes (`table '%s'`), while lines 489 and 493 emit the joined list with no surrounding quotes (`table %s ...`). The values on 489/493 are `ForKey()` strings that already embed double quotes (`"schema"."table"`), so leaving them bare is defensible, but the mixed convention within one small diff reads as accidental. Minor, purely cosmetic — user-facing error text is not a compatibility surface.

Suggested fix: pick one convention for how a table identifier is delimited in these messages — e.g. keep the already-quoted `ForKey()` values bare on 489/493 and drop the single quotes here, or vice-versa. No functional change either way.

---

makalaaneesh (commented):
[AI] Automated review — 1 finding (0 critical, 0 warning, 1 suggestion). See inline comments.

---

makalaaneesh (approved):
LGTM. pls resolve the AI comments wherever necessary

PR #3820, opened 2026-09-17
Add integration test cases for custom key on partitioned tables and one case for NULL values in custom key

priyanshi-yb:
Adds live-migration integration tests for custom CDC partition keys on partitioned tables. Test-only — no product code changes.

- **Partition-column key with row movement** — custom key equals the LIST partition column; verifies moving a row across partitions (emitted as delete+insert, not an update) does not trip the immutability guard.
- **Leaf-local unique index, key = partition column** — a per-leaf unique index routed by the partition column produces no cross-channel conflicts.
- **Leaf-local unique index, different custom key** — the same unique value coexisting in two leaves under a non-partition custom key is not flagged as a false conflict.
- **Multi-level + cross-schema partitions** — LIST→RANGE→HASH partitioning with leaf partitions spread across several schemas migrates correctly under a custom key.
- **NULL custom-key values** — rows whose custom key is NULL route via the sentinel channel and stream consistently alongside non-NULL rows.

No user-facing changes.

Five new `TestLiveMigrationCustomCdcPartitionKey*` tests in `live_migration_uk_conflict_test.go` (build tag `failpoint_import`). Run with `failpoint-ctl enable` then `go test -tags failpoint_import -run TestLiveMigrationCustomCdcPartitionKey ./yb-voyager/src/testlivemigration/`.

No.

No.


yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,730 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+// under custom-key routing a different column of the table - conflicts on the uk column should still be detected
+//
+// test_live is LIST-partitioned by region and routed by the custom key (region), so every leaf's
+// events carry a custom key. Each leaf has its
+// OWN unique index on uk_val. Two facts are exercised:
+//   - Coexistence: the SAME uk_val (777) lives in both r1 and r2 at once (snapshot rows) because
+//     the indexes are independent per leaf; the snapshot must import both.
+//   - Conflicts should be detected as the custom key is different so same uk val across leaves still be detected.
+//
+// The count failpoint must therefore record ZERO conflicts, and the target must stay consistent.
```
makalaaneesh: [AI][Warning] **Doc comment states the opposite of what the test asserts.**

The doc comment for `TestLiveMigrationCustomCdcPartitionKeyLeafLocalUniqueIndexAcrossLeavesWithDifferentCustomKey` ends with `// The count failpoint must therefore record ZERO conflicts, and the target must stay consistent.` — but this test's whole purpose is that conflicts ARE detected, and the body asserts `require.Greater(t, conflictStats.Total, 0, ...)` (line 2291). This line is a stale copy-paste from the preceding `...AcrossLeaves` test (its identical line ~2025 is correct there).

`yb-voyager/src/testlivemigration/AGENTS.md`: *"Each test states in a doc comment which invariant it pins and why the chosen data pattern triggers it."* A doc that pins the opposite invariant actively misleads a maintainer trying to understand or fix the test.

Suggested fix:
```go
// The count failpoint must therefore record NON-ZERO conflicts (same uk_val on different
// channels), and the target must stay consistent.
```

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,730 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+	testutils.FatalIfError(t, err, "failed to initiate cutover")
+
+	err = lm.WaitForCutoverComplete(0, 30)
+	testutils.FatalIfError(t, err, "failed to wait for cutover complete")
+}
+
+// Test pins that a leaf-local unique index (defined independently on each partition, not on the root) is handled correctly
+// under custom-key routing a different column of the table - conflicts on the uk column should still be detected
+//
+// test_live is LIST-partitioned by region and routed by the custom key (region), so every leaf's
```
makalaaneesh: [AI][Warning] **Doc comment claims routing by `region`, but the test routes by `id`.**

This line says `// test_live is LIST-partitioned by region and routed by the custom key (region), so every leaf's`, and the "Coexistence" paragraph that follows is copy-pasted from the previous test. This test actually routes by `id`: the override is `test_schema.test_live:(id)` (line ~2258) and it asserts `Columns == []string{"id"}`. The distinct routing column (`id`, different from both the partition column `region` and the unique column `uk_val`) is the entire point of this variant — it is what causes same-`uk_val` churn to land on different channels and produce detectable conflicts. "routed by region" contradicts both the code and this test's own title.

Violates the same `testlivemigration/AGENTS.md` doc-comment rule.

Suggested fix: rewrite the comment to state routing is by `id` (custom key distinct from the partition column `region` and the leaf-local unique column `uk_val`), so same-`uk_val` events on different `id`s hash to different channels and conflicts must be detected. Drop or adapt the "Coexistence" paragraph to this test's actual data.

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,730 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+
+	require.False(t, lm.GetImportRunner().IsStopped(),
+		"import should keep running during count failpoint mode")
+
+	conflictStats, err := testutils.ReadUniqueKeyConflictStats(uniqueKeyConflictStatsPath)
+	testutils.FatalIfError(t, err, "failed to read unique key conflict stats")
+	require.Greater(t, conflictStats.Total, 0, "partial unique delta should produce UK conflicts")
+	require.Greater(t, conflictStats.ByTable[`"test_schema"."test_live"`], 0, "test_live should produce UK conflicts")
+
+	require.LessOrEqual(t, conflictStats.Total, 6, "partial unique delta should produce UK conflicts")
```
makalaaneesh: [AI][Warning] **Conflict-count upper bound is unexplained and the assertion message is a wrong copy-paste.**

`yb-voyager/src/testlivemigration/AGENTS.md`: *"Conflict/event/row counts from streaming are timing-dependent: assert a justified range (both lower and upper bound), derived from the workload and explained in a comment."* Here the range is `Total > 0` (line 2291) and `Total <= 6` (this line) with no comment justifying why 6 is the bound. Additionally the upper-bound messages are verbatim copies of the lower-bound messages — this line reads `"partial unique delta should produce UK conflicts"` and line 2295 reads `"test_live should produce UK conflicts"` — so a failure of the *upper* bound (over-triggering) prints a message describing under-triggering.

Suggested fix: add a one-line comment deriving the bound and correct the messages, e.g.:
```go
// The delta reclaims a freed uk_val 6 times across different id channels; each is at most
// one cross-channel conflict, so 0 < conflicts <= 6.
require.Greater(t, conflictStats.Total, 0, "cross-channel uk reclaim should produce UK conflicts")
require.LessOrEqual(t, conflictStats.Total, 6, "at most one conflict per cross-channel uk reclaim (6)")
```

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,730 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+	testutils.FatalIfError(t, err, "target does not match source after streaming")
+
+	err = lm.InitiateCutoverToTarget(false, nil)
+	testutils.FatalIfError(t, err, "failed to initiate cutover")
+
+	err = lm.WaitForCutoverComplete(0, 30)
+	testutils.FatalIfError(t, err, "failed to wait for cutover complete")
+}
+
+// Test pins that a leaf-local unique index (defined independently on each partition, not on the root) is handled correctly
```
makalaaneesh: [AI][Suggestion] **Doc comment does not start with the function name.**

The comment begins `// Test pins that a leaf-local unique index ...` rather than the standard Go doc-comment form starting with the identifier. Every sibling test in this file (and Go convention) starts with `// TestName ...`. Minor, but it breaks `go doc`/tooling association and the file's own convention.

Suggested fix: start the comment `// TestLiveMigrationCustomCdcPartitionKeyLeafLocalUniqueIndexAcrossLeavesWithDifferentCustomKey pins that ...`.

---

makalaaneesh (commented):

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+			BEGIN
+				-- r1 churn on uk_val=1 (custom key 'r1' -> one channel)
+				DELETE FROM test_schema.test_live WHERE id = 100 AND region = 'r1';
+				INSERT INTO test_schema.test_live (id, region, uk_val) VALUES (1, 'r2', 1);
+				DELETE FROM test_schema.test_live WHERE id = 1 AND region = 'r2';
+				INSERT INTO test_schema.test_live (id, region, uk_val) VALUES (2, 'r1', 1);
+				DELETE FROM test_schema.test_live WHERE id = 2 AND region = 'r1';
+				INSERT INTO test_schema.test_live (id, region, uk_val) VALUES (3, 'r2', 1);
+
+				-- r2 churn on uk_val=10001 (custom key 'r2' -> a different channel)
```
makalaaneesh, line 2229: [AI][Warning] **Stale copy-paste comments in this `SourceDeltaSQL` describe the opposite routing mechanism.**

The doc-comment header of `TestLiveMigrationCustomCdcPartitionKeyLeafLocalUniqueIndexAcrossLeavesWithDifferentCustomKey` was corrected to state routing by `id`, but the comments inside its `SourceDeltaSQL` were not swept — they are verbatim copies of the sibling `...AcrossLeaves` test (which routes by `region`) and now assert the opposite of what this test does:

- line 2215: "all events for a leaf share that leaf's region custom key"
- line 2216: "they route to the same channel"
- lines 2216-2217: "disjoint uk_val ranges ... so no cross-leaf bucket can coincide"
- line 2221: "r1 churn ... one channel"
- line 2229: "r2 churn ... a different channel"

This test routes by `id`, the delta interleaves both regions, and the whole point is that same-`uk_val` events land on different id-channels and DO conflict (`require.Greater(conflictStats.Total, 0)`).

This violates root `CLAUDE.md` ("When a refactor renames or restructures something, sweep the surrounding comments... A stale comment is worse than no comment") and `testlivemigration/AGENTS.md` ("Each test states in a doc comment ... why the chosen data pattern triggers it").

Failure case: a maintainer reading the delta comments concludes events serialize on one channel and no conflict can occur — the opposite of the assertion — and could "fix" the test in the wrong direction.

Suggested fix: rewrite the `SourceDeltaSQL` comments to state that routing is by the custom key `id`, so events for the same `uk_val` but different `id`s hash to different channels and produce cross-channel UK conflicts that must be detected.

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+	testutils.FatalIfError(t, err, "target does not match source after streaming")
+
+	err = lm.InitiateCutoverToTarget(false, nil)
+	testutils.FatalIfError(t, err, "failed to initiate cutover")
+
+	err = lm.WaitForCutoverComplete(0, 30)
+	testutils.FatalIfError(t, err, "failed to wait for cutover complete")
+}
+
+// TestLiveMigrationCustomCdcPartitionKeyLeafLocalUniqueIndexAcrossLeavesWithDifferentCustomKey pins that a leaf-local unique index (defined independently on each partition, not on the root) is handled correctly
```
makalaaneesh, line 2160: [AI][Suggestion] **Grammatical errors in this test's doc-comment header.**

Line 2160-2161 reads "under custom-key routing a different column of the table" (missing "by"); line 2168 "same uk val across leaves still be detected" is ungrammatical; and line 2160 is a single unwrapped long line unlike the sibling tests' ~100-col wrapped comments. The header was already touched by the prior fix, so cleanup here is in scope.

Suggested fix: reword to "... under custom-key routing BY a column distinct from both the partition column and the unique column — conflicts on the uk column must still be detected." and fix line 2168 to "Conflicts must be detected: because the custom key (`id`) differs from the unique column, the same uk_val on different ids lands on different channels."

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+	testutils.FatalIfError(t, err, "failed to wait for snapshot complete")
+
+	err = lm.InitMetaDB()
+	testutils.FatalIfError(t, err, "failed to initialize meta db")
+	importDataStatus, err := lm.GetMetaDB().GetImportDataStatusRecord()
+	testutils.FatalIfError(t, err, "failed to get import data status record")
+	assert.Equal(t, cmd.PARTITION_BY_CUSTOM, importDataStatus.TableToCDCPartitionKey[`"test_schema"."test_live"`].Strategy,
+		"test_live should use the custom partition strategy")
+	assert.Equal(t, []string{"id"}, importDataStatus.TableToCDCPartitionKey[`"test_schema"."test_live"`].Columns,
+		"test_live custom key columns should be persisted")
```
makalaaneesh, line 2277: [AI][Suggestion] **This test omits `--cdc-partition-key: "auto"` that its 4 sibling tests pass explicitly.**

`StartImportDataWithEnv` here passes only `--cdc-partition-key-overrides`, unlike the other four new tests (and the pre-existing `TestLiveMigrationWithCustomCdcPartitionKey`) which also pass `"--cdc-partition-key": "auto"`. Harmless (the flag defaults to "auto") but the inconsistency invites a reader to wonder whether it is significant.

Suggested fix: add `"--cdc-partition-key": "auto",` to the flags map for parity with the sibling tests.

---

makalaaneesh (commented):
[AI] Automated review — 3 findings (0 critical, 1 warning, 2 suggestion). See inline comments.

---

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+	testutils.FatalIfError(t, err, "failed to wait for forward streaming complete")
+
+	require.False(t, lm.GetImportRunner().IsStopped(),
+		"import should keep running during count failpoint mode")
+
+	conflictStats, err := testutils.ReadUniqueKeyConflictStats(uniqueKeyConflictStatsPath)
+	testutils.FatalIfError(t, err, "failed to read unique key conflict stats")
+	// The delta reclaims a freed uk_val 6 times across different id channels; each is at most
+	// one cross-channel conflict, so 0 < conflicts <= 6.
+	require.Greater(t, conflictStats.Total, 0, "cross-channel uk reclaim should produce UK conflicts")
```
makalaaneesh: [AI][Warning] **`...WithDifferentCustomKey` asserts cross-leaf false positives as expected, the opposite of what the PR body claims**

Import merges the leaf-local indexes `idx_test_live_r1_uk` and `idx_test_live_r2_uk` into a single root-level index by column signature (`dedupeUniqueIndexes`, `yb-voyager/src/tgtdb/postgres.go:573`). The delta alternates leaves, so each INSERT reclaims a `uk_val` freed in the **other** leaf: `del(100,r1)→ins(1,r2)`, `del(1,r2)→ins(2,r1)`, `del(2,r1)→ins(3,r2)`, plus the same three for 10001. That makes all six countable pairs cross-leaf, and neither source nor target has a constraint that spans leaves. The test asserts `Total > 0`, and its doc comment calls these conflicts that "should still be detected". The PR body says this case "is **not flagged as a false conflict**". The test also never shows that a real same-leaf reclaim is caught: each same-leaf predecessor DELETE has already been drained by the earlier cross-leaf wait. Per `src/testlivemigration/AGENTS.md`, the doc comment must state which invariant the test pins, and here the stated invariant, the body and the assertion disagree.

**Breaks when:** Suppose someone correctly scopes conflict buckets per leaf, which removes spurious flushes and waits on partitioned tables with leaf-local indexes. `Total` then drops to 0 and this test fails, even though the new behaviour is better. Meanwhile a regression that stops detecting real same-leaf reclaims across channels would still pass.

**Suggested fix:** Decide which invariant this test should pin. (a) Merged root-level detection is the accepted conservative behaviour: say that in the doc comment and the PR body, since the counted conflicts are cross-leaf false positives that cost waits but not correctness, and link a ticket if per-leaf scoping is wanted later. (b) The test should prove real conflicts are caught: reclaim inside one leaf across different `id` values and derive the bounds from that delta.

```sql
-- same-leaf reclaim across id channels: a real conflict on idx_test_live_r1_uk
DELETE FROM test_schema.test_live WHERE id = 100 AND region = 'r1';
INSERT INTO test_schema.test_live (id, region, uk_val) VALUES (1, 'r1', 1);
DELETE FROM test_schema.test_live WHERE id = 1 AND region = 'r1';
INSERT INTO test_schema.test_live (id, region, uk_val) VALUES (2, 'r1', 1);
```

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+		InitialDataSQL: []string{
+			// r1 and r2 each hold uk_val=777 simultaneously (leaf-local indexes) plus disjoint
+			// value ranges. 8 snapshot rows total.
+			`INSERT INTO test_schema.test_live (id, region, uk_val) VALUES
+				(100, 'r1', 1), (101, 'r1', 2), (150, 'r1', 777),
+				(200, 'r2', 10001), (201, 'r2', 10002), (202, 'r2', 10003), (250, 'r2', 777),
+				(102, 'r1', 3);`,
+		},
+		SourceDeltaSQL: []string{
+			// Single transaction. Within each leaf, free and reclaim one leaf-local uk_val across
```
makalaaneesh: [AI][Warning] **Comments copied from the sibling test are wrong for `...WithDifferentCustomKey`**

The SourceDeltaSQL comment and the inline SQL comments came over from `...LeafLocalUniqueIndexAcrossLeaves` unchanged and now contradict the code. They say "all events for a leaf share that leaf's region custom key, so they route to the same channel", "The two leaves use disjoint uk_val ranges (r1: 1, r2: 10001) so no cross-leaf bucket can coincide", "-- r1 churn on uk_val=1 (custom key 'r1' -> one channel)" and "-- r2 churn ... (custom key 'r2' -> a different channel)". In fact the key is `id`, the inserts alternate between r1 and r2, and 10001 is inserted into r1. The doc comment has problems too. A blank line at 2172 detaches it from the func, so it is no longer a Go doc comment. Line 2160 runs past 200 characters. The "Coexistence" bullet describes rows the delta never touches. Root `CLAUDE.md` says a stale comment is worse than none because it reads as authoritative.

**Suggested fix:** Rewrite the delta comment for the `(id)` key: each INSERT reclaims a `uk_val` freed by a different `id`, and so on a different channel, in the other leaf. Drop the per-block `custom key 'r1'/'r2'` lines, remove the blank line before `func`, and wrap the doc comment to the file's ~100-column style.

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+		`"public"."customers"`: {Inserts: 100, Updates: 50, Deletes: 0},
+	}, 120, 5)
+	testutils.FatalIfError(t, err, "failed to wait for forward streaming complete")
+
+	// custom_key is immutable and no primary key is recycled, so no conflicts should be detected.
+	conflicts, err := testutils.ReadUniqueKeyConflictStats(uniqueKeyConflictStatsPath)
+	if err != nil && !os.IsNotExist(err) {
+		testutils.FatalIfError(t, err, "failed to read unique key conflict stats")
+	}
+	require.Nil(t, conflicts, "multi-level cross-schema custom-key migration must not produce conflicts")
```
makalaaneesh: [AI][Warning] **No-conflict assertions in the row-movement and multi-level tests pass under any routing strategy**

In `...MultiLevelCrossSchemaPartitions`, the only conflict bucket is the leaf PK `(id)`, added by `addPrimaryKeyToConflictSetForCustomTables`, and the delta never reuses an `id`. In `...WithRowMovement` (line 1998), the PK is `(id, region)`, there is no unique index, and every moved row lands under a new PK value. So no event can ever match a cached bucket in either test, and `require.Nil(t, conflicts, ...)` would pass with custom-key routing, PK routing, or no exclusion at all. The comment at 2440 ("custom_key is immutable ... so no conflicts should be detected") suggests the custom key is what prevents conflicts, but nothing in the workload could conflict. `src/testlivemigration/AGENTS.md` requires each test's data pattern to actually trigger the invariant it names.

**Breaks when:** Break custom-key routing for multi-level or cross-schema roots, for example by resolving the partition key from a leaf tuple instead of the root. Both tests still pass their conflict assertion. Only data consistency and the persisted-strategy check remain as real signals.

**Suggested fix:** Either delete the conflict-stats read and the failpoint env from these two tests so they don't claim to check conflicts, or add a workload that can conflict. For multi-level: DELETE `id=1001` and re-INSERT `id=1001` with a different `custom_key`, which recycles the same leaf PK across channels, then assert a justified `1 <= Total <= 1`. For row movement, see S1.

```sql
DELETE FROM public.customers WHERE id = 1001;
INSERT INTO public.customers (id, custom_key, status, arr, email)
    VALUES (1001, 9001, 'ACTIVE', 50, 'recycled@example.com');  -- same leaf PK, new custom key -> different channel
```

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+			DatabaseName: "test_custom_key_eq_part_col_row_movement",
+		},
+		SchemaNames: []string{"test_schema"},
+		SchemaSQL: []string{
+			`CREATE SCHEMA IF NOT EXISTS test_schema;
+			CREATE TABLE test_schema.test_live (
+				id int,
+				region text,
+				val int,
+				PRIMARY KEY (id, region)
```
makalaaneesh: [AI][Suggestion] **The riskier row-movement case, with a leaf-only PK not including the partition column, is not covered**

`...WithRowMovement` uses `PRIMARY KEY (id, region)`, so a moved row's DELETE and INSERT carry different PK values and cannot race. The hazardous variant is a partitioned root with no PK and leaves owning `PRIMARY KEY (id)`, imported with `--use-partition-root false`. That is the same shape as the new multi-level test and the existing `TestLiveMigrationPartitionedTableChildPKWithCustomCdcPartitionKeyPKRecycleConflict`. There, moving `id=i` from r1 to r2 emits DELETE `(i)` routed on the `'r1'` channel and INSERT `(i)` on the `'r2'` channel with the **same** PK. If the INSERT applies first it hits a duplicate key, unless conflict detection flags the pair. That is the case where "custom key = partition column + row movement" can actually go wrong.

**Suggested fix:** Add a variant, or change this test, to use leaf-only `PRIMARY KEY (id)` and `--use-partition-root false`. Assert `1 <= Total <= 50` (at most one pending DELETE per moved row), plus `Updates: 0` and data consistency.

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+	require.False(t, lm.GetImportRunner().IsStopped(),
+		"a nullable custom key with NULL values must not stop the import")
+
+	// No conflicts: the uk_val=1 free/reclaim is entirely among NULL-keyed rows, which all share the
+	// sentinel partition key (same channel) and are therefore applied in commit order.
+	conflicts, err := testutils.ReadUniqueKeyConflictStats(uniqueKeyConflictStatsPath)
+	if err != nil && !os.IsNotExist(err) {
+		testutils.FatalIfError(t, err, "failed to read unique key conflict stats")
+	}
+	require.Nil(t, conflicts, "NULL-keyed rows share the sentinel channel, so no cross-channel conflict should occur")
```
makalaaneesh: [AI][Suggestion] **NULL-key test is one-sided: nothing proves NULL-keyed events are still conflict-checked, or that the sentinel stays distinct from real keys**

Every reclaim in the delta is NULL-key to NULL-key, and the only assertion is `require.Nil(conflicts)`. The test would still pass if NULL-keyed events were skipped by conflict detection altogether, or if the NULL sentinel collided with a real key value. Either of those is exactly the kind of NULL-handling regression this test is meant to catch. `src/testlivemigration/AGENTS.md` asks count assertions to have justified bounds on both sides.

**Suggested fix:** Add one cross-key reclaim where a NULL-keyed DELETE frees a `uk_val` and a non-NULL-keyed INSERT reclaims it. These route to different channels, so assert exactly one detected conflict. The expected total goes to 6 ins / 1 upd / 3 del.

```go
-- NULL-keyed row frees uk_val=100 (sentinel channel); a non-NULL key reclaims it on another channel
DELETE FROM test_schema.test_live WHERE id = 100;          -- custom_key NULL, uk_val 100
INSERT INTO test_schema.test_live (id, custom_key, uk_val, val) VALUES (400, 77, 100, 400);
...
require.NotNil(t, conflicts)
require.Equal(t, 1, conflicts.Total, "NULL-key delete vs non-NULL-key insert on the same uk_val must be flagged")
```

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+
+	// Root-level counts: 100 inserts (50 explicit + 50 from row movement), 0 updates, 50 deletes.
+	err = lm.WaitForForwardStreamingComplete(map[string]ChangesCount{
+		`"test_schema"."test_live"`: {Inserts: 100, Updates: 0, Deletes: 50},
+	}, 120, 5)
+	testutils.FatalIfError(t, err, "failed to wait for forward streaming complete")
+
+	// The custom key equals the partition column and "changes" via row movement, but because that
+	// surfaces as DELETE+INSERT (not an UPDATE), the immutability guard must not stop the import.
+	require.False(t, lm.GetImportRunner().IsStopped(),
```
makalaaneesh: [AI][Suggestion] **`IsStopped` checks after streaming completes never catch the failure they describe**

When the immutability guard fires, `hashEvent` returns an error and the import process exits. Because of that, `WaitForForwardStreamingComplete` never reaches the expected counts and fails after 120s with the generic "streaming phase did not complete". The following `require.False(t, lm.GetImportRunner().IsStopped(), "... must not trip the immutability guard")` is never the failing line, and the same applies at 2288 and 2577. A reader debugging a guard regression gets a timeout, not the intended message.

**Suggested fix:** Before the streaming wait, poll with `require.Never` or `Eventually` on `IsStopped()` alongside the wait, or check the import's stderr for `is required to be immutable`, so a guard trip fails fast with the specific message. Otherwise drop the redundant checks.

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+	err := lm.SetupContainers(context.Background())
+	testutils.FatalIfError(t, err, "failed to setup containers")
+
+	err = lm.SetupSchema()
+	testutils.FatalIfError(t, err, "failed to setup schema")
+
+	err = lm.StartExportData(true, nil)
+	testutils.FatalIfError(t, err, "failed to start export data")
+
+	uniqueKeyConflictCountFailpointEnv := testutils.GetFailpointEnvVar(
```
makalaaneesh: [AI][Suggestion] **Around 45 lines of setup and assertion boilerplate are duplicated five times**

Each new test repeats the same pieces: the failpoint env string, the stats path, the persisted-strategy `assert.Equal` pair on `TableToCDCPartitionKey[...]`, the `ReadUniqueKeyConflictStats` + `os.IsNotExist` tolerance, and the cutover tail. Root `CLAUDE.md` asks for duplicate logic to be consolidated into shared helpers. Two smaller issues come with the copies. The persisted-strategy check uses `assert`, not `require`, even though it is the precondition that makes the rest of each test meaningful (`yb-voyager/AGENTS.md`: use `require` for setup steps that must succeed). And the `...WithDifferentCustomKey` test silently omits `--cdc-partition-key auto` where its siblings pass it.

**Suggested fix:** Add small helpers in this file and use them in the five new tests.

```go
func requireCustomCdcPartitionKeyPersisted(t *testing.T, lm *LiveMigrationTest, table string, cols []string) {
	require.NoError(t, lm.InitMetaDB())
	status, err := lm.GetMetaDB().GetImportDataStatusRecord()
	require.NoError(t, err)
	key, ok := status.TableToCDCPartitionKey[table]
	require.True(t, ok, "no persisted cdc partition key for %s", table)
	require.Equal(t, cmd.PARTITION_BY_CUSTOM, key.Strategy)
	require.Equal(t, cols, key.Columns)
}

func readUniqueKeyConflictStatsOrNil(t *testing.T, path string) *testutils.UniqueKeyConflictStats {
	stats, err := testutils.ReadUniqueKeyConflictStats(path)
	if os.IsNotExist(err) {
		return nil
	}
	require.NoError(t, err)
	return stats
}
```

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+	testutils.FatalIfError(t, err, "failed to get import data status record")
+	assert.Equal(t, cmd.PARTITION_BY_CUSTOM, importDataStatus.TableToCDCPartitionKey[`"test_schema"."test_live"`].Strategy,
+		"test_live should use the custom partition strategy")
+	assert.Equal(t, []string{"region"}, importDataStatus.TableToCDCPartitionKey[`"test_schema"."test_live"`].Columns,
+		"test_live custom key columns should be persisted")
+
+	err = lm.ExecuteSourceDelta()
+	testutils.FatalIfError(t, err, "failed to execute source delta")
+
+	// Root-level counts: 6 inserts, 0 updates, 6 deletes.
```
makalaaneesh: [AI][Suggestion] **Narration comments that restate the next line**

Several comments paraphrase the literal directly below them: `// Root-level counts: 6 inserts, 0 updates, 6 deletes.` above `{Inserts: 6, Updates: 0, Deletes: 6}` (also at 1982, 2282, 2434, 2569), and `// 8 snapshot rows, including uk_val=777 present in both leaves at once.` above `: 8` (2118, 2264). Root `CLAUDE.md` says to never write a line that paraphrases the code it sits on. The derivations inside the SQL blocks (e.g. "100 inserts (50 explicit + 50 from row movement)") do earn their place.

**Suggested fix:** Delete the restating comments above the `WaitForForwardStreamingComplete` / `WaitForSnapshotComplete` maps. Keep the derivation that lives next to the delta SQL.

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+// test_live is LIST-partitioned by region and routed by the custom key (region). A row's region
+// is immutable in place: changing it moves the row to another partition, which PostgreSQL logical
+// decoding emits at the leaf level as DELETE (old leaf) + INSERT (new leaf), never as an UPDATE.
+// So even though the custom key "changes" from the user's point of view, the importer never sees
+// an UPDATE event whose custom-key column changes between before- and after-image -- the
+// immutability guard (which only inspects UPDATE events) must NOT fire, and the migration must
+// stay consistent. The DELETE (before-image region=r1) routes to the r1 channel and the INSERT
+// (region=r2) to the r2 channel; they carry different primary keys ((id, r1) vs (id, r2)), so no
+// primary-key or unique-key conflict arises.
+func TestLiveMigrationCustomCdcPartitionKeyEqualsPartitionColumnWithRowMovement(t *testing.T) {
```
makalaaneesh: [AI][Suggestion] **PR body gives the wrong build tag, so its run command executes zero tests**

The PR body says the tests use build tag `failpoint_import` and to run `go test -tags failpoint_import -run TestLiveMigrationCustomCdcPartitionKey ...`. The file's constraint is `//go:build integration_live_migration_with_failpoint`, which is what `.github/workflows/integration-tests.yml` runs. With `-tags failpoint_import` the file is excluded and the command passes vacuously. CI does run the tests under the correct tag.

**Suggested fix:** Change the "How was this tested" section to `failpoint-ctl enable && go test -tags integration_live_migration_with_failpoint -run TestLiveMigrationCustomCdcPartitionKey ./src/testlivemigration/` (from `yb-voyager/`), then `failpoint-ctl disable`.

makalaaneesh (commented):
[AI] Automated review — 9 findings (0 critical, 3 warning, 6 suggestion). See inline comments.

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
makalaaneesh: this seems like a false positive then?

yb-voyager/src/testlivemigration/live_migration_uk_conflict_test.go
```diff
@@ -1870,6 +1870,731 @@ func TestLiveMigrationWithCoveringUniqueKeyIndex(t *testing.T) {
...
+//
+// test_live is routed by the custom key (custom_key), which is left NULL for some rows. The delta
+// keeps custom_key immutable per row (NULL stays NULL, so the immutability guard never fires) and:
+//   - free/reclaims a unique index value (uk_val=1) across different primary keys where every
+//     involved row has custom_key=NULL. Because all NULL-keyed events share the sentinel partition
+//     key, they route to the same channel and apply in commit order, so this must produce NO
+//     cross-channel conflict.
+//   - streams an in-place non-key UPDATE on a NULL-keyed row (routing off the NULL before-image)
+//     and inserts of non-NULL-keyed rows, to prove mixed NULL/non-NULL flow stays consistent.
+func TestLiveMigrationCustomCdcPartitionKeyWithNullValues(t *testing.T) {
```
makalaaneesh: If not already covered, we should write a test that asserts that NULL -> NOT NULL for the custom partition key fails because we don't allow mutation of custom partition key.

PR #3823, opened 2026-09-21
Update migtest expectations for pg_parquet and postgres_fdw on YB 2026.1.2

Abhi-yb:

Release qualification of YugabyteDB 2026.1.2.0-b137 (Jenkins 8888, 8889) produced 8 migtest failures. All are stale expectations on our side — no migration step errors and no product defect. DB-23675.

**`postgres_fdw` (7 tests)** — from 2026.1.2 a newly created cluster ships with `postgres_fdw` already installed in `pg_catalog` (global views, yugabyte-db#30591). Two symptoms:

- `osm`, `stackexchange`, `sample-is`, `adventureworks`, `hasura-ecommerce` compare the reported extensions against a fixed set. They never create `postgres_fdw`, but it is now present.
- `omnibus` and `pgtbrus` do create it in their own schema. Voyager exports `CREATE EXTENSION IF NOT EXISTS ... WITH SCHEMA <schema>`, so the import only logs `NOTICE: extension "postgres_fdw" already exists, skipping`. The extension stays in `pg_catalog` and its five routines are absent from the target schema, dropping the per-schema counts by exactly 5 (omnibus `foreign_db_example` 7→2, pgtbrus `public` 9→4).

Whether the extension is present depends on when the cluster was created, not on the running binary — a pre-2026.1.2 cluster restarted on new binaries does not have it. The new checks therefore read the schema `postgres_fdw` is installed in rather than gating on the YB version. `get_extension_schema` in `migtests/lib/yb.py` provides that lookup.

**`pg_parquet` (1 test)** — unrelated and pre-existing. The gate in `multiple-schemas/validate` was `if "2025.2" in tgtVersion`, a substring match that can never hold for `2026.1.2.0`, so `pg_parquet` was never added to the expected list even though the build ships it. Replaced with a version range.

None. Test-only changes.

Ran the affected migtests locally against 2026.1.2.0-b137.

On a freshly created cluster (`postgres_fdw` pre-installed, matching Jenkins):

| Test | Before | After |
| --- | --- | --- |
| `pg/pgtbrus` | `AssertionError` (9 vs 4) | pass |
| `pg/stackexchange` | `AssertionError` on extension set | pass |
| `pg/omnibus` | `AssertionError` (7 vs 2) | pass |
| `pg/multiple-schemas` | `Unexpected extensions found: {'pg_parquet'}` | pass |

On a cluster created before 2026.1.2 (`postgres_fdw` absent), `pg/pgtbrus` and `pg/stackexchange` still pass, exercising the other branch of the new checks.

No.

No.

pgcomplex_schemas -- https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/8910/
pg tests -- https://jenkins.dev.yugabyte.com/job/users/job/yb-voyager-testing/job/yb-voyager-testing-pipeline/8909/parameters/


migtests/tests/pg/adventureworks/validate
```diff
@@ -129,7 +129,12 @@ def migration_completed_checks(tgt):
 
 	extension_list = tgt.fetch_all_pg_extension("public")
 	print(f"extension_list: {extension_list}\n EXPECTED_EXTENSION_SET: {EXPECTED_EXTENSION_SET}")
-	assert extension_list == EXPECTED_EXTENSION_SET
+	expected_extension_set = set(EXPECTED_EXTENSION_SET)
```
makalaaneesh: [AI][Warning] **Duplicated pre-install adjustment block across 5 validate scripts.**

`migtests/AGENTS.md` (Validation Scripts) requires reusable validation patterns to be library functions in `lib/yb.py`, not copy-pasted across validate scripts. This "add `("postgres_fdw",)` to the expected set when pre-installed" block is duplicated verbatim in `adventureworks`, `hasura-ecommerce`, `osm`, `sample-is`, and `stackexchange`. Only the raw lookup (`get_extension_schema`) was extracted; the adjustment pattern was not, so a future change to the rule (see the `pg_catalog`-narrowness comment on line 135) has to be made in five places and will drift.

Suggested — add a helper to `lib/yb.py`:
```python
def with_preinstalled_fdw(self, expected_set):
    result = set(expected_set)
    if self.get_extension_schema("postgres_fdw") is not None:
        result.add(("postgres_fdw",))
    return result
```
and call `expected_extension_set = tgt.with_preinstalled_fdw(EXPECTED_EXTENSION_SET)` in each script.

---

migtests/tests/pg/adventureworks/validate
```diff
@@ -129,7 +129,12 @@ def migration_completed_checks(tgt):
...
+	# Some YB cluster setups pre-install postgres_fdw into pg_catalog; it is not
+	# created by this migration and depends on cluster setup, not the YB version.
+	if tgt.get_extension_schema("postgres_fdw") == "pg_catalog":
```
makalaaneesh: [AI][Warning] **`== "pg_catalog"` guard is narrower than the schema-agnostic assert it protects.**

`fetch_all_pg_extension()` in `lib/yb.py` ignores its `schema_name` argument and returns every extension via `SELECT extname FROM pg_extension`, so `extension_list` contains `postgres_fdw` whenever it exists in *any* schema. These five tests never create `postgres_fdw` themselves, so its only source is a pre-install. Gating the expected-set augmentation on the schema being exactly `pg_catalog` means that a cluster which pre-installs `postgres_fdw` into some other schema makes `assert extension_list == expected_extension_set` fail even though nothing is actually wrong.

Suggested — gate on presence, not schema: `if tgt.get_extension_schema("postgres_fdw") is not None:` (folds naturally into the shared helper suggested on line 132).

---

migtests/tests/pg/omnibus/validate
```diff
@@ -236,6 +238,12 @@ def migration_completed_checks(tgt):
 	    for objects, new_value in changes.items():
 	        EXPECTED_COUNTS[schema][objects] = new_value
 
+	# Some YB cluster setups pre-install postgres_fdw into pg_catalog, which turns the
+	# migrated CREATE EXTENSION into a no-op and leaves its routines out of the target
+	# schema. This depends on cluster setup, not on the YB version.
+	if tgt.get_extension_schema("postgres_fdw") != "foreign_db_example":
+		EXPECTED_COUNTS["foreign_db_example"]["functions_procedures_count"] -= POSTGRES_FDW_ROUTINE_COUNT
+
```
makalaaneesh, line 245: [AI][Warning] **`POSTGRES_FDW_ROUTINE_COUNT = 5` can drive the expected count negative on a non-PG15 cluster.**

The comment above states the pre-install "depends on cluster setup, not on the YB version" — i.e. it can happen on any version. But `5` is the PG15/PG16 `postgres_fdw` routine count; older `postgres_fdw` ships only 2 functions. Here `foreign_db_example.functions_procedures_count` has base `4` and is overridden to `7` only under `has_pg15_merge`, while this subtraction is *not* gated on PG15. On a non-PG15 cluster that pre-installs `postgres_fdw`, the count becomes `4 - 5 = -1`, so `assert len(...) == -1` can never pass. `pgtbrus/validate:59` has the same coupling (`6 - 5 = 1` — non-negative, but still the wrong count on a non-PG15 pre-install).

Suggested — either gate the subtraction to PG15 clusters, or derive `POSTGRES_FDW_ROUTINE_COUNT` from `has_pg15_merge(tgtVersion)` (5 for PG15, 2 otherwise) and guard with `max(0, …)`. If in practice the pre-install is only a PG15 (2026.1.x) phenomenon, say that in the comment instead of "not version-dependent," so the coupling is documented.

---

migtests/lib/yb.py
```diff
@@ -338,6 +338,12 @@ def fetch_all_pg_extension(self, schema_name = "public") -> set[str]:
 		cur.execute(f"SELECT extname FROM pg_extension")
 		return set(cur.fetchall())
 
+	def get_extension_schema(self, extension_name) -> str:
```
makalaaneesh: [AI][Suggestion] **`get_extension_schema` return annotation should be `Optional[str]`.**

`def get_extension_schema(self, extension_name) -> str:` can return `None` (`return row[0] if row else None`). The `-> str` annotation is inaccurate and misleads callers into assuming a non-null string — the two callers in this PR both compare against string literals, which is exactly the case that silently misbehaves if a reader trusts the annotation.

Suggested — `from typing import Optional` and annotate `-> Optional[str]`.

---

makalaaneesh (commented):

---

migtests/tests/pg/multiple-schemas/validate
```diff
@@ -196,6 +190,11 @@ def migration_completed_checks(tgt):
 				if age_ext in extensions:
 					expectedExtensions.append(age_ext)
 
+		# pg_parquet ships in 2025.2.0-2025.2.2 and 2025.2.5+, is dropped again in
+		# 2026.1.0-2026.1.1 and returns in 2026.1.2. Assert it only when reported.
+		if "pg_parquet" in extensions:
+			expectedExtensions.append("pg_parquet")
```
priyanshi-yb: what does this mean? We should assert this only when the target version is something we expect?

migtests/tests/pg/adventureworks/validate
```diff
@@ -127,7 +127,9 @@ def migration_completed_checks(tgt):
 	check_validations_per_schema(tgt, "sa")
 	check_validations_per_schema(tgt, "sales")
 
-	extension_list = tgt.fetch_all_pg_extension("public")
+	# YB pre-installs postgres_fdw from 2026.1.2 (yugabyte-db#30591); this migration
+	# does not create it, so it is not part of what this check validates.
+	extension_list = tgt.fetch_all_pg_extension("public") - {("postgres_fdw",)}
```
makalaaneesh: [AI][Suggestion] **postgres_fdw-exclusion idiom copy-pasted across validate scripts**

`migtests/AGENTS.md` (Validation Scripts) says reusable validation patterns should be library functions in `lib/yb.py`, "not copy-pasted across test validate scripts." The `- {("postgres_fdw",)}` set-subtraction is repeated verbatim (with the identical 2-line comment) in 5 scripts (adventureworks, hasura-ecommerce, osm, sample-is, stackexchange), and the `[p for p in ... if not p.startswith("postgres_fdw")]` filter in 2 more (omnibus, pgtbrus).

The block is much smaller now, so this is a judgment call — a one-liner set-difference is close to the root `AGENTS.md` "favor simplicity / YAGNI" line, but the reuse pattern is explicit in the standard and a future pre-install change still means editing 7 sites.

Suggested fix: add a small helper next to `fetch_all_pg_extension`, e.g.

```python
def fetch_all_pg_extension_excluding_preinstalled(self, schema_name="public") -> set:
    # postgres_fdw is pre-installed in pg_catalog from YB 2026.1.2 (yugabyte-db#30591);
    # migrations under test do not create it.
    return self.fetch_all_pg_extension(schema_name) - {("postgres_fdw",)}
```

and call it from the five extension-check scripts.

(Separate pre-existing note, out of scope: `fetch_all_pg_extension`'s `schema_name` argument is ignored — the query has no schema filter.)

---

makalaaneesh (commented):
[AI] Automated review (incremental re-review of new commits since the last AI review) — 1 finding (0 critical, 0 warning, 1 suggestion). See inline comment. The previously posted [AI] findings are all addressed by the new commits.

---

migtests/tests/pg/pgtbrus/validate
```diff
@@ -38,21 +38,12 @@ def migration_completed_checks(tgt):
...
-	# for PG15 merged versions (preview >= 2.25 or stable >= 2025.1), some extensions create more functions/procedures than in earlier versions.
-	# Example hstore
-	# 2024.2: 57 functions / procedures
-	# 2.25+: 60 functions / procedures
-	# Thus the change in count.
-
-	if yb.has_pg15_merge(tgtVersion):
-		assert len(fetched_procedures_functions) == 9
-	else:
-		assert len(fetched_procedures_functions) == 6
```
priyanshi-yb: why have we removed this?

PR #3825, opened 2026-09-21
Add callhome metrics for the conflict detection

priyanshi-yb:

---

| Component        


yb-voyager/src/callhome/import_data_metrics_collector.go
```diff
@@ -66,3 +74,9 @@ func (c *ImportDataMetricsCollector) GetSnapshotTotalBytes() int64 {
 	defer c.RUnlock()
 	return c.snapshotTotalBytes
 }
+
+func (c *ImportDataMetricsCollector) GetCdcConflictCountPerTable() map[string]int64 {
+	c.RLock()
+	defer c.RUnlock()
+	return c.cdcConflictCountPerTable
```
makalaaneesh: [AI][Critical] **`GetCdcConflictCountPerTable` returns the live internal map by reference — concurrent map read/write panic during live migration.**

This getter returns `c.cdcConflictCountPerTable` (the internal map) and releases `RLock` on return. `packAndSendImportDataToTargetPayload` (`importData.go:1983`) assigns this reference into the payload, which is later `json.Marshal`'ed inside `SendPayload` with no lock held. Meanwhile `IncrementConflictCountForTable` (line 51) keeps writing to the same map under `Lock` on the streaming/conflict-detection path.

`sendCallhomePayloadAtIntervals` runs pack+send in a background goroutine every 15 minutes (started at `cmd/root.go:177`) while streaming is active, so the marshal iterates the map concurrently with writes → Go runtime `fatal error: concurrent map read and map write`, which is unrecoverable and aborts the migration. The other getters return scalars (value copies) and are safe; only this one leaks a reference.

Trigger: live migration with SendDiagnostics on, at least one CDC conflict recorded; the 15-minute callhome send marshals the map while a new conflict is being counted.

Suggested fix — return a copy under the lock:
```go
func (c *ImportDataMetricsCollector) GetCdcConflictCountPerTable() map[string]int64 {
	c.RLock()
	defer c.RUnlock()
	out := make(map[string]int64, len(c.cdcConflictCountPerTable))
	for k, v := range c.cdcConflictCountPerTable {
		out[k] = v
	}
	return out
}
```

---

yb-voyager/src/callhome/import_data_metrics_collector.go
```diff
@@ -26,13 +26,15 @@ type ImportDataMetricsCollector struct {
...
 	currentParallelConnections int
+	cdcConflictCountPerTable   map[string]int64
 }
 
 func NewImportDataMetricsCollector() *ImportDataMetricsCollector {
 	return &ImportDataMetricsCollector{
 		snapshotTotalRows:          0,
 		snapshotTotalBytes:         0,
 		currentParallelConnections: 0,
+		cdcConflictCountPerTable: make(map[string]int64),
```
makalaaneesh: [AI][Warning] **File is not gofmt-clean.**

`gofmt -l` flags this file: the added struct-literal line `cdcConflictCountPerTable: make(map[string]int64),` breaks the colon alignment of the surrounding fields. Root `AGENTS.md` requires new/edited files to be gofmt-clean, and lint mirrors CI.

Suggested fix: run `gofmt -w`; the field should align as `cdcConflictCountPerTable:   make(map[string]int64),`.

---

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -335,6 +340,22 @@ func (c *ConflictDetectionCache) WaitUntilNoConflict(incomingEvent *tgtdb.Event)
 // failing the migration. Caller must hold the lock.
 func (c *ConflictDetectionCache) recordConflictMetricLocked(incomingEvent *tgtdb.Event) {
 	metrics.Get().RecordImportCDCConflict(c.importerRole, incomingEvent.TableNameTup.ForOutput())
```
makalaaneesh: [AI][Warning] **Prometheus conflict metric still receives the raw table name, contradicting this PR's own updated comment.**

`recordConflictMetricLocked` passes `incomingEvent.TableNameTup.ForOutput()` (raw `schema.table`) to `metrics.Get().RecordImportCDCConflict`, whose parameter is named `anonymizedTableName` and is used verbatim as a Prometheus label on `importCDCConflictsTotal` (`src/metrics/prometheus.go:365`). The struct comment this PR edits (lines 138–141) now claims `anonymizedTableNames` ensures "no raw table name reaches the metrics endpoint" — but the `/metrics` Prometheus endpoint still gets the raw identifier.

The raw call predates this PR, but the PR modifies this exact function, adds the anonymized-name map here, and makes the comment assert the opposite; this is the natural place to complete the anonymization.

Trigger: `--metrics-port` enabled; `yb_voyager_import_data_cdc_conflicts_total` exposes real user table names under a label documented as anonymized.

Suggested fix — pass the precomputed anonymized name (same fallback) to the Prometheus call:
```go
anonymizedTableName, ok := c.anonymizedTableNames.Get(incomingEvent.TableNameTup)
if !ok {
	anonymizedTableName = "XXX"
}
metrics.Get().RecordImportCDCConflict(c.importerRole, anonymizedTableName)
```
(then keep the callhome guard/increment below using the same value). If the raw name is intentional for the local Prometheus endpoint, reword the struct comment so it only claims anonymization for the callhome metric.

---

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -335,8 +335,9 @@ Version History:
 1.4: Added CutoverTimings field
 1.5: Added table list count to ImportDataMetrics
 1.6: Added iterative cutover enabled and next iteration migration UUID fields
+1.7: Added CdcConflictCountPerTable field to ImportDataMetrics
 */
-var IMPORT_DATA_CALLHOME_PAYLOAD_VERSION = "1.6"
+var IMPORT_DATA_CALLHOME_PAYLOAD_VERSION = "1.7"
```
makalaaneesh: [AI][Warning] **Expected callhome migtest payload not updated for the version bump and new field.**

This is anchored to the payload-version bump because the affected expected file (`migtests/tests/pg/datatypes/expected_callhome_payloads/import_data_callhome.json`) is not part of this PR's diff — and that is exactly the problem.

The callhome `AGENTS.md` testing rule requires updating expected migtest payload files when payload fields change. The comparison (`migtests/scripts/functions.sh:normalize_json`) does a sorted full diff and does not ignore `payload_version` or `data_metrics`. This PR bumps `payload_version` 1.6 → 1.7 and, because `CdcConflictCountPerTable` has no `omitempty` and the collector initializes it non-nil, the payload now always emits `"cdc_conflict_count_per_table": {}`. The expected file still shows `"payload_version": 1.6` and lacks the field, so the callhome migtest will fail.

Suggested fix: update the expected `import_data_callhome.json` (and any other `import_data_callhome` expected payloads) to `payload_version` 1.7 and add `"cdc_conflict_count_per_table": {}` under `data_metrics`.

---

yb-voyager/cmd/conflictDetectionCache.go
```diff
@@ -335,6 +340,22 @@ func (c *ConflictDetectionCache) WaitUntilNoConflict(incomingEvent *tgtdb.Event)
...
+	// The callhome conflict metric is best-effort: record it only when diagnostics are enabled
+	// AND the collector has been initialized. callhomeMetricsCollector is a global set up only
+	// by a full import run (see importData.go), so it is nil on paths/tests that never initialize
+	// it (e.g. import-data-to-source/source-replica, unit tests). This mirrors the nil guards at
+	// every other callhomeMetricsCollector call site.
+	if !callhome.SendDiagnostics || callhomeMetricsCollector == nil {
+		return
+	}
+	anonymizedTableName, ok := c.anonymizedTableNames.Get(incomingEvent.TableNameTup)
+	if !ok {
```
makalaaneesh: [AI][Suggestion] **`"XXX"` is an in-band sentinel that collapses distinct tables in the metric.**

When no anonymized name is precomputed, all such tables are counted under the literal `"XXX"`, merging distinct tables into one bucket. Root `AGENTS.md` discourages in-band sentinels.

This is best-effort observability and documented, so it is minor, but a more explicit marker (e.g. `"__unanonymized__"`) or omitting the count would avoid silently conflating tables and reads less like a placeholder that could occur in real data.

---

yb-voyager/src/callhome/diagnostics.go
```diff
@@ -373,6 +374,8 @@ type ImportDataMetrics struct {
 	SnapshotTotalBytes      int64 `json:"snapshot_total_bytes"`
 	CdcEventsImportRate3min int64 `json:"cdc_events_import_rate_3min"`
 
+	CdcConflictCountPerTable map[string]int64 `json:"cdc_conflict_count_per_table"`
```
makalaaneesh: [AI][Suggestion] **Consider `omitempty` for `CdcConflictCountPerTable`.**

Without `omitempty`, offline and no-conflict runs always emit `"cdc_conflict_count_per_table": {}`, adding noise to every import payload and forcing all expected migtest payloads to carry it. Adding `omitempty` would omit the empty map.

Weigh against local consistency — most sibling fields don't use `omitempty`; author's call.

---

makalaaneesh (commented):
[AI] Automated review — 6 findings (1 critical, 3 warning, 2 suggestion). See inline comments.

---

PR #3827, opened 2026-09-21
callhome: report schema detect-drift runs

ShivanshGahlot:

`schema detect-drift` registered `--send-diagnostics` but sent nothing, so we have no signal on a new feature: not whether anyone runs it, and not whether it ever finds drift.

- **New phase payload** — `schema-detect-drift`, `SchemaDriftPhasePayload` v1.0. A new phase, so no existing payload version moves.
- **Counts and two histograms** — changes found, intervals compared, captures stored, whether the live read was compared, and drift grouped by type and by severity.
- **No identifiers** — schema, table and column names are not sent. The anonymized schema names already travel in `SourceDBDetails`.
- **`drifts_by_type` is the point** — which DDL changes actually happen to a source mid-migration is the thing we guess at today.
- **Failed runs report too** — a run that compared nothing still sends, since "the feature was unusable here" is the signal we most need.
- **`utils.ErrExitErr`** is now set on the operational-error path, so the atexit handler files those runs as ERROR instead of a clean exit.

`MigrationType` is left unset: detect-drift reads an export directory, and nothing in it says whether the migration is offline or live.

`--send-diagnostics` already existed on this command but did nothing. It now governs a real payload.

`go test -tags unit ./src/callhome/` pins the payload's JSON shape against an expected struct, matching how every other phase payload is covered. `TestCountDriftsBy` covers the histograms. Gap: nothing asserts the payload is actually sent on each exit path; that needs the end-to-end harness in the last PR of this stack.

Yes — new `SchemaDriftPhasePayload` at version 1.0. No existing payload version changed, because this is a new phase rather than an edit to an existing one.

No. The payload is sent, not persisted.


yb-voyager/cmd/detectDrift.go
```diff
@@ -688,6 +699,79 @@ func nothingComparedError(r schemadrift.Report) error {
...
+	if err := callhome.SendPayload(&payload); err == nil && (status == COMPLETE || status == ERROR) {
+		callHomeErrorOrCompletePayloadSent = true
+	}
+}
+
+func buildSchemaDriftPayload(errorMsg error, report *schemadrift.Report) callhome.SchemaDriftPhasePayload {
+	driftPayload := callhome.SchemaDriftPhasePayload{
+		PayloadVersion:   callhome.SCHEMA_DRIFT_CALLHOME_PAYLOAD_VERSION,
+		OutputFormats:    utils.CsvStringToSlice(driftOutputFormat),
+		Error:            callhome.SanitizeErrorMsg(errorMsg, anonymizer),
```
makalaaneesh, line 735: [AI][Suggestion] **Anonymization of the nothing-compared error relies on `SanitizeErrorMsg`'s colon-truncation, which can silently leak schema names.**

`buildSchemaDriftPayload` sets `Error: callhome.SanitizeErrorMsg(errorMsg, anonymizer)`. On the nothing-compared path (`detectDrift.go:603`) the error is `nothingComparedError(report)`, whose default branch (`:695-698`) interpolates `strings.Join(reasons, "; ")`, and those reasons are built from `CapturePoint.Excluded`, which embeds raw schema names (`schemadrift/build.go:82-83`: `"captured only %s, so it cannot answer for %s"`). Today those names aren't sent only because `SanitizeErrorMsg` keeps just `strings.Split(err.Error(), ":")[0]` (`diagnostics.go:627`) and the names happen to fall after the first colon. That's a fragile, non-obvious guarantee: a reword that moves an identifier ahead of the first colon, or any change to the sanitizer, would silently leak schema names — which `src/callhome/AGENTS.md` forbids.

Failure case: source captured `public,sales` but scope expects `hr` → `Excluded = "captured only public, sales, so it cannot answer for hr"` → carried into the error; safe today purely by colon position.

Suggested fix: don't hand the reason-bearing error to the payload. On the nothing-compared path send a stable, identifier-free error and return the descriptive one to the caller — the histogram/count fields already convey the signal:
```go
nothingCompared := nothingComparedError(report)
packAndSendSchemaDriftPayload(ERROR, errors.New("nothing to compare"), &report)
return false, nothingCompared
```
so anonymization no longer depends on where a colon lands in a user-facing message.

---

makalaaneesh (commented):
[AI] Automated review — 1 finding (0 critical, 0 warning, 1 suggestion). See inline comment.

---

